PR #1, opened 2019-12-10
Initial implementation

imeckler:
Creating on behalf of @Ltvvlz to have a PR for the initial implementation


marlin-succinct/circuits/src/compiled.rs
```diff
@@ -0,0 +1,117 @@
...
+
+*****************************************************************************************************************/
+
+use sprs::CsMat;
+use commitment::urs::URS;
+use algebra::{Field, PrimeField, PairingEngine};
+use ff_fft::{DensePolynomial, Evaluations, EvaluationDomain};
+pub use super::index::Index;
+
+pub struct Compiled<E: PairingEngine>
```
imeckler: We will need the row/col/val polynomials evaluated on multiple domains. Check out the `Index` type [here](https://github.com/scipr-lab/marlin/blob/master/src/ahp/indexer.rs). I think it should be factored better into 3 records, one for each of A, B, C,  but I do believe we need all the values included in that type.

marlin-succinct/protocol/src/prover.rs
```diff
@@ -0,0 +1,599 @@
...
+        // the first part of the argument needs the following oracles:
+        // alpha, betta[0] in order to compute RA and evaluate Z, ZA & ZB polynomials
+
+        let mut argument = RandomOracleArgument::<E::Fr>::new(index.oracle_params.clone());
+        
+        argument.commit_scalar(&E::Fr::one()); // this commit comes from the previous proof context
+        oracles.alpha = argument.challenge();
+        // commit the public input into the argument
+        argument.commit_slice(&witness.0[0..witness.1]);
+        oracles.betta[0] = argument.challenge();
```
imeckler: Just a note on spelling, should be `beta`.

marlin-succinct/protocol/src/prover.rs
```diff
@@ -0,0 +1,599 @@
...
+
+        // evaluate Z, ZA & ZB polynomials
+        w_eval = w.evaluate(oracles.betta[0]);
+        za_eval = za.evaluate(oracles.betta[0]);
+        zb_eval = zb.evaluate(oracles.betta[0]);
+
+        // run the random oracle argument further to query
+        // oracles gamma[0..3]
+        argument.commit_scalar(&w_eval);
+        oracles.gamma[0] = argument.challenge();
```
imeckler: Can we use `eta_A, eta_B, eta_C`?

Ltvvlz: Sure. You prefer not to use oracle arrays like eta[0..3] but rather eta_A, eta_B, eta_C ?

marlin-succinct/protocol/src/prover.rs
Ltvvlz: Thank you for spotting...

marlin-succinct/circuits/src/compiled.rs
Ltvvlz: OK, investigating this.

marlin-succinct/protocol/src/batch_verifier.rs
```diff
@@ -0,0 +1,101 @@
...
+    (
+        &self,
+        index: &Index<E>,
+        rng: &mut dyn RngCore
+    ) -> Result<bool, ProofError>
+    {
+        let mut batch = vec![Vec::new(), Vec::new(), Vec::new()];
+        for proof in self.batch.iter()
+        {
+            match proof
```
imeckler: I think you can use the `?` operator to make this code less nested: https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html

marlin-succinct/protocol/src/prover.rs
```diff
@@ -0,0 +1,599 @@
...
+            }
+        ).collect();
+        r.reverse();
+        let ra = DensePolynomial::<E::Fr>::from_coefficients_vec(r);
+
+        // compute commitments, evaluations and openings for the first sumcheck argument
+        // proceed with the random oracle argument further to query additional oracles
+        // betta[1], batch[0]
+
+        match Self::sumcheck_1_compute (index, &ra, &zv, &z, &oracles)
```
imeckler: I believe you can similarly use the `?` operator here https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html

marlin-succinct/protocol/src/prover.rs
imeckler: Maybe a record with `a, b, c` fields? I would rather the not leave the mapping {0, 1, 2} <-> {A, B, C} implicit.

marlin-succinct/protocol/src/prover.rs
```diff
@@ -134,29 +134,29 @@ impl<E: PairingEngine> ProverProof<E>
...
+        w_eval = w.evaluate(oracles.beta[0]);
+        za_eval = za.evaluate(oracles.beta[0]);
+        zb_eval = zb.evaluate(oracles.beta[0]);
 
         // run the random oracle argument further to query
-        // oracles gamma[0..3]
+        // oracles eta[0..3]
         argument.commit_scalar(&w_eval);
-        oracles.gamma[0] = argument.challenge();
+        oracles.eta[0] = argument.challenge();
```
imeckler: You can absorb the commitments to `w, z_A, z_B` before squeezing `alpha` and the `eta` values -- it's better for efficiency to not switch between absorbing and squeezing.

imeckler: Oh actually I don't quite understand what's going on, it seems things are not in the correct order here.

imeckler: I believe the order should be as in https://github.com/CodaProtocol/coda/blob/double-marlin-verifier/src/lib/double_marlin/double_marlin.ml#L202. And it is not necessary to absorb any polynomial evalutions

marlin-succinct/protocol/src/prover.rs
```diff
@@ -172,189 +172,171 @@ impl<E: PairingEngine> ProverProof<E>
...
+        sigma2 = g2.coeffs[0];
+        g2.coeffs.remove(0);
+
+        // evaluate H2, G2 polynomials
+        h2_eval = h2.evaluate(oracles.beta[1]);
+        g2_eval = g2.evaluate(oracles.beta[1]);
+
+        // query batch[1], beta[2] random oracles from the argument context
+        argument.commit_scalar(&h2_eval);
+        oracles.batch[1] = argument.challenge();
```
imeckler: There is no need to have separate challenges for all the batches. You can just sample one at the end of the prover sending all the polynomial commitments and messages (I call it [xi here](https://github.com/CodaProtocol/coda/blob/double-marlin-verifier/src/lib/double_marlin/double_marlin.ml#L239))

marlin-succinct/protocol/src/prover.rs
```diff
@@ -187,13 +187,13 @@ impl<E: PairingEngine> ProverProof<E>
             zb_comm,
             h1_comm,
             g1_comm,
-            h2_comm : index.urs.commit(&h2, index.h_group.size())?,
-            g2_comm : index.urs.commit(&g2, index.h_group.size()-1)?,
-            h3_comm : index.urs.commit(&h3, index.h_group.size()*6)?,
-            g3_comm : index.urs.commit(&g3, index.h_group.size()-1)?,
+            h2_comm: index.urs.commit(&h2, index.h_group.size())?,
+            g2_comm: index.urs.commit(&g2, index.h_group.size()-1)?,
+            h3_comm: index.urs.commit(&h3, index.h_group.size()*6)?,
```
imeckler: should be `index.k_group.size()` I think?

marlin-succinct/protocol/src/prover.rs
```diff
@@ -187,13 +187,13 @@ impl<E: PairingEngine> ProverProof<E>
...
+            g3_comm: index.urs.commit(&g3, index.h_group.size()-1)?,
```
imeckler: same here

PR #2, opened 2019-12-22
This commit implements the following:

Ltvvlz:
URS G2 elements trimming. The set of URS G2 elements needed by the protocol is determined by the set of maximal degrees of the polynomials that are committed. This set of polynomials is determined by the set of circuits supported by the same instance of the URS. This commit removes all other URS G2 elements.

NOTE: the Index "constructor" creates single URS instance to support the protocol for that index. Next commits should generalize this to one URS instance being able to support multiple indexes/circuits.

### DIFF
--- a/marlin-succinct/circuits/src/compiled.rs
+++ b/marlin-succinct/circuits/src/compiled.rs
@@ -36,7 +36,7 @@ pub struct Compiled<E: PairingEngine>
 impl<E: PairingEngine> Compiled<E>
 {
     // this function compiles the constraints
-    //  urs: univarsal reference string
+    //  urs: universal reference string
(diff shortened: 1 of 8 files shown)


Ltvvlz:
As of now, the Index "constructor" creates single URS instance to support the protocol for that index. Next commits should generalize this for one URS instance to support multiple indexes/circuits.

imeckler:
Is the trimming problematic from the point of view of supporting multiple circuits? In practice usage will look like the following:

1. Create a single URS for a maximum size of N (which will be created using an MPC).
2. Over time, create indexes for circuits of depth < N using the previously created URS.

imeckler:
Going to merge this but in the new year let's sync on the above.

PR #4, opened 2019-12-31
Fix degree bounding in the prover

imeckler:
This PR fixes how degree bounding is handled in polynomial commitments. I believe what was there previously was not sound.

Suppose we have fixed the Kate URS (x^0 * g, x^1 * g, x^2 * g, ..., x^N * g). For a polynomial f = (f_0, f_1, ..), let [f] denote the multiexp f_0 * (x^0 * g) + f_1 * (x^1 * g) + ...

In this PR, commitments are made as follows (following the marlin paper as mentioned in [this issue](https://github.com/o1-labs/marlin/issues/3)):

- For a polynomial f whose degree does not need to be bounded, the commitment to f is [f].
- For a polynomial f whose degree needs to be bounded by d, the commitment to f is the pair ([f], [x^{N - d} * f]).

When opening a commitment (U, V) to a polynomial f bounded by d to a value y at a point a, we check that U opens to y and V opens to a^{N - d} * y.

I had some trouble understanding the verifier code so I did not update that.

**EDIT** going to do this differently, see the below comment.


imeckler:
I just realized this method of enforcing degree bounds does something bad when producing the batched opening proofs. Namely it includes the shifted polynomials x^{N - d} f(x) in the sum for the batching. This is bad because it means that sum then has degree N, and so every opening proof takes the maximal amount of time.

What we can do instead is the following (thanks to Ariel Gabizon for [the idea on Twitter](https://twitter.com/relgabizon/status/1212440419235176448)): 

To prove a degree bound of d on f, provide U := commit(f) and V := commit(x^{N - d} f).
Only include the evaluation of f in the batched evaluation sum. Then, separately check the degree bound on f by checking the pairing e(U, (beta^{N-d}) h) = e(V, h).

PR #8, opened 2019-12-31
this commits the initial implementation of D-log-based Marlin protocol

Ltvvlz:
1. D-log polycommitment with degree bounds with shifted and unshifted commitments as per Marlin
2. Batching d-log polycommitment openings across multiple evaluation points
3. D-log polycommitment tests
4. Protocol with d-log polycommitment unit tests


marlin-dlog/protocol/src/prover.rs
```diff
@@ -0,0 +1,483 @@
...
+            zb_comm,
+            h1_comm,
+            g1_comm,
+            h2_comm: index.srs.commit(&h2, index.h_group.size()-1)?,
+            g2_comm: index.srs.commit(&g2, index.h_group.size()-1)?,
+            h3_comm: index.srs.commit(&h3, index.k_group.size()*6-6)?,
+            g3_comm: index.srs.commit(&g3, index.k_group.size()-1)?,
+
+            // polynomial commitment batched opening proofs
+            proof1: index.srs.open
```
imeckler: All the opening proofs should be batched into one.

The way to do this is described [here in the first section](https://www.overleaf.com/project/5ddbf8f6b95ef30001b7a31d). The idea is you take all the polynomials (w, za, zb, h1, g1, g2, h2, g3, h3, matrix polynomials) and sum them with the Horner linear combination like F := sum_i xi^i * f_i. You also pick a random r_j for each evaluation point (the betas).

Then you use the inner product argument on < F_i, sum_j r_j pows(beta_j) > where pows(x) = (1, x, x^2, ...).

Ltvvlz: The following has been committed into this branch:

1. D-log polycommitment with degree bouns with shifted and unshifted commitments as per Marlin
2. Batching d-log polycommitment openings across multiple evaluation points
3. D-log polycommitment tests
4. Protocol with d-log polycommitment unit tests

PR #11, opened 2020-01-06
Pass in URS to index create function

imeckler:
This PR makes it possible to pass in a previously generated URS into the index create function.


imeckler:
Whoops. Thanks so much!

PR #12, opened 2020-01-09
Tweaks

imeckler:
This PR has a few tweaks, most of which I needed to make the prover line up with the constraint system verifier. Namely:

- reversing the order of the polynomial batching and making the first coefficient 1 rather than the batching scalar
- increasing the number of rounds in poseidon
- fixing the marlin sponge
- changing the order of some things in the prover
- adding a few "oracle" values

I also 
- added serialization functions for the URS
- trimmed unnecessary G2 elements from the verifier index


Ltvvlz:
1. Reviewed the changes.

The following difficulties have been observed and potential corrections offered in commit 60e50741c292085f650671926d4dc35162a5780a.

1. Correct polynomial commitment unit test
2. Reverse the changes relevant to the commitment to x_hat polynomial. The inconsistency with the employed commitment is with the commitment being done over one subset of urs.gp group elements whereas the commitment opening proof is done over another subset of urs.gp group elements. The other approach would be to use exponentiate_sub_domain function for commitment and do "dilute" x_hat polynomial upon commitment opening (according to how the public input is mixed into the witness) and recalculating the "diluted" evaluation upon the commitment opening verification.

Ltvvlz:
commit b352ccad80608eb607e70e4390b497a9781694ab add additional possible change remove not needed G2 elements from prover and verifier indexes

imeckler:
Ah thank you for your fixes -- I don't know what I was thinking using the spread out CRS, silly of me :)

PR #17, opened 2020-01-15
this commits 6K->3K polynomial degree reduction

ghost:
This PR proposes implementation of 6K->3K polynomial degree reduction in both Marlins

### DIFF
--- a/dlog/circuits/src/compiled.rs
+++ b/dlog/circuits/src/compiled.rs
@@ -22,8 +22,10 @@ pub struct Compiled<G: AffineCurve>
     pub col_comm: G,
     pub row_comm: G,
     pub val_comm: G,
+    pub rc_comm: G,
 
     // compiled polynomials and evaluations
+    pub rc      : DensePolynomial<Fr<G>>,
     pub row     : DensePolynomial<Fr<G>>,
     pub col     : DensePolynomial<Fr<G>>,
     pub val     : DensePolynomial<Fr<G>>,
@@ -33,6 +35,7 @@ pub struct Compiled<G: AffineCurve>
     pub row_eval_b: Evaluations<Fr<G>>,
     pub col_eval_b: Evaluations<Fr<G>>,
     pub val_eval_b: Evaluations<Fr<G>>,
+    pub rc_eval_b : Evaluations<Fr<G>>,
 }
 
 impl<G: AffineCurve> Compiled<G>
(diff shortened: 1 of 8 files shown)


dlog/circuits/src/compiled.rs
```diff
@@ -114,11 +121,14 @@ impl<G: AffineCurve> Compiled<G>
...
     ) -> Vec<Fr<G>>
     {
-        self.row_eval_b.evals.iter().zip(self.col_eval_b.evals.iter()).map
+        self.row_eval_b.evals.iter().
+            zip(self.col_eval_b.evals.iter()).
+            zip(self.rc_eval_b.evals.iter()).
+            map
         (
-            |(row, col)|
+            |((row, col), rc)|
```
imeckler: I think the previous implementation of this way more efficient (it used only 1 multiplication rather than the 3 used here).

dlog/circuits/src/index.rs
```diff
@@ -82,7 +82,7 @@ impl<G: AffineCurve> Index<G>
         (
             EvaluationDomain::<Fr<G>>::new(h_group_size),
             EvaluationDomain::<Fr<G>>::new(k_group_size),
-            EvaluationDomain::<Fr<G>>::new(k_group_size * 6 - 6),
+            EvaluationDomain::<Fr<G>>::new(k_group_size * 3 - 3),
```
imeckler: :grinning:

pairing/circuits/src/compiled.rs
```diff
@@ -112,11 +119,14 @@ impl<E: PairingEngine> Compiled<E>
...
     ) -> Vec<E::Fr>
     {
-        self.row_eval_b.evals.iter().zip(self.col_eval_b.evals.iter()).map
+        self.row_eval_b.evals.iter().
+            zip(self.col_eval_b.evals.iter()).
+            zip(self.rc_eval_b.evals.iter()).
+            map
         (
-            |(row, col)|
+            |((row, col), rc)|
```
imeckler: same remark about efficiency

dlog/circuits/src/compiled.rs
ValarDragon: But they're not equivalent as polynomials though. Letting `f,g` be defined as follows:
`f(x) = (row(x) - alpha) * (col(x) - beta)`
`g(x) = row(x)*col(x) - alpha * col(x) - beta*row(x) + alpha*beta`

`f(x) != g(x)` for any x not in `K`. `x` can't be in `K` due to zero-knowledge (and completeness IIRC?)

imeckler: Sorry, you're right -- I got confused and thought it was just evaluating over a domain where they agreed.

PR #25, opened 2020-01-24
optimize commitment opening a bit, add sg to opening proofs

imeckler:
This PR

1. Optimizes the computation of the "g" array in the bulletproofs opening.
2. Adds the "sg" polynomial to opening proofs so that we don't have to recompute them individually.

### DIFF
--- a/dlog/commitment/Cargo.toml
+++ b/dlog/commitment/Cargo.toml
@@ -15,3 +15,4 @@ colored = "1.9.2"
 rand = "0.7.3"
 rayon = { version = "1" }
 blake2 = "0.7"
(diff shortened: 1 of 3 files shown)


ltvvlz:
1. Reviewed the windowed Shamir optimization
2. Reviewed parallelization optimizations with "rayon"

3. Reviewing the addition of sg polynomial in opening proofs

ltvvlz:
Why do we use blinding factors in the polycommitment schema? They are used for hiding, we care for soundness. Do we need it ? It seems we would get faster schema without hiding especially, since we do not use zero-knowledge in the core protocol.

PR #28, opened 2020-01-31
splitting polynomials with dlog-based protocol

ltvvlz:
This completes the Marlin optimizations with the splitting of polynomials for the dlog-based protocol.

NOTE: the commits to this branch also contain the feature of the polynomial degree reduction for the pairing-base Marlin protocol.


dlog/commitment/src/commitment.rs
```diff
@@ -125,51 +132,51 @@ fn shamir_sum<G: AffineCurve>(
...
-    //     RETURN: commitment group element
-    pub fn commit_no_degree_bound(&self, plnm: &DensePolynomial<Fr<G>>) -> Result<G, ProofError> {
-        if self.g.len() < plnm.coeffs.len() {
-            return Err(ProofError::PolyCommit);
+        max: Option<usize>,
+    ) -> PolyComm<G>
+    {
+        PolyComm::<G>
+        {
+            // committing all the segments without shifting
```
imeckler: Can you separate some of the subexpressions out into separate bindings? I think it would make this a bit more readable. Like

```rust
let n = self.g.len();
let segments = (plnm.len() + (n-1)) / n;
let unshifted = (0..segments).map(|i| ...);
```

dlog/commitment/src/commitment.rs
```diff
@@ -125,51 +132,51 @@ fn shamir_sum<G: AffineCurve>(
...
+            unshifted: (0..plnm.len()/self.g.len() + if plnm.len()%self.g.len() != 0 {1} else {0}).map
+            (
+                |i|
+                {
+                    VariableBaseMSM::multi_scalar_mul
+                    (
+                        &self.g,
+                        &plnm.coeffs[i*self.g.len()..if (i+1)*self.g.len() > plnm.coeffs.len() {plnm.coeffs.len()} else {(i+1)*self.g.len()}]
```
imeckler: Does it not work to just do `[i*n..(i + 1) * n]` ? Will it not just truncate as needed?

dlog/commitment/src/commitment.rs
```diff
@@ -125,51 +132,51 @@ fn shamir_sum<G: AffineCurve>(
...
+                Some(max) =>
+                {
+                    if max % self.g.len() == 0 {None}
+                    else
+                    {
+                        // committing only last segment shifted to the right edge of SRS
+                        Some(VariableBaseMSM::multi_scalar_mul
+                        (
+                            &self.g[self.g.len() - (max%self.g.len())..],
+                            &plnm.coeffs[max-(max%self.g.len())..if max < plnm.coeffs.len() {max} else {plnm.coeffs.len()}]
```
imeckler: Same remark about the right boundary

dlog/commitment/src/commitment.rs
```diff
@@ -196,20 +203,27 @@ impl<G: AffineCurve> SRS<G> {
         let p = {
             let mut p = DensePolynomial::<Fr<G>>::zero();
             let mut scale = Fr::<G>::one();
+
```
imeckler: Can you declare a function above this code like

```rust
// TODO: Figure out how to return iterator here
fn segments<F:Field>(n: usize, p : DensePolynomial<F>, degree_bound: Option<usize>) -> Vec<DensePolynomial<F>> {
    let d = p.coeffs.len();
    let segments = (d + (n - 1)) / n;

    let mut res : Vec<_> =
        (0..segments).map(|i| {
            DensePolynomial::from_coefficients_slice(
                &p.coeffs[(i*n)..((i+1)*n)])
        }).collect() ;

    match degree_bound {
        Some(bound) => res.push(
            DensePolynomial::from_coefficients_slice(
                &p.coeffs[(segments - 1)*n..]).shiftr(n - (bound % n))),
        None => ()
    };

    res
}
```

and then in here just do

```rust
            for (p_i, degree_bound) in plnms.iter() {
                for seg in segments(self.g.len(), p_i, *degree_bound) {
                    p += &seg.scale(scale);
                    scale *= &polyscale;
                }
            }
```

dlog/commitment/src/commitment.rs
```diff
@@ -364,11 +378,11 @@ impl<G: AffineCurve> SRS<G> {
             Fr<G>,      // scaling factor for polynoms
             Fr<G>,      // scaling factor for evaluation point powers
             Vec<(
-                G,                  // unshifted polycommitment
-                Vec<Fr<G>>,         // vector of evaluations
-                Option<(G, usize)>, // shifted polycommitment and degree bound
+                &PolyComm<G>,       // polycommitment
+                Vec<&Vec<Fr<G>>>,   // vector of evaluations
+                Option<usize>,      // optional degree bound
```
imeckler: Maybe the degree bound should be inside the PolyComm struct along with the optional group element?

pairing/circuits/src/compiled.rs
```diff
@@ -17,10 +17,10 @@ pub struct Compiled<E: PairingEngine>
     pub constraints: CsMat<E::Fr>,
 
     // compiled polynomial commitments
-    pub col_comm: E::G1Affine,
-    pub row_comm: E::G1Affine,
-    pub val_comm: E::G1Affine,
-    pub rc_comm: E::G1Affine,
+    pub col_comm: PolyComm<E::G1Affine>,
```
imeckler: Can you remove the commits which do this on the pairing version so we can make sure they work one at a time?

dlog/commitment/src/commitment.rs
ltvvlz: This has been addressed.

dlog/commitment/src/commitment.rs
ltvvlz: This has been addressed.

dlog/commitment/src/commitment.rs
ltvvlz: This has been addressed.

dlog/commitment/src/commitment.rs
ltvvlz: This doesn't quite seem to work (unless I misunderstand):

1. p.coeffs[(i*n)..((i+1)*n)]) could result in boundary access violation
2. This function puts the shifted segment into the same vector/iterator. However it has to be handled separately since it has to be mixed into the accumulator poly shifted to the rigjt edge of SRS..

dlog/commitment/src/commitment.rs
ltvvlz: I struggle to see the benefit.
Upon serialization, PolyComm has what needs to be serialized. Degree bound would be discarded since it is part of the protocol.  Why to carry it around ?

pairing/circuits/src/compiled.rs
ltvvlz: This is completed: state of the pairing-based Marlin is returned to that committed in the Master branch at the point and off which this branch has been forked.

ltvvlz:
The fixes/improvements (in response to the comments) have been committed in  a6dece1 commit 2e7413a1705e5dd56243b035eaef01ac6781d51a

PR #29, opened 2020-02-06
Various fixes 

imeckler:
My apologies for the ill-defined PR. This contains various fixes to achieve consistency with the verifier circuit, mostly in changing the order in which certain values are combined.

It also implements an improved algorithm for computing the function (g1, x1, g2, x2) -> x1*g1 + x2*g2 using 2 bit windows.


ltvvlz:
1. Reviewed the changes for every commit
2. Loaded the branch, run the Marlin tests

PR #32, opened 2020-04-12
Modified scalar challenge

imeckler:
This PR defines an injective map f: 2^128 -> F, and replaces a subset of the challenges `c` in marlin with `f(c)`. 

The reason is that for these challenges, the verifier previously computed `c * P` (for some point `P`), but, it is much cheaper to compute `f(c) * P` inside the circuit. Therefore, one thinks of the value `c` squeezed from the sponge as an encoding of the field element `f(c)`.

Or another way of thinking about it is that instead of sampling challenges as a uniform random 128 bit string, we select a uniformly random element from the set { f(c) | c a 128 bit challenge } and since f is injective this is ok.

The technique (which uses an endomorphism on the curve) is described in the halo paper.

### DIFF
--- a/dlog/commitment/src/commitment.rs
+++ b/dlog/commitment/src/commitment.rs
@@ -18,7 +18,7 @@ use algebra::{
     UniformRand, VariableBaseMSM, SWModelParameters
 };
 use ff_fft::DensePolynomial;
-use oracle::FqSponge;
+use oracle::{FqSponge, marlin_sponge::ScalarChallenge};
 use rand_core::RngCore;
 use rayon::prelude::*;
 use std::iter::Iterator;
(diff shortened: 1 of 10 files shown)


ValarDragon:
Minor point, as I couldn't tell from the PR whats being done. 128 bits of randomness isn't sufficient for soundness in a lot of the protocol. Most of your interactive soundness errors will be `poly_degree / {size of the set randomness is sampled from}`. e.g. for Marlin, the soundness of the first round by using 128 bits challenges would be `H/ (2^128)`, which for circuits of interest is  `~= 2^110`. The same is true of random linear combinations of polynomials.

imeckler:
> Minor point, as I couldn't tell from the PR whats being done. 128 bits of randomness isn't sufficient for soundness in a lot of the protocol. Most of your interactive soundness errors will be `poly_degree / {size of the set randomness is sampled from}`. e.g. for Marlin, the soundness of the first round by using 128 bits challenges would be `H/ (2^128)`, which for circuits of interest is `~= 2^110`. The same is true of random linear combinations of polynomials.

That's a good point and definitely affects the implementation as it exists!

PR #33, opened 2020-05-05
Affine "combine" for dlog commitment

imeckler:
This PR implements the affine version of the combining function for the dlog commitment. Runtime is about 20-25% better than the previous best.

E.g.,

```
$ cargo test bench_combine --release -- --nocapture

combine 100000 7.490999722s
affine_window_combine 100000 5.190810545s
affine_combine 100000 6.07763902s
window_combine 100000 6.652958733s
```

The current best is `affine_window_combine`, the previous was `window_combine`.

### DIFF
--- a/dlog/commitment/src/combine.rs
+++ b/dlog/commitment/src/combine.rs
@@ -1,0 +1,727 @@
+use algebra::{
+    curves::models::short_weierstrass_jacobian::{GroupAffine as SWJAffine},
+    AffineCurve, BitIterator, Field, PrimeField, ProjectiveCurve,
+    SWModelParameters
+};
+use itertools::Itertools;
+use rayon::prelude::*;
+
+fn add_pairs_in_place<P:SWModelParameters>(p: &mut Vec<SWJAffine<P>>) {
+    let len = if p.len()%2 == 0 {p.len()} else {p.len()-1};
+    let mut denominators = p.chunks_exact_mut(2).map
+    (
+        |p|
+        if p[0].x == p[1].x
+        {
+            if p[1].y == P::BaseField::zero() {P::BaseField::one()} else {p[1].y.double()}
+        } else {p[0].x - &p[1].x}
+    ).collect::<Vec<_>>();
+
+    algebra::fields::batch_inversion::<P::BaseField>(&mut denominators);
+
+    for (i, d) in (0..len).step_by(2).zip(denominators.iter())
+    {
+        let j = i/2;
+        if p[i+1].is_zero() == true
+        {
+            p[j] = p[i];
+        }
+        else if p[i].is_zero() == true
+        {
+            p[j] = p[i+1];
+        }
+        else if p[i+1].x == p[i].x && (p[i+1].y != p[i].y || p[i+1].y == P::BaseField::zero())
+        {
+            p[j] = SWJAffine::<P>::zero();
+        }
+        else if p[i+1].x == p[i].x && p[i+1].y == p[i].y
+        {
+            let sq = p[i].x.square();
+            let s = (sq.double() + &sq + &P::COEFF_A) * d;
+            let x = s.square() - &p[i].x.double();
+            let y = -p[i].y - &(s * &(x - &p[i].x));
+            p[j].x = x;
+            p[j].y = y;
+        }
+        else
+        {
+            let s = (p[i].y - &p[i+1].y) * d;
+            let x = s.square() - &p[i].x - &p[i+1].x;
+            let y = -p[i].y - &(s * &(x - &p[i].x));
+            p[j].x = x;
+            p[j].y = y;
+        }
+    }
+
+    let len = p.len();
+    if len % 2 == 1
+    {
+        p[len/2] = p[len-1];
+        p.truncate(len/2+1);
+    }
+    else
+    {
+        p.truncate(len/2);
+    }
+}
+
+fn batch_add_assign<P:SWModelParameters>(
+    mut denominators: &mut [P::BaseField],
+    v0: &mut [SWJAffine<P>],
+    v1: &[SWJAffine<P>] ) {
+    let n = v0.len();
+
+    for i in 0..n {
+        let p0 = v0[i];
+        let p1 = v1[i];
+        let d = 
+            if p0.x == p1.x
+            {
+                if p1.y == P::BaseField::zero() {P::BaseField::one()} else {p1.y.double()}
+            } else {p0.x - &p1.x};
+        denominators[i] = d;
+    }
+
+    algebra::fields::batch_inversion::<P::BaseField>(&mut denominators);
+
+    for (i, d) in (0..n).zip(denominators.iter())
+    {
+        let p0 = v0[i];
+        let p1 = v1[i];
+
+        if p1.is_zero() == true {
+        }
+        else if p0.is_zero() == true
+        {
+            v0[i] = p1;
+        }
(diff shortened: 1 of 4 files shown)


ltvvlz:
`pub fn affine_combine<P: SWModelParameters>(g1: &Vec<SWJAffine<P>>, g2: &Vec<SWJAffine<P>>, x1: P::ScalarField, x2: P::ScalarField) -> Vec<SWJAffine<P>> {
    const CHUNK_SIZE : usize = 10_000;
    let b : Vec<_> = g1.chunks(CHUNK_SIZE).zip(g2.chunks(CHUNK_SIZE)).collect();
    let v : Vec<_> = b.into_par_iter().map(|(v1, v2)| affine_combine_base(v1, v2, x1, x2)).collect();
    v.concat()
}`

CHUNK_SIZE=10_000. Too small array/chunk size can degrade the batched Affine addition benefit since it is best for the longer arrays. Also, too small thread jobs may result in suboptimal CPU parallelism since the jobs would be going up and down too often. It could be better to parallelize elsewhere and do the batched addition on longer arrays/chunks.

ltvvlz:
`#![feature(test)]

extern crate test;
mod combine;
pub mod srs;
pub mod commitment;
`
This does not work for relase builds, looks to work only for noghtly.

imeckler:
> `pub fn affine_combine<P: SWModelParameters>(g1: &Vec<SWJAffine<P>>, g2: &Vec<SWJAffine<P>>, x1: P::ScalarField, x2: P::ScalarField) -> Vec<SWJAffine<P>> { const CHUNK_SIZE : usize = 10_000; let b : Vec<_> = g1.chunks(CHUNK_SIZE).zip(g2.chunks(CHUNK_SIZE)).collect(); let v : Vec<_> = b.into_par_iter().map(|(v1, v2)| affine_combine_base(v1, v2, x1, x2)).collect(); v.concat() }`
> 
> CHUNK_SIZE=10_000. Too small array/chunk size can degrade the batched Affine addition benefit since it is best for the longer arrays. Also, too small thread jobs may result in suboptimal CPU parallelism since the jobs would be going up and down too often. It could be better to parallelize elsewhere and do the batched addition on longer arrays/chunks.

Hmm, given the structure of the prover and the commitment opening function, I'm not sure there is another place to parallelize. Do you have an idea in mind?

This chunk size had good performance in the 100,000 - 1,000,000 range which is of interest. Also, 10,000 seemed to be big enough that processing the whole array at once had similar performance to running this code with RAYON_NUM_THREADS=1.

Do you think we should set the chunk size dynamically somehow based on the input size?

imeckler:
> `#![feature(test)]
> 
> extern crate test;
> mod combine;
> pub mod srs;
> pub mod commitment;
> `
> This does not work for relase builds, looks to work only for nightly.

Ah ok, I will move the tests around to get rid of this.

ltvvlz:
> pub fn affine_combine<P: SWModelParameters>(g1: &Vec<SWJAffine<P>>, g2: &Vec<SWJAffine<P>>, x1: P::ScalarField, x2: P::ScalarField) -> Vec<SWJAffine<P>> { const CHUNK_SIZE : usize = 10_000; let b : Vec<_> = g1.chunks(CHUNK_SIZE).zip(g2.chunks(CHUNK_SIZE)).collect(); let v : Vec<_> = b.into_par_iter().map(|(v1, v2)| affine_combine_base(v1, v2, x1, x2)).collect(); v.concat() }
> CHUNK_SIZE=10_000. Too small array/chunk size can degrade the batched Affine addition benefit since it is best for the longer arrays. Also, too small thread jobs may result in suboptimal CPU parallelism since the jobs would be going up and down too often. It could be better to parallelize elsewhere and do the batched addition on longer arrays/chunks.
> 
> Hmm, given the structure of the prover and the commitment opening function, I'm not sure there is another place to parallelize. Do you have an idea in mind?
> This chunk size had good performance in the 100,000 - 1,000,000 range which is of interest. Also, 10,000 seemed to be big enough that processing the whole array at once had similar performance to running this code with RAYON_NUM_THREADS=1.
> Do you think we should set the chunk size dynamically somehow based on the input size?

My general understanding as well as the observations from the experimentation with multiexp suggest that it is more optimal to parallelize over larger jobs where threads execute as long as possible rather than over short jobs. Overall CPU utilization may drop when jobs are small. I do not know how the prover will be organized in the system. Do we consider a setup where prover has to compute few proofs in parallel ? If so, this could be a place to parallelize. Otherwise, I agree, it is not clear where else to parallelize.

In regards to the chunk size, how about using a chunk size which is the array length divided by the number of CPUs (let cpus = num_cpus::get(); from crate "num_cpus") ? If the number of CPUs is too large, then, possibly, set then set chunk size to something specific, like you do.

imeckler:
> Do we consider a setup where prover has to compute few proofs in parallel ? If so, this could be a place to parallelize. Otherwise, I agree, it is not clear where else to parallelize.

The key speed metric is in just the single proof setting. The other setting is important as well, but the most important thing is driving down the single proof generation time.

> In regards to the chunk size, how about using a chunk size which is the array length divided by the number of CPUs (let cpus = num_cpus::get(); from crate "num_cpus") ?

Doesn't this have the potential to underutilize the CPU? When I run it, the CPU utilization appears to be total, and I think the fixed cost per chunk is pretty low if the chunk is at least a few thousand.

The theoretical best we could hope for is

(2 * affine-double + 1 * affine-add) / (2 * jacobian-double + 1 * jacobian-mixed-add)
=
(2 * 6 + 1 * 6) / (2 * 7 + 1 * 11) = 0.72.

We get pretty close to this so I think the parallelization strategy is good. 

Maybe more convincingly, the single-threaded performance advantage is essentially the same as the parallel one, so I don't think any issues are coming from the parallelization strategy.

ltvvlz:
> > Do we consider a setup where prover has to compute few proofs in parallel ? If so, this could be a place to parallelize. Otherwise, I agree, it is not clear where else to parallelize.
> 
> The key speed metric is in just the single proof setting. The other setting is important as well, but the most important thing is driving down the single proof generation time.
> 
> > In regards to the chunk size, how about using a chunk size which is the array length divided by the number of CPUs (let cpus = num_cpus::get(); from crate "num_cpus") ?
> 
> Doesn't this have the potential to underutilize the CPU? When I run it, the CPU utilization appears to be total, and I think the fixed cost per chunk is pretty low if the chunk is at least a few thousand.
> 
> The single-threaded performance advantage (when we don't do any chunking at all) is essentially the same as the parallel one, so I don't think any issues are coming from the parallelization strategy or chunking.
> 
> Just for the record, the theoretical best we could hope for is roughly
> 
> ```
> (2 * affine-double + 1 * affine-add) / (2 * jacobian-double + 1 * jacobian-mixed-add)
> =
> (2 * 6 + 1 * 5) / (2 * 7 + 1 * 11) = 0.68.
> ```
> 
> (although this ignores the cost in producing the window tables, where I think we probably have an even bigger advantage)
> 
> The empirical advantage was 5.19 / 6.65 ~= 0.78 so definitely room for improvement, I just don't think it's coming from the parallelization strategy.

OK, I understand, makes sense. Since you have seen the benefit close to that of the singlethreaded case, it looks like it work well. As for underutilisation of CPU, as long as the chunk number is divisible by CPU number (even if one chunk per CPU) it does not look to be underutilised, especially, if the chunk length is, with this setup, longer. But again, since you saw the multithr  performance close to the optimal, it looks like your implementation is, at least, one of the optimal ones.

PR #43, opened 2020-05-27
Plonk

ltvvlz:
This PR is to review the pairing-based Plonk implementation.

The following protocols are implemented:
1. Compile Index. This includes:
   a. the compilation of the Constraint System
   b. interpolating polynomials needed by prover and verifier
   c. computation of the commitments to the Index polynomials
2. Prover
3. Verifier

Verification of the implemented functionality is implemented with plonk_group_addition.rs:pairing_plonk_group_addition test. To run the test, execute the following command:
cargo test --release -- --nocapture pairing_plonk_group_addition

This test verifies the following:

1. indexer/prover/verifier protocols of Plonk.
2. public input enforcement
3. wire permutation constraints

The diagram, depicting the circuit of the proved/verified computation statement is attached as diagram.jpg for reference.


circuits/plonk/src/gate.rs
```diff
@@ -0,0 +1,65 @@
...
+This source file implements Plonk computation wire index primitive.
+
+*****************************************************************************************************************/
+
+use algebra::Field;
+
+#[derive(Clone)]
+pub struct CircuitGate<F: Field>
+{
+    pub l: (usize, usize),   // left input wire index and its permutation
```
imeckler, line 12: Could you improve the comment here? It's not clear to me what the meaning of the `(usize, usize)` pair is. Don't we only need a single `usize` per each of `l, r, o` saying e.g., the equivalence class of variable that it belongs to?

dlog/plonk/src/index.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    pub fn create<'b>(size: usize, spec : SRSSpec<'a, G>) -> SRSValue<'a, G>{
+        match spec {
+            SRSSpec::Use(x) => SRSValue::Ref(x),
+            SRSSpec::Generate => SRSValue::Value(Self::generate(size))
+        }
+    }
+}
+
+pub struct Index<'a, G: CommitmentCurve>
```
imeckler, line 50: Nice, happy we could reuse our infrastructure!

dlog/plonk/src/index.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl<'a, G: CommitmentCurve> Index<'a, G> where G::BaseField: PrimeField
+{
+    pub fn verifier_index(&self) -> Result<VerifierIndex<G>, ProofError> {
+        Err(ProofError::ProofCreation)
+    }
+
+    // this function compiles the circuit from constraints
+    pub fn create
```
imeckler, line 89: Todo

dlog/plonk/src/plonk_sponge.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    fn challenge(&mut self) -> ScalarChallenge<Fr> {
+        ScalarChallenge(self.squeeze(oracle::sponge::CHALLENGE_LENGTH_IN_LIMBS))
+    }
+
+    fn absorb_evaluations(&mut self, x_hat: &[Fr], e: &ProofEvaluations<Fr>) {
+        self.last_squeezed = vec![];
+        self.sponge.absorb(&self.params, x_hat);
+
+        let points = [
```
imeckler, line 37: Guessing all the dlog stuff is TODO? It would be good to move it into another PR so we can review and then merge this one.

pairing/plonk/src/prover.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        let a = &Evaluations::<E::Fr>::from_vec_and_domain(index.cs.gates.iter().map(|gate| witness[gate.l.0]).collect(), index.cs.domain).interpolate()
+            + &DensePolynomial::rand(1, &mut OsRng).mul_by_vanishing_poly(index.cs.domain);
+        let b = &Evaluations::<E::Fr>::from_vec_and_domain(index.cs.gates.iter().map(|gate| witness[gate.r.0]).collect(), index.cs.domain).interpolate()
+            + &DensePolynomial::rand(1, &mut OsRng).mul_by_vanishing_poly(index.cs.domain);
+        let c = &Evaluations::<E::Fr>::from_vec_and_domain(index.cs.gates.iter().map(|gate| witness[gate.o.0]).collect(), index.cs.domain).interpolate()
+            + &DensePolynomial::rand(1, &mut OsRng).mul_by_vanishing_poly(index.cs.domain);
+
+        // commit to the a, b, c wire values
+        let a_comm = index.urs.get_ref().commit(&a)?;
```
imeckler, line 75: Maybe we can call them either `a, b, c` or `l, r, o` everywhere for consistency?

pairing/plonk/src/prover.rs
```diff
@@ -0,0 +1,293 @@
...
+        (
+            |j| z[j] *=
+                &((witness[j] + &(index.cs.sid[j] * &oracles.beta) + &oracles.gamma) *&
+                (witness[j+n] + &(index.cs.sid[j] * &oracles.beta * &index.cs.r) + &oracles.gamma) *&
+                (witness[j+2*n] + &(index.cs.sid[j] * &oracles.beta * &index.cs.o) + &oracles.gamma))
+        );
+
+        (1..n).for_each(|i| {let x = z[i-1]; z[i] *= &x});
+        if z.pop().unwrap() != E::Fr::one() {return Err(ProofError::ProofCreation)};
+        z.insert(0, E::Fr::one());
```
imeckler, line 109: Not a big deal (maybe it's cleaner this way?) but you could reindex things by 1 so you don't have to insert at the front of the vector (I think that's what's happening here?) which is O(n)

pairing/plonk/src/prover.rs
```diff
@@ -0,0 +1,293 @@
...
+        let z_comm = index.urs.get_ref().commit(&z)?;
+
+        // absorb the z commitment into the argument and query alpha
+        fq_sponge.absorb_g(&[z_comm]);
+        oracles.alpha = fq_sponge.challenge();
+        let alpsq = oracles.alpha.square();
+
+        // compute quotient polynomial
+
+        let t1 =
```
imeckler, line 123: Is there some way to make this more readable? I think some of these borrows are not necessary (I believe only the RHS of all the operations needs to be a borrow).

pairing/plonk/src/prover.rs
```diff
@@ -0,0 +1,293 @@
...
+            (
+                (evals.a + &(oracles.beta * &evals.sigma1) + &oracles.gamma) *
+                &(evals.b + &(oracles.beta * &evals.sigma2) + &oracles.gamma) *
+                &(oracles.beta * &evals.z * &oracles.alpha)
+            );
+        let r4 = z.scale(alpsq * &index.cs.l0.evaluate(oracles.zeta));
+        let r = &(&(&r1 + &r2) - &r3) + &r4;
+        evals.r = r.evaluate(oracles.zeta);
+
+        // query opening scaler challenge
```
imeckler, line 209: Should be `scalar`

circuits/plonk/src/gate.rs
ltvvlz: The second part of the tuple, defines the index of the wire permutation. Permutation sets (or equivalence classes) have to be defined in some way. Defining a permutation as decomposition of the set of wires into the disjoint permutation sets is equivalent to defining a bijective map of a set of wires onto itself. Thus, defining where each wire goes under the permutation. The left part of the tuple is the wire index, the right part the index of the "permuted wire" index. This mapping is then used, upon the creation of the ConstraintSystem struct, to build the permutation polynomials. Another  way would be to define the permutation polys directly but it is possibly more elegant to define everything, including the permutation map, in the set of the gates that completely determine the constraint system. Will enhance the comment,

dlog/plonk/src/index.rs
ltvvlz: Good to hear.

dlog/plonk/src/index.rs
ltvvlz: Current work for the next PR

dlog/plonk/src/plonk_sponge.rs
ltvvlz: Agreed. Will create another PR over the same branch for dlog-based Plonk implementation..

pairing/plonk/src/prover.rs
ltvvlz: Agreed, will do calling them all l, r, o

pairing/plonk/src/prover.rs
ltvvlz: I will think of something.

pairing/plonk/src/prover.rs
ltvvlz: I tried few ways of making it less horrible. Unfortunately, + operator's add function takes both left as well as right sides as references. And, although we can remove few references from these expressions, the primary inconvenience is with expressions consisting of many arith operations which add up the references. Therefore, not referencing some of the operands will not make it much better. I may have misunderstood your suggestion though. If so, please clarify.

pairing/plonk/src/prover.rs
ltvvlz: Scaler deriving from "to scale" with DensePolynomial::scale function.

PR #44, opened 2020-06-03
Plonk implementation with updates to Algebra

ltvvlz:
This commits the Plonk implementation for review.

This PR is for the review of Plonk implementation that has the following:

1.
     Plonk with pairing-based commitment schema supporting polynomial splitting.
2.
     Plonk with dlog-based commitment schema supporting polynomial splitting. This implementation, in order to use polysplitting together with the linearization optimization, adds additional polyevaluation (for the quotient polynomial) to the proof. Alternative implementation, not using the linearization optimization, has been committed (with commit b51ffa113234db7324c6576f46b645f2ed32d353) into "plonk" branch. The alternative implementation has proof size larger than the proof size of this implementation.
3.
     Unit tests for both Plonk implementations.


jon-chuang:
Hi, is this standard plonk or turboPlonk? 

Implementing turboplonk in it's most general form requires ripping out the foundations of the circuit builder, I think.

circuits/plonk/src/constraints.rs
```diff
@@ -0,0 +1,173 @@
+/*****************************************************************************************************************
```
kobigurk, line 1: What about using //! comments  so the docs will be nicer?

circuits/plonk/src/constraints.rs
```diff
@@ -0,0 +1,173 @@
...
+        mut gates: Vec<CircuitGate<F>>,
+        public: usize,
+    ) -> Option<Self>
+    {
+        let domain = Domain::<F>::new(Domain::<F>::compute_size_of_domain(gates.len())?)?;
+        let mut sid = domain.elements().map(|elm| {elm}).collect::<Vec<_>>();
+        let r =
+        {
+            let mut r = domain.sample_element_outside_domain(&mut OsRng);
+            while r.legendre().is_qnr() == false {r = domain.sample_element_outside_domain(&mut OsRng)}
```
kobigurk, line 56: It might be better to pass down an implementer of an Rng trait rather than using OsRng directly.

dlog/plonk/src/verifier.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::plonk_sponge::{FrSponge};
+use commitment_dlog::commitment::{CommitmentCurve, Utils, PolyComm};
+use rand_core::OsRng;
+
+type Fr<G> = <G as AffineCurve>::ScalarField;
+type Fq<G> = <G as AffineCurve>::BaseField;
+
+impl<G: CommitmentCurve> ProverProof<G>
+{
+    // This function verifies the batch of zk-proofs
```
kobigurk, line 22: /// style comments look nicer in docs

dlog/tests/marlin_group_addition.rs
```diff
@@ -38,7 +38,7 @@ type Fr = <Affine as AffineCurve>::ScalarField;
 const MAX_SIZE: usize = 8;
 
 #[test]
-fn group_addition_dlog()
+fn dlog_marlin_group_addition()
 {
     let rng = &mut OsRng;
```
kobigurk, line 43: Same comment as rng before

ltvvlz: Many thanks for the suggestions..!

PR #47, opened 2020-07-17
Review for turbo plonk

imeckler:
Just making this so I can leave some comments


dlog/plonk/src/index.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+pub struct VerifierIndex<'a, G: CommitmentCurve>
+{
+    pub domain: D<Fr<G>>,          // evaluation domain
+    pub max_poly_size: usize,           // maximal size of polynomial section
+    pub srs: SRSValue<'a, G>,           // polynomial commitment keys
+
+    // index polynomial commitments
+    pub sigma_comm: [PolyComm<G>; 3],   // permutation commitment array
```
imeckler, line 73: Since we'll have accounts with verification indexes in them, we have to optimize the size a little bit. 

`sigma_comm` and `sid_comm` depend only on the `domain` right? At least on the OCaml side we will want to factor this out of the verifier index.

dlog/plonk/src/index.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub sid_comm:   PolyComm<G>,        // SID commitment
+    pub ql_comm:    PolyComm<G>,        // left input wire commitment
+    pub qr_comm:    PolyComm<G>,        // right input wire commitment
+    pub qo_comm:    PolyComm<G>,        // output wire commitment
+    pub qm_comm:    PolyComm<G>,        // multiplication commitment
+    pub qc_comm:    PolyComm<G>,        // constant wire commitment
+
+    // poseidon polynomial commitments
+    pub rcm_comm:   [PolyComm<G>; 3],   // round constant polynomia commitment array
+    pub fpm_comm:   PolyComm<G>,        // full/partial round indicator polynomial commitment
```
imeckler, line 83: Do we need the full/partial indicators? We are going to have 0 partial rounds. This is again just in the interest of optimizing verifier index size.

dlog/plonk/src/index.rs
```diff
@@ -0,0 +1,166 @@
...
+    // ECC arithmetic polynomial commitments
+    pub add_comm:   PolyComm<G>,        // EC addition selector polynomial commitment
+    pub mul1_comm:  PolyComm<G>,        // EC variable base scalar multiplication selector polynomial commitment
+    pub mul2_comm:  PolyComm<G>,        // EC variable base scalar multiplication selector polynomial commitment
+    pub emul1_comm: PolyComm<G>,        // endoscalar multiplication selector polynomial commitment
+    pub emul2_comm: PolyComm<G>,        // endoscalar multiplication selector polynomial commitment
+    pub emul3_comm: PolyComm<G>,        // endoscalar multiplication selector polynomial commitment
+
+    pub r:          Fr<G>,              // coordinate shift for right wires
+    pub o:          Fr<G>,              // coordinate shift for output wires
```
imeckler, line 96: Does this depends only on the domain? We will want to factor it out on the OCaml side if so.

dlog/plonk/src/verifier.rs
```diff
@@ -0,0 +1,214 @@
...
+type Fr<G> = <G as AffineCurve>::ScalarField;
+type Fq<G> = <G as AffineCurve>::BaseField;
+
+impl<G: CommitmentCurve> ProverProof<G> where G::ScalarField : QnrField
+{
+    // This function verifies the batch of zk-proofs
+    //     proofs: vector of Plonk proofs
+    //     index: Index
+    //     RETURN: verification status
+    pub fn verify
```
imeckler, line 26: We'll have to do the same thing we're doing in marlin with the commitment to the public input polynomial right?

dlog/plonk/src/index.rs
ltvvlz: > Since we'll have accounts with verification indexes in them, we have to optimize the size a little bit.
> 
> `sigma_comm` and `sid_comm` depend only on the `domain` right? At least on the OCaml side we will want to factor this out of the verifier index.

sigma_comm depends on
1. domain
2. copy permutations
3. srs

sid_comm depends on
1. domain
3. srs

dlog/plonk/src/index.rs
ltvvlz: This has been committed in f14103f2eb3d0e7e4c531f688185ebf46bcbe214

dlog/plonk/src/index.rs
ltvvlz: Yes, these values depend on the domain in the sense that they are sampled outside of it.

dlog/plonk/src/verifier.rs
ltvvlz: Makes sense. Will do in one of the next commits.

dlog/plonk/src/index.rs
ltvvlz: sid_comm is not needed by the verifier !

removed it from verifier index in 21763b0f5986edc65822ff9dc366d4ad0c7cec87

Thanks for finding.

dlog/plonk/src/verifier.rs
ltvvlz: This has been completed with commit 3f8916cde2e52bfc15680586a8a153db1c4529e4

oracle/src/sponge.rs
```diff
@@ -149,55 +149,57 @@ where
         self.last_squeezed = vec![];
         for g in g.iter()
         {
-            if g.infinity {
-                panic!("marlin sponge got zero curve point");
-            } else {
-                self.sponge.absorb(&self.params, &[g.x]);
-                self.sponge.absorb(&self.params, &[g.y]);
-            }
+            self.sponge.absorb(&self.params, &[g.x]);
```
imeckler, line 152: Shouldn't we still panic here? The x and y coordinates are not valid for points which are infinity.

imeckler: If we cannot panic, we would like to have some kind of sentinel affine value for points at infinity, like (0, 0). Then we can replicate that behavior in the circuit.

ltvvlz: These comments have been addressed in commit 861b022a66a331a41407b79882295ca86b18a464 by restoring the original logic of Rust runtime panic in case of the absorption of the Point at Infinity.

PR #55, opened 2020-09-08
batch verification across different vks

imeckler:
This PR changes the marlin verify function to batch verification across proofs with different verification indexes. It requires that the indexes all have the same CRS length. This limitation shouldn't be necessary but I couldn't get it to work for now.

It also reorganizes the CRS so that if `n < m` then `create(n, 0, 0).g` is a prefix of `create(m, 0, 0).g` and `create(n, 0, 0).h == create(m, 0, 0).h`. This property is useful for batching verification across indexes of varying CRS length.


ltvvlz (approved):
1. Relieved the code changes for the verification functionality
2. Corrected the test marlin_reduction.rs (attached below)
3. Approved

[marlin_reduction.zip](https://github.com/o1-labs/marlin/files/5189260/marlin_reduction.zip)

PR #61, opened 2020-10-16
Bazel initial support:  Marlin

mobileink:
Tested on MacOS and Debian 9.

This and Zexe are mutually dependent; neither will build until both are merged.  PR for Zexe is [37](https://github.com/o1-labs/zexe/pull/37)


bkase (changes requested):
I'm worried about the duplicate Cargo.toml files inside the `cargo/` folder and the ones in the repo in the different directories. Is it possible to either unify these or make one derivable from the other mechanically? If we don't, they will certainly get out of sync and bitrot.

mobileink:
Point taken.  It's surely possible to derive cargo/Cargo.toml from the other cargo files, but I looked around online and don't see any tools for that, and writing one doesn't look entirely trivial to me.  But I'm not sure this is a major problem - the likelihood that cargo/Cargo.toml will get out of sync is probably no higher than the probability that Marlin and Zexe will get out of sync, no?  Could we write a githook that prevents divergence?

mobileink:
Fixed an oversight: put cargo in top-level bzl dir, to segregate pure bazel stuff and minimize top-level clutter.

bkase (approved):
@mobileink can follow up with a fix in a later PR

mobileink:
Obsoleted by #71

PR #62, opened 2020-10-19
Simplified inner product argument

imeckler:
This PR modifies the inner product argument so that the log-sized verification multiexponentiation involves no square challenges. This is advantageous inside the circuit because it allows us to avoid computing the expensive deterministic square root. The scheme is described in [A2 here](https://eprint.iacr.org/2020/499.pdf).

Our implementation differs from the paper in two ways:

1. We use the hiding technique employed by the Halo authors (since we already have that implemented) rather than the hiding technique proposed there. 
2. Instead of randomizing H by scaling it by a challenge as in this paper, we continue to use an H which is the hash-to-curve of a challenge. This is mostly because we already have that implemented, but it's also more efficient inside the circuit.


ltvvlz (approved):
reviewed the changes. some of the minor ones mat have been duplicated in pr-63

PR #63, opened 2020-10-28
latest changes

ltvvlz:
this implements the following:

1. zk for polys before interpolation
2. fix some broken tests
3. correct commitment schema
4. implement the use the correct endoscalar
5. correct scaling of custom constraint checks with the correct powers of alpha

### DIFF
--- a/circuits/plonk/src/constraints.rs
+++ b/circuits/plonk/src/constraints.rs
@@ -66,15 +66,15 @@ pub struct ConstraintSystem<F: FftField>
     pub ps8:    Evaluations<F, D<F>>,       // poseidon selector over domain.d8
 
     // ECC arithmetic selector polynomials
-    pub addl3:  Evaluations<F, D<F>>,       // EC point addition selector evaluations w over domain.d4
-    pub addl4:  Evaluations<F, D<F>>,       // EC point addition selector evaluations w over domain.d8
+    pub addl4:  Evaluations<F, D<F>>,       // EC point addition selector evaluations w over domain.d4
     pub mul1l:  Evaluations<F, D<F>>,       // scalar multiplication selector evaluations over domain.d4
     pub mul2l:  Evaluations<F, D<F>>,       // scalar multiplication selector evaluations over domain.d8
     pub emul1l: Evaluations<F, D<F>>,       // endoscalar multiplication selector evaluations over domain.d4
     pub emul2l: Evaluations<F, D<F>>,       // endoscalar multiplication selector evaluations over domain.d4
     pub emul3l: Evaluations<F, D<F>>,       // endoscalar multiplication selector evaluations over domain.d8
 
-    pub l0:     Evaluations<F, D<F>>,       // 0-th Lagrange evaluated over domain.d8
+    pub l04:    Evaluations<F, D<F>>,       // 0-th Lagrange evaluated over domain.d4
+    pub l08:    Evaluations<F, D<F>>,       // 0-th Lagrange evaluated over domain.d8
     pub l1:     Evaluations<F, D<F>>,       // 1-st Lagrange evaluated over domain.d8
     pub zkpl:   Evaluations<F, D<F>>,       // zero-knowledge polynomial over domain.d8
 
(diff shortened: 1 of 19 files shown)


dlog/commitment/src/commitment.rs
```diff
@@ -352,20 +353,30 @@ impl<G: CommitmentCurve> SRS<G> where G::ScalarField : CommitmentField {
...
-                while offset < p_i.coeffs.len() {
-                    let segment = DensePolynomial::<Fr<G>>::from_coefficients_slice
-                        (&p_i.coeffs[offset..if offset+self.g.len() > p_i.coeffs.len() {p_i.coeffs.len()} else {offset+self.g.len()}]);
-                    // always mixing in the unshifted segments
-                    p += &segment.scale(scale);
-                    scale *= &polyscale;
-                    offset += self.g.len();
-                    if offset > p_i.coeffs.len() {
-                        if let Some(m) = degree_bound {
+                if let Some(m) = degree_bound {
```
imeckler, line 356: What was the mistake here? Hard to parse out why this is changed in this way

circuits/plonk/src/polynomials/varbasemul.rs
```diff
@@ -38,49 +36,49 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F>
...
+    pub fn vbmul_scalars(evals: &Vec<ProofEvaluations<F>>, alpha: &[F]) -> Vec<F>
     {
         // 2*xP - λ1^2 + xT
         let tmp = evals[0].l.double() - &evals[0].r.square() + &evals[1].r;
 
         vec!
         [
             // verify booleanity of the scalar bit
-            (evals[0].r.square() - &evals[0].r) * &alpha[1]
+            (evals[0].r.square() - &evals[0].r) * &alpha[0]
```
imeckler, line 61: Why the change with the alphas? Were we using higher powers of alpha than necessary?

dlog/plonk/src/range.rs
```diff
@@ -0,0 +1,7 @@
+use std::ops::Range;
+
+pub const PERM:     Range<usize> = 0..2;
```
imeckler, line 3: Why do you have them using non-overlapping ranges? Isn't it ok for them to use the same alphas?

circuits/plonk/src/constraints.rs
```diff
@@ -65,16 +66,17 @@ pub struct ConstraintSystem<F: FftField>
...
-    pub addl4:  Evaluations<F, D<F>>,       // EC point addition selector evaluations w over domain.d8
+    pub addl4:  Evaluations<F, D<F>>,       // EC point addition selector evaluations w over domain.d4
     pub mul1l:  Evaluations<F, D<F>>,       // scalar multiplication selector evaluations over domain.d4
     pub mul2l:  Evaluations<F, D<F>>,       // scalar multiplication selector evaluations over domain.d8
     pub emul1l: Evaluations<F, D<F>>,       // endoscalar multiplication selector evaluations over domain.d4
     pub emul2l: Evaluations<F, D<F>>,       // endoscalar multiplication selector evaluations over domain.d4
     pub emul3l: Evaluations<F, D<F>>,       // endoscalar multiplication selector evaluations over domain.d8
 
-    pub l0:     Evaluations<F, D<F>>,       // 0-th Lagrange evaluated over domain.d8
+    pub l04:    Evaluations<F, D<F>>,       // 0-th Lagrange evaluated over domain.d4
```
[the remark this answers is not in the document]
ltvvlz: no, it wouldn't help memory if we avoided holding onto this separately. It would make the use of memory/performance worse with allocating it on stack every rime used.

dlog/commitment/src/commitment.rs
ltvvlz: this change has been pr'd few weeks ago to go to the master. it is the same change.

circuits/plonk/src/polynomials/varbasemul.rs
ltvvlz: the reason is same as for why we check different equations scaling them with different powers of random value

dlog/plonk/src/range.rs
ltvvlz: see the previous comment

circuits/plonk/src/constraints.rs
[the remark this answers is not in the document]
imeckler: I was suggesting we just use the subset of l08 and not reallocate a new vector. It would admittedly be awkward because we would have to use special operations rather than the ones implemented for the `Evaluations` struct.

circuits/plonk/src/polynomials/varbasemul.rs
imeckler: For the record I do not think this is necessary given that the different types of equations are applied on disjoint subsets of the rows. But since it barely affects performance it seems fine to me.

imeckler:
Merging but with the note that this does not (as far as I can tell) randomize the column polynomials. For now I will do this on the OCaml side (by just adding some random rows) because I had trouble getting the tests to pass with doing it on the rust side.

imeckler (approved):

PR #83, opened 2021-01-27
Feature/five wires

mrmr1993:
This contains the substantial changes from @ltvvlz's #79. The differences from there are:
* the changes are split into #81, #82 (the parent PR of this) and this PR
* the 5-wires implementation exists alongside the original one, instead of replacing it
  - the modified poseidon oracle is tweaked to share more code with the existing one, to avoid duplicating more code in its dependents
* some accidental changes (e.g. removing rayon parallelism in the oracle library) have been reverted.

This compiles and all tests pass locally. Also tested a build of mina and the transaction snark tests there, which were both successful.

PR #82, opened 2021-01-27
Add tweedle fp5 and fq5 parameters for oracle

mrmr1993:
This breaks part of df6b372 from @ltvvlz's #79.

PR #84, opened 2021-02-03
Plonk with Plookup

ltvvlz:
This PR adds the implementation of the Plookup protocol (https://eprint.iacr.org/2020/315) into Plonk so as to optimize the proofing system for the computations that involve various lookup functions. This is valuable for constraining the computations of functions/operations that are, otherwise, expensive to implement in a circuit over a prime field.

The proposed implementation adds support for a single lookup table to be looked up for a witness vector which is the 4-th column (for columns 0..=4) vector of Plonk witness. The feature is implemented with the following two mechanisms:

1. Addition of LOOKUP custom Plonk constraint. This constrains the column 0..=3 values to that of the column 4 as encoding of the constrained operation operand and result values, along with the supported operation opcodes, into the lookup value to be checked against the constraint system lookup table. For example, LOOKUP constraint for the 8-bit XOR operation constrains the values of the XOR opcode, two input operands and one output into the lookup value with the bitwise shifts and addition. Similarly for AND, OR (with different opcode values). For NOR, the gate constrains the opcode, input and output operands to the lookup value. Other types of lookup functions can also be constrained with this gate type to encode inputs and outputs into the lookup values.

2. Addition of Plookup index/prover/verifier functionality into Plonk protocol. This addition is implemented as per the protocol specification of https://eprint.iacr.org/2020/315. It is important to note here that Plookup protocol, as integrated into Plonk, bears some burden on the Plonk performance as prover/verifier functionality has to deal with the additional polynomials and their evaluations/commitments. Nevertheless, for a significant class of computations which depend on SNARK-inconvenient operations, the resulting proofing system provides the performance benefit of the smaller circuit, constraining the computation.

The implementation of Plookup in Plonk has been augmented with the test of this functionality. The turbo_plonk test has been enhanced with the lookup constraints for the 8-bit binary operations such as XOR, AND, OR, NOR.

This implementation is the step in the design/implementation of the custom Plonk constraints for the computation that involves encryption of HTTPS data with TLSNotary that computes irrefutable evidence of the session.


mrmr1993:
This seems to include the changes from #75 (which are now split into #81, #82, #83 on top of current master), and modifies the 3-wires constraint system.

@ltvvlz can you change this to create a new constraint system for 5-wires-plonk+plookup instead of replacing the 3-wires one, and update it to merge cleanly to master? It may make sense to base it upon #82, which introduces the 5-wires poseidon parameters, or perhaps even #83 if it makes sense to share some gates etc. with that change instead of duplicating them.

imeckler:
Plus 1 to @mrmr1993's comments. It will make review (and merging) a lot more straightforward. Looking forward to reading this!

PR #87, opened 2021-02-04
Plonk with Plookup

ltvvlz:
This PR is derived from PR-83 in order to address the comments of PR-84.

### DIFF
--- a/circuits/plonk-plookup/src/constraints.rs
+++ b/circuits/plonk-plookup/src/constraints.rs
@@ -76,8 +76,8 @@ pub struct ConstraintSystem<F: FftField>
     pub emull:  E<F, D<F>>,             // endoscalar multiplication selector evaluations over domain.d4
 
     // lookup table
-    pub table4w:E<F, D<F>>,             // shifted lookup table polynomial over domain.d4
-    pub table4: E<F, D<F>>,             // lookup table polynomial over domain.d4
+    pub table8w:E<F, D<F>>,             // shifted lookup table polynomial over domain.d8
+    pub table8: E<F, D<F>>,             // lookup table polynomial over domain.d8
     pub table1: E<F, D<F>>,             // lookup table polynomial over domain.d1
 
     // constant polynomials
@@ -179,7 +179,7 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F>
         tbl.resize(n, F::zero());
         let table1 = E::<F, D<F>>::from_vec_and_domain(tbl, domain.d1); 
         let tablem = table1.clone().interpolate();
-        let table4 = tablem.evaluate_over_domain_by_ref(domain.d4);
+        let table8 = tablem.evaluate_over_domain_by_ref(domain.d8);
 
         // constant polynomials
         let l18 = DP::from_coefficients_slice(&[F::zero(), F::one()]).evaluate_over_domain_by_ref(domain.d8);
@@ -230,8 +230,8 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F>
 
             // lookup table polynonial
             table1,
-            table4w: table4.shift(4),
-            table4,
+            table8w: table8.shift(8),
+            table8,
             tablem,
 
             // constant polynomials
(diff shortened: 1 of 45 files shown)


circuits/plonk-plookup/src/constraints.rs
```diff
@@ -0,0 +1,386 @@
...
+use array_init::array_init;
+
+#[derive(Clone)]
+pub struct ConstraintSystem<F: FftField>
+{
+    pub public: usize,                  // number of public inputs
+    pub domain: EvaluationDomains<F>,   // evaluation domains
+    pub gates:  Vec<CircuitGate<F>>,    // circuit gates
+
+    // POLYNOMIALS OVER THE MONOMIAL BASE
```
imeckler, line 25: nit: should be 'monomial basis'

circuits/plonk-plookup/src/constraints.rs
```diff
@@ -0,0 +1,386 @@
...
+        while r.legendre().is_qnr() == false || domain.evaluate_vanishing_polynomial(r).is_zero()
+        {
+            let mut h = Blake2b::new();
+            h.input(&{*i += 1; *i}.to_be_bytes());
+            r = F::from_random_bytes(&h.result()[..31]).unwrap();
+        }
+        r
+    }
+
+    pub fn sample_shifts(domain: &D<F>, len: usize) -> Vec<F>
```
imeckler, line 295: Can we share this code with the non-plookup implementation? Maybe pull out into a utils library or something?

circuits/plonk-plookup/src/domains.rs
```diff
@@ -0,0 +1,24 @@
+use algebra::FftField;
+use ff_fft::{EvaluationDomain, Radix2EvaluationDomain as D};
+
+#[derive(Debug, Clone, Copy)]
```
imeckler, line 4: Can we reuse this file with PLONK?

circuits/plonk-plookup/src/gate.rs
```diff
@@ -0,0 +1,123 @@
...
+    Double,     // EC point doubling in Affine form
+    Vbmul1,     // EC variable base scalar multiplication 
+    Vbmul2,     // unpacking EC variable base scalar multiplication 
+    Endomul,    // EC variable base scalar multiplication with group endomorphim optimization
+    Pack,       // packing
+    Lookup,     // lookup
+}
+
+#[derive(Clone)]
+pub struct CircuitGate<F: FftField>
```
imeckler, line 32: Let's reuse this and the ToBytes/FromBytes impls with non-plookup plonk. I think a good way to do it would be to parametrize `CircuitGate` over `<Gt : GateType>` where `GateType` is a trait

```
trait GateType {
     fn verify(&self, witness: &[Vec<F>; COLUMNS], cs: &ConstraintSystem<F>) -> bool 
}
```

and then in the implementation for this trait, you just match on the `self` parameter and do the appropriate thing.

Then, for plookup plonk, we can have the GateType be an enum which is either a GateType value from regular plonk, or a lookup gate. That we, we can also easily reuse all the verification code with PLONK.

circuits/plonk-plookup/src/gates/addition.rs
```diff
@@ -0,0 +1,79 @@
+/*****************************************************************************************************************
+
+This source file implements non-special point (with distinct abscissas) Weierstrass curve addition
+
+    ADD gate constrains
```
imeckler, line 5: nit: constrains -> constraints, here and everywhere else it appears

imeckler (changes requested):

imeckler:
Left some comments -- will resume reviewing tomorrow!

circuits/plonk-plookup/src/gates/packing.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    pub fn verify_pack(&self, witness: &[Vec<F>; COLUMNS]) -> bool
+    {
+        let this: [F; COLUMNS] = array_init(|i| witness[i][self.row]);
+        let next: [F; COLUMNS] = array_init(|i| witness[i][self.row+1]);
+
+        self.typ == GateType::Pack
+        &&
+        next[4] ==
```
imeckler, line 36: I guess it doesn't really make a difference, but is there a reason it's more convenient to split the bits across the 2 rows in this way?

circuits/plonk-plookup/src/polynomials/addition.rs
```diff
@@ -0,0 +1,80 @@
...
+*****************************************************************************************************************/
+
+use algebra::{FftField, SquareRootField};
+use ff_fft::{Evaluations, DensePolynomial, Radix2EvaluationDomain as D};
+use crate::polynomial::WitnessOverDomains;
+use oracle::utils::{EvalUtils, PolyUtils};
+use crate::constraints::ConstraintSystem;
+use crate::scalars::ProofEvaluations;
+
+impl<F: FftField + SquareRootField> ConstraintSystem<F>
```
imeckler, line 40: Again wondering if all this code can be shared with the PLONK implementation

PR #88, opened 2021-03-02
code corrections for updated Rust toolchain

ltvvlz:
After updating Rust toolchain with "rustup update", building this repo results in compilation errors. This PR updates the source code to correct compilation.


mrmr1993:
This appears to be due to a breaking change in the rust typechecker rust-lang/rust#79904. I don't think we should take this change until we're set up to migrate our infrastructure to rust >= 1.49.

For now, you should downgrade to a version less than 1.49 (our infra uses 1.45.2) with e.g. `rustup toolchain install 1.45.2 && rustup default 1.45.2`).

PR #102, opened 2021-05-11
Refactor all poseidon instances into single implementation and add new poseidon "3" hash function

jspada:
* Params: rounds=54, width=3, rate=2, alpha=7
    * Added unit tests


oracle/src/poseidon_3.rs
```diff
@@ -0,0 +1,174 @@
...
+    const SPONGE_RATE: usize = 2;
+    const SPONGE_BOX: usize = 7;
+    const FULL_MDS: bool = true;
+}
+
+pub fn sbox<F : Field, SC: SpongeConstants>(x: F) -> F {
+    x.pow([SC::SPONGE_BOX as u64])
+}
+
+#[derive(Clone)]
```
imeckler, line 29: Can you share the code with the other poseidon implementations? Shouldn't it just be a matter of specifying different constants rather than repeating the implementation?

imeckler (changes requested):

oracle/src/poseidon_3.rs
jspada: Sure, no problem.  (I only cloned it this way because I saw that's how it was done between poseidon 3-wire and poseidon 5-wire.)

Just to confirm, you want it refactored a bit so they can all share the same implementation, but just use different parameters and different permutation functions?

oracle/tests/poseidon_tests.rs
```diff
@@ -169,15 +168,15 @@ mod tests {
...
+            Fp::from_repr(BigInteger256([12625032309730357895, 3881775963142723428, 1948451027071626224, 400220142328418896]))
         );
 
         // _rand_fields(1);
         assert_poseidon_3_eq!(
             &[
                 Fp::from_repr(BigInteger256([7268460211608788188, 10132480989041334579, 2339874299280274918, 194293202993774285]))
             ],
-            Fp::from_repr(BigInteger256([2558090059455896147, 7307347974437946425, 4536174362957322879, 2630632011732286938]))
+            Fp::from_repr(BigInteger256([13319422177750956895, 2619256142820001370, 15974443115283230879, 4444502174303366803]))
```
imeckler, line 179: Why did these values change out of curiosity? Shouldn't the values for 3 wires PLONK be the same?

jspada: In my previous commit I had accidentally defined `Parameters3` from `oracle::pasta::fp5 ` params! 
 
Near the top of this file...
```
-    use oracle::pasta::fp5 as Parameters3;
+    use oracle::poseidon::PlonkSpongeConstants3;
+    use oracle::pasta::fp_3 as Parameters3;
```

imeckler: ah, got it!

imeckler (approved):

PR #103, opened 2021-05-13
[oracle] adding test vector generator

mimoo:
This PR combines work from @jspada and I.
It allows us to create exportable (in JSON format) test vectors for the three-wire and five-wire version of Poseidon.

You can generate the test vector files yourself to check that they match the ones in `test_vectors/`:

```
cargo run --bin export_test_vectors --no-default-features --features three_wire -- three_wire.json
cargo run --bin export_test_vectors --no-default-features --features five_wire -- five_wire.json
```

This also simplifies our test to just make use the test vector files.


jspada:
Nice!

I prefer the idea of JSON + bytes representation, but the only problem is that it breaks from compatibility with other implementations.

Specifically, the c reference signer can't support json and doesn't have a bytes interface and in the OCaml implementation we use base 10.

The reason we can't use JSON in the c implementation is because (as a reference implementation) it's required to have no dependencies on external libraries (no native json in C).

Some ideas on way forward
1) Adding an output format option `[c, rust, ocaml]` that allows it to be output in the desired format (c-struct, bytes, base-10)
2) Adding a bytes interface to both the c reference signer and OCaml and also adding an output format option `[c,json]` that outputs either c-struct or json (both using bytes representation for field elements).

What do you think?  I prefer (2) and sort-a feel like there should always be a bytes interface for these sort of things.

jspada:
Here's a format that could work for the c-implementation after adding the supporting code there.

```c
typedef struct poseidon_test {
  int   input_len;
  char *input[10];
  char *output;
} PoseidonTest;

PoseidonTest tests[] = {
  {
    0,
    {
    },
    "01ca17b5913ca7a42adfd9353676cd9a939effb71e3246fd78a2d9283c664529"
  },
  {
    1,
    {
      "df698e389c6f1987ffe186d806f8163738f5bf22e8be02572cce99dc6a4ab030"
    },
    "e7d279a184604c5ae038be9d7e62a23b75373825614f3b8175db6378e4fbc622"
  },
  {
    2,
    {
      "df698e389c6f1987ffe186d806f8163738f5bf22e8be02572cce99dc6a4ab030",
      "f26a8a03d9c9bbd9c6b2a1324d2a3f4d894bafe25a7e4ad1a498705f4026ff2f"
    },
    "e7d279a184604c5ae038be9d7e62a23b75373825614f3b8175db6378e4fbc622"
  }
}
```

mimoo:
I used json + hex because this is usually the more portable / used format for test vectors. I don't think it's a problem to generate it in different formats but I'm not sure why hex/json are bad ideas.

> The reason we can't use JSON in the c implementation is because (as a reference implementation) it's required to have no dependencies on external libraries (no native json in C).

Does it matter what the test dependencies are? It looks like you can simply drop this file in your codebase to parse JSON in C: https://github.com/zserge/jsmn/blob/master/jsmn.h

jspada:
It's definitely not a bad idea.  On the contrary, it's a good idea. 👍🏼 Definitely more portable.

All I'm saying is that we need to update the other implementations so that they can use it.

In the C implementation we we just want to avoid linking to anything, but if there's some small code (with the appropriate license) we can drop in, then that would be fine too.

In any case, we need to update the other bits of code (not just the rust) to use the JSON + bytes/hex version.

jspada:
So it sounds like we are aiming for

3) Adding a bytes/hex + JSON interface to both the c reference signer and OCaml.

?

mimoo:
oh yeah, that makes sense. I'm looking into adding these to the OCaml implementation at the moment!

jspada:
Sounds good - I'll have a look at the C side.

mimoo:
Test vector-wise, I was thinking if we wanted to generate more test vectors, and potentially bad test vectors. For example, bytearrays that are larger than a field element (0xfffff...)

jspada:
Another idea.  If we had this JSON+hex interface working on c reference signer and OCaml it could also be good for specifying parameters.  Historically those are also in 3 different formats.

jspada:
Looking at the C part, I'm not sure I have the bandwidth for this change right now, with the poseidon + tls-notary prover work as high priority. 

Perhaps we should plan to do this after the tls-notary stuff is done?  That will be a better time.

We could then start with the test vectors, then parameters, and parameter generator (sage).

It would be a great improvement to have everything in a portable format, represented the same in each implementation.

mimoo:
btw any blockers for this PR @jspada ?

jspada:
> btw any blockers for this PR @jspada ?

Yes - updating the other implementations to support JSON.

jspada:
Added base-10 mode `B10` that prints out the field elements in base-10.  It's still JSON so you'll need to make some manual edits, but at least they're the format the OCaml code expects.  Also added a stdout mode (just use `-` for the output file name).
```
Example runs
cargo run --bin export_test_vectors ... -- B10 -
or
cargo run --bin export_test_vectors ... -- Hex -
```
Although not ideal, hopefully this is useful.

oracle/Cargo.toml
```diff
@@ -12,6 +12,9 @@ ff-fft = { path = "../zexe/ff-fft", features = [ "parallel"] }
 ocaml = { version = "0.18.1", optional = true }
 rand = "0.7.3"
 rayon = { version = "1" }
+num-bigint = "0.3"
+custom_derive = ""
```
mimoo, line 16: wait, this works o_O?

jspada: Yeah, I tested it by
* Serializing field element to base10 string using this method
* Deserializing this base10 string back to a field element (using `Fp::from_str`)
* Checked that it's equal to what I started with

I saw that in other places in zexe they used `BigInt::parse_bytes` and `BigInt::from_bytes_le`

jspada: It's not great, but it gives you what you need to create the OCaml test vectors.

mimoo: I didn't know you could include a library without a version number. It also looks like it was superseded by another library? https://crates.io/crates/custom_derive

mimoo:
cc @jspada the test vectors work on javascript: https://github.com/MinaProtocol/mina/pull/8913

time to merge this one :D?

oracle/Cargo.toml
jspada: add version number, make dev dependency & update readme

oracle/Cargo.toml
```diff
@@ -14,6 +14,28 @@ ocaml = { version = "0.22.0", optional = true }
 rand = "0.7.3"
 rayon = { version = "1" }
 
-[features]
+# for tests
+serde = { version = "1.0", features = ["derive"] }
+serde_json = { version = "1.0" }
+hex = { version = "0.4" }
```
mimoo, line 20: nit: this is for export_test_vectors, if it was for tests then we would have put that in dev-dependencies

jspada: These are required for running the tests
For example, `cargo test --test poseidon_tests` runs the tests in `poseidon_tests.rs`, which calls `serde_json::from_reader` then  `test_vectors -> hex_to_field -> hex::decode`

mimoo: >  I tried moving them to dev-dependencies before, but it was not happy. Feel free to update if you want to have a go at it.

that's because they are required for export_test_vectors which is not a test

jspada: Quickly found a way to do it.

oracle/Cargo.toml
```diff
@@ -14,21 +14,24 @@ ocaml = { version = "0.22.0", optional = true }
...
 custom_derive = { version = "0.1.7", optional = true }
 enum_derive = { version = "0.1.7", optional = true }
+serde = { version = "1.0", features = ["derive"], optional = true }
+serde_json = { version = "1.0", optional = true }
+hex = { version = "0.4", optional = true }
+
+[dev-dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = { version = "1.0" }
+hex = { version = "0.4" }
```
mimoo, line 28: you don't need to re-add them here since they are already listed as normal dependencies above

oracle/Cargo.toml
mimoo: that's cleaner, nice!

imeckler (approved):

PR #107, opened 2021-06-02
[step 1] cleaning up before upgrade to arkworks

mimoo:
This PR adds the missing crates to the workspace (in the root `Cargo.toml`), remove some unused dependencies, and runs `cargo fmt`


mrmr1993 (approved):
The changes look good, and I verified locally that the changes match what `cargo fmt` spits out. Nice :)

PR #109, opened 2021-06-02
[step 2] moving the pasta curves from zexe to this workspace

mimoo:
builds on https://github.com/o1-labs/marlin/pull/107 (so you can ignore commits from #107)

this PR moves the pasta curves code from our fork of zexe to this repository (in a root crate called `curves`).

Note that some tests have to be commented out as we can't import the test crates from zexe anymore. This is fixable, as arkworks export these tests in a public crate: https://github.com/arkworks-rs/algebra/blob/HEAD/test-templates/Cargo.toml so we can fix this later once we upgrade.


mrmr1993 (approved):
Verified manually with `diff -r curves/src/pasta zexe/algebra/src/pasta` that this diff only includes
* reformatting of the constants to add line-breaks
* removal of `#[cfg(feature = "pasta")]` annotations
* commenting out of temporarily unavailable tests
  - @mimoo can you open an issue to re-enable these when you're merging this PR?

PR #112, opened 2021-06-08
Upgrade ocaml-rs to 0.22.0

mrmr1993:
This depends on o1-labs/zexe#50; we should land that first.


mrmr1993:
This has already been merged to the release branch on the mina side. Merging.

PR #114, opened 2021-06-16
[step 5] zexe -> arkworks

mimoo:
step 4 needs to land first before this one (https://github.com/o1-labs/marlin/pull/113)

This PR makes the final move from zexe to arkworks. There's a lot of name import changes, and fixes to make things work. What is important is that the fixes don't fix things incorrectly.

### DIFF
--- a/curves/src/pasta/curves/pallas.rs
+++ b/curves/src/pasta/curves/pallas.rs
@@ -1,12 +1,9 @@
 use crate::pasta::*;
-use algebra::{
-    biginteger::BigInteger256,
-    curves::{
-        models::short_weierstrass_jacobian::{GroupAffine, GroupProjective},
-        ModelParameters, SWModelParameters,
-    },
-    field_new, Zero,
+use ark_ec::{
+    models::short_weierstrass_jacobian::{GroupAffine, GroupProjective},
+    ModelParameters, SWModelParameters,
 };
+use ark_ff::{biginteger::BigInteger256, field_new, Zero};
 
 #[derive(Copy, Clone, Default, PartialEq, Eq)]
 pub struct PallasParameters;
--- a/curves/src/pasta/curves/pallas.rs
+++ b/curves/src/pasta/curves/pallas.rs
@@ -18,32 +18,16 @@ pub type Projective = GroupProjective<PallasParameters>;
 
 impl SWModelParameters for PallasParameters {
     /// COEFF_A = 0
-    const COEFF_A: Fp = field_new!(Fp, BigInteger256([0x0, 0x0, 0x0, 0x0]));
+    const COEFF_A: Fp = field_new!(Fp, "0");
 
     /// COEFF_B = 5
-    const COEFF_B: Fp = field_new!(
-        Fp,
-        BigInteger256([
-            0xa1a55e68ffffffed,
-            0x74c2a54b4f4982f3,
-            0xfffffffffffffffd,
-            0x3fffffffffffffff
-        ])
-    );
+    const COEFF_B: Fp = field_new!(Fp, "5");
 
     /// COFACTOR = 1
     const COFACTOR: &'static [u64] = &[0x1];
 
     /// COFACTOR_INV = 1
-    const COFACTOR_INV: Fq = field_new!(
-        Fq,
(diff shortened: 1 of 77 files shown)


Pratyush:
btw once this is merged, you might be able to also drop the re-implementation of the pasta curves here; we have an implementation in [`ark-pallas`](https://crates.io/crates/ark-pallas) and [`ark-vesta`](https://crates.io/crates/ark-vesta)

mimoo:
@Pratyush thanks for chiming in! The problem is that we use different generators, but I was wondering if we could import the rest while still keeping our modifications o.o

Pratyush:
As in the elliptic curve prime subgroup generators? Is there a reason why they were chosen differently from Zcash? (I believe we just have the Zcash generators)

mimoo:
(rebased)

imeckler:
> As in the elliptic curve prime subgroup generators? Is there a reason why they were chosen differently from Zcash? (I believe we just have the Zcash generators)

I don't think there's a good reason, we just had chosen them before the ZCash team had chosen theirs. Ours use 1 as the x coordinate and theirs uses -1.

oracle/src/sponge.rs
```diff
@@ -155,28 +155,24 @@ where
         let total_length = P::ScalarField::size_in_bits();
 
         x.iter().for_each(|x| {
-            // Big endian
-            let mut bits: Vec<bool> = x.into_repr().to_bits();
-            // Little endian
-            bits.reverse();
-            let mut bits: Vec<_> = (0..total_length)
+            // Padding
+            let bits: Vec<bool> = x.into_repr().to_bits_be();
```
imeckler, line 159: I think may be wrong

imeckler (changes requested):
Looks good, just left one comment on a bit of code where I think you may have changed the semantics.

curves/src/pasta/curves/pallas.rs
```diff
@@ -18,32 +18,16 @@ pub type Projective = GroupProjective<PallasParameters>;
```
mimoo, line 18: I verified that the comments were correct with

```

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn print_out_parameters() {
        println!("COEFF_B: {}", VestaParameters::COEFF_B);
        println!("COFACTOR_INV: {}", VestaParameters::COFACTOR_INV);
        println!("G_GENERATOR_X: {}", G_GENERATOR_X);
        println!("G_GENERATOR_Y: {}", G_GENERATOR_Y);
    }
}
```

and

```

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn print_out_parameters() {
        println!("COEFF_B: {}", PallasParameters::COEFF_B);
        println!("COFACTOR_INV: {}", PallasParameters::COFACTOR_INV);
        println!("G_GENERATOR_X: {}", G_GENERATOR_X);
        println!("G_GENERATOR_Y: {}", G_GENERATOR_Y);
    }
}
```

oracle/src/sponge.rs
Pratyush: Yeah the old code inserted the `0` values at the beginning of the `Vec`. Something like the following should work:
```rust
            let bits: Vec<bool> = x.into_repr().to_bits_be();
            let mut padding = vec![false, total_length - bits.len()]; // maybe `checked_sub`?
            padding.extend(bits);
			bits = padding;
```

mimoo: good catch! for some reason I remember simplifying this part but it looks like I forgot some code >.<

mimoo: fixed in https://github.com/o1-labs/marlin/pull/114/commits/25bcb80a24c10ebead8ed1cfa578450f9b07572c

mimoo: don't think your code would work @pratyush as total_length and bits.len() are not known at compile time

Pratyush: The `vec![]` macro doesn't require compile-time lengths; run-time is fine =)

mimoo: woot! TIL :D

mimoo:
rebased

Issue #116, opened 2021-06-18
Simplify sponge interface


jspada:
While working on the new 3-wire Poseidon Matthew suggested that it would be a good improvement if the sponge interface (e.g. `absorb` and `squeeze`) did not always require the `ArithmeticSpongeParams` passed to them.  It should always be the same parameters, so this makes sense.

Instead of having to always pass them, we could just add the params as an argument to `Sponge::new` and store them in struct.

PR #117, opened 2021-06-18
Simplified sponge interface

jspada:
This simplifies the sponge interface so we don't need to repeatedly pass the `ArithmeticSpongeParams` everywhere.

See #116


mrmr1993 (approved):
Really nice, this is much cleaner!

PR #119, opened 2021-07-02
[step 6] refactoring the ocaml bindings to accomodate arkworks

mimoo:
**step 5 (https://github.com/o1-labs/marlin/pull/114) needs to land first**

The upgrade from zexe to arkworks has removed the `ocaml::FromValue` and `ocaml::ToValue` implementations on a number of foreign types. Because of this, we're now working on wrapper types that don't necessarily implement the same traits, which makes things like this not work:

```rust
use ark_ec::AffineCurve;

// ...

#[cfg_attr(feature = "ocaml_types", derive(ocaml::ToValue, ocaml::FromValue))]
struct CamlProverProof<G: AffineCurve> {
    pub commitments: ProverCommitments<G>,
    pub proof: OpeningProof<G>,
    // OCaml doesn't have sized arrays, so we have to convert to a tuple..
    pub evals: (ProofEvaluations<Vec<Fr<G>>>, ProofEvaluations<Vec<Fr<G>>>),
    pub public: Vec<Fr<G>>,
    pub prev_challenges: Vec<(Vec<Fr<G>>, PolyComm<G>)>,
}

```

This PR creates (or refactor) equivalent OCaml-compatible types for all the structures that we need in Mina.


mimoo:
@mrmr1993 this works, but the 5-wire stuff is now broken. IIUC 5-wire is the future, and so I should fix it as well right?

mimoo:
rebased in https://github.com/o1-labs/marlin/pull/121

PR #122, opened 2021-07-12
Bump zexe

mrmr1993:
The zexe submodule somehow got out of sync. This PR updates it to the latest master, which fixes a compile error.


mrmr1993:
This is trivial, just propagating a fix from the zexe repo so that we can fix mina's CI. Merging without review.

PR #128, opened 2021-07-20
[easy] added rust documentation in CI

mimoo:


mimoo:
test is failing because of the zexe submodule which will disappear once we move to arkworks, I'll wait for that to rebase

imeckler (approved):

PR #132, opened 2021-08-04
Return Option<_>s or Result<_>s instead of panicking

mrmr1993:
This PR tweaks the behaviour of some methods to ensure that we avoid panicking wherever possible.

This is motivated by compilation to WASM: there is no unwinding primitive there (yet), and so any panic is treated as an 'abort' instead, halting the WASM process.

### DIFF
--- a/circuits/plonk/src/constraints.rs
+++ b/circuits/plonk/src/constraints.rs
@@ -84,22 +84,23 @@ pub struct ConstraintSystem<F: FftField> {
     pub fr_sponge_params: ArithmeticSpongeParams<F>,
 }
 
-pub fn zk_w<F: FftField>(domain: D<F>) -> F {
-    domain.group_gen.pow(&[domain.size - 3])
+pub fn zk_w<F: FftField>(domain: D<F>) -> Option<F> {
+    if domain.size < 3 { return None }
+    Some(domain.group_gen.pow(&[domain.size - 3]))
 }
 
-pub fn zk_polynomial<F: FftField>(domain: D<F>) -> DensePolynomial<F> {
+pub fn zk_polynomial<F: FftField>(domain: D<F>) -> Option<DensePolynomial<F>> {
     // x^3 - x^2(w1+w2+w3) + x(w1w2+w1w3+w2w3) - w1w2w3
-    let w3 = zk_w(domain);
+    let w3 = zk_w(domain)?;
     let w2 = domain.group_gen * w3;
     let w1 = domain.group_gen * w2;
 
-    DensePolynomial::from_coefficients_slice(&[
+    Some(DensePolynomial::from_coefficients_slice(&[
         -w1 * &w2 * &w3,
         (w1 * &w2) + &(w1 * &w3) + &(w3 * &w2),
         -w1 - &w2 - &w3,
         F::one(),
-    ])
+    ]))
 }
 
 impl<F: FftField + SquareRootField> ConstraintSystem<F> {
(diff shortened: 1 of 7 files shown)


dlog/commitment/src/commitment.rs
```diff
@@ -60,19 +60,16 @@ pub fn shift_scalar<F: PrimeField>(x: F) -> F {
...
                 } else {
                     let points = com
                         .iter()
-                        .map(|c| {
-                            assert!(c.shifted.is_some());
-                            c.shifted.unwrap()
-                        })
-                        .collect::<Vec<_>>();
+                        .map(|c| { c.shifted })
+                        .collect::<Option<Vec<_>>>()?;
```
imeckler, line 72: What is this call to collect doing?

mrmr1993: It's using [this impl](https://doc.rust-lang.org/std/option/enum.Option.html#impl-FromIterator%3COption%3CA%3E%3E). It's basically returning `None` if any of the values is `None`, or `Some(_)` if they're all `Some`s.

dlog/plonk/src/index.rs
```diff
@@ -107,7 +107,9 @@ where
             emul2_comm: srs.get_ref().commit_non_hiding(&self.cs.emul2m, None),
             emul3_comm: srs.get_ref().commit_non_hiding(&self.cs.emul3m, None),
 
-            w: zk_w(self.cs.domain.d1),
+            // Safe to unwrap here because we've already successfully called this when creating the
+            // constraint system.
```
mimoo, line 111: better to use except than unwrap, the comment could move in there perhaps?

dlog/plonk/src/prover.rs
```diff
@@ -299,7 +301,7 @@ where
         fq_sponge.absorb_g(&t_comm.unshifted);
         fq_sponge.absorb_g(&vec![dummy; max_t_size - t_comm.unshifted.len()]);
         {
-            let s = t_comm.shifted.unwrap();
+            let s = t_comm.shifted.unwrap(); // Unwrap is fine here, we've just created the value.
```
mimoo, line 304: sam here

circuits/plonk/src/constraints.rs
```diff
@@ -84,22 +84,23 @@ pub struct ConstraintSystem<F: FftField> {
     pub fr_sponge_params: ArithmeticSpongeParams<F>,
 }
 
-pub fn zk_w<F: FftField>(domain: D<F>) -> F {
-    domain.group_gen.pow(&[domain.size - 3])
+pub fn zk_w<F: FftField>(domain: D<F>) -> Option<F> {
+    if domain.size < 3 { return None }
+    Some(domain.group_gen.pow(&[domain.size - 3]))
```
mimoo, line 89: I think it's cleaner to use `checked_sub` for overflows/underflows as you don't need to think about the edge case:

```rust
domain.size.checked_sub(3).map(|idx| 
  domain.group_gen.pow(&[idx])
);
```

dlog/commitment/src/commitment.rs
mimoo: Shouldn't we return an Error here instead of an Option? It seems like you would never want this behavior to not end up being an error.

circuits/plonk/src/constraints.rs
jspada: this seems better to me also

PR #134, opened 2021-08-23
Reduce rounds in proofs when possible

imeckler:
This changes the inner product opening prover to reduce the number of rounds in an opening proof to the minimum possible, instead of treating every polynomial as one of maximum degree (possibly "padded" with 0 coefficients).

The key idea of the implementation is just to change the `open` method to be on a [`TrimmedSRS` struct](https://github.com/o1-labs/proof-systems/pull/134/files#diff-bf2e54871ef20ba4fd6906cd31a6992f6dc9978c9af23b6eb5c82ec431f7fde0R25) (which represents an SRS which is shortened to some smaller power of 2) and then just expose basically a slice into the SRS array rather than the whole thing.


dlog/commitment/src/srs.rs
```diff
@@ -22,6 +22,40 @@ pub struct SRS<G: CommitmentCurve> {
     pub endo_q: G::BaseField,
 }
 
+pub struct TrimmedSRS<'a, G: CommitmentCurve> {
+    pub srs: &'a SRS<G>,
+    pub trimmed_length_log2: usize
+}
```
mimoo, line 28: I'd be good to comment anything public so we can generate some rustdoc at some point :o

dlog/commitment/src/srs.rs
```diff
@@ -22,6 +22,40 @@ pub struct SRS<G: CommitmentCurve> {
...
+
+impl<G: CommitmentCurve> SRS<G> {
+    pub fn trim<'a>(&'a self, trimmed_length_log2: usize) -> TrimmedSRS<'a, G> {
+        assert!((1 << trimmed_length_log2) <= self.g.len());
+        TrimmedSRS { srs: self, trimmed_length_log2 }
+    }
+}
```
mimoo, line 35: it'd be good to have a test to exercise the new functionality

dlog/plonk/src/index.rs
```diff
@@ -131,7 +131,7 @@ where
         srs: SRSSpec<'a, G>,
     ) -> Self {
         let srs = SRSValue::create(srs);
-        let max_poly_size = srs.get_ref().g.len();
+        let max_poly_size = cs.domain.d1.size as usize;
```
mimoo, line 134: can you explain this change? My understanding is that this is used to know what is the threshold before we have to split our polynomials. How is that related with d1's size?

dlog/commitment/src/srs.rs
jspada: Happy to help with this if you want to suggest the test details.

imeckler: I edited the dlog/tests/batch.rs test so that it uses trimming

dlog/plonk/src/index.rs
imeckler: added a comment explaining

mimoo: I'm still not sure I understand, you're assuming:

* `cs.domain.d1.size` is going to be smaller than the SRS
* `cs.domain.d1.size` is going to be a perfect log2

In that case, why aren't we creating an SRS of this size, or trusting that it was created with the correct size in mind? (and keep using `srs.get_ref().g.len()`)

dlog/commitment/src/commitment.rs
```diff
@@ -441,13 +441,13 @@ where
         mut sponge: EFqSponge, // sponge
         rng: &mut dyn RngCore,
     ) -> OpeningProof<G> {
-        let rounds = ceil_log2(self.g.len());
```
mimoo, line 444: I think you can get rid of ceil_log2 in the codebase now

dlog/plonk/src/index.rs
mimoo: (also the assert is useless right below if you set `max_poly_size = cs.domain.d1.size`

dlog/plonk-5-wires/src/index.rs
```diff
@@ -82,28 +82,31 @@ where
...
 {
     pub fn verifier_index(&self) -> VerifierIndex<G> {
-        let srs = match &self.srs {
+        let srs0 = match &self.srs {
             SRSValue::Value(s) => SRSValue::Value(s.clone()),
             SRSValue::Ref(x) => SRSValue::Ref(x),
         };
 
+        let trimmed_length = (self.cs.domain.d1.size as u64).trailing_zeros() as usize;
+        let srs = srs0.get_ref().trim(trimmed_length);
```
mimoo, line 91: here I'm confused as well. My understand here is that we should trim only when the polynomial we want to commit to fits in `n` such that `n < g.len()` and `n` is a power of 2. So why trim the SRS for all the future commitments here?

mimoo: oh never mind, we only use the trimmed SRS to commit to the pre-computed polynomials, but store the non-trimmed "srs0" in the verifier index.

dlog/commitment/src/commitment.rs
```diff
@@ -10,7 +10,7 @@ The folowing functionality is implemented
 
 *****************************************************************************************************************/
 
-use crate::srs::SRS;
+use crate::srs::TrimmedSRS;
```
mimoo, line 13: test

dlog/plonk-5-wires/src/prover.rs
```diff
@@ -189,12 +189,15 @@ where
             .interpolate()
         });
 
+        let trimmed_length = (index.cs.domain.d1.size as u64).trailing_zeros() as usize;
+        let srs = index.srs.get_ref().trim(trimmed_length);
```
mimoo, line 193: here I'm confused as well, why not just use the default SRS? And what if the domain size is not a perfect power of 2?

dlog/plonk-5-wires/src/prover.rs
```diff
@@ -232,7 +235,7 @@ where
             .interpolate();
 
         // commit to z
-        let z_comm = index.srs.get_ref().commit(&z, None, rng);
+        let z_comm = srs.commit(&z, None, rng);
```
mimoo, line 238: there's a number of `self.srs` below that can be simplified

dlog/plonk/src/prover.rs
```diff
@@ -97,6 +97,20 @@ unsafe impl<'a, G: AffineCurve + ocaml::FromValue<'a>> ocaml::FromValue<'a> for
...
+    let mut i = c.unshifted.len() - 1;
+    let zero = C::zero();
+    loop {
+        if c.unshifted[i] != zero {
+            break;
+        }
+        i -= 1;
+    };
+    c.unshifted.truncate(i + 1);
+    c
```
mimoo, line 111: I think it'd be cleaner to write it as:

```rust
for (i, u) in c.unshifted.iter().rev().enumerate() {
  if u != C::zero() {
      return PolyComm {
          unshifted: c.unshifted.truncate(i);
          ..c
      }
  }
}
c
```

mimoo: (where unshifted.truncate is implemented to truncate `i` entries)

mimoo: also, isn't this something we should implement automatically whenever we commit to something? (instead of here as a separate function)

dlog/tests/batch.rs
```diff
@@ -31,8 +31,14 @@ where
...
 
-    let size = 1 << 7;
+    let size_log2 = 7;
+    let padding_to_test_trimming = 2;
+    let size = 1 << (size_log2 + padding_to_test_trimming);
+
+    // We make the srs larger than the "trimmed" size in order to
+    // test the trimming functionality.
     let srs = SRS::<Affine>::create(size);
+    let srs = srs.trim(size_log2);
```
mimoo, line 41: at this point wouldn't it be better if `SRS::create` returned a trimmed SRS?

dlog/commitment/src/srs.rs
```diff
@@ -22,6 +22,42 @@ pub struct SRS<G: CommitmentCurve> {
     pub endo_q: G::BaseField,
 }
 
+/// A trimmed reference string. It contains a reference to an SRS, along with a
+/// trimmed power-of-2 length.
+pub struct TrimmedSRS<'a, G: CommitmentCurve> {
+    srs: &'a SRS<G>,
```
mrmr1993, line 28: This is accessed in `commitment.rs`, it should probably be public.

dlog/tests/batch.rs
imeckler: The goal with the trimming is to have one big global SRS (say 2^17 or something) and then trim it down for smaller circuits so that they all share the memory.

dlog/plonk/src/prover.rs
imeckler: Hmm.. you know now that I think about it, I think this is incorrect as it breaks zero-knowledge (it leaks that the top coefficients are zero). Will fix

dlog/tests/batch.rs
mimoo: Oh! I see, that makes more sense.

mimoo:
Thinking a bit more about this, would it make sense to change the API so that a proof creation or verification takes an SRS and per-circuit parameters as two different arguments, as opposed to an index that combines both? Here `self` could be the SRS, and the per-circuit parameter a normal argument.

dlog/commitment/src/srs.rs
mimoo: (getter here is better as you don't want to let someone set that manually.)

dlog/plonk/src/prover.rs
mimoo: I don't think it leaks anything as these zeros we're trimming are public

mimoo: I guess you could further improve the code as:

```rust
let unshifted = c.unshifted.clone();
let mut truncated: Vec<_> = unshifted.into_iter().rev().skip_while(|&num| num == 0u8).collect();
truncated.reverse();
PolyComm {
  unshifted: truncated;
  ..c
}
```

mimoo:
btw @imeckler what happened to this one?

PR #139, opened 2021-09-01
Commit evaluations directly using Lagrange basis

imeckler:
This PR makes it possible to commit to evaluations directly using precomputed commitments to Lagrange polynomials.

Currently, this should improve prover performance if many of the values in the columns are 0 or small. When committing using the standard basis, we don't get to exploit that as once we interpolate all the coefficients become kind of random.

In the future, it should be possible to remove the calls to `interpolate` completely, which will improve performance further by getting rid of FFTs.


dlog/commitment/src/fft.rs
```diff
@@ -0,0 +1,122 @@
...
+        // Perform sub-FFT
+        serial_fft(tmp, new_omega, new_two_adicity);
+    });
+
+    a.iter_mut()
+        .enumerate()
+        .for_each(|(i, a)| *a = tmp[i % num_chunks][i / num_chunks]);
+}
+
+pub fn serial_group_fft<G: ProjectiveCurve>(a: &mut [G], omega: G::ScalarField, log_n: u32) {
```
mimoo, line 85: pub functions should have comments for the rustdoc: https://o1-labs.github.io/proof-systems/

dlog/commitment/src/srs.rs
```diff
@@ -6,16 +6,22 @@ This source file implements the Marlin structured reference string primitive
...
 use std::io::{Read, Result as IoResult, Write};
+use std::collections::HashMap;
+use crate::fft::group_fft;
+use ark_poly::{EvaluationDomain, Radix2EvaluationDomain as D};
 
 #[derive(Debug, Clone)]
 pub struct SRS<G: CommitmentCurve> {
     pub g: Vec<G>, // for committing polynomials
     pub h: G,      // blinding
+    pub lagrange_bases: HashMap<usize, Vec<G>>, // Lagrange bases, per domain size
```
mimoo, line 24: btw, as these are all `pub`, each field can be commented with `///` comments and they'll appear in rustdoc

dlog/commitment/src/fft.rs
```diff
@@ -0,0 +1,122 @@
+use ark_ec::ProjectiveCurve;
+use ark_ff::{Field, One, FftField};
+use ark_poly::domain::DomainCoeff;
+use rayon::prelude::*;
+use std::ops::MulAssign;
```
mimoo, line 5: btw, it looks like most of this code is taken from https://github.com/arkworks-rs/algebra/blob/86d804bbe6a75f876dddc29dd7973724c84d1a03/poly/src/domain/utils.rs#L78 is there a reason not to use the functions provided by arkworks?

dlog/commitment/src/commitment.rs
```diff
@@ -380,52 +380,32 @@ where
...
-                }
-            }
-        };
+    pub fn commit_evaluations_non_hiding(
+        &self,
+        domain: D<Fr<G>>,
+        plnm: &Evaluations<Fr<G>, D<Fr<G>>>,
+        max: Option<usize>,
+    ) -> PolyComm<G> {
+        let is_zero = plnm.evals.iter().all(|x| x.is_zero());
```
mimoo, line 392: might be a good idea to implement `is_zero()` on the [ExtendedEvaluations](https://github.com/o1-labs/proof-systems/blob/mimoo/15-wire-plonk-fixes-3/utils/src/evaluations.rs#L6) trait

dlog/commitment/src/srs.rs
```diff
@@ -70,6 +76,38 @@ where
...
+        let mut lg: Vec<<G as AffineCurve>::Projective> =
+            self.g[0..n].iter().map(|g| g.into_projective()).collect();
+
+        group_fft::<<G as AffineCurve>::Projective>(
+            lg.as_mut_slice(),
+            domain.group_gen_inv,
+            domain.log_size_of_group,
+        );
+
+        let n_inv = <G::ScalarField as From<u64>>::from(n as u64)
```
mimoo, line 100: can't you write `let: G::ScalarField = (n as u64).inverse().unwrap().into();` ?

dlog/commitment/src/fft.rs
imeckler: I didn't use it because it's private -- I guess we could open a PR over there to expose it publicly if @Pratyush et al are open to it being public.

dlog/commitment/src/srs.rs
imeckler: What is inverse as u64?

dlog/commitment/src/fft.rs
Pratyush: What is the context for using these? You can directly invoke these FFT methods via `EvaluationDomain::fft`

imeckler: Oh cool -- for whatever reason I didn't realize I could just call those. Deleted this file and switched to calling `ifft_in_place` directly.

dlog/commitment/src/srs.rs
mimoo: oops, 
```
let n: G::ScalarField = (n as u64).into();
let n_inv = n.inverse().except("n was 0");
```

dlog/commitment/src/srs.rs
```diff
@@ -70,9 +77,27 @@ where
...
+        if self.lagrange_bases.contains_key(&n) {
+            return;
+        }
+
+        let mut lg: Vec<<G as AffineCurve>::Projective> =
+            self.g[0..n].iter().map(|g| g.into_projective()).collect();
+        domain.ifft_in_place(&mut lg);
+
+        <G as AffineCurve>::Projective::batch_normalization(lg.as_mut_slice());
+        self.lagrange_bases.insert(n, lg.iter().map(|g| g.into_affine()).collect());
```
mimoo, line 97: I'm not sure I really understand what is happening here tbh. Also, a test would be nice to make sure that committing in evaluation forms is equivalent to committing in coeff form

dlog/commitment/src/commitment.rs
```diff
@@ -926,6 +906,60 @@ impl<F: Field> Utils<F> for DensePolynomial<F> {
...
+        (0..p / n + if p % n != 0 { 1 } else { 0 })
+            .map(|i| {
+                VariableBaseMSM::multi_scalar_mul(
+                    basis,
+                    &scalars[i * n..p]
+                        .iter()
+                        .map(|s| s.into_repr())
+                        .collect::<Vec<_>>(),
+                )
+                .into_affine()
```
Pratyush, line 931: You might want to do a batch inversion here, much faster.

mimoo (approved):
Thanks for quickly addressing the comments. The test made it really clear what was going on, and the explanation is top notch!

dlog/commitment/src/commitment.rs
imeckler: You mean "batch normalization" right (ofc it involves the batch inversion...) -- yeah it would be better. This vector is quite small but still worth doing.

Issue #135, opened 2021-08-23
List of improvements


mimoo:
common:

- [x] implement testing coverage
- [ ] implement property testing
- [ ] implement fuzzing?
- [x] benches for every PR in CI

poly commitments:

- [ ] create a better API for polynomial commitments (ideally, there should be different types for commitments, hiding commitments, and batches commitments, with an API to easily create batch of commitments based on several commitments)
- [ ] prevent user to use SRS2 to create an evaluation proof for a polycomm that was created with SRS1
- [ ] document
- [ ] flamegraphs and benches

15-wires circuits:

- [ ] write tests for all gates (gate, polynomial, as well as negative test (e.g. this witness shouldn't work))
- [ ] [constraint-system] make sure that there's always enough rows for the public inputs
- [ ] CircuitGate should use an enum for separate different gates, instead of using the same format (c, wires, row, etc.)
- [ ] document

15-wires protocol:

- [ ] document
- [ ] give a better type to the recursion stuff (`prev_challenges`) and perhaps make it optional?
- [ ] create benchmarks and make sure that we get CI feedback on each PR to understand how changes affect proving and verifying performances. Same for proof size.
- [ ] use OsRng instead of ThreadRng
- [ ] The proverIndex and verifierIndex should have a commonIndex that stores all the stuff they share
- [ ] [prover] pass the public arguments as a list of public values instead of the first columns of the witness
- [x] use a single type for the alphas (not an alpha + powers of alpha starting at 2). 
- [ ] better abstractions for chunked polynomials, ideally objects that contain chunked polynomials or not should have different types.
- [ ] flamegraphs and benches
- [ ] create more nested types for large structs like ConstraintSystem
- [ ] there's no need to have a "nolookup" folder now (there used to be lookup and nolookup modules)

oracle:

- [ ] do not make the state public
- [ ] do we really need parts of the config to be a generic argument instead of being passed as a normal argument at creation?
- [ ] document

PR #140, opened 2021-09-02
[15-wires] some bug fixes

mimoo:
In this PR I fix three issues:

* when we absorb a curve point with Poseidon, we panic if the curve point is zero
* when we use the Fq sponge to create a digest for the Fr sponge we panic if the value is too large.

### DIFF
--- a/oracle/src/sponge.rs
+++ b/oracle/src/sponge.rs
@@ -191,7 +191,13 @@ where
 
     fn digest(mut self) -> P::ScalarField {
         let x: <P::BaseField as PrimeField>::BigInt = self.squeeze_field().into_repr();
-        P::ScalarField::from_repr(x.into()).expect("the sponge code has a bug")
+        // Returns zero for values that are too large.
+        // This means that there is a bias for the value zero (in one of the curve).
+        // An attacker could try to target that seed, in order to predict the challenges u and v produced by the Fr-Sponge.
+        // This would allow the attacker to mess with the result of the aggregated evaluation proof.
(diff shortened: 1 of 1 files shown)


dlog/plonk-15-wires/src/prover.rs
```diff
@@ -333,18 +336,18 @@ where
             println!("p{} f_comm {:?}", line!(), index.srs.get_ref().commit_non_hiding(&f, None));
         } */
 
-        let f = &(&(&(&(&(&index.cs.gnrc_lnrz(&e[0].w)
+        let f = &(&(&(&(&(&index.cs.gnrc_lnrz(&evals[0].w)
```
jspada, line 339: Would aid readability to add a comment with the equation and description above this

mimoo: yeah I think readability of the prover/verifier could be improved :o I've started listing ideas of future improvements we could work on here: https://github.com/o1-labs/proof-systems/issues/135

oracle/src/sponge.rs
```diff
@@ -142,7 +142,9 @@ where
         self.last_squeezed = vec![];
         for g in g.iter() {
             if g.infinity {
-                panic!("sponge got zero curve point");
+                // absorb a fake point (0, 0)
```
imeckler, line 145: In the future -- I think we should guarantee that commitments are not point at infinity by blinding them. Fine for now though

imeckler (approved):

PR #144, opened 2021-09-03
[commitment] refactored the batch commitment

mimoo:
This commit cleans the batch commitment test to make it clearer. It also creates types in order to make it easier to create aggregated proofs and pass them to the verify API.

I also added a simpler unit test.


imeckler (approved):

mimoo:
rebasing

jspada (approved):

PR #151, opened 2021-09-08
[15-wires][easy] clean generic gate scalars

mimoo:


dlog/plonk-15-wires/src/verifier.rs
```diff
@@ -387,9 +387,11 @@ where
                 // generic
                 commitments_part.push(&index.qm_comm);
                 commitments_part.extend(index.qw_comm.iter().map(|c| c).collect::<Vec<_>>());
-                commitments_part.push(&index.qc_comm);
                 scalars_part.extend(&ConstraintSystem::gnrc_scalars(&evals[0].w));
 
+                commitments_part.push(&index.qc_comm);
+                scalars_part.push(Fr::<G>::one());
+
```
jspada, line 394: Why do you put one here?

jspada (approved):

dlog/plonk-15-wires/src/verifier.rs
mimoo: cf. the last line in the equation here: https://o1-labs.github.io/mina-book/specs/15_wires/generic.html

Screen Shot 2021-09-09 at 5 03 54 PM

PR #155, opened 2021-09-16
[easy][commitment] rephrase what the linter doesn't like

mimoo:
For some reason the linter in vscode is confused by the `max()` function there. This makes the code uglier, but fixes the only lint issue.


dlog/commitment/src/commitment.rs
```diff
@@ -80,7 +83,7 @@ impl<C: AffineCurve> PolyComm<C> {
                 if com.len() == 0 || elm.len() == 0 {
                     Vec::new()
                 } else {
-                    let n = com.iter().map(|c| c.unshifted.len()).max().unwrap();
+                    let n = Iterator::max(com.iter().map(|c| c.unshifted.len())).unwrap();
```
mimoo, line 86: here is the change. The file was not properly formatted hence the other no-op changes

jspada (approved):
lgtm!

PR #121, opened 2021-07-12
[step 6] refactoring the ocaml bindings to accomodate arkworks 

mimoo:
The upgrade from zexe to arkworks has removed the `ocaml::FromValue` and `ocaml::ToValue` implementations on a number of foreign types (e.g.`Fp`). Because of this, we're now working on wrapper types (e.g. `CamlFp(Fp)`) that don't necessarily implement the same traits, which makes things like this not work:

```rust
use ark_ec::AffineCurve;
// ...
#[cfg_attr(feature = "ocaml_types", derive(ocaml::ToValue, ocaml::FromValue))]
struct CamlProverProof<G: AffineCurve> {
    pub commitments: ProverCommitments<G>,
    pub proof: OpeningProof<G>,
    // OCaml doesn't have sized arrays, so we have to convert to a tuple..
    pub evals: (ProofEvaluations<Vec<Fr<G>>>, ProofEvaluations<Vec<Fr<G>>>),
    pub public: Vec<Fr<G>>,
    pub prev_challenges: Vec<(Vec<Fr<G>>, PolyComm<G>)>,
}

```

This PR creates (or refactor) equivalent OCaml-compatible types for all the structures that we need in Mina.

It also:

* moves the ocaml code into its own module, it's cleaner
* updates to the latest version ocaml-rs (so `ToValue` becomes `IntoValue`)

PR #156, opened 2021-09-17
[ocaml] generate bindings from Rust code

mimoo:
The idea of this PR is to avoid writing the OCaml bindings by hand, and generate them automatically from the Rust code. ([I also recorded myself going over the code a bit if that helps](https://www.loom.com/share/a96000d653614e248f6b7639838db9fb).)

In the end, you end up writing this kind of code (taken from https://github.com/MinaProtocol/mina/pull/9525):

```rust
fn main() {
    println!("(* this file is generated automatically *)\n");
    let env = &mut Env::default();

    decl_fake_generic!(T1, 0);
    decl_fake_generic!(T2, 1);

    decl_module!(env, "Types", {
        decl_type!(env, CamlScalarChallenge::<T1> => "scalar_challenge");
        decl_type!(env, CamlRandomOracles::<T1> => "random_oracles");
        decl_type!(env, CamlProofEvaluations::<T1> => "proof_evaluations");
        decl_type!(env, CamlPolyComm::<T1> => "poly_comm");
        decl_type!(env, CamlOpeningProof::<T1, T2> => "opening_proof");
        decl_type!(env, CamlProverCommitments::<T1> => "prover_commitments");
        decl_type!(env, CamlProverProof<T1, T2> => "prover_proof");
    });

    decl_module!(env, "BigInt256", {
        decl_type!(env, CamlBigInteger256 => "t");
        decl_func!(env, caml_bigint_256_of_numeral => "of_numeral");
    });

    decl_module!(env, "Fp", {
        decl_type!(env, CamlFp => "t");
        decl_func!(env, caml_pasta_fp_size_in_bits => "size_in_bits");
        decl_func!(env, caml_pasta_fp_size);
        decl_func!(env, caml_pasta_fp_add);
```

to generate a single OCaml file containing:

```ocaml
(* this file is generated automatically *)

module Types = struct
  type 'CamlF scalar_challenge = 'CamlF

  type 'CamlF random_oracles = {
    beta : 'CamlF;
    gamma : 'CamlF;
    alpha_chal : 'CamlF scalar_challenge;
    alpha : 'CamlF;
    zeta : 'CamlF;
    v : 'CamlF;
    u : 'CamlF;
    zeta_chal : 'CamlF scalar_challenge;
    v_chal : 'CamlF scalar_challenge;
    u_chal : 'CamlF scalar_challenge;
  }

  type 'F proof_evaluations = {
    l : 'F array;
    r : 'F array;
    o : 'F array;
    z : 'F array;
    t : 'F array;
    f : 'F array;
    sigma1 : 'F array;
    sigma2 : 'F array;
  }

  type 'CamlG poly_comm = { unshifted : 'CamlG array; shifted : 'CamlG option }

  type ('G, 'F) opening_proof = {
    lr : 'G * 'G array;
    delta : 'G;
    z1 : 'F;
    z2 : 'F;
    sg : 'G;
  }

  type 'CamlG prover_commitments = {
    l_comm : 'CamlG poly_comm;
    r_comm : 'CamlG poly_comm;
    o_comm : 'CamlG poly_comm;
    z_comm : 'CamlG poly_comm;
    t_comm : 'CamlG poly_comm;
  }

  type ('CamlG, 'CamlF) prover_proof = {
    commitments : 'CamlG prover_commitments;
    proof : ('CamlG, 'CamlF) opening_proof;
    evals : 'CamlF proof_evaluations * 'CamlF proof_evaluations;
    public : 'CamlF array;
    prev_challenges : 'CamlF array * 'CamlG poly_comm array;
  }
end
module BigInt256 = struct
  type t

  external of_numeral : bytes -> int -> int -> t = "caml_bigint_256_of_numeral"
end

module Fp = struct
  type t

  external size_in_bits : unit -> int = "caml_pasta_fp_size_in_bits"

  external caml_pasta_fp_size : unit -> BigInt256.t = "caml_pasta_fp_size"

  external caml_pasta_fp_add : t -> t -> t = "caml_pasta_fp_add"
end
```

Essentially, it creates a number of macros to help derive the OCaml bindings.

* `#[derive(OcamlGen)]` derive macro for types. This implements the `OCamlDesc` and `OCamlBinding` traits.
* `#[derive(OcamlCustomTypes)]` derive macro for custom types. This also implements the `OCamlDesc` and `OCamlBinding` traits.
* `#[ocaml_gen]` attribute macro for functions, which creates a `*_to_ocaml()` function returning the generated OCaml binding

The traits it implements on types are the following ones:

```rust
/// OCamlBinding is the trait implemented by types to generate their OCaml bindings.
/// It is usually derived automatically via the [OcamlGen] macro,
/// or the [OCamlCustomType] macro for custom types.
/// For functions, refer to the [ocaml_gen] macro.
pub trait OCamlBinding {
    /// will generate the OCaml bindings for a type (called root type).
    /// It takes the current environment [Env],
    /// as well as an optional name (if you wish to rename the type in OCaml).
    fn ocaml_binding(env: &mut Env, rename: Option<&'static str>) -> String;
}

/// OCamlDesc is the trait implemented by types to facilitate generation of their OCaml bindings.
/// It is usually derived automatically via the [OcamlGen] macro,
/// or the [OCamlCustomType] macro for custom types.
pub trait OCamlDesc {
    /// describes the type in OCaml, given the current environment [Env]
    /// and the list of generic type parameters of the root type
    /// (the type that makes use of this type)
    fn ocaml_desc(env: &Env, generics: &[&str]) -> String;

    /// Returns a unique ID for the type. This ID will not change if concrete type parameters are used.
    fn unique_id() -> u128;
}
```

Essentially, the `ocaml_binding()` function is called to generate bindings for a type, and it'll call `ocaml_desc()` on all of its fields (recursively). The `ocaml_desc()` function is aware of the current environment (via the `Env` var), and can thus re-write types that have been renamed or relocated in some other module.

In order to avoid mixing types, a `unique_id()` function is also implemented, which returns a unique id per-type (regardless of the concrete type parameters used)

### DIFF
--- a/circuits/plonk-15-wires/src/lookup/scalars.rs
+++ b/circuits/plonk-15-wires/src/lookup/scalars.rs
@@ -35,8 +35,9 @@ impl<F: FftField> ProofEvaluations<Vec<F>> {
     }
 }
 
+/*
 #[derive(Clone)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::ToValue, ocaml::FromValue))]
+#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
 pub struct CamlProofEvaluations<Fs> {
     pub w: (Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs),
     pub z: Fs,
@@ -51,9 +52,9 @@ pub struct CamlProofEvaluations<Fs> {
 }
 
 #[cfg(feature = "ocaml_types")]
-unsafe impl<Fs: ocaml::ToValue> ocaml::ToValue for ProofEvaluations<Fs> {
-    fn to_value(self) -> ocaml::Value {
-        ocaml::ToValue::to_value(CamlProofEvaluations {
+unsafe impl<Fs: ocaml::IntoValue> ocaml::IntoValue for ProofEvaluations<Fs> {
+    fn into_value(self) -> ocaml::Value {
+        ocaml::IntoValue::to_value(CamlProofEvaluations {
             w: {
                 let [w0, w1, w2, w3, w4, w5, w6, w7, w8, w9, w10, w11, w12, w13, w14] = self.w;
                 (
@@ -102,9 +103,10 @@ unsafe impl<Fs: ocaml::FromValue> ocaml::FromValue for ProofEvaluations<Fs> {
         }
     }
 }
+*/
 
 #[derive(Clone, Debug)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::ToValue, ocaml::FromValue))]
+#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
 pub struct RandomOracles<F: Field> {
     // Plonk oracles
     pub po: RO<F>,
--- a/circuits/plonk-15-wires/src/lookup/scalars.rs
+++ b/circuits/plonk-15-wires/src/lookup/scalars.rs
@@ -35,78 +35,7 @@ impl<F: FftField> ProofEvaluations<Vec<F>> {
     }
 }
 
-/*
-#[derive(Clone)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
-pub struct CamlProofEvaluations<Fs> {
-    pub w: (Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs, Fs),
-    pub z: Fs,
-    pub t: Fs,
-    pub f: Fs,
-    pub s: (Fs, Fs, Fs, Fs, Fs),
-    pub l: Fs,
-    pub lw: Fs,
-    pub h1: Fs,
-    pub h2: Fs,
-    pub tb: Fs,
-}
-
-#[cfg(feature = "ocaml_types")]
-unsafe impl<Fs: ocaml::IntoValue> ocaml::IntoValue for ProofEvaluations<Fs> {
-    fn into_value(self) -> ocaml::Value {
-        ocaml::IntoValue::to_value(CamlProofEvaluations {
-            w: {
-                let [w0, w1, w2, w3, w4, w5, w6, w7, w8, w9, w10, w11, w12, w13, w14] = self.w;
-                (
-                    w0, w1, w2, w3, w4, w5, w6, w7, w8, w9, w10, w11, w12, w13, w14,
-                )
-            },
-            z: self.z,
-            t: self.t,
-            f: self.f,
-            s: {
-                let [s0, s1, s2, s3, s4] = self.s;
-                (s0, s1, s2, s3, s4)
-            },
-            l: self.l,
-            lw: self.lw,
-            h1: self.h1,
-            h2: self.h2,
-            tb: self.tb,
-        })
-    }
-}
-
-#[cfg(feature = "ocaml_types")]
-unsafe impl<Fs: ocaml::FromValue> ocaml::FromValue for ProofEvaluations<Fs> {
-    fn from_value(v: ocaml::Value) -> Self {
-        let evals: CamlProofEvaluations<Fs> = ocaml::FromValue::from_value(v);
-        ProofEvaluations {
-            w: {
-                let (w0, w1, w2, w3, w4, w5, w6, w7, w8, w9, w10, w11, w12, w13, w14) = evals.w;
-                [
-                    w0, w1, w2, w3, w4, w5, w6, w7, w8, w9, w10, w11, w12, w13, w14,
-                ]
-            },
-            z: evals.z,
-            t: evals.t,
-            f: evals.f,
-            s: {
-                let (s0, s1, s2, s3, s4) = evals.s;
-                [s0, s1, s2, s3, s4]
-            },
-            l: evals.l,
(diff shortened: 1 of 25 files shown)


circuits/plonk/src/scalars.rs
```diff
@@ -69,3 +67,126 @@ impl<F: Field> RandomOracles<F> {
...
+pub mod caml {
+    use super::*;
+    use ocaml_gen::OcamlGen;
+    use oracle::sponge::caml::CamlScalarChallenge;
+
+    //
+    // ProofEvaluations<F> <-> CamlProofEvaluations<CamlF>
+    //
+
+    #[derive(Clone, ocaml::IntoValue, ocaml::FromValue, OcamlGen)]
```
mrmr1993, line 85: Can you check the generated assembly code and confirm that we're not doing extra 'deep deconstruct-reconstruct' clones due to this? If we are, lets just make the type parameter represent `'f array` instead of `'f` and deal with this on the OCaml end.

mimoo: I think you're catching a potential performance regression from this PR: https://github.com/o1-labs/proof-systems/pull/121/

IIRC, I didn't get why the types were not clearly defined if they were always vectors. 

I can create an issue for this (as this seems unrelated to this PR) if that's OK? (Also I'm a bit worried about  handling `Vec<F>` as a generic type parameter within ocaml-gen.)

mimoo:
had some trouble with CI, but all good now!

mimoo:
rebased

circuits/plonk/src/scalars.rs
```diff
@@ -38,7 +37,6 @@ impl<F: FftField> ProofEvaluations<Vec<F>> {
 }
 
 #[derive(Clone, Debug)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
```
mrmr1993, line 41: Can you un-delete this, so that the existing bindings will still compile while you finish converting the new ones?

dlog/plonk/src/prover.rs
```diff
@@ -27,7 +27,6 @@ type Fr<G> = <G as AffineCurve>::ScalarField;
 type Fq<G> = <G as AffineCurve>::BaseField;
 
 #[derive(Clone)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
```
mrmr1993, line 30: Similarly here, would be better not to delete this and break the bindings.

dlog/plonk/src/prover.rs
```diff
@@ -36,16 +35,6 @@ pub struct ProverCommitments<G: AffineCurve> {
     pub t_comm: PolyComm<G>,
 }
 
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
```
mrmr1993, line 39: Likewise here.

dlog/plonk/src/prover.rs
```diff
@@ -64,48 +53,6 @@ pub struct ProverProof<G: AffineCurve> {
     pub prev_challenges: Vec<(Vec<Fr<G>>, PolyComm<G>)>,
 }
 
-#[cfg(feature = "ocaml_types")]
```
mrmr1993, line 67: And here.

oracle/src/sponge.rs
```diff
@@ -11,7 +11,6 @@ const HIGH_ENTROPY_LIMBS: usize = 2;
 // TODO: move to a different file / module
 /// A challenge which is used as a scalar on a group element in the verifier
 #[derive(Clone, Copy, Debug)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
```
mrmr1993, line 14: I think the existing bindings still rely on this; if so, undelete?

circuits/plonk/src/scalars.rs
```diff
@@ -10,7 +10,6 @@ use o1_utils::ExtendedDensePolynomial as _;
 use oracle::sponge::ScalarChallenge;
 
 #[derive(Clone)]
-#[cfg_attr(feature = "ocaml_types", derive(ocaml::IntoValue, ocaml::FromValue))]
```
mrmr1993, line 13: Undelete, as elsewhere

mrmr1993 (commented):
This lgtm in general. I'm keen to not break the existing code while the migration isn't complete yet; I've commented where you've removed `derive` statements that will do so.

circuits/plonk-15-wires/src/lookup/scalars.rs
```diff
@@ -35,8 +35,9 @@ impl<F: FftField> ProofEvaluations<Vec<F>> {
     }
 }
 
+/*
```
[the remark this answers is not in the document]
mimoo: this is required to build, and I thought the lookup folder would disappear in https://github.com/o1-labs/proof-systems/pull/158 but it didn't. @imeckler ?

circuits/plonk/src/scalars.rs
mimoo: I don't think this will work as the existing bindings use a different Field trait (from our zexe fork)

mrmr1993: Ah, of course.

circuits/plonk-15-wires/src/lookup/scalars.rs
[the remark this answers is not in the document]
mimoo: deleted!

mrmr1993 (approved):

mimoo:
make CI happy

PR #158, opened 2021-09-27
Lookup

imeckler:
This PR implements the lookup constraints and computation of the sorted-tables and aggregated lookup quotient. The implementation differs from the [lookup paper](https://eprint.iacr.org/2020/315.pdf) in a few key ways:

1. It allows for 4 lookups per row.
2. Each of these lookups can be a linear combination of columns, rather than just one of the columns itself.
3. Because of how we do zero-knowledge masking of polynomials, we cannot handle the overlap between adjacent chunks of sorted-lookup-table columns in the same "wrap around" way as the paper. The alternative is explained in the comment at the top of `polynomials/lookup.rs`.


circuits/plonk-15-wires/src/gate.rs
```diff
@@ -35,6 +35,11 @@ pub enum GateType {
     Endomul,
     /// lookup
     Lookup,
+    /// ChaCha
+    ChaCha0,
+    ChaCha1,
+    ChaCha2,
+    ChaChaFinal,
```
mimoo, line 42: :o

circuits/plonk-15-wires/src/lookup_spec.rs
```diff
@@ -0,0 +1,211 @@
+use ark_ff::{Field, FftField};
+use std::collections::{HashMap, HashSet};
+use ark_poly::{Radix2EvaluationDomain as D, Evaluations as E};
+use crate::domains::EvaluationDomains;
+use crate::gate::{GateType, CurrOrNext, CircuitGate};
+
+#[derive(Clone, Copy, Debug)]
+pub struct LocalPosition {
```
mimoo, line 8: If we could have any public type / field documented with `///` it would be swell :D

mimoo:
If I look at this PR individually, it looks like it just adds a gatetype for lookup, but it doesn't actually add lookups to the protocol right? It'd be good to have a test to show how this lookup gate works as well

PR #160, opened 2021-10-04
implementing serialization of types via Serde

mimoo:
We currently use serialization of our types by hand: https://github.com/MinaProtocol/mina/blob/develop/src/lib/marlin_plonk_bindings/stubs/src/index_serialization.rs

This is error-prone, and hard to maintain. As types and structs change, we need to make sure to change the serialization logic. The Rust-way is to use [Serde](https://serde.rs/), which offers a set of macros that we can use to implement serialization and deserialization traits for all of our types.

One of the problem we have is that [arkworks](http://arkworks.rs/), for some [yet-unknown reason](https://github.com/arkworks-rs/algebra/issues/178), doesn't implement `Serde::{Serialize, Deserialize}` on its types, but instead its own set of [ark-serialize](https://docs.rs/ark-serialize/0.3.0/ark_serialize/) traits. This means that we need to tell serde how to use these arkworks-specific traits to serialize and deserialize.

No biggy, the [serde_with](https://docs.rs/serde_with/1.10.0/serde_with) crate comes to the rescue.

### DIFF
--- a/circuits/plonk-15-wires/src/gate.rs
+++ b/circuits/plonk-15-wires/src/gate.rs
@@ -8,10 +8,17 @@ use crate::{nolookup::constraints::ConstraintSystem, wires::*};
 use ark_ff::bytes::{FromBytes, ToBytes};
 use ark_ff::FftField;
 use num_traits::cast::{FromPrimitive, ToPrimitive};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 use std::io::{Error, ErrorKind, Read, Result as IoResult, Write};
 
 #[repr(C)]
-#[derive(Clone, Debug, PartialEq, FromPrimitive, ToPrimitive)]
+#[derive(Clone, Debug, PartialEq, FromPrimitive, ToPrimitive, Serialize, Deserialize)]
+#[cfg_attr(
+    feature = "ocaml_types",
+    derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::OcamlEnum)
+)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub enum GateType {
     /// zero gate
     Zero,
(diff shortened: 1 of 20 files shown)


utils/src/dense_polynomial.rs
```diff
@@ -1,5 +1,12 @@
+//! This adds a few utility functions for the [DensePolynomial] arkworks type.
```
mimoo, line 1: squeezing in this commit, it's a no-op

circuits/plonk-15-wires/src/gate.rs
```diff
@@ -13,7 +13,7 @@ use serde_with::serde_as;
 use std::io::{Error, ErrorKind, Read, Result as IoResult, Write};
```
mimoo, line 13: squeezing this commit as well, just adding some ocaml types that I need

ocaml-gen/derive/src/lib.rs
```diff
@@ -139,7 +139,7 @@ pub fn derive_ocaml_enum(item: TokenStream) -> TokenStream {
```
mimoo, line 139: squeezing this one as well. This is to get better errors when using ocaml-gen

circuits/plonk-15-wires/src/nolookup/constraints.rs
```diff
@@ -72,63 +88,85 @@ pub struct ConstraintSystem<F: FftField> {
...
 
     /// wire coordinate shifts
+    #[serde_as(as = "[o1_utils::serialization::SerdeAs; PERMUTS]")]
     pub shift: [F; PERMUTS],
     /// coefficient for the group endomorphism
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub endo: F,
 
     /// random oracle argument parameters
+    #[serde(skip)]
```
mimoo, line 169: I created an issue about skipping fields: https://github.com/o1-labs/proof-systems/issues/161

PR #180, opened 2021-10-12
Lookup, chacha, scalar mul, poseidon

imeckler:
This PR implements 

- lookup polynomials
- chacha constraints
- poseidon constraints with expr
- scalar multiplication with expr


mimoo:
btw why am I still seeing 100 commits here o_O ?

PR #181, opened 2021-10-12
Complete EC addition gate

imeckler:
This PR replaces the gates

- EC double
- EC add

with a single gate `CompleteAdd` that performs whichever of doubling or adding is appropriate. The interface (accessible in the permutation columns) for this gate is, you store values as

```
0  1  2  3  4  5  6
x1 y1 x2 y2 x3 y3 inf
```

and if `(x1, y1) + (x2, y2)` is not the point at infinity, `(x3, y3)` is their sum (even if `(x1, y1) == (x2, y2)`).

Also in this case `inf = 0`.

If the sum is the point at infinity, `inf = 1` and `(x3, y3) = (x1, y1) + (x1, y1)`.

When using this gate, if the user expects the addition to succeed, `inf` should be constrained to be equal to `0`, perhaps using the permutation argument to constrain it to be equal to a variable already known to be zero.


mimoo:
btw why no gate file?

Issue #172, opened 2021-10-08
[ocaml-gen] issue with re-declaring custom types


mimoo:
Sometimes in OCaml, we can use the same custom type in different context.

For example, we use the `CamlPointer` custom type in different modules:

```rust
    decl_module!(w, env, "FieldVectors", {
        decl_module!(w, env, "Fp", {
            decl_type!(w, env, CamlPointer<T1> => "t");
            decl_type_alias!(w, env, "elt" => CamlFp);

            decl_func!(w, env, caml_pasta_fp_vector_create => "create");
            decl_func!(w, env, caml_pasta_fp_vector_length => "length");
            decl_func!(w, env, caml_pasta_fp_vector_emplace_back => "emplace_back");
            decl_func!(w, env, caml_pasta_fp_vector_get => "get");
        });

        decl_module!(w, env, "Fq", {
            decl_type!(w, env, CamlPointer<T1> => "t");
            decl_type_alias!(w, env, "elt" => CamlFq);

            decl_func!(w, env, caml_pasta_fq_vector_create => "create");
            decl_func!(w, env, caml_pasta_fq_vector_length => "length");
            decl_func!(w, env, caml_pasta_fq_vector_emplace_back => "emplace_back");
            decl_func!(w, env, caml_pasta_fq_vector_get => "get");
        });
    });
```

converts to

```ocaml
module FieldVectors = struct 

  module Fp = struct 
    type t
    type elt = Foundations.Fp.t
    external create : unit -> t = "caml_pasta_fp_vector_create"
    external length : t -> int = "caml_pasta_fp_vector_length"
    external emplace_back : t -> Foundations.Fp.t -> unit = "caml_pasta_fp_vector_emplace_back"
    external get : t -> int -> Foundations.Fp.t = "caml_pasta_fp_vector_get"
  end


  module Fq = struct 
    type t
    type elt = Foundations.Fq.t
    external create : unit -> t = "caml_pasta_fq_vector_create"
    external length : t -> int = "caml_pasta_fq_vector_length"
    external emplace_back : t -> Foundations.Fq.t -> unit = "caml_pasta_fq_vector_emplace_back"
    external get : t -> int -> Foundations.Fq.t = "caml_pasta_fq_vector_get"
  end

end
```

[...]

PR #195, opened 2021-10-21
[ocaml-gen] rename macros and revert multi-declaration of custom types

mimoo:
see https://github.com/o1-labs/proof-systems/issues/172#issuecomment-948145633

tl;dr is that this is that declaring the same custom type multiple times is error-prone. I'm investigating another solution.

as for name changes:

* OcamlGen -> ocaml_gen::Struct
* ocaml_gen -> ocaml_gen::func
* OCamlCustomType -> ocaml_gen::CustomType
* OcamlEnum -> ocaml_gen::Enum


mimoo:
rebased

mimoo:
rebased

mrmr1993 (approved):

PR #197, opened 2021-10-21
[kimchi][proof] add zero-knowledge rows in witness

mimoo:
We currently add zero-knowledge rows to the witness on the OCaml side. This is not great as this library might be used by all sorts of people who might not think about adding the zero-knowledge stuff.


mimoo:
rebased, fixing remaining tests now

mimoo:
Screen Shot 2021-10-29 at 1 28 04 PM

just so that there are no surprised, added an error in ProofError

PR #204, opened 2021-10-29
add round constants for round 55

imeckler:
Simply adds new round constants (generated using the same invocation of the `params.sage` script) for one more round.


mimoo (approved):
LGTM

PR #205, opened 2021-10-29
[kimchi] choosing messagepack as our serialization format

mimoo:
Quickly went through the different serialization format that we can use, and msgpack seems to be one of the smallest/fastest/most well-supported format.

The only other contestant is cbor, which seems to be originated from msgpack, and has an RFC. People have mix feelings, and the two standards seem close enough, that I chose the one with the best Rust support.

### DIFF
--- a/circuits/plonk-15-wires/Cargo.toml
+++ b/circuits/plonk-15-wires/Cargo.toml
@@ -30,7 +30,7 @@ ocaml = { version = "0.22.2", optional = true }
 ocaml-gen = { path = "../../ocaml/ocaml-gen", optional = true }
 
 [dev-dependencies]
-bincode = "1.3.3"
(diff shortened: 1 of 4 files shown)


mrmr1993 (approved):

mimoo:
rebasing

mimoo:
@mrmr1993 isn't there an option in the settings to allow people to rebase through github when merging?

mrmr1993:
> isn't there an option in the settings to allow people to rebase through github when merging?

There is, yes. We deliberately disabled it because rebases were causing us hell, we always prefer merges

mimoo:
ah interesting, what was the issue :o ? I always saw rebasing as cleaner as it doesn't pollute the commit history

PR #207, opened 2021-11-01
[kimchi] adding benchmarks code

mimoo:
This PR adds a benchmark using:

* [criterion.rs](https://github.com/bheisler/criterion.rs) a benchmark crate
* [iai](https://github.com/bheisler/iai) a benchmark crate better suited for CI

@mrmr1993 @imeckler is this enough to benchmark what we want? I'm not sure how you've run the bench so far. This benchmark only uses a circuit of a few thousands of generic gates. We can add other gates into the mix, Ideally we would test all gates here. Also, this only benches:

- the witness creation (very basic one, since we don't do what's being done on the OCaml side like the creation of the permutation)
- the proof creation
- the verification of the proof

we could also separate the proof creation from the verification of the proof in two different benchmarks.

### DIFF
--- a/circuits/kimchi/src/gates/generic.rs
+++ b/circuits/kimchi/src/gates/generic.rs
@@ -27,18 +27,10 @@ impl<F: FftField> CircuitGate<F> {
     }
 
     /// creates an addition gate
-    pub fn create_generic_add(row: usize, wires: GateWires) -> Self {
+    pub fn create_generic_add(row: usize, wires: GateWires, left_coeff: F, right_coeff: F) -> Self {
         let on = F::one();
         let off = F::zero();
-        let qw: [F; GENERICS] = array_init(|col| {
-            if col < 2 {
-                on
-            } else if col == 2 {
-                -on
-            } else {
-                off
-            }
-        });
+        let qw: [F; GENERICS] = [left_coeff, right_coeff, -on];
         Self::create_generic(row, wires, qw, off, off)
     }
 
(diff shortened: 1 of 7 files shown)


circuits/kimchi/src/gates/generic.rs
```diff
@@ -27,27 +27,19 @@ impl<F: FftField> CircuitGate<F> {
     }
 
     /// creates an addition gate
-    pub fn create_generic_add(row: usize, wires: GateWires) -> Self {
```
mimoo, line 30: I was missing the fact that we can do `left_coeff * w0 + right_coeff * w1`, not just `w0 + w1`

circuits/kimchi/src/gates/generic.rs
```diff
@@ -27,27 +27,19 @@ impl<F: FftField> CircuitGate<F> {
...
     }
 
     /// creates a multiplication gate
     pub fn create_generic_mul(row: usize, wires: GateWires) -> Self {
         let on = F::one();
         let off = F::zero();
-        let qw: [F; GENERICS] = array_init(|col| if col == 2 { on } else { off });
-        Self::create_generic(row, wires, qw, -on, off)
+        let qw: [F; GENERICS] = [off, off, -on];
+        Self::create_generic(row, wires, qw, on, off)
```
mimoo, line 42: cleaner to have a negation of the output selector coeff, instead of doing it on the multiplication selector coeff

circuits/kimchi/src/gates/generic.rs
```diff
@@ -94,7 +86,7 @@ impl<F: FftField> CircuitGate<F> {
         ensure_eq!(
             zero,
             sum + &mul + &constant_selector,
-            "generic: incorrect sum"
+            "generic: incorrect sum or mul"
```
mimoo, line 89: better error

imeckler (approved):

imeckler (approved):

mimoo:
rebase

PR #216, opened 2021-11-05
[kimchi] introduce alpha iterator

mimoo:
This PR introduces a few types to handle the powers of alpha. Please look at the commits independently when reviewing as I tried to split the changes in a clean way: https://github.com/o1-labs/proof-systems/pull/216/commits

closes https://github.com/o1-labs/proof-systems/issues/219

---

Most of the magic is explained in the documentation of the alpha module, but let's repeat it here:

This module implements an abstraction to keep track of the powers of alphas. As a recap, alpha is a challenge sent by the verifier in PLONK, and is used to segregate different constraints. It is important that different constraints use different powers of alpha, as otherwise they can interact and potentially cancel one another. (The proof is in the use of the Schwartz-Zipple lemma.)

We use powers of alpha in two different places in the codebase:

- when creating the index, we do not know alpha at this point so we simply keep track of what constraints will use what powers
- when creating a proof or verifying a proof, at this point we know alpha so we can use the mapping we created during the creation of the index.

For this to work, we use two types:

- [Builder], which allows us to map constraints to powers
- [Alphas], which you can derive from [Builder] and an `alpha`

Both construction will enforce that you use all the powers of alphas that you register for constraint. This allows us to
make sure that we compute the correct amounts, without reusing powers of alphas between constraints.


circuits/kimchi/src/expr.rs
```diff
@@ -1047,10 +1048,10 @@ impl<F: Field> Expr<ConstantExpr<F>> {
 
     /// Combines multiple constraints `[c0, ..., cn]` into a single constraint
     /// `alpha^alpha0 * c0 + alpha^{alpha0 + 1} * c1 + ... + alpha^{alpha0 + n} * cn`.
-    pub fn combine_constraints(alpha0: usize, cs: Vec<Self>) -> Self {
+    pub fn combine_constraints(alphas: &mut impl Iterator<Item = usize>, cs: Vec<Self>) -> Self {
         let zero = Expr::<ConstantExpr<F>>::zero();
         cs.into_iter()
-            .zip(alpha0..)
+            .zip_eq(alphas)
```
mimoo, line 1054: this is to make sure that we don't create LESS constraints if we set `alphas` to a range that's too small. (This will panic if we do this).

circuits/kimchi/src/polynomials/chacha.rs
```diff
@@ -342,71 +342,81 @@ pub fn chacha20(mut s: Vec<u32>) -> Vec<u32> {
     s
 }
 
-// TODO: Assert somehow that we're using the right number of alphas.
-pub fn constraint<F: FftField>(alpha0: usize) -> E<F> {
-    let w = |i: usize, r: CurrOrNext| E::<F>::cell(Column::Witness(i), r);
+struct Helper<F>(PhantomData<F>);
```
mimoo, line 345: converted a lot of the closures to functions. (No need to use closures as we don't capture outside values.)

mimoo: (I think ideally we should return all the constraints and combine them at the end of the main function `constraint`)

circuits/kimchi/src/polynomials/permutation.rs
```diff
@@ -93,27 +97,37 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
...
-        -e[0]
+        let alpha0 = alphas
+            .next()
+            .expect("not enough powers of alpha for permutation");
+        let _alpha1 = alphas
+            .next()
+            .expect("not enough powers of alpha for permutation");
+        let _alpha2 = alphas
+            .next()
+            .expect("not enough powers of alpha for permutation");
```
mimoo, line 122: I need to consume the iterator otherwise I get a warning

dlog/kimchi/src/index.rs
```diff
@@ -242,41 +256,70 @@ where
...
-            expr.linearize(evaluated_cols).unwrap()
+            // generic gate
+            let mut alphas = powers_of_alpha.register(ConstraintType::Gate(GateType::Generic), 1);
+            alphas.next();
+
+            // permutation
+            let mut alphas = powers_of_alpha.register(ConstraintType::Permutation, 3);
+            alphas.next();
+            alphas.next();
+            alphas.next();
```
mimoo, line 268: the generic and permutation gates are the only ones that don't use the expression framework, so I just need to register enough powers here and consume the iterator to avoid an error.

mimoo: the rest of the code below registers the right number of alphas for each gate, if it registers too many the expression framework will panic when it tries to call `combine_constraints`. If it doesn't register enough, same.

dlog/kimchi/src/prover.rs
```diff
@@ -417,77 +386,180 @@ where
...
+            t8 += &emulscalar8;
+
+            if cfg!(test) {
+                let (_, res) = emulscalar8
+                    .clone()
+                    .interpolate()
+                    .divide_by_vanishing_poly(index.cs.domain.d1)
+                    .unwrap();
+                assert!(res.is_zero());
+            }
```
mimoo, line 488: added these tests as I was dealing with another issue that was hard to debug. These tests were such a time saver that I strongly think we should keep them in.

dlog/kimchi/src/verifier.rs
```diff
@@ -344,7 +349,7 @@ where
             fq_sponge,
             digest,
             oracles,
-            alphas,
```
mimoo, line 347: looks like we don't use this `alphas` anywhere (at least not in proof-systems, and not in ocaml-land). Since we need to create the `Alphas` structure within the `oracles()` function we need to return it here. BTW I'd really like to get rid of this `oracles()` function and this `Oracles` struct.

mimoo:
cc @imeckler I changed the code so that now we reuse powers of alphas across mutually exclusive constraints.
So:

- the permutation argument, the lookup argument, and any gate, use a different range of alphas.
- gates all reuse the same alphas

circuits/kimchi/src/polynomials/chacha.rs
imeckler: What is the point of this struct rather than just having functions?

circuits/kimchi/src/polynomials/chacha.rs
```diff
@@ -422,20 +432,20 @@ pub fn constraint<F: FftField>(alpha0: usize) -> E<F> {
...
+        constraints.push(Helper::combine_nybbles(y_xor_xprime_rotated) - yprime);
+        E::combine_constraints(&mut alphas.by_ref().take(9), constraints)
     };
 
     let index = |g: GateType| E::cell(Column::Index(g), Curr);
     use GateType::*;
     vec![
         // a += b; d ^= a; d <<<= 16 (=4*4);
-        index(ChaCha0) * line(4),
+        index(ChaCha0) * Helper::line(&mut alphas.by_ref().take(5), 4),
```
imeckler: Guessing in the follow up these will all use the same alpha powers?

circuits/kimchi/src/polynomials/poseidon.rs
```diff
@@ -65,11 +75,10 @@ pub fn constraint<F: FftField + SquareRootField>() -> E<F> {
 /// ...
 /// The rth position in this array contains the alphas used for the equations that
 /// constrain the values of the (r+1)th state.
+pub fn constraint<F: FftField + SquareRootField>(alphas: &mut impl Iterator<Item = usize>) -> E<F> {
```
imeckler: It's important for the efficiency of the verifier circuit that the poseidon gate has the powers of alpha starting at 0. Can you enforce this somehow?

circuits/kimchi/src/polynomials/poseidon.rs
```diff
@@ -78,96 +87,28 @@ pub fn constraint<F: FftField + SquareRootField>() -> E<F> {
...
-        for eq in ROUND_EQUATIONS.iter() {
-            let ss = sboxed[eq.source];
-            let aa = alp[eq.source];
-            for (i, p) in ss.iter().enumerate() {
-                // Each of these contributes to the right hand side of SPONGE_WIDTH cell equations
-                let coeff =
-                    (0..SPONGE_WIDTH).fold(F::zero(), |acc, j| acc + aa[j] * params.mds[j][i]);
-                rhs += coeff * p;
-            }
+    for (round, eq) in ROUND_EQUATIONS.iter().enumerate() {
```
imeckler, line 90: The existing implementation was written carefully to ensure efficiency. If you want to include this rewrite can you make sure there was no degradation? If it's difficult to tell please revert otherwise

dlog/kimchi/src/index.rs
```diff
@@ -246,29 +256,56 @@ where
...
+
+            // poseidon gate
+            let mut alphas = powers_of_alpha.register(ConstraintType::Gate(GateType::Poseidon), 15);
+            let mut expr = poseidon::constraint(&mut alphas);
```
imeckler: Poseidon should be using [0, 15)

imeckler (changes requested):
Since there's a bunch of changes to the prover and verifier, can you check there have been no performance regressions?

circuits/kimchi/src/polynomials/chacha.rs
mimoo: oh there's no real benefit besides to avoid having to declare the generic parameters in the functions every time

circuits/kimchi/src/polynomials/chacha.rs
mimoo: yeah, I created functions for each chacha constraint as they are seen as different gates from the outside

circuits/kimchi/src/polynomials/poseidon.rs
mimoo: yeah! We just have to declare it as the first constraint in index :) I can add a commit for that

circuits/kimchi/src/polynomials/poseidon.rs
mimoo: the refactor is doing exactly the same thing (just in a cleaner way), unless I made a mistake somewhere,

mimoo: let me double check by printing though

mimoo:
@imeckler on my mac:

```
Benchmarking Proof creation/proof creation (2^16): Collecting 10 samples in estimated 61.648 s (10 iterat                                                                                                         Proof creation/proof creation (2^16)                        
                        time:   [5.5767 s 5.6278 s 5.6751 s]
                        change: [-2.7577% -1.4365% -0.2437%] (p = 0.06 > 0.05)

Benchmarking Proof verification/proof verification: Collecting 10 samples in estimated 5.8235 s (60 itera                                                                                                         Proof verification/proof verification                        
                        time:   [98.986 ms 102.51 ms 107.04 ms]
                        change: [-17.347% -13.928% -9.7095%] (p = 0.00 < 0.05)
                        Performance has improved.
```

dlog/kimchi/src/index.rs
mimoo: btw, this is because poseidon is the largest poly?

dlog/kimchi/src/index.rs
```diff
@@ -246,29 +256,56 @@ where
...
+            alphas.next();
+
+            // permutation
+            let mut alphas = powers_of_alpha.register(ConstraintType::Permutation, 3);
+            alphas.next();
+            alphas.next();
+            alphas.next();
+
+            // poseidon gate
+            let mut alphas = powers_of_alpha.register(ConstraintType::Gate(GateType::Poseidon), 15);
```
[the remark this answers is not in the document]
mimoo: I think you're looking at an outdated version of the code :o 

there were some limitations with things we could do prior to the big cleanup PR because kimchi-circuits was a dependency of kimchi and thus couldn't use kimchi types as a dependency either. But this limitation is lifted now, so I think there are nicer things we could do, like passing the `alphas::Builder` type directly to the constraint functions so that they can use the number of powers of alpha they need directly (and potentially report it back so that we can calculate that!)

But I propose to do this in another PR and note this as future improvement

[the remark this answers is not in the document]
jspada: 👍🏼  Created an issue to track this https://github.com/o1-labs/proof-systems/issues/272

kimchi/src/tests/endomul.rs
```diff
@@ -1,9 +1,12 @@
+<<<<<<< HEAD:kimchi/src/src/tests/endomul.rs
+=======
```
imeckler, line 2: Looks like you missed a file when merging

dlog/kimchi/src/prover.rs
jspada: sgtm!

circuits/kimchi/src/polynomials/poseidon.rs
jspada: Interesting @imeckler - why is that important for efficiency?

dlog/kimchi/src/verifier.rs
```diff
@@ -531,7 +531,7 @@ where
                                     ChaChaFinal => &index.chacha_comm.as_ref().unwrap()[3],
                                 };
                                 scalars.push(scalar);
-                                commitments.push(col);
+                                commitments.push(c);
```
jspada: don't prefer 1 character variable name - switch back to `col`

circuits/kimchi/src/polynomials/chacha.rs
```diff
@@ -514,7 +514,7 @@ mod tests {
             h.insert(Column::Index(GateType::Generic));
             h
         };
-        let mut powers = 0..20;
+        let mut powers = 0..24;
```
jspada: It would be cool if we could do something like `chacha::constraint_count()` or something like that.

circuits/kimchi/src/gate.rs
```diff
@@ -100,6 +100,11 @@ impl<F: Field> JointLookup<F> {
     }
 }
 
+/// The different types of gates the system supports.
+/// Note that all the gates are mutually exclusive:
+/// they cannot be used at the same time on single row.
```
jspada: The new double generic gate optimization is safe about this?

dlog/kimchi/src/alphas.rs
```diff
@@ -4,6 +4,10 @@
 //! It is important that different constraints use different powers of alpha,
 //! as otherwise they can interact and potentially cancel one another.
 //! (The proof is in the use of the Schwartz-Zipple lemma.)
+//! As such, we want two properties from this:
+//!
+//! - we should keep track of a mapping between type of constraint and range of powers
+//! - when powers of alphas are used, we should ensure that no more no less are used
```
jspada: I like your way of engineering this safety check in.

For unit tests are there even more exhaustive checks we could do?

jspada (changes requested):
Awesome work David!  Seems like a great improvement.  Put some suggestions in there.

circuits/kimchi/src/expr.rs
```diff
@@ -1841,6 +1841,16 @@ impl<F: Zero> Add<Expr<F>> for Expr<F> {
     }
 }
 
+impl<F: Zero + Clone> AddAssign<Expr<F>> for Expr<F> {
```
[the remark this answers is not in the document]
mimoo: you don't comment trait implementations in rust :o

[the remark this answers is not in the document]
jspada: Oh, why not?

I guess I should remove mine here?  https://github.com/o1-labs/proof-systems/blob/194b0748bc8a674f9453d11be8342e233e070752/signer/src/schnorr.rs#L64

I think sometimes can be useful.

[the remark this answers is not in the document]
mimoo: yours is not a rustdoc comment though. The reason is that you document the trait, not the implementation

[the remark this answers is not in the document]
jspada: Ah, of course, i just meant "comment" comment :-)

circuits/kimchi/src/polynomials/poseidon.rs
mimoo: I believe it's from this snippet in the verifier circuit:

```ocaml
    (* note that for all gates, the powers of alpha start at a^0 = 1 *)
    let poseidon =
      let (pn :: ps) = Vector.rev m.coefficients_comm in
      let res =
        Vector.fold ~init:pn ps ~f:(fun acc c -> c + endoscale acc alpha)
      in
      scale res plonk.poseidon_selector |> negate
```

you have one less endoscale.

circuits/kimchi/src/expr.rs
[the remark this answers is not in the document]
mimoo: tbh I'm not sure it makes sense to add any comment in this function :o it's pretty straight forward and follows the same logic as the Add implementation above

dlog/kimchi/src/alphas.rs
```diff
@@ -0,0 +1,291 @@
...
+
+impl Builder {
+    /// Registers a new [ConstraintType],
+    /// associating it a number `powers` of powers of alpha.
+    /// The function returns an iterator of powers that must be used.
+    #[must_use]
+    pub fn register(
+        &mut self,
+        ty: ConstraintType,
+        powers: usize,
```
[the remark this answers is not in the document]
mimoo: I'm not sure what the subgroups of the pasta curves are, but I would expect that with high probability alpha has a huge order, so no need to constrain. In practice we would never reach a large number anyway as adding constraints have cost

circuits/kimchi/src/polynomials/chacha.rs
```diff
@@ -342,71 +342,81 @@ pub fn chacha20(mut s: Vec<u32>) -> Vec<u32> {
...
-    };
+    }
 
-    // Constraints for the line L(x, x', y, y', z, k), where k = 4 * nybble_rotation
-    let line = |nybble_rotation: usize| -> E<F> {
-        let y_xor_xprime_nybbles = chunks_over_2_rows(3);
-        let x_plus_z_nybbles = chunks_over_2_rows(7);
-        let y_nybbles = chunks_over_2_rows(11);
+    /// Constraints for the line L(x, x', y, y', z, k), where k = 4 * nybble_rotation
+    fn line(alphas: &mut impl Iterator<Item = usize>, nybble_rotation: usize) -> E<F> {
```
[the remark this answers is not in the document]
mimoo: no they should use the same alphas now as well :o since they are used by different gates

circuits/kimchi/src/polynomials/chacha.rs
```diff
@@ -342,71 +342,81 @@ pub fn chacha20(mut s: Vec<u32>) -> Vec<u32> {
...
                 // y' = (y ^ x') <<< 4 * nybble_rotation
-                combine_nybbles(y_xor_xprime_rotated) - yprime,
+                Helper::combine_nybbles(y_xor_xprime_rotated) - yprime,
             ],
         )
-    };
+    }
+}
 
+pub fn constraint<F: FftField>(alphas: &mut impl Iterator<Item = usize>) -> E<F> {
```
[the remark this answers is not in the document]
mimoo: yeah chacha_final has 9 alphas

dlog/kimchi/src/verifier.rs
mimoo: it actually previously was `c` so I just reverted since `col` already exists and `c` is in scope only for a few lines

dlog/kimchi/src/alphas.rs
mimoo: it'd be nice to use proptests to test different configurations in tests

dlog/kimchi/src/alphas.rs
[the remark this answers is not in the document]
jspada: I guess it's 2^32 (from your earlier investigation)

dlog/kimchi/src/alphas.rs
jspada: 👍🏼

mrmr1993 (changes requested):
This PR seems to have a large number of unrelated changes. Can you rework this to use the minimal set of changes needed?

dlog/kimchi/src/verifier.rs
querolita: would you like `com` better?

jspada: i would

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
+//! This module implements an abstraction to keep track of the powers of alphas.
+//! As a recap, alpha is a challenge sent by the verifier in PLONK,
+//! and is used to segregate different constraints.
+//! It is important that different constraints use different powers of alpha,
+//! as otherwise they can interact and potentially cancel one another.
+//! (The proof is in the use of the Schwartz-Zipple lemma.)
```
querolita, line 6: _Zippel_

querolita (commented):
Very nice work! I have left a few comments here and there to clarify some goals of the code. Overall it looks really good, an elegant abstraction for the current `alpha0` parameter :) It also clarifies some parts of our discussion from last week

imeckler (approved):

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
...
+};
+
+// ------------------------------------------
+
+/// A constraint type represents a polynomial that will be part of the final equation f (the circuit equation)
+#[derive(PartialEq, Eq, Clone, Hash, Debug, Serialize, Deserialize)]
+pub enum ConstraintType {
+    /// gates in the PLONK constraint system.
+    /// As gates are mutually exclusive (only a single selector polynomial
+    /// will be non-zero for a given point),
```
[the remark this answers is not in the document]
mimoo: Screen Shot 2022-02-07 at 2 53 37 PM

I refer to the gate selectors, the second table from the left. You see that each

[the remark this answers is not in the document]
mimoo: Screen Shot 2022-02-07 at 3 06 48 PM

sigmas should only have 7 columns :D

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// we can reuse the same power of alphas accross gates.
```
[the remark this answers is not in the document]
mimoo: yes!

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
...
+// ------------------------------------------
+
+/// This type can be used to create a mapping between powers of alpha and constraint types.
+/// See [Builder::default] to create one,
+/// and [Builder::register] to register a new mapping.
+/// Once you know the alpha value, you can convert this type to a [Alphas].
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct Builder {
+    /// The next power of alpha to use
+    next_power: usize,
```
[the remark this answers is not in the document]
mimoo: I thought the comment and the "next" were enough to clarify that :o otherwise I would have called that "last"

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+        self.next_power = new_power;
+        range
+    }
+
+    /// Creates a new instance of [Alphas] via a [Builder] and value `alpha`.
+    pub fn into_alphas<F: Field>(self, alpha: F) -> Alphas<F> {
+        let mut last_power = F::one();
+        let mut alphas = Vec::with_capacity(self.next_power);
+        alphas.push(F::one());
```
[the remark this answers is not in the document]
mimoo: you're right! There is an off-by-one here, I pre-compute one unnecessary power of alpha

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
...
+            alphas,
+            mapping: powers.mapping.clone(),
+        }
+    }
+
+    /// This returns a range of powers (the exponents), upperbounded by `num`
+    #[must_use]
+    pub fn get_powers(
+        &mut self,
+        ty: ConstraintType,
```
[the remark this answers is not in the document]
mimoo: It's actually really annoying, because both `ty` and `type` are commonly used in code that deals with types.  This is because `type` is often a restricted keyword. I actually like `type_` better.

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,276 @@
...
+    I: Iterator<Item = T>,
+    T: std::fmt::Display,
+{
+    type Item = I::Item;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
+}
+
+impl<I, T> Drop for MustConsumeIterator<I, T>
```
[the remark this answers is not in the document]
mimoo: This is because they are pre-computed powers, not the actual range tracked

kimchi/src/index.rs
```diff
@@ -148,28 +156,34 @@ pub fn constraints_expr<F: FftField + SquareRootField>(
...
-        let constraints = lookup::constraints(dummy, domain);
-        let combined = Expr::combine_constraints(2 + super::range::CHACHA.end, constraints);
-        expr + combined
-    } else {
-        expr
-    };
+    // gates
+    let alphas = powers_of_alpha.register(ConstraintType::Gate, 21);
+
+    let mut expr = poseidon::constraint(alphas.clone().take(15));
```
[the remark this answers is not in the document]
mimoo: I was pretty limited previously because alpha was defined in the protocol crate, and since the circuit crate was a dependency it couldn't depend on the alpha type directly. Now that I moved the circuit crate into the kimchi crate, this should not be a limitation anymore. You can chime in https://github.com/o1-labs/proof-systems/issues/272, but I think this could be a good change in the future

kimchi/src/prover.rs
```diff
@@ -419,79 +389,214 @@ where
...
-            .divide_by_vanishing_poly(index.cs.domain.d1)
-            .map_or(Err(ProofError::PolyDivision), Ok)?;
-        if !res.is_zero() {
-            return Err(ProofError::PolyDivision);
-        }
+            drop(mul8);
 
-        t += &bnd;
+            // endoscaling
+            let alphas = all_alphas.get_powers(ConstraintType::Gate, 11);
```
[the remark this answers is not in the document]
mimoo: yeah I think it'd be nice if we didn't need to pass these values again. https://github.com/o1-labs/proof-systems/issues/272 might be the right place to discuss this

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -94,27 +98,37 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
...
+        alphas: impl Iterator<Item = F>,
     ) -> DensePolynomial<F> {
         let zkpm_zeta = self.zkpm.evaluate(&zeta);
-        let scalar = Self::perm_scalars(e, beta, gamma, alpha, zkpm_zeta);
+        let scalar = Self::perm_scalars(e, beta, gamma, alphas, zkpm_zeta);
         self.sigmam[PERMUTS - 1].scale(scalar)
     }
 
     pub fn perm_scalars(
         e: &[ProofEvaluations<F>],
```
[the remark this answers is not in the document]
mimoo: that's for another PR :)

kimchi/src/alphas.rs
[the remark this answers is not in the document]
querolita: aha cool because the sigmas refer to the IO registers I suppose
Ok, this diagram clarifies many things, but also compliments what I said above I think. Particularly, it is good to clarify that these things are known to the verifier (as you say, it is the circuit specification / relation to be proven). 
Now for the column sum you refer to for the gates, I think it is the opposite: I would say it is row-wise what has an all-1 sum. But anyway, I know what you mean. And yeah, I guess these selector polynomials are computed using lagrange, what will give them this linearly independence property. Together with the fact that the verifier has access to these, it all makes sense to me, thanks 👍

kimchi/src/alphas.rs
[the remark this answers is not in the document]
querolita: Ok nice, in any case this was just a triviality :)

kimchi/src/alphas.rs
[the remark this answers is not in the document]
querolita: Oooh I see! good clarification

kimchi/src/alphas.rs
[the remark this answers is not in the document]
querolita: Ah, "pre-compute", but not actually using it for the range... Cool, in any case I guess it is fine to avoid the unnecessary additional one (which I think you already fixed in the new commit 👍 )

kimchi/src/alphas.rs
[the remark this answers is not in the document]
querolita: mmm well perhaps this is just me? the name felt a bit confusing to me. So if you say "last" to me it will sound like "the very last one that is included".

kimchi/src/index.rs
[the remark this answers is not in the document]
querolita: Oooh that makes sense! Dependencies huh.... Fine, I'll check out that issue :)

kimchi/src/prover.rs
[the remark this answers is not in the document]
querolita: Thanks!

kimchi/src/circuits/polynomials/permutation.rs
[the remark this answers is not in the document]
querolita: oh cool, yeah I tend to address too many things in the same PR... my bad

kimchi/src/prover.rs
```diff
@@ -1,10 +1,5 @@
-/********************************************************************************************
+//! This module implements prover's zk-proof primitive.
```
mrmr1993, line 1: Spurious

kimchi/src/prover.rs
```diff
@@ -419,79 +389,214 @@ where
...
-                alpha,
-                alphas[alphas.len() - 1],
-                lookup::constraints(&index.cs.dummy_lookup_values[0], index.cs.domain.d1)
-                    .iter()
-                    .map(|e| e.evaluations(&env))
-                    .collect(),
-            ),
-        };
+            // scalar multiplication
+            let alphas = all_alphas.get_powers(ConstraintType::Gate, 21);
```
mrmr1993, line 448: `21` here is a 'magic constant', repeated here and in `index.rs`. It would be good to factor it out, as we had with range.

mrmr1993: Similarly with the others

kimchi/src/tests/chacha.rs
```diff
@@ -71,9 +71,13 @@ fn chacha_prover() {
 
     // create the index
     let fp_sponge_params = oracle::pasta::fp::params();
-    let cs =
-        ConstraintSystem::<Fp>::create(gates, vec![chacha::xor_table()], fp_sponge_params, PUBLIC)
-            .unwrap();
+    let cs = ConstraintSystem::<Fp>::create(
+        gates,
+        vec![chacha::testing::xor_table()],
```
mrmr1993, line 76: The `xor_table` is an inherent part of the `chacha` gate, not just for testing. It shouldn't have been moved to a testing module (and this PR isn't the place to do that anyway).

mrmr1993 (changes requested):

kimchi/src/circuits/polynomials/complete_add.rs
```diff
@@ -41,7 +41,7 @@ fn zero_check<F: Field>(z: E<F>, z_inv: E<F>, r: E<F>) -> Vec<E<F>> {
 /// for doubling.
 ///
 /// See [here](https://en.wikipedia.org/wiki/Elliptic_curve#The_group_law) for the formulas used.
-pub fn constraint<F: Field>(alpha0: usize) -> (usize, E<F>) {
+pub fn constraint<F: Field>(alphas: impl Iterator<Item = usize>) -> E<F> {
```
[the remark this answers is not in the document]
mimoo: I believe they are equivalent, cf https://doc.rust-lang.org/reference/types/impl-trait.html#anonymous-type-parameters

kimchi/src/circuits/gate.rs
```diff
@@ -106,6 +106,11 @@ impl<F: Field> JointLookup<F> {
     }
 }
 
+/// The different types of gates the system supports.
```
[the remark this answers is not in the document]
mimoo: I'll do this, but just wanted to share why I add these:

* imo we should take the opportunity to document undocumented stuff in our PRs
* in this particular example, I wanted to emphasis that they are mutually exclusive (which is related to the PR)

kimchi/src/circuits/gate.rs
```diff
@@ -427,7 +432,8 @@ impl<F: FftField> CircuitGate<F> {
             VarBaseMul => self.verify_vbmul(row, witness),
             EndoMul => self.verify_endomul(row, witness, cs),
             EndoMulScalar => self.verify_endomul_scalar(row, witness, cs),
-            ChaCha0 | ChaCha1 | ChaCha2 | ChaChaFinal => panic!("todo"),
+            // TODO: implement the verification for chacha
+            ChaCha0 | ChaCha1 | ChaCha2 | ChaChaFinal => Ok(()),
```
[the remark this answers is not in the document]
mimoo: IIRC some test was panicking, can't remember where

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -1,73 +1,84 @@
-/*****************************************************************************************************************
-
-This source file implements Posedon constraint polynomials.
-
-*****************************************************************************************************************/
+//! This module implements the Poseidon constraint polynomials.
```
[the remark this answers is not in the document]
mimoo: what do you mean by this? I'm turning this comment into a module-level rustdoc comment, which will get exported to our html generated docs

kimchi/src/circuits/gate.rs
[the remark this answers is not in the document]
mrmr1993: > imo we should take the opportunity to document undocumented stuff in our PRs

IMO documentation also needs review, and reviewing documentation of unrelated parts of the code while trying to understand a change is worse for both the quality of the review and the reviewer.

kimchi/src/circuits/polynomials/poseidon.rs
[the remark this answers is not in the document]
mrmr1993: I mean, this PR has an inflated diff due to changes like this, which are unrelated and act to obscure the actual meaningful logic changes. I would like to see this change in a different PR.

mimoo:
extracted what I could in other PRs, which have been merged (except for https://github.com/o1-labs/proof-systems/pull/293). This PR should look much leaner now!

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,261 @@
...
+//!   simply keep track of what constraints will use what powers
+//! - when creating a proof or verifying a proof, at this point we know alpha
+//!   so we can use the mapping we created during the creation of the index.
+//!
+//! For this to work, we use two types:
+//!
+//! - [Builder], which allows us to map constraints to powers
+//! - [Alphas], which you can derive from [Builder] and an `alpha`
+//!
+//! Both construction will enforce that you use all the powers of
```
mrmr1993, line 24: `constructions`

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// The next power of alpha to use
+    /// (the end result will be [1, alpha^next_power[)
+    next_power: usize,
+    /// The mapping between constraint types and powers of alpha
+    mapping: HashMap<ConstraintType, Range<usize>>,
+}
+
+impl Builder {
+    /// Registers a new [ConstraintType],
+    /// associating it a number `powers` of powers of alpha.
```
mrmr1993, line 72: `associating it with`

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// The function returns an iterator of powers that must be used.
```
mrmr1993, line 73: `This function returns an iterator that produces the number of powers registered.`

The `must be used` here appears not to be true

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,261 @@
...
+    #[must_use]
+    pub fn register(&mut self, ty: ConstraintType, powers: usize) -> Range<usize> {
+        let new_power = self.next_power + powers;
+        let range = self.next_power..new_power;
+        if self.mapping.insert(ty.clone(), range.clone()).is_some() {
+            panic!("you are attempting to register {:?} twice", ty);
```
mrmr1993, line 79: `cannot re-register {:?}`?

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,261 @@
...
+// ------------------------------------------
+
+/// This type can be used to create a mapping between powers of alpha and constraint types.
+/// See [Builder::default] to create one,
+/// and [Builder::register] to register a new mapping.
+/// Once you know the alpha value, you can convert this type to a [Alphas].
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct Builder {
+    /// The next power of alpha to use
+    /// (the end result will be [1, alpha^next_power[)
```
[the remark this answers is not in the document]
mimoo: that's not-inclusive

kimchi/src/alphas.rs
```diff
@@ -0,0 +1,261 @@
...
+
+impl Builder {
+    /// Registers a new [ConstraintType],
+    /// associating it a number `powers` of powers of alpha.
+    /// The function returns an iterator of powers that must be used.
+    #[must_use]
+    pub fn register(&mut self, ty: ConstraintType, powers: usize) -> Range<usize> {
+        let new_power = self.next_power + powers;
+        let range = self.next_power..new_power;
+        if self.mapping.insert(ty.clone(), range.clone()).is_some() {
```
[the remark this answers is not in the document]
mimoo: for `ty` yeah, I should derive Copy, but for `range` I do need it otherwise I can't move it afterwards

kimchi/src/prover.rs
mimoo: I have some ideas I want to explore on how to tackle this. I think it'd be nice if we didn't have to calculate this and we could directly get it from calling the linearization functions, but this should be in another PR.

Note that at the moment, if you enter a number too low or too high it will either crash due to the `zip_eq` (when combining) or because you didn't register enough. The only thing that can go wrong is that you register too many powers of alpha, which is not really a security issue.

mrmr1993 (approved):

mimoo:
@jspada looks like your requested change is still blocking me from merging this one

jspada (commented):
lgtm

Issue #211, opened 2021-11-03
[kimchi] absorb new evaluations


mimoo:
Since we added evaluations in the proof, I believe we need to add them here as well: https://github.com/o1-labs/proof-systems/blob/master/dlog/kimchi/src/plonk_sponge.rs#L38

PR #220, opened 2021-11-08
Backend modifications for pickles compatibility

imeckler:
This PR makes a few changes to ensure compatibility with the pickles verifier circuits and the rest of the OCaml code. Specifically

- adds OCaml types for expressions so that we can pass the `linearization` field of the index back and forth (it was previously just getting clobbered to the default value)
- removes some of the DensePolynomials from the constraint system and calculates using evaluation forms instead
- Makes it so that the T commitment is always 7 group elements, by padding it with commitments to the 0 polynomial if need be
- When absorbing field elements from the bigger field, splitting them as (packed_high_bits, low_bit) instead of (packed_low_bits, high_bit). The former is compatible with the decomposition we need for performing scalar multiplications inside the circuit.


mimoo:
Is this normal? I see a bunch of my commits here

circuits/kimchi/src/expr.rs
```diff
@@ -1947,3 +1951,308 @@ impl<F: fmt::Display + Clone> fmt::Display for Expr<F> {
...
+                    )
+                }
+                PolishToken::Store => CamlPolishToken::Store,
+                PolishToken::Load(x) => {
+                    CamlPolishToken::Load((*x).try_into().expect("isize -> usize"))
+                }
+            }
+        }
+    }
+}
```
mimoo, line 2258: so, do we really need to expose these types to the OCaml side? What do we really need on the OCaml side?

mimoo (commented):
some feedback/questions

circuits/kimchi/src/gates/varbasemul.rs
```diff
@@ -121,6 +121,8 @@ impl<F: FftField> CircuitGate<F> {
     }
 
     pub fn verify_vbmul(&self, witness: &[Vec<F>; COLUMNS]) -> Result<(), String> {
+        return Ok(());
```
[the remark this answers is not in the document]
mimoo: probably add a TODO + create an issue if this is too much work?

dlog/commitment/src/commitment.rs
```diff
@@ -75,9 +77,22 @@ where
     }
 }
 
-pub fn shift_scalar<F: PrimeField>(x: F) -> F {
-    let two: F = 2_u64.into();
-    x - two.pow(&[F::Params::MODULUS_BITS as u64])
+pub fn shift_scalar<G: AffineCurve>(x: G::ScalarField) -> G::ScalarField
```
mimoo, line 80: can you add some doc here?

mimoo:
closes https://github.com/o1-labs/proof-systems/issues/211

dlog/kimchi/src/prover.rs
```diff
@@ -487,13 +489,23 @@ where
         t += &bnd;
 
         // commit to t
-        let t_comm = index.srs.commit(&t, None, rng);
+        let t_comm = {
+            let (mut t_comm, mut omega_t) = index.srs.commit(&t, None, rng);
+
+            let expected_t_size = 7;
```
mimoo, line 495: nit: use constant COLUMNS value

dlog/kimchi/src/prover.rs
```diff
@@ -487,13 +489,23 @@ where
...
+            let dummies = expected_t_size - t_comm.unshifted.len();
+            // Add some hiding commitments to 0
+            for _ in 0..dummies {
+                use ark_ec::ProjectiveCurve;
+                let w = Fr::<G>::rand(rng);
+                t_comm.unshifted.push(index.srs.h.mul(w).into_affine());
+                omega_t.unshifted.push(w);
+            }
+            (t_comm, omega_t)
```
mimoo, line 504: can you explain that change? The comment seems to imply that the padding can only exist when the commitment is to 0

mimoo (commented):
more comments

dlog/commitment/src/commitment.rs
mrmr1993: Probably it would be sufficient to summarise [this](https://github.com/MinaProtocol/mina/blob/39dd86401bb56ee06a6004fe8ac4f952050d4a6d/src/lib/pickles_types/shifted_value.ml#L34) and [this](https://github.com/MinaProtocol/mina/blob/39dd86401bb56ee06a6004fe8ac4f952050d4a6d/src/lib/pickles_types/shifted_value.ml#L136) comment.

dlog/kimchi/src/index.rs
```diff
@@ -144,6 +144,55 @@ pub struct VerifierIndex<G: CommitmentCurve> {
     pub fq_sponge_params: ArithmeticSpongeParams<Fq<G>>,
 }
 
+pub fn expr_linearization<F: FftField + SquareRootField>(
+    domain: D<F>,
+    lookup_used: bool,
+    chacha: bool,
+    dummy_lookup_value: Option<&[F]>,
```
mrmr1993, line 151: Nit: use `dummy_loolup_value` as the signaller for `lookup_used`.

oracle/src/sponge.rs
```diff
@@ -153,37 +153,32 @@ where
...
-            bits.extend_from_slice(&unpadded_bits);
+            let bits = x.into_repr().to_bits_le();
 
             // absorb
             if <P::ScalarField as PrimeField>::Params::MODULUS
                 < <P::BaseField as PrimeField>::Params::MODULUS.into()
             {
                 self.sponge.absorb(&[P::BaseField::from_repr(
-                    <P::BaseField as PrimeField>::BigInt::from_bits_be(&bits),
+                    <P::BaseField as PrimeField>::BigInt::from_bits_le(&bits),
```
mrmr1993, line 165: This would similarly benefit from the `BigInt` equality constraint.

mrmr1993 (approved):
LGTM. I've left some nits.

For the lookup argument to be sound, I still think we need an equivalent to #245 (and similarly on the pickles side), but happy to add that separately.

circuits/kimchi/src/gates/endosclmul.rs
```diff
@@ -53,78 +55,48 @@ impl<F: FftField> CircuitGate<F> {
     ) -> Result<(), String> {
         let this: [F; COLUMNS] = array_init(|i| witness[i][self.row]);
         let next: [F; COLUMNS] = array_init(|i| witness[i][self.row + 1]);
-        let xq1 = (F::one() + ((cs.endo - F::one()) * next[12])) * this[0];
-        let xq2 = (F::one() + ((cs.endo - F::one()) * next[14])) * this[0];
 
-        ensure_eq!(self.typ, GateType::Endomul, "endomul: incorrect gate");
```
mimoo, line 59: I reverted this, since we check this for every gate. Not sure if it's useful but if you want to remove it remove all of them :)

circuits/kimchi/src/gates/endosclmul.rs
```diff
@@ -53,78 +55,48 @@ impl<F: FftField> CircuitGate<F> {
...
+            },
+            ProofEvaluations {
+                w: next,
+                z: F::zero(),
+                s: array_init(|_| F::zero()),
+                lookup: None,
+                generic_selector: F::zero(),
+                poseidon_selector: F::zero(),
+            },
+        ];
```
mimoo, line 87: perhaps we could use a function to initialize evals quickly if we're going to repeat this pattern across all the verify functions

circuits/kimchi/src/expr.rs
```diff
@@ -1800,7 +1800,7 @@ impl<F: Zero> Zero for Expr<F> {
     }
 }
 
-impl<F: One + PartialEq> One for Expr<F> {
+impl<F: Zero + One + PartialEq> One for Expr<F> {
```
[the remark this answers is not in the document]
mimoo: It is, but it's not immediately obvious why: `One` here requires `Mul` to be implemented on `Expr`.

And `Mul` is implemented as such:

```
impl<F: Zero + One + PartialEq> Mul<Expr<F>> for Expr<F> {
```

circuits/kimchi/src/wires.rs
```diff
@@ -19,12 +19,31 @@ pub const WIRES: [usize; COLUMNS] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 1
...
+)]
+impl Wire {
+    #[cfg_attr(
+        feature = "wasm_types",
+        wasm_bindgen::prelude::wasm_bindgen
+    )]
+    pub fn create(row: i32, col: i32) -> Self {
+        Self { row: row as usize, col: col as usize }
+    }
+}
```
mimoo, line 45: can't this function exist on the javascript side? It's pretty simple

dlog/kimchi/src/index.rs
mimoo: signaller?

mimoo:
rebased, fixed CI, addressed some of the comments  here: https://github.com/o1-labs/proof-systems/pull/257

dlog/commitment/src/commitment.rs
mimoo: not sure I understand the second comment, why is the constant 5 used? And shouldn't the result of the scalar mul also be split between high bit and low bits?

dlog/commitment/src/commitment.rs
```diff
@@ -75,9 +77,22 @@ where
...
+where
+    G::BaseField: PrimeField,
+{
+    let n1 = <G::ScalarField as PrimeField>::Params::MODULUS;
+    let n2 = <G::ScalarField as PrimeField>::BigInt::from_bits_le(
+        &<G::BaseField as PrimeField>::Params::MODULUS.to_bits_le()[..],
```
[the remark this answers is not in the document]
mimoo: also not to clear what's happening to me, is `from_bits_le` truncating its argument to chop the high bit?

dlog/commitment/src/commitment.rs
```diff
@@ -75,9 +77,22 @@ where
...
+    );
+    let two: G::ScalarField = (2 as u64).into();
+    if n1 < n2 {
+        (x - (two.pow(&[<G::ScalarField as PrimeField>::Params::MODULUS_BITS as u64])
+            + G::ScalarField::one()))
```
mimoo, line 91: from https://github.com/MinaProtocol/mina/blob/39dd86401bb56ee06a6004fe8ac4f952050d4a6d/src/lib/pickles_types/shifted_value.ml#L34 my understanding is that it should be `- G::ScalarField::one()` not `+`

mrmr1993: The parentheses handle negation, this is correct as written.

circuits/kimchi/src/wires.rs
mrmr1993: Javascript needs helpers like this to create rust objects in WASM, otherwise there's no way to construct a wire that could be passed back into rust calls.

circuits/kimchi/src/gates/endosclmul.rs
mrmr1993: A `dummy` function (or perhaps an implementation of `Default`) would be good here.

dlog/kimchi/src/index.rs
mrmr1993: Infer `lookuo_used = true` when it's `Some`, `false` when it's `None`.

circuits/kimchi/src/expr.rs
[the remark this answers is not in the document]
mrmr1993: `One: Mul<Self, Output = Self>`. What an odd design choice.

dlog/commitment/src/commitment.rs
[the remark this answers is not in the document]
mrmr1993: Potentially, yes, but I think we're implicitly assuming that the `BigInt`s are the same size. Maybe we should bite the bullet and assert equality everywhere.

dlog/commitment/src/commitment.rs
mimoo: oh, damn parenthesis :)

mimoo:
merged master instead of rebasing, cc @mrmr1993

dlog/commitment/src/commitment.rs
imeckler: Added a comment

dlog/kimchi/src/prover.rs
imeckler: Yes, the padding is some commitments to 0, since we only have to add padding when the upper degree coefficients of t are 0.

dlog/kimchi/src/index.rs
```diff
@@ -187,7 +170,37 @@ pub fn expr_linearization<F: FftField + SquareRootField>(
...
+    }
+    h.insert(Z);
+    h.insert(LookupAggreg);
+    h.insert(LookupTable);
+    h.insert(Index(GateType::Poseidon));
+    h.insert(Index(GateType::Generic));
+    h
+}
+
+pub fn expr_linearization<F: FftField + SquareRootField>(
```
mimoo, line 194: missing doc :D

dlog/kimchi/src/index.rs
```diff
@@ -187,7 +170,37 @@ pub fn expr_linearization<F: FftField + SquareRootField>(
         expr
     };
 
-    expr.linearize(evaluated_cols)
+    expr
+}
+
+pub fn linearization_columns<F: FftField + SquareRootField>() -> std::collections::HashSet<Column> {
```
mimoo, line 176: no doc here

PR #233, opened 2021-11-25
Removing field row from CircuitGate struct

querolita:
I removed the field row from the CircuitGate struct. It was redundant information, as we are creating circuits by adding CircuitGate elements to an array (so their row is their position in that array). Apart from the definitions of the struct, there are changes in the verification functions for each CircuitGate, which now include a row parameter that is used to index the witness (as it was used before, but now the row is no longer a field in CircuitGate, but another parameter in those functions). This provides a less prone error instantiation of tests, as we don't need to keep track of the row index for each gate we create (in fact, I think we can still remove some variables such as first_row, last_row that appear in some tests, but I kept them for now until this change is accepted).

### DIFF
--- a/circuits/kimchi/src/gate.rs
+++ b/circuits/kimchi/src/gate.rs
@@ -368,7 +368,7 @@ impl GateType {
 pub struct CircuitGate<F: FftField> {
     /// row position in the circuit
     // TODO(mimoo): shouldn't this be u32 since we serialize it as a u32?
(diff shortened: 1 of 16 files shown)


mimoo (changes requested):
This looks great! Looks like you still have some issues passing CI (e.g. `cargo fmt`)

mimoo (approved):
LGTM!

PR #238, opened 2021-12-02
Implemented extensible Schnorr signer

### LINKED ISSUE #237: Schnorr signer crate
There are a couple projects requiring a rust mina signer: websnapps and chainsafe

In websnapps the TLS notary needs to sign notarizations. Ideally, these should use the standard mina schnorr signature algorithm.

Instead of using the c reference signer (which was never intended to be a library) and would require many changes, we should implement a library for signing transactions and notarizations.

Since Chainsafe's mina-rust implementation also needs a mina signature library it makes the most sense to do this in rust so that both projects can reuse the same implementation.

More details: https://github.com/MinaProtocol/mina/issues/9594

jspada:
* Mina's Schnorr signature scheme
    * Random oracle input serializer
    * Hashable and Signable interfaces
    * Standard signature unit tests

Closes #237


signer/README.md
```diff
@@ -0,0 +1,82 @@
...
+This crate provides an API and framework for Mina signing.  It follows the algorithm outlined in the [Mina Signature Specification](https://github.com/MinaProtocol/mina/blob/develop/docs/specs/signatures/description.md).
+
+## Simple interface
+
+The simple interface uses the default signer configuration compatible with mainnet and testnet transaction signatures.
+
+```rust
+use rand;
+use mina_signer::{NetworkId, Keypair, Signer};
+use mina_signer::NetworkId;
```
mimoo, line 12: importing NetworkId twice

signer/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+let mut ctx = mina_signer::create(NetworkId::TESTNET);
+let sig = ctx.sign(key_pair, transaction);
+
+assert_eq!(ctx.verify(sig, key_pair.public, transaction), true);
```
mimoo, line 17: you can just write `assert!(ctx.verify(sig, key_pair.public, transaction))`

signer/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+The advanced interface allows specification of an alternative cryptographic sponge and parameters, for example, in order to create signatures that can be verified more efficiently using the Kimchi proof system.
+
+```rust
+use rand;
+use mina_signer::{NetworkId, Keypair, Signer};
+use oracle::{pasta, poseidon};
+
+let mut ctx = mina_signer::custom::<poseidon::PlonkSpongeConstants5W>(
+    pasta::fp5::params(),
```
mimoo, line 30: why is poseidon stuff being passed twice, once as generic param and once as argument?

signer/README.md
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl Hashable for Foo {
+    fn to_roinput(self) -> ROInput {
+        let mut roi = ROInput::new();
+
+        roi.append_u32(self.foo);
+        roi.append_u64(self.bar);
+
+        roi
```
mimoo, line 60: did you create an issue to use serde for that :o ? This manual redefinition of the struct really scares me as we change these structs all the time

signer/README.md
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+impl Signable for Foo {
+    fn domain_string(network_id: NetworkId) -> &'static str {
+       match network_id {
+           NetworkId::MAINNET => "FooSigMainnet",
+           NetworkId::TESTNET => "FooSigTestnet",
```
mimoo, line 68: it feels a bit tedious to define a domain separation string for each network, why is this needed? Aren't we using the network id elsewhere to create different hash functions?

signer/src/domain.rs
```diff
@@ -0,0 +1,134 @@
...
+//!
+//! Shorthands and helpers for base and scalar field elements
+
+use ark_ec::AffineCurve;
+use ark_ff::PrimeField; // for into_repr()
+
+use mina_curves::pasta::pallas as Pallas;
+
+/// Affine curve point type
+pub use Pallas::Affine as CurvePoint;
```
mimoo, line 11: why re-export this type?

signer/src/domain.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Base field element type
+pub type BaseField = <CurvePoint as AffineCurve>::BaseField;
+/// Scalar field element type
+pub type ScalarField = <CurvePoint as AffineCurve>::ScalarField;
+
+use ark_serialize::CanonicalSerialize;
```
mimoo, line 17: move all the `use` above :o

signer/src/domain.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Field element helpers
+pub trait FieldHelpers<F: PrimeField> {
```
mimoo, line 20: can you move this to o1-utils?

signer/src/signature.rs
```diff
@@ -0,0 +1,33 @@
+//! Mina signature structure and associated helpers
+
+use std::fmt;
+
+use crate::{BaseField, FieldHelpers, ScalarField};
+
+/// Signature structure
+#[derive(Clone, Copy, Eq, fmt::Debug, PartialEq)]
+pub struct Signature {
+    /// Field component
```
mimoo, line 10: you mean _base_ field component

signer/src/signature.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub rx: BaseField,
+
+    /// Scalar component
+    pub s: ScalarField,
```
mimoo, line 14: do these fields need to be public outside of the crate? If not you can set `pub(crate)`

signer/src/seckey.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use crate::ScalarField;
+
+/// Secret key
+#[derive(Clone, Copy, PartialEq, Eq)] // No Debug nor Display
+pub struct SecKey(ScalarField);
+
+impl SecKey {
+    /// Create a secret key from scalar field element
+    pub fn new(scalar: ScalarField) -> Self {
```
mimoo, line 11: no function to generate one randomly?

signer/src/seckey.rs
```diff
@@ -0,0 +1,19 @@
...
+        Self(scalar)
+    }
+
+    /// Convert secret key into scalar field element
+    pub fn to_scalar(self) -> ScalarField {
```
mimoo, line 16: since this consumes self, this should be called `into_scalar` (rust convention)

signer/src/lib.rs
```diff
@@ -0,0 +1,173 @@
+#![warn(missing_docs)]
+
+//! Mina signer library for verification and signing
```
mimoo, line 3: btw, if you don't want to repeat yourself between the README and the doc here, you can also include some of the README content with this: https://github.com/o1-labs/proof-systems/blob/master/ocaml/ocaml-gen/src/lib.rs#L3

signer/src/keypair.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn rand(rng: &mut (impl RngCore + CryptoRng)) -> Self {
+        let secret: ScalarField = ScalarField::rand(rng);
+        let public: CurvePoint = CurvePoint::prime_subgroup_generator()
+            .mul(secret)
+            .into_affine();
+
+        Keypair {
+            secret: SecKey::new(secret),
+            public: PubKey::new(public),
+        }
```
mimoo, line 40: or you can call `Self::new(secret, public)`

signer/src/keypair.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::{CurvePoint, FieldHelpers, PubKey, ScalarField, SecKey};
+use ark_ec::{AffineCurve, ProjectiveCurve};
+use ark_ff::UniformRand;
+use rand::{self, CryptoRng, RngCore};
+
+/// Keypair structure
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct Keypair {
+    /// Secret key
+    pub secret: SecKey,
```
mimoo, line 16: any reason this is `pub`?

README.md
```diff
@@ -16,5 +16,6 @@ proof-systems/
 ├── groupmap/ # TODO: description
 ├── ocaml-gen/ # ocaml bindings generator tool
 ├── oracle/ # implementation of the poseidon hash function
+├── signer/ # implementation of schnorr signature scheme
```
mimoo, line 19: 👍

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+use std::ops::Neg;
+
+use crate::{BaseField, CurvePoint, FieldHelpers};
+
+/// Length of Mina addresses
+pub const MINA_ADDRESS_LEN: usize = 55;
+
+/// Public key
+#[derive(Copy, Clone, fmt::Debug, PartialEq, Eq)]
+pub struct PubKey(CurvePoint);
```
mimoo, line 18: 👍

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+use sha2::{Digest, Sha256};
+use std::ops::Neg;
+
+use crate::{BaseField, CurvePoint, FieldHelpers};
+
+/// Length of Mina addresses
+pub const MINA_ADDRESS_LEN: usize = 55;
+
+/// Public key
+#[derive(Copy, Clone, fmt::Debug, PartialEq, Eq)]
```
mimoo, line 17: `Debug` is already in scope (no need for `fmt::`)

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+pub struct PubKey(CurvePoint);
+
+impl PubKey {
+    /// Create a public key from curve point
+    pub fn new(point: CurvePoint) -> Self {
```
mimoo, line 22: is it important to do any validation check at this point? what if that pubkey is untrusted?

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn from_address(address: &str) -> Result<Self, &'static str> {
+        if address.len() != MINA_ADDRESS_LEN {
+            return Err("Invalid address length");
+        }
+
+        let bytes = bs58::decode(address)
+            .into_vec()
+            .map_err(|_| "Invalid address encoding")?;
+
+        let (raw, checksum) = (&bytes[..bytes.len() - 4], &bytes[bytes.len() - 4..]);
```
mimoo, line 36: if bs58 is variable-length encoding, don't you want to check the length of `bytes` instead of `address`? As bytes might be different length eventhough address is MINA_ADDRESS_LEN?

signer/src/pubkey.rs
mimoo: also it feels like this function should be renamed `from` or something with `from_` as this is external input

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+            return Err("Invalid address version info");
+        }
+
+        let x = BaseField::from_bytes(x_bytes).map_err(|_| "invalid x-coordinate bytes")?;
+        let mut pt =
+            CurvePoint::get_point_from_x(x, y_parity).ok_or("Invalid address x-coordinate")?;
+
+        if pt.y.into_repr().is_even() == y_parity {
+            pt.y = pt.y.neg();
+        }
```
mimoo, line 57: this public point verification logic should live in Pubkey::new

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        Ok(PubKey::new(pt))
+    }
+
+    /// Convert public key into curve point
+    pub fn to_point(self) -> CurvePoint {
```
mimoo, line 63: `into_point`

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+        self.0
+    }
+
+    /// Convert public key into compressed public key
+    pub fn to_compressed(self) -> CompressedPubKey {
```
mimoo, line 68: `into_compressed`

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+        let point = self.to_point();
+        CompressedPubKey {
+            x: point.x,
+            is_odd: !point.y.into_repr().is_even(),
```
mimoo, line 72: why not use `is_odd`?

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    /// Serialize public key into corresponding Mina address
+    pub fn to_address(self) -> String {
```
mimoo, line 77: into_adress

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
...
+#[derive(Clone, Copy)]
+pub struct CompressedPubKey {
+    /// X-coordinate
+    pub x: BaseField,
+
+    /// Parity of y-coordinate
+    pub is_odd: bool,
+}
+
+fn to_address(x: BaseField, is_odd: bool) -> String {
```
mimoo, line 105: is there a rationale for implementing this outside of the `impl PublicKey`?

signer/src/roinput.rs
```diff
@@ -0,0 +1,867 @@
...
+///         roi
+///     }
+/// }
+/// ```
+/// **Details:** For technical reasons related to our proof system and performance, fields are
+/// serialized for signing differently than other types. Additionally, during signing all members
+/// of the random oracle input get serialized together in two different ways: both as *bytes* and
+/// as a vector of *field elements*.  The random oracle input encapsulates and automates this
+/// complexity.
+
```
mimoo, line 47: remove space

signer/src/roinput.rs
```diff
@@ -0,0 +1,867 @@
...
+            acc
+        });
+
+        bits.extend(&self.bits);
+
+        bits.into()
+    }
+
+    /// Serialize random oracle input to vector of base field elements
+    pub fn to_fields(&self) -> Vec<BaseField> {
```
mimoo, line 110: I think I pointed that before, but this will have collisions if you don't include the length of bits somewhere

signer/src/schnorr.rs
```diff
@@ -0,0 +1,156 @@
...
+        let mut bytes = format!("{:*<20}", domain_string).as_bytes().to_vec();
+        bytes.resize(32, 0);
+
+        bytes
+    }
+
+    // This function uses a cryptographic hash function to create a uniformly and
+    // randomly distributed nonce.  It is crucial for security that no two different
+    // messages share the same nonce.
+    fn blinding_hash<H>(&self, kp: &Keypair, input: H) -> ScalarField
```
mimoo, line 98: can you call this something else? like `derive_nonce` or something :D?

signer/src/schnorr.rs
```diff
@@ -0,0 +1,156 @@
...
+        bytes[bytes.len() - 1] &= 0b0011_1111;
+
+        ScalarField::from_random_bytes(&bytes[..]).expect("failed to create scalar from bytes")
+    }
+
+    // This function uses a cryptographic hash function (based on a sponge construction) to
+    // convert the message to be signed (and some other information) into a uniformly and
+    // randomly distributed scalar field element.  It uses Mina's variant of the Poseidon
+    // SNARK-friendly cryptographic hash function.
+    // Details: <https://github.com/o1-labs/cryptography-rfcs/blob/httpsnapps-notary-signatures/mina/001-poseidon-sponge.md>
```
mimoo, line 127: you should still use `///` btw, even on internal functions (as we can export doc for that too)

signer/tests/notarization.rs
```diff
@@ -0,0 +1 @@
+
```
mimoo, line 1: delete this file

signer/src/schnorr.rs
```diff
@@ -0,0 +1,156 @@
...
+
+use crate::{
+    BaseField, CurvePoint, FieldHelpers, Hashable, Keypair, NetworkId, PubKey, ROInput,
+    ScalarField, Signable, Signature, Signer,
+};
+
+/// Schnorr signer context for the Mina signature algorithm
+///
+/// For details about the signature algorithm please see [crate::schnorr]
+pub struct Schnorr<SC: SpongeConstants> {
```
mimoo, line 35: `struct SignerCtx` perhaps?

signer/src/schnorr.rs
```diff
@@ -0,0 +1,156 @@
...
+        roi.append_bytes(&[self.network_id.into()]);
+
+        hasher.update(roi.to_bytes());
+
+        let mut bytes = [0; 32];
+        hasher.finalize_variable(|out| bytes.copy_from_slice(out));
+        // Drop the top two bits to convert into a scalar field element
+        //   N.B. Since the order p is very close to 2^m for some m, truncating only creates
+        //   a tiny amount of bias that should be insignificant and keeps the implementation
+        //   simple by avoiding reduction modulo p.
```
mimoo, line 117: two other things you can point out:

- that this is true for Pallas' scalar field at least (not necessarily true if we change curve)
- that doing a modular reduction would be worse

signer/src/schnorr.rs
```diff
@@ -0,0 +1,156 @@
...
+        S: Signable,
+    {
+        let mut roi: ROInput = input.to_roinput();
+        roi.append_field(pub_key.to_point().x);
+        roi.append_field(pub_key.to_point().y);
+        roi.append_field(rx);
+
+        // Set sponge initial state (explicitly init state so signer context can be reused)
+        // N.B. Mina sets the sponge's initial state by hashing the input's domain bytes
+        self.sponge.state = vec![BaseField::zero(); self.sponge.state.len()];
```
mimoo, line 139: can you implement a `reset` function on `ArithmeticSponge` instead?

signer/src/schnorr.rs
```diff
@@ -0,0 +1,156 @@
...
+        self.sponge.sponge_state = SpongeState::Absorbed(0);
+        self.sponge
+            .absorb(&[
+                BaseField::from_bytes(&Schnorr::<SC>::domain_bytes::<S>(self.network_id))
+                    .expect("invalid domain bytes"),
+            ]);
+        self.sponge.squeeze();
+
+        // Absorb random oracle input
+        self.sponge.absorb(&roi.to_fields());
```
mimoo, line 149: do you think it'd make sence to have `to_fields()` consume `roi` so you can only call it once?

mimoo: I really don't like the ROI approach btw because it feels prone to misuse, it looks like we're just shoving bits and field elements together and there might be collisions or ambiguity in the way we do this.

I think a better design would do this:

* absorb a field just absorbs a field element
* if you absorb bits/bytes, it caches them in a queue (and also prepends the length of the bits/bytes every time you call these functions in the queue as well)
* if you absorb bits/bytes when the queue gets bigger than the size of a field element, you convert some of the queue to a field element
* if you absorb field (or call `to_fields()`) when the bits queue is not empty, then you first convert your bit queue into field, stick that into the field array, and then absorb your field

mimoo (approved):
good job on that implementation! Left a bunch of comments :)

signer/README.md
jspada: This is just from how proof-systems sponge works.  They are two different things.  The generic param is the `SpongeConstants`, which configures the sponge capacity, rate, rounds, s-box alpha, width and a couple of other things.  The argument is the `ArithmeticSpongeParams` which just contains the round constants and MDS matrix.  If the sponge, it's consts and params are ever refactored into a single parameter then we can change this API to just use the single thing.  For now, unfortunately they come as two things from two different places.

signer/README.md
jspada: Unfortunately there's no network id in the signed data, which is why mina does this.  There are some interesting historic reasons for this.  We discussed improving this previously, but it would entail a bunch of 3rd parties having to update their implementations (perhaps some are even on custom hardware), which was not ideal.

signer/README.md
jspada: Yes.  https://github.com/o1-labs/proof-systems/issues/241

Agreed!

For now this should be OK because for tx signatures it doesn't change.  For backward compatibility, each version of tx/signatures is a new thing (i.e. new tx version and structure), rather than updating the old one.

Anyway, it's way better to do it automatically and I will try to do this with subsequent ticket 241 :-)

signer/src/domain.rs
jspada: Yes, I plan to do this in this ticket https://github.com/o1-labs/proof-systems/issues/242

signer/src/signature.rs
jspada: Y

signer/src/signature.rs
jspada: Some things in another crate (the tests and the websnapps prover) need to access them individually.   Seems like something that should be public.

signer/src/seckey.rs
jspada: Wish granted :-)

signer/src/lib.rs
```diff
@@ -0,0 +1,173 @@
+#![warn(missing_docs)]
```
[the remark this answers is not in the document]
jspada: 👍🏼

signer/src/lib.rs
jspada: rust is awesome

signer/src/pubkey.rs
```diff
@@ -0,0 +1,152 @@
+//! Public key structures and algorithms
+//!
+//! Definition of public key structure and helpers
+
+use ark_ff::{BigInteger, PrimeField};
+use bs58;
```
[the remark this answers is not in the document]
jspada: haha!

signer/src/keypair.rs
jspada: Code in schnorr.rs needs access to secret.

signer/README.md
mimoo: I think this should be documented in the trait, to make sure people choose different domain string depending on the network

signer/src/keypair.rs
jspada: I'll `pub(crate)` it

mimoo: Ideally this should not be public IMO. Two things you could do:

* make it public only to the crate with `pub(crate)`, but you can still write code that will extract the secret key within the crate
* make it public only to this file, by removing `pub` and implementing signing on `Keypair`.

signer/src/pubkey.rs
jspada: AFAIK MINA_ADDRESS_LEN is on the b58 encoding, but I can add an extra check for bytes.

signer/src/pubkey.rs
jspada: So it can be shared by `PubKey` and `CompressedPubKey`.

signer/src/roinput.rs
jspada: Yes, but I can't change this or it will not be compatible with existing mainnet/testnet signatures.  If we want to fix this we should introduce tx version 2 and definitely pad with the length.

signer/src/schnorr.rs
jspada: how about `nonce_hash` ?

signer/src/schnorr.rs
jspada: Schnorr implements Signer interface, so it can be called whatever it wants as long as it implements the Signer interface.  I thought this is nice, so we can at this level just call things what they are under the hood, which isn't really visible to the user of the API

signer/src/schnorr.rs
jspada: Agreed - can this be part of this PR?

signer/src/schnorr.rs
jspada: 👍🏼

signer/src/schnorr.rs
jspada: I agree with your suggestions, but I think this should be done in another ticket.  The reason is because if we cannot make changes to this without breaking compatibility with mainnet mina transaction signatures, which is one of my main objectives (to be the same as the OCaml code).  That said, these are very good suggestions -- especially the length-padding part.  Thus, I've opened up a separate ticket for improving and simplifying the ROI structure and length padding things.  https://github.com/o1-labs/proof-systems/issues/244

signer/src/pubkey.rs
jspada: Good one!  Agreed -- thanks.

mrmr1993 (approved):

oracle/src/poseidon.rs
```diff
@@ -84,6 +84,7 @@ pub trait Sponge<Input: Field, Digest> {
     fn new(params: ArithmeticSpongeParams<Input>) -> Self;
     fn absorb(&mut self, x: &[Input]);
     fn squeeze(&mut self) -> Digest;
+    fn reset(&mut self);
```
mimoo, line 87: missing some doc ;)

PR #239, opened 2021-12-02
Remove unnecessary lookup_table_lengths field in index

mrmr1993:
This PR removes `lookup_table_lengths` from the index, and tweaks the construction of the sorted table so that it handles duplicate values correctly.

This PR also fixes the `ValueNotInTable` error: this error now triggers when a looked-up value is absent from the source table, where previously it only fired in an edge-case.

### DIFF
--- a/circuits/kimchi/src/polynomials/lookup.rs
+++ b/circuits/kimchi/src/polynomials/lookup.rs
@@ -336,7 +336,7 @@ pub fn sorted<
 
     let sorted = {
         let mut sorted: Vec<Vec<E>> =
-            vec![Vec::with_capacity(lookup_rows + 1); max_lookups_per_row];
+            vec![Vec::with_capacity(lookup_rows + 1); max_lookups_per_row + 1];
 
         let mut i = 0;
         for t in lookup_table().take(lookup_rows) {
--- a/circuits/kimchi/src/polynomials/lookup.rs
+++ b/circuits/kimchi/src/polynomials/lookup.rs
@@ -25,6 +25,100 @@
 //! last element of LookupSorted(i) = last element of LookupSorted(i + 1),
 //! and if i % 2 = 1, we enforce that the
 //! first element of LookupSorted(i) = first element of LookupSorted(i + 1)
+//!
+//! Overview of the protocol
+//! ========================
+//! * We have our initial table `lookup_table`, with our desired values listed.
+//! * We have the implicit table `lookups(witness)` representing the values looked up in each row
+//!   of the witness.
+//!   - This table is initially variable-width, where some rows have no lookups, and others have
+//!     several.
+//!   - We explicitly compute this table, and where the width for a particular row is less than the
+//!     maximum width, we insert a 'dummy' lookup value as many times as we need to to give every
+//!     row the same number of lookups.
+//!   - We'll call this padded table `witness_lookups`.
+//! * We want to generate a `sorted_table` that contains every entry from the concatenated table
+//! `lookup_table||witness_lookups`, where values are in the same order as `lookup_table`, with all
+//! duplicates placed next to each other.
+//!   - There's an edge case around duplicate values in the `lookup_table` itself: these should
+//!     appear in `sorted_table` at least once each time they appeared in the `lookup_table`.
+//!   - This ensures that, for any `beta` and for each `i`, the pair `lookup_table[i] + beta *
+//!     lookup_table[i+1]` corresponds to some distinct `j` such that `sorted_table[j] + beta *
+//!     sorted_table[j+1]`.
+//!   - For all other values of `j`, `sorted_table[j] = sorted_table[j+1]`: since we've dealt with
+//!     all of the 'different' pairs corresponding from moving from one value in `lookup_table` to
+//!     the next, the only remaining pairs are those corresponding to the duplicates provided by the
+//!     lookups in `witness_lookups`.
+//!   - For example, if `lookup_table` is `[0, 1, 2, 3, 4, 5]` and `witness_lookups` is
+//!     `[0, 0, 0, 2, 2, 4]`, then `sorted_table` is `[0, 0, 0, 0, 1, 2, 2, 2, 3, 4, 4, 5]`, and
+//!     the differences are
+//!     `[(0, 0), (0, 0), (0, 0), (0, 1), (1, 2), (2, 2), (2, 2), (2, 3), (3, 4), (4, 4), (4, 5)]`.
+//!     The entries where the pairs are different are those that match with the `lookup_table`, and
+//!     the equal pairs can be paired with the `witness_lookups`. This `sorted_table` is computed
+//!     by the `sorted` function.
+//! * in order to check the multiset inclusion, we calculate the product over our sorted table:
+//!   `gamma * (1 + beta) + sorted_table[i] + beta * sorted_table[i+1]`
+//!   - again, when the adjacent terms `sorted_table[i]` and `sorted_table[i+1]` are equal, this
+//!     simplifies to `(gamma + sorted_table[i]) * (1 + beta)`
+//!   - when they are different, there is some `j` such that it equals `gamma * (1 + beta) +
+//!     lookup_table[i] + beta * lookup_table[i+1]`
+//!   - using the example above, this becomes
+//!     ```
+//!         gamma * (1 + beta) + 0 + beta * 0
+//!       * gamma * (1 + beta) + 0 + beta * 0
+//!       * gamma * (1 + beta) + 0 + beta * 0
+//!       * gamma * (1 + beta) + 0 + beta * 1
+//!       * gamma * (1 + beta) + 1 + beta * 2
+//!       * gamma * (1 + beta) + 2 + beta * 2
+//!       * gamma * (1 + beta) + 2 + beta * 2
+//!       * gamma * (1 + beta) + 2 + beta * 3
+//!       * gamma * (1 + beta) + 3 + beta * 4
+//!       * gamma * (1 + beta) + 4 + beta * 4
+//!       * gamma * (1 + beta) + 4 + beta * 5
+//!     ```
+//!     which we can simplify to
+//!     ```
+//!         (gamma + 0) * (1 + beta)
+//!       * (gamma + 0) * (1 + beta)
+//!       * (gamma + 0) * (1 + beta)
+//!       * gamma * (1 + beta) + 0 + beta * 1
+//!       * gamma * (1 + beta) + 1 + beta * 2
+//!       * (gamma + 2) * (1 + beta)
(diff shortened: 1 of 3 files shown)


circuits/kimchi/src/polynomials/lookup.rs
```diff
@@ -314,33 +313,41 @@ pub fn sorted<
     let by_row = lookup_info.by_row(gates);
     let max_lookups_per_row = lookup_info.max_per_row;
 
+    for t in lookup_table().take(lookup_rows) {
+        // Don't multiply-count duplicate values in the table, or they'll be duplicated for each
+        // duplicate! E.g. 3 values the same would insert the value 3 times for each instance.
```
mimoo, line 318: "3 values the same" -> typo?

mrmr1993: It's what I meant, but perhaps it could be clearer? If the same value `x` were to appear 3 times in the lookup table, we would count the 3, and then insert 3 `x`s for each time we encounter `x` in the table (which is 3 times). This comes into play with the dummy value, since we now iterate over the whole table and encounter it multiple times.

Any suggestions on how to express it more clearly?

mimoo: I'm actually not sure what we're trying to do exactly here. Might have to spend some time reverse engineering the protocol we're implementing :o but in english you meant "3 of the same values" no?

circuits/kimchi/src/polynomials/lookup.rs
```diff
@@ -314,33 +313,41 @@ pub fn sorted<
...
+        counts.entry(t).or_insert(1);
+    }
+
     for (i, row) in by_row.iter().enumerate().take(lookup_rows) {
         let spec = row;
         let padding = max_lookups_per_row - spec.len();
         for joint_lookup in spec.iter() {
             let table_entry = E::evaluate(&params, joint_lookup, witness, i);
```
mimoo, line 326: off-scope for your PR, but I think we ought to specify what all these terms mean. Here `table_entry` makes you think that it's an entry in the table, but it's not, it's a joint tuple of values you want to lookup in the table.

circuits/kimchi/src/polynomials/lookup.rs
```diff
@@ -314,33 +313,41 @@ pub fn sorted<
...
-    for t in lookup_table().take(lookup_rows) {
-        let count = counts.entry(t).or_insert(0);
-        *count += 1;
-    }
-
     let sorted = {
         let mut sorted: Vec<Vec<E>> = vec![];
         for _ in 0..max_lookups_per_row + 1 {
             sorted.push(Vec::with_capacity(lookup_rows + 1))
         }
```
mimoo, line 339: that should be better:

```rust
let mut sorted = vec![Vec::with_capacity(lookup_rows + 1); max_lookups_per_row + 1];
```

circuits/kimchi/src/polynomials/lookup.rs
```diff
@@ -314,33 +313,41 @@ pub fn sorted<
...
-                Some(x) => *x,
+        for t in lookup_table().take(lookup_rows) {
+            let t_count = match counts.get_mut(&t) {
+                None => panic!("Value has disappeared from count table"),
+                Some(x) => {
+                    let res = *x;
+                    // Reset the count, any duplicate values should only appear once from now on.
+                    *x = 1;
+                    res
+                }
```
mimoo, line 350: this is not doing the same thing though :o now you're going through `lookup_rows` instead of `lookup_table_entries` (which is 4 for XOR)

mimoo, line 350: OK so I spent like 5 hours reverse engineering what's happening here xD I think I get it now. Essentially, we're just concatenating the values of `lookup_table` and `joint_lookup(witness)`. Then snake-ifying the table (and repeating the element linking each columns). At this point I'm willing to accept that this is working if the tests pass, but I didn't manage to convince myself that this is correct.

mimoo: I think there ought to be a simpler way to implement and document this.

circuits/kimchi/src/polynomials/lookup.rs
mrmr1993: I've reworded this, hopefully it's clearer now.

circuits/kimchi/src/polynomials/lookup.rs
mrmr1993: You read here is right, but to explicitly spell out my understanding:
* We have our initial table `lookup_table`, with our desired values listed
* We have the implicit table `lookups(witness)` representing the values looked up in each row of the witness.
  - This table is initially variable-width, where some rows have no lookups, and others have several.
  - We explicitly compute this table, and where the width for a particular row is less than the maximum width, we insert a 'dummy' lookup value as many times as we need to to give every row the same number of lookups.
  - We'll call this padded table `witness_lookups`
* We want to generate a `sorted_table` that contains every entry from the concatenated table `lookup_table||witness_lookups`, where values are in the same order as `lookup_table`, with all duplicates placed next to each other
  - There's an edge case around duplicate values in the `lookup_table` itself: these should appear in `sorted_table` at least once each time they appeared in the `lookup_table`
  - This ensures that, for any `beta` and for each `i`, the pair `lookup_table[i] + beta * lookup_table[i+1]` corresponds to some distinct `j` such that `sorted_table[j] + beta * sorted_table[j+1]`.
  - For all other values of `j`, `sorted_table[j] = sorted_table[j+1]`: since we've dealt with all of the 'different' pairs corresponding from moving from one value in `lookup_table` to the next, the only remaining pairs are those corresponding to the duplicates provided by the lookups in `witness_lookups`.
  - For example, if `lookup_table` is `[0, 1, 2, 3, 4, 5]` and `witness_lookups` is `[0, 0, 0, 2, 2, 4]`, then `sorted_table` is `[0, 0, 0, 0, 1, 2, 2, 2, 3, 4, 4, 5]`, and the differences are `[(0, 0), (0, 0), (0, 0), (0, 1), (1, 2), (2, 2), (2, 2), (2, 3), (3, 4), (4, 4), (4, 5)]`. The entries where the pairs are different are those that match with the `lookup_table`, and the equal pairs can be paired with the `witness_lookups`. **This `sorted_table` is what this part of the code generates.**
* in order to check the multiset inclusion, we calculate the product over our sorted table: `gamma * (1 + beta) + sorted_table[i] + beta * sorted_table[i+1]`
  - again, when the adjacent terms `sorted_table[i]` and `sorted_table[i+1]` are equal, this simplifies to `(gamma + sorted_table[i]) * (1 + beta)`
  - when they are different, there is some `j` such that it equals `gamma * (1 + beta) + lookup_table[i] + beta * lookup_table[i+1]`
  - using the example above, this becomes
    ```
        gamma * (1 + beta) + 0 + beta * 0
      * gamma * (1 + beta) + 0 + beta * 0
      * gamma * (1 + beta) + 0 + beta * 0
      * gamma * (1 + beta) + 0 + beta * 1
      * gamma * (1 + beta) + 1 + beta * 2
      * gamma * (1 + beta) + 2 + beta * 2
      * gamma * (1 + beta) + 2 + beta * 2
      * gamma * (1 + beta) + 2 + beta * 3
      * gamma * (1 + beta) + 3 + beta * 4
      * gamma * (1 + beta) + 4 + beta * 4
      * gamma * (1 + beta) + 4 + beta * 5
    ```
    which we can simplify to
    ```
        (gamma + 0) * (1 + beta)
      * (gamma + 0) * (1 + beta)
      * (gamma + 0) * (1 + beta)
      * gamma * (1 + beta) + 0 + beta * 1
      * gamma * (1 + beta) + 1 + beta * 2
      * (gamma + 2) * (1 + beta)
      * (gamma + 2) * (1 + beta)
      * gamma * (1 + beta) + 2 + beta * 3
      * gamma * (1 + beta) + 3 + beta * 4
      * (gamma + 4) * (1 + beta)
      * gamma * (1 + beta) + 4 + beta * 5
    ```
* because we said before that each pair corresponds to either a pair in the `lookup_table` or a duplicate from the `witness_table`, the product over the sorted table should equal the product of `gamma * (1 + beta) + lookup_table[i] + beta * lookup_table[i+1]` multiplied by the product of `(gamma + witness_table[i]) * (1 + beta)`, since each term individually cancels out.
  - using the example above, the `lookup_table` terms become
    ```
        gamma * (1 + beta) + 0 + beta * 1
      * gamma * (1 + beta) + 1 + beta * 2
      * gamma * (1 + beta) + 2 + beta * 3
      * gamma * (1 + beta) + 3 + beta * 4
      * gamma * (1 + beta) + 4 + beta * 5
    ```
    and the `witness_table` terms become
    ```
        (gamma + 0) * (1 + beta)
      * (gamma + 0) * (1 + beta)
      * (gamma + 0) * (1 + beta)
      * (gamma + 2) * (1 + beta)
      * (gamma + 2) * (1 + beta)
      * (gamma + 4) * (1 + beta)
    ```

There is some nuance around table lengths; for example, notice that `witness_table` need not be the same length as `lookup_table` (and indeed is not in our implementation, due to multiple lookups per row), and that `sorted_table` will always be longer than `lookup_table`, which is where we require 'snakifying' to check consistency. Happily, we don't have to perform snakifying on `witness_table`, because its contribution above only uses a single term rather than a pair of terms.

mimoo: This is great! Thank-you so much for the write up :)

Would you mind copy/pasting this into the module doc of a `plookup/mod.rs` or `polynomials/plookup.rs`? I think this is very valuable information :o 

> There's an edge case around duplicate values in the lookup_table itself: these should appear in sorted_table at least once each time they appeared in the lookup_table

Why would a lookup table have duplicate values though? The only thing I can think of is if the challenge is zero or if we're talking about different lookup tables

mrmr1993: I've placed my comment (with slight modification) as exposition at the top of this `lookup.rs` file.

> Why would a lookup table have duplicate values though?

If the lookup table doesn't fill the entire `n - ZK_ROWS - 1` available in the polynomial, we need to fill the gaps with some value, but we don't want to introduce a new value (since that could be used for unintended lookups). Instead, we pad the table to the end with a 'dummy' value selected from one of the existing values. However, this means that that value appears `n - ZK_ROWS - 1 - len(table)` times, so we need to be careful that these duplicates aren't either erased or multiply-counted.

circuits/kimchi/src/polynomials/lookup.rs
mrmr1993: I've renamed `table_entry` to `joint_lookup_evaluation`. Hopefully that's clearer?

mimoo (approved):
nice one!

PR #248, opened 2021-12-09
[kimchi] better benchmark

mimoo:
Hopefully this will allow benchmarking to work in CI as well. The benchmark is still not great for reasons listed in https://github.com/o1-labs/proof-systems/issues/249

### DIFF
--- a/dlog/kimchi/src/bench.rs
+++ b/dlog/kimchi/src/bench.rs
@@ -1,4 +1,7 @@
-use crate::{index::Index, prover::ProverProof};
+use crate::{
+    index::{Index, VerifierIndex},
+    prover::ProverProof,
+};
 use ark_ff::{One, UniformRand, Zero};
 use ark_poly::{univariate::DensePolynomial, UVPolynomial};
 use array_init::array_init;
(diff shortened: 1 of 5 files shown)


dlog/kimchi/src/bench.rs
```diff
@@ -84,35 +107,33 @@ pub fn proof(num: usize) {
...
         // previous opening for recursion
         let prev = {
-            let k = ceil_log2(index.srs.g.len());
+            let k = ceil_log2(self.index.srs.g.len());
             let chals: Vec<_> = (0..k).map(|_| Fp::rand(rng)).collect();
             let comm = {
                 let coeffs = b_poly_coefficients(&chals);
                 let b = DensePolynomial::from_coefficients_vec(coeffs);
-                index.srs.commit_non_hiding(&b, None)
+                self.index.srs.commit_non_hiding(&b, None)
```
jspada, line 115: does it make sense to also have benchmark with hiding commitments?

jspada (approved):

dlog/kimchi/src/bench.rs
mimoo: I'm ignoring the recursion stuff

PR #250, opened 2021-12-13
RFC 0001: Extended lookup tables

mrmr1993:
This RFC proposes an extension to our use of lookup tables using the PLOOKUP multiset inclusion argument, so that values within lookup tables can be chosen after the constraint system for a circuit has been fixed.


jspada:
I'm wondering if the RFC should go into the https://github.com/o1-labs/cryptography-rfcs repo instead?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,495 @@
+## Summary
```
jspada, line 1: I'm wondering if the RFC should go into the https://github.com/o1-labs/cryptography-rfcs repo instead?

mimoo: haven't had the time to go through it yet, but I've put mine on https://o1-labs.github.io/mina-book/crypto/plonk/maller_15.html because I could write LaTeX there.

querolita: Don't know about you guys, but I get a 404 error when trying to access [https://github.com/o1-labs/cryptography-rfcs](https://github.com/o1-labs/cryptography-rfcs)

querolita: Not sure about you, but I get a 404 error when trying to access the link [ https://github.com/o1-labs/cryptography-rfcs ](https://github.com/o1-labs/cryptography-rfcs )

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+will reference as `FixedTable(i)` for `0 <= i < W` and `TableID`.
+
+### Representing the runtime table
+
+The runtime table can be considered as a 'mask' that we apply to the fixed
+table to introduce proof-specific data at runtime.
+
+We make the simplifying assumption that an entry in the runtime table has
+exactly 1 'fixed' entry, which can be used for indexing (or set to `0` if
+unused), and that the table has a fixed width of `X`. We can pad any narrower
```
jspada, line 157: Is the width the number of columns?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+entries in the table to the full width `X` in the same way as the fixed tables
+above.
+
+We represent the masked values in the runtime table as `RuntimeTable(i)` for
+`1 <= i < X`.
+
+To specify whether a runtime table entry is applicable in the `i`th row, we use
+a selector polynomial `RuntimeTableSelector`. In order to reduce the polynomial
+degree of later checks, we add a constraint to check that the runtime entry is
+the zero vector wherever the selector is zero:
```
jspada, line 167: an application of the Lagrange technique from Plonk?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+SideLoadedCombined(i)
+  = sum[j=1, ?, SideLoadedTable(i, j) * joint_combiner^(j-1)]
+```
+and check for consistency of the evaluations and openings when verifying the
+proof.
+
+The `SideLoadedTable` polynomials are included in the proof, but we do so 'as
+if' they had been part of the verification index (ie. without any
+proof-specific openings), so that they may be reused across multiple different
+proofs without modification.
```
jspada, line 223: Would be nice to have a note about the size of this extra proof data.  For example, as a function of number of tables, max table size, etc?  also how many extra commitments and evaluations this adds could be useful

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+
+The `joint_combiner` should be sampled using randomness that depends on the
+individual `SideLoadedTable(i,j)`s, to ensure that the side-loaded values
+cannot be chosen with prior knowledge `joint_combiner`.
+
+*Note: it may be useful for the verifier to pass the `SideLoadedTable(i,j)`s as
+public inputs to the proof, so that the circuit may assert that the table it
+had side-loaded was signed or otherwise correlated with a cryptographic
+commitment that it had access to. The details of this are deliberately elided
```
jspada, line 232: It would be very helpful to have a high-level use case example demonstrating the life cycle of a side-loaded table in a proof, the circuit creation, proof generation, commitment and signature of side-loaded data, and finally verification.

jspada: In the form of a motivating example to set the stage very clearly about what we are trying to do with an illustration of everything coming together on a high-level user activity

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+ =
+   LookupPermutation[prev]
+   * ((LookupTable - FixedAndRuntimeTable) * joint_combiner^(-1)
+       + gamma + x * beta * shift[7])
+   * (SideLoadedCombined(0) + gamma + x * beta * shift[8])
+   * (SideLoadedCombined(1) + gamma + x * beta * shift[9])
+   * (SideLoadedCombined(2) + gamma + x * beta * shift[10])
+   * (SideLoadedCombined(3) + gamma + x * beta * shift[11])
+```
+where `shift[0..7]` is the existing `shift` vector used for the witness
```
jspada, line 285: Is there a max number of tables limit somewhere?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+```
+Lagrange(n-ZK_ROWS) * (LookupPermutation - 1) = 0
+```
+to ensure that the permuted values cancel.
+
+Note also that the permutation argument allows for interaction **between**
+tables, as well as injection of values into the combined lookup table. As a
+result, the permutation argument can be used to 'copy' parts of one table into
+another, which is likely to be useful in examples like the 'fold' one above, if
+some data is expected to be shared between `t_in` and `t_out`.
```
jspada, line 305: Similar to our main permutation argument which only acts on the first 7 columns, is there a limit on the table columns for which the permutation argument works?  Is there any advantage in doing a similar thing here?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+
+### Permutation argument for cheap array lookups
+
+It would be useful to add data to the runtime table using the existing
```
jspada, line 309: I can imagine how to add data using a lookup, but it would be helpful to get an quick high-level intuition how to add data using permutation :-)

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+   * ((LookupTable - FixedAndRuntimeTable) * joint_combiner^(-1)
+       + gamma + x * beta * shift[7])
+   * (SideLoadedCombined(0) + gamma + x * beta * shift[8])
+   * (SideLoadedCombined(1) + gamma + x * beta * shift[9])
+   * (SideLoadedCombined(2) + gamma + x * beta * shift[10])
+   * (SideLoadedCombined(3) + gamma + x * beta * shift[11])
+   * (RuntimeTable(1) + gamma + x * beta * shift[12])
+```
+to allow interaction between any of the first 7 witness rows and the first row
+of the runtime table. Note that the witness rows can only interact with the
```
jspada, line 340: Cool

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+    * (LookupPermutation * (0 + gamma + beta * zero_sigma)
+      - (0 + gamma + beta * shift[len(shift)-1]))
+    = 0
+  ```
+* lookup permutation finalizer
+  ```
+  Lagrange(n-ZK_ROWS) * (LookupPermutation - 1) = 0
+  ```
+
+This results in 5 new checks, and the removal of 1 old check.
```
jspada, line 434: Do we only pay the cost of these when specific tables are present?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+gives the maximum flexibility for circuit designers, and each primitive has
+uses for projects that we are persuing.
+
+This design has been refined over several iterations to reduce the number of
+components exposed in the proof and the amount of computation that the verifier
+requires. The largest remaining parts represent the data itself and the method
+for combining them, where simplifying assumptions have been made to minimize
+their impact while still satisfying the design constraints.
+
+### What other designs have been considered and what is the rationale for not choosing them?
```
jspada, line 459: Do cairo / STARKs support some of this functionality?  If so, how does it compare?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+* Confirmation of the design constraints, or simplifications thereto.
+* Necessity / utility of the permutation argument.
+
+### What parts of the design do you expect to resolve through the implementation of this feature before merge?
+
+None, this RFC has been developed in part from an initial design and iteration
+on implementations of parts of that design.
+
+### What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+
```
jspada, line 500: What design choices for this system or it's parameters could be impacted by the design of these future use-cases?

jspada (approved):

jspada:
Very good RFC.  A couple of high-level thoughts.
* Could it be useful for runtime tables to have `ZK` elements?
* Could `range-queries` on tables be an demanding use-case?
* A `recorded presentation` on this design would be very helpful
* Would be good to have a `summary table of features, limitations and parameters` (even if they aren't decided).  I think the footer already sortof has this.

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+These goals support 5 major use-cases:
+* allow the verifier circuit to support 'custom' or user-specified gates
+  - using array indexing, the use of constants and commitments/evaluations can
+    be determined by the verifier index, instead of being fixed by the
+    permutation argument as currently.
+* allow random-access lookups in arrays as a base-primitive for user programs
+* allow circuits to use low-cost branching and other familiar programming
+  primitives
+  - this depends on the development of a bytecode interpreter on top of the
+    lookup-table primitives.
```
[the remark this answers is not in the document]
jspada: It's interesting to think how this could link-up with what Anais is doing.  I wonder if there are any design considerations that could come out of her work that could impact the design here.

querolita (commented):
A few comments to improve readability and clarifications

rfcs/0001-extended-lookup-tables.md
querolita: In STARKs they use a huge table that they call the execution trace where they load all needed witness values, but it can also store flags or other elements that may be needed to represent the relation. Then each column of the table is converted into polynomials and their constraints work for both secret witnesses and the values these auxiliary columns take.

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+  - e.g. to run a small bytecode program provided in a transaction at
+    block-production time.
+* allow zkOracles to load a full HTTPS transcript without using an expensive
+  in-circuit commitment.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+In order to support the desired goals, we first define 3 types of table:
+* **fixed tables** are tables declared as part of the constraint system, and
```
querolita, line 41: So these are tables that contain public elements that somehow describe the relation to be proven. Meaning prover and verifier agree on these tables prior the proof generation?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+  are the same for every proof of that circuit.
+* **runtime tables** are tables whose contents are determined at proving time,
+  to be used for array accesses to data from the witness.
```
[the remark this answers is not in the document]
querolita: Perhaps these include derivations from the witness. Meaning, computations on the witness data that are not per-se the witness, but some values that depend on the witness (and thus should still remain secret), and you may want to store them for efficiency reasons? Ideas?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+* **side-loaded tables** are tables that are committed to in a
+  proof-independent way, so that they may be reused across proofs and/or signed
+  without knowledge of the specific proof, but may be different for each
+  execution of the circuit.
```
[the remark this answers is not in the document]
querolita: This reminds me of the commit-and-prove approach, right? So this is like some information that is committed to even before the proof system is decided. In CP-SNARKs the idea is that you have a private data, to which you commit to. Think about a commitment to your bank balance or something like that. And perhaps next month you want to prove your bank account satisfies some property. Then you can have proofs FOR that commitment, which was created before. So it brings interoperability across proof systems, but also time. In my head these tables look a lot like this.

rfcs/0001-extended-lookup-tables.md
querolita: I would say so

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+### Representing the runtime table
+
+The runtime table can be considered as a 'mask' that we apply to the fixed
+table to introduce proof-specific data at runtime.
+
+We make the simplifying assumption that an entry in the runtime table has
+exactly 1 'fixed' entry, which can be used for indexing (or set to `0` if
+unused), and that the table has a fixed width of `X`. We can pad any narrower
+entries in the table to the full width `X` in the same way as the fixed tables
+above.
```
querolita, line 159: Meaning, padding with zeros (or a different value if needed, as @jspada suggested

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+We already solve a similar problem with 'vector' tables -- tables where a row
+contains multiple values -- where we want to ensure that the prover used an
+entry `(x, y, z)` from the table, and not some parts of different entries, or
+some other value entirely. In order to do this, we generate a randomising field
+element `joint_combiner`, where the randomness depends on the circuit witness,
+and then compute `x + y * joint_combiner + z * joint_combiner^2`. Notice that,
+if the prover wants to select some `(a, b, c)` not in the table and claim that
+it in fact was, they are not able to use the knowledge of the `joint_combiner`
+to choose their `(a, b, c)`, since their choice will change the value of the
+`joint_combiner`.
```
mimoo, line 83: I really like that these RFCs are opportunities to define terms that we can use to communicate more clearly (and even reuse in the implementation!), you're doing a lot of this here which I appreciate :) I suggest adding terms to distinguish the different types of look up tables between transformation. At this point I think these two terms would be really useful (in the implementation as well):

* lookup table (perhaps shortened as LT?) -> the actual table with columns
* combined lookup table (CLT) -> the table with a single column (after using `joint_combiner`)

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+
+This is a standard technique for checking multiple equalities at once by
+checking a single equality; to use it, we must simply ensure that the
+randomness depends on all of the (non-constant) values in the equality.
+
+We propose extending each table with an extra column, which we will call the
+`TableID` column, which associates a different number with each table. Since
+the sizes of all tables must be known at constraint system time, we can treat
+this as a fixed column. For example, this converts a lookup `(a, b, c)` in the
+table with ID `1` into a lookup of `(a, b, c, 1)` in the combined table, and
```
mimoo, line 93: you mean the "non-combined" table here (or LT)

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+|0     |?runtime|
+|1     |?runtime|
+|2     |?runtime|
+
+where the `value1` entries are fixed in the constraint system. This ensure that
+a malicious prover is not free to create multiple values with the same index,
+such as `[(0,a), (0,x), (1,y), (2,z)]`, otherwise the value `a` might be used
+where the value `x` was intended.
+
+### Combining multiple tables
```
mimoo, line 66: imo diagrams would be really useful in these types of section :)

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+ensures that any value `(x, y, z)` from table ID `2` will not match that
+lookup, since its representative is (x, y, z, 2)`.
+
+To avoid any interaction between the `TableID` column and any other data in
+the tables, we use `TableID * joint_combiner^max_joint_size` to mix in the
+`TableID`, where `max_joint_size` is the maximum length of any lookup or table
+value. This also simplifies the 'joining' of the `table_id` polynomial
+commitment, since we can straightforwardly scale it by the constant
+`joint_combiner^max_joint_size` to find its contribution to every row in the
+combined table.
```
mimoo, line 103: I'm always scared of off-by-one's. What about scaling it with `joint_combiner^0 = 1`?

So the previous example would look like `2 + x * joint_combiner + y * joint_combiner^2 +
x * joint_combiner^3`

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+
+### Sampling the `joint_combiner`
+
+Currently, the `joint_combiner` is sampled using randomness that depends on the
+witness. However, where we have runtime or side-loaded tables, a malicious
+prover may be able to select values in those tables that absue knowledge of
```
mimoo, line 109: minor: I had to lookup "absue", maybe use a simpler word :D

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+type of table, so that the verifier can confirm the correct construction of the
+combined table.
+
+### Representing the combined fixed table
+
+The fixed table can be constructed at constraint system generation time, by
+extending all of the constituent tables to the same width, generating the
+appropriate table IDs array, and concatenating all of the tables. Concretely:
+
+Let `t[i][j][k]` be the `k`th element of the `j`th entry in the `i`th fixed
```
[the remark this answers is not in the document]
mimoo: +1

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+runtime or side-loaded tables, so that the values in these tables cannot be
+chosen using knowledge of the `joint_combiner`. To do this, we must use the
+commitments to the values in these tables as part of the source of randomness
+for `joint_combiner` in the prover and verifier.
+
+In particular, this means that we must have a commitment per column for each
+type of table, so that the verifier can confirm the correct construction of the
+combined table.
+
+### Representing the combined fixed table
```
mimoo, line 126: I don't think the formulas are correct in this section. I think it'd be good to have diagram to match the explanation. I made this but I'm confused as to what happens after the padding step

IMG_5275

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+This can be encoded as `W+1` polynomials in the constraint system, which we
+will reference as `FixedTable(i)` for `0 <= i < W` and `TableID`.
+
+### Representing the runtime table
+
+The runtime table can be considered as a 'mask' that we apply to the fixed
+table to introduce proof-specific data at runtime.
+
+We make the simplifying assumption that an entry in the runtime table has
+exactly 1 'fixed' entry, which can be used for indexing (or set to `0` if
```
mimoo, line 156: > an entry in the runtime table has exactly 1 'fixed' entry

you mean a fixed column no?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+
+The side-loaded tables can be considered as additional 'masks' that we apply to
+the combined fixed and runtime table, to introduce this side-loaded data.
+
+Again, we make the simplifying assumption that an entry from a side-loaded
+table has exactly 1 'fixed' entry, which can be used for indexing (or set to
+`0` if unused). We also assume that at most 1 side-loaded table contributes to
+each entry.
+
+In order to compute the lookup multiset inclusion argument, we also need the
```
mimoo, line 202: multiset equality argument?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+individual `RuntimeTable(i)`s, to ensure that the runtime values cannot be
+chosen with prior knowledge `joint_combiner`.
+
+### Representing the side-loaded tables
+
+The side-loaded tables can be considered as additional 'masks' that we apply to
+the combined fixed and runtime table, to introduce this side-loaded data.
+
+Again, we make the simplifying assumption that an entry from a side-loaded
+table has exactly 1 'fixed' entry, which can be used for indexing (or set to
```
mimoo, line 198: 'fixed' entry -> 'fixed' column

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+LookupTable - FixedAndRuntimeTable
+```
+
+We will represent the polynomials for the side-loaded tables as
+`SideLoadedTable(i, j)`, where `i` identifies the table, and `j` identifies the
+column within that table. We also include polynomials `SideLoadedCombined(i)`
+in the proof, along with their openings, where
+```
+SideLoadedCombined(i)
+  = sum[j=1, ?, SideLoadedTable(i, j) * joint_combiner^(j-1)]
```
mimoo, line 215: why `j-1` and not `j`?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+```
+and check for consistency of the evaluations and openings when verifying the
+proof.
+
+The `SideLoadedTable` polynomials are included in the proof, but we do so 'as
+if' they had been part of the verification index (ie. without any
```
mimoo, line 221: verification index -> verifier index (that's how it's called in the codebase)

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+  to be used for array accesses to data from the witness.
+* **side-loaded tables** are tables that are committed to in a
+  proof-independent way, so that they may be reused across proofs and/or signed
+  without knowledge of the specific proof, but may be different for each
+  execution of the circuit.
+
+These distinct types of tables are a slight fiction: often it will be desirable
+to have some fixed part of runtime or side-loaded tables, e.g. to ensure that
+indexing is reliable. For example, a table representing an array of values
+`[x, y, z]` in the proof might be laid out as
```
mimoo, line 53: IIUC we also need to have different gates to support these: side-loaded table lookup gate, runtime table lookup gate, and as many fixed table lookup gate as we want to have.

Perhaps built-in table is a better name than fixed table?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
+## Summary
+[summary]: #summary
+
+This RFC proposes an extension to our use of lookup tables using the PLOOKUP
+multiset inclusion argument, so that values within lookup tables can be chosen
+after the constraint system for a circuit has been fixed.
```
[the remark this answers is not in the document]
mrmr1993: Correct. I've tweaked the description of the current implementation in the `Combining multiple tables` section below to mention that they are 'fixed' tables.

rfcs/0001-extended-lookup-tables.md
mrmr1993: This RFC deliberately doesn't mention the actual gates, but we should be able to use the same gate for lookups in any of the tables, since they'll all be combined into 1 table.

For me, 'built-in' seems to imply something stronger, like it being an inherent part of kimchi itself. 'Baked-in' sits slightly better with me, but I think I still prefer 'fixed table'.

rfcs/0001-extended-lookup-tables.md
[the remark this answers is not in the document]
mrmr1993: The witness is pretty much all we know of the transcript at the time we need to select the `joint_combiner`. I've added a comment explicitly mentioning that we'll use the transcript.

mrmr1993: There are a few different styles of 'combining' at play here:
* we 'combine' the fixed tables into a single fixed table
* we 'combine' the fixed table with the runtime and/or side-loaded tables, where the latter are used as masks
* we 'combine' any of the 4 different tables above with the `table_id`s column
* we 'combine' the values in the table above with the `joint_combiner` to create a single value representing each row.

I'll try to think of and define some reasonable terms. At the moment, I'm leaning towards
* 'fixed tables'
* 'combined fixed table'
* 'fixed+runtime table'
* 'lookup table' (fixed+runtime+side-loaded)
* an 'identified' table is a table considered along with its `table_id`
  - e.g. 'identified lookup table', 'identified fixed+runtime table', etc.
* a 'collapsed' table is an 'identified' table where the rows are collapsed into a single value using the `joint_combiner`

rfcs/0001-extended-lookup-tables.md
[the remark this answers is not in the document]
mrmr1993: Added a clarifying comment.

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+
+The fixed table can be constructed at constraint system generation time, by
+extending all of the constituent tables to the same width, generating the
+appropriate table IDs array, and concatenating all of the tables. Concretely:
+
+Let `t[i][j][k]` be the `k`th element of the `j`th entry in the `i`th fixed
+table.
+Let `W` be the maximum width of all of the `t[i][j]`s.
+For any `t[i][j]` whose width `w` is less than `W`, pad it to width `W` by
+setting `t[i][j][k] = 0` for all `w < k <= W`.
```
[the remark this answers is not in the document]
mrmr1993: Added a suggestion for using `TableID = -1` where we don't intend to use the entry.

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+### Sampling the `joint_combiner`
+
+Currently, the `joint_combiner` is sampled using randomness that depends on the
+witness. However, where we have runtime or side-loaded tables, a malicious
+prover may be able to select values in those tables that absue knowledge of
+`joint_combiner` to create collisions.
+
+For example, to maliciously insert a value `(x, y, z)` for table with ID 1, the
+prover could select the value `x + y * joint_combiner + z * joint_combiner^2 +
+-1 * joint_combiner^max_joint_size` for an entry in table 2.
```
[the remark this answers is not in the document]
mrmr1993: It's right, but it was confusing. I've added some more exposition, hopefully it's clearer now.

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,503 @@
...
+For any `t[i][j]` whose width `w` is less than `W`, pad it to width `W` by
+setting `t[i][j][k] = 0` for all `w < k <= W`.
+Form the combined table by concatenating all of these tables
+`FixedTable = t[0] || t[1] || t[2] || ...`
+and store in `TableID` the table ID that the corresponding row came from.
+Then, for every `x` and `y`, we have
+```
+FixedTable[x][k] = t[i][j][k]`
+TableID[x] = i
+where x = len(t[0]) + len(t[1]) + ... + len(t[i-1]) + j
```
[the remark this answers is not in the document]
mrmr1993: This is formalising the concatenation `t[0] || t[1] || ...`, so that `output_row` represents the current row in the concatenated table. `len(t[id])` is supposed to be the number of rows in the table `t[id]`, so we might have e.g.
```
t[0] = [[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]] // len(t[0]) = 4
t[1] = [[0], [1], [2], [3], [4], [5], [6], [7]] // len(t[1]) = 8
let output_row = 7
FixedTable[output_row][col] = t[1][3][col]
output_row = len(t[0]) + 3 = 4 + 3 = 7
```

Any suggestions on how to make this clearer?

rfcs/0001-extended-lookup-tables.md
mrmr1993: Clarified at first usage.

rfcs/0001-extended-lookup-tables.md
mrmr1993: :man_shrugging:

rfcs/0001-extended-lookup-tables.md
mrmr1993: This is explicitly laid out in the section `Full list of polynomials and constraints` below.

rfcs/0001-extended-lookup-tables.md
mrmr1993: You could imagine a primitive `store_array(idx, [val_0, val_1, ..., val_n])` in snarky that creates a permutation between the previous position of `val_i` and the `idx+i`th entry of the table, in the same way that we create a permutation between subsequent uses of a value `val` between different rows in the main witness. This would be functionally the same as `store(idx, val_0)`, `store(idx+1, val_1)`, ..., `store(idx+n, val_n)` where `store` uses lookups, but doing it this way lets us avoid using lots of lookups across lots of rows to initialise the storage.

For an explicit example, consider
```
store_array(idx, [{row= 155, col= 2}, {row= 205, col=7}])
```
By considering the values in the runtime table as 'column 12', we can build the equivalence sets as usual, and then we get the permutations
```
[{row= 155, col= 2}, {row= idx, col= 12}, ....]
[{row= 205, col= 7}, {row= idx+1, col= 12}, ...]
```

Naturally this is only really usable when `idx` is fixed by the constraint system, but it will usually be possible to do this by 'allocating some blocks' in the runtime table corresponding to each array that you want to load.

rfcs/0001-extended-lookup-tables.md
mrmr1993: Yes, I've added a comment detailing this.

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+setting `t[id][row][col] = 0` for all `w < col <= W`.
+Form the combined table by concatenating all of these tables
+`FixedTable = t[0] || t[1] || t[2] || ...`
+and store in `TableID` the table ID that the corresponding row came from.
+Then, for every `output_row`, we have
+```
+FixedTable[output_row][col] = t[id][row][col]`
+TableID[output_row] = id
+where output_row = len(t[0]) + len(t[1]) + ... + len(t[id-1]) + row
+```
```
mimoo, line 164: I read this over and over and it still doesn't make sense to me. For example, how do you compute the value of fixedtable[0][0] ? Also, what if t[0] and t[1] don't have the same number of rows/entries?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+these components -- as the seed for this randomness.
+
+### Representing the combined fixed table
+
+The fixed table can be constructed at constraint system generation time, by
+extending all of the constituent tables to the same width (i.e. number of
+columns), generating the appropriate table IDs array, and concatenating all of
+the tables. Concretely:
+
+Let `t[id][row][col]` be the `col`th element of the `row`th entry in the `id`th
```
mimoo, line 151: btw you use the terms "row" and "entry" interchangeably, I think we should stick to a single term or at least define that somewhere

rfcs/0001-extended-lookup-tables.md
mimoo: also I do not see that fixed column anywhere in the formulas

rfcs/0001-extended-lookup-tables.md
mimoo: re-reading this it still doesn't make sense to me, j should start at 0 here

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+in this RFC, since the principle and implementation are both fairly
+straightforward.*
+
+### Position-independence of side-loaded tables
+
+We want to ensure that side-loaded tables are maximally re-usable across
+different proofs, even where the 'position' of the same side-loaded table in
+the combined table may vary between proofs. For example, we could consider a
+'folding' circuit `C` which takes 2 side-loaded tables `t_in` and `t_out`,
+where we prove
```
mimoo, line 264: the phrasing is very difficult to follow, I'm not sure if I understood this correctly but does `C(t_0, t_1)` check that `t_0` is in `t_in` and `t_1` is in `t_out`?

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+```
+C(t_0, t_1)
+C(t_1, t_2)
+C(t_2, t_3)
+...
+```
+and the 'result' of the fold is the final `t_n`. In each pair of executions in
+the sequence, we find that `t_i` is used as both `t_in` and `t_out`, but we
+would like to expose the same table as the value for each.
```
mimoo, line 273: > expose the same table as the value for each.

not sure I understand that paragraph either

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+
+To achieve this, we use a permutation argument to 'place' the values from each
+side-loaded table at the correct position in the final table. Recall that every
+value
+```
+LookupTable - FixedAndRuntimeTable
+```
+is either 0 (where no side-loaded table contributes a value) or the value of
+some row in `SideLoadedCombined(i) * joint_combiner` from the table `i`.
```
mimoo, line 282: `SideLoadedCombined(i) * joint_combiner` -> `SideLoadedCombined(i)`

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+ * (SideLoadedCombined(3) + gamma + beta * Sigma(11))
+ =
+   LookupPermutation[prev]
+   * ((LookupTable - FixedAndRuntimeTable) * joint_combiner^(-1)
+       + gamma + x * beta * shift[7])
+   * (SideLoadedCombined(0) + gamma + x * beta * shift[8])
+   * (SideLoadedCombined(1) + gamma + x * beta * shift[9])
+   * (SideLoadedCombined(2) + gamma + x * beta * shift[10])
+   * (SideLoadedCombined(3) + gamma + x * beta * shift[11])
+```
```
mimoo, line 306: maybe it'd be nicer to show the accumulator before the polynomialized version

* `acc_0 = i`
* `acc_i = acc_{i-1} * f /g `

where

* `f = ((LookupTable - FixedAndRuntimeTable) * joint_combiner^(-1)
       + gamma + x * beta * shift[7])
   * (SideLoadedCombined(0) + gamma + x * beta * shift[8])
   * (SideLoadedCombined(1) + gamma + x * beta * shift[9])
   * (SideLoadedCombined(2) + gamma + x * beta * shift[10])
   * (SideLoadedCombined(3) + gamma + x * beta * shift[11])`
* `g = ((LookupTable - FixedAndRuntimeTable) * joint_combiner^(-1)
     + gamma + beta * Sigma(7))
 * (SideLoadedCombined(0) + gamma + beta * Sigma(8))
 * (SideLoadedCombined(1) + gamma + beta * Sigma(9))
 * (SideLoadedCombined(2) + gamma + beta * Sigma(10))
 * (SideLoadedCombined(3) + gamma + beta * Sigma(11))`

rfcs/0001-extended-lookup-tables.md
mrmr1993: The idea behind not doing this was that we can avoid paying the cost of computing that `joint_combiner` highest power if we only have a single table with ID 0.

rfcs/0001-extended-lookup-tables.md
mrmr1993: Correct, yes. The intention of the first column is as a mix-in so that you have (the possibility of) an indexed lookup a la memory accesses. You could set it to `0` if you don't care about the contents.

By doing it this way, we can load values into the side-loaded table without needing to care about their fixed argument -- since it's already implicit in the permutation argument.

rfcs/0001-extended-lookup-tables.md
mimoo: Yes

rfcs/0001-extended-lookup-tables.md
mimoo: I added markdown diagram in https://github.com/o1-labs/proof-systems/pull/536, you might want to steal this to make this part clearer :o

rfcs/0001-extended-lookup-tables.md
```diff
@@ -0,0 +1,530 @@
...
+FixedTable[output_row][col] = t[id][row][col]`
+TableID[output_row] = id
+where output_row = len(t[0]) + len(t[1]) + ... + len(t[id-1]) + row
+```
+
+This can be encoded as `W+1` polynomials in the constraint system, which we
+will reference as `FixedTable(col)` for `0 <= col < W` and `TableID`.
+
+For any unused entries, we use `TableID = -1`, to avoid collisions with any
+tables.
```
mimoo, line 170: IIUC you need to update this with `TableID = 0`

mimoo, line 170: IIUC you need to update this with `TableID = 0`

mimoo:
@mrmr1993 we should move that to the mina book, I can do this if you don't wanna do it?

mimoo (approved):

PR #257, opened 2021-12-18
Backend modifications for pickles compatibility (rebased)

mimoo:
Same as https://github.com/o1-labs/proof-systems/pull/220 but rebased. Not sure if @imeckler is working on the other one so I avoided force pushing there.


circuits/kimchi/src/nolookup/constraints.rs
```diff
@@ -605,7 +605,17 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
             // check if wires are connected
             for col in 0..PERMUTS {
                 let wire = gate.wires[col];
-                assert!(wire.col < PERMUTS);
+
+                if wire.col < PERMUTS {
```
mrmr1993: This inequality is the opposite of what it should be

mrmr1993:
Please don't rebase commits that have already been reviewed and/or referenced by other repos.

circuits/kimchi/src/nolookup/constraints.rs
mimoo: this is fixed in a later commit :o

mimoo:
ah yeah that makes sense, I figured it would make reviews easier as they were a bunch of unrelated commits initially

mimoo:
merged master into https://github.com/o1-labs/proof-systems/pull/220 instead

PR #261, opened 2022-01-07
Update OCamlgen test for mina compatibility

mrmr1993:
This PR
* removes the `+nightly` flag to `cargo build`, which is currently unsupported on mina's CI (and unneeded for these tests)
* removes the fake `inline_tests` stanza from the dune file, replaces it with an executable and a `runtest` rule.


mimoo (approved):
LGTM, I'm wondering why add a test executable though and not keep it as a lib that has inline tests?

mrmr1993:
> LGTM, I'm wondering why add a test executable though and not keep it as a lib that has inline tests?

It doesn't really have inline tests, it's a bit of an abuse of that dune rule even if it does work.

PR #263, opened 2022-01-12
Fix performance bug

imeckler:
I accidentally committed some debug lines which were computing completely unused polynomial commitments inside the opening function. This removes that and also optimizes the code for combining polynomials to an improved version. In testing this improved prover performance by about 15-20% :sweat_smile:

### DIFF
--- a/dlog/commitment/src/commitment.rs
+++ b/dlog/commitment/src/commitment.rs
@@ -481,6 +481,11 @@ pub fn combined_inner_product<G: CommitmentCurve>(
     res
 }
 
+enum OptShiftedPolynomial<P> {
+    Unshifted(P),
+    Shifted(P, usize),
+}
+
 impl<G: CommitmentCurve> SRS<G>
 where
     G::ScalarField: CommitmentField,
@@ -660,16 +665,11 @@ where
         let mut g = self.g.clone();
         g.extend(vec![G::zero(); padding]);
 
-        let mut combined_polynomial_terms = vec![];
-
-        // scale the polynoms in accumulator shifted, if bounded, to the end of SRS
         let (p, blinding_factor) = {
-            let mut p = DensePolynomial::<Fr<G>>::zero();
-
-            let mut omega = Fr::<G>::zero();
+            let start = std::time::Instant::now();
             let mut scale = Fr::<G>::one();
-
-            // iterating over polynomials in the batch
+            let mut omega = Fr::<G>::zero();
+            let mut plnm_chunks : Vec<(Fr::<G>, OptShiftedPolynomial<_>)> = vec![];
             for (p_i, degree_bound, omegas) in plnms.iter().filter(|p| !p.0.is_zero()) {
                 let mut offset = 0;
                 let mut j = 0;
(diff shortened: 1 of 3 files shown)


utils/src/dense_polynomial.rs
```diff
@@ -28,9 +29,9 @@ pub trait ExtendedDensePolynomial<F: Field> {
 impl<F: Field> ExtendedDensePolynomial<F> for DensePolynomial<F> {
     fn scale(&self, elm: F) -> Self {
         let mut result = self.clone();
-        for coeff in &mut result.coeffs {
+        result.coeffs.par_iter_mut().for_each(|coeff| {
```
mimoo: I think `for coeff in result.coeffs.par_iter_mut() {` should work no?

dlog/commitment/src/commitment.rs
```diff
@@ -660,62 +665,50 @@ where
...
         g.extend(vec![G::zero(); padding]);
 
-        let mut combined_polynomial_terms = vec![];
-
-        // scale the polynoms in accumulator shifted, if bounded, to the end of SRS
         let (p, blinding_factor) = {
-            let mut p = DensePolynomial::<Fr<G>>::zero();
-
-            let mut omega = Fr::<G>::zero();
+            let start = std::time::Instant::now();
```
mimoo: btw it feels weird to have these timers in prod code. Wouldn't it be better to extract this logic in its own function and then write a bench for it?

dlog/commitment/src/commitment.rs
```diff
@@ -677,45 +679,44 @@ where
                 if let Some(m) = degree_bound {
                     assert!(p_i.coeffs.len() <= m + 1);
                     while j < omegas.unshifted.len() {
-                        let segment = DensePolynomial::<Fr<G>>::from_coefficients_slice(
-                            &p_i.coeffs[offset..if offset + self.g.len() > p_i.coeffs.len() {
+                        let segment = &p_i.coeffs[offset
+                            ..if offset + self.g.len() > p_i.coeffs.len() {
```
mimoo, line 683: unrelated to the PR, but this inlined `if` should really be rewritten

dlog/commitment/src/commitment.rs
```diff
@@ -724,7 +725,33 @@ where
...
+                            res.append(&mut v);
+                            DensePolynomial::from_coefficients_vec(res)
+                        }
+                    }
+                })
+                .collect();
+
+            for p in scaled {
+                res += &p;
+            }
```
mimoo, line 752: could this logic of reconstructing a polynomial from its chunks be abstracted in `utils/src/polynomials.rs` or something like that? 

```rust
pub trait ChunkedPolynomial: Default {
  fn add_unshifted(&mut self, scale: F, coeffs: Vec<F>)
  fn add_shifted(&mut self, scale: F, coeffs: Vec<F>)
  fn reconstruct(self) -> DensePolynomial
}
```

mimoo (approved):
LGTM, left you a few comments

dlog/commitment/src/commitment.rs
imeckler: Hm I think I have deleted this in a later commit

jspada:
nice catch and nice performance jump

utils/src/dense_polynomial.rs
imeckler: Doesn't work actually, only works for `Iterator`s, not the parallel iterator trait from rayon.

dlog/commitment/src/commitment.rs
imeckler: Thanks, did something like this

PR #264, opened 2022-01-13
2x-ish speedup from endomorphism

imeckler:
This PR uses the endomorphism to compute the `g1 + x * g2` step of the commitment opener. The result is a roughly 2x overall speedup in prover time. E.g., with domain size 2^14, prover time has gone from 2.830285628s to 1.440663255s.

Algorithm 1 of https://eprint.iacr.org/2019/1021.pdf (on page 19) was used as a reference.

### DIFF
--- a/dlog/commitment/src/combine.rs
+++ b/dlog/commitment/src/combine.rs
@@ -311,14 +311,14 @@ fn affine_window_combine_one_endo_base<P: SWModelParameters>(
     // acc = 2 (phi(g2) + g2)
     let mut points = g2.to_vec();
     batch_endo_in_place(endo_coeff, &mut points);
-    batch_add_assign_no_branch(&mut denominators, &mut points, &g2);
+    batch_add_assign_no_branch(&mut denominators, &mut points, g2);
     batch_double_in_place(&mut denominators, &mut points);
 
     let mut tmp_s = g2.to_vec();
     let mut tmp_acc = g2.to_vec();
     for i in (0..(128 / 2)).rev() {
         // s = g2
-        assign(&mut tmp_s, &g2);
+        assign(&mut tmp_s, g2);
         // tmp = acc
         assign(&mut tmp_acc, &points);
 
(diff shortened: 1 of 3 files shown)


dlog/commitment/src/combine.rs
```diff
@@ -64,6 +65,36 @@ fn add_pairs_in_place<P: SWModelParameters>(pairs: &mut Vec<SWJAffine<P>>) {
     }
 }
 
+fn batch_add_assign_no_branch<P: SWModelParameters>(
+    denominators: &mut [P::BaseField],
```
mrmr1993, line 69: It seems scary/error-prone to use `denominators` as a shared buffer like this.
Can you rework this as a private type `FieldBuffer<T>` that has a method `use::<A, F: Fn<&mut T, A>>(mut self: Buffer<T>, f: F) -> A`?  The emitted code should be the same, but it would be more self-explanatory and less prone to errors after sharing.

mrmr1993 (approved):
Nice, and looks correct, but left a small safety nit

dlog/commitment/src/combine.rs
```diff
@@ -64,6 +65,36 @@ fn add_pairs_in_place<P: SWModelParameters>(pairs: &mut Vec<SWJAffine<P>>) {
     }
 }
 
+fn batch_add_assign_no_branch<P: SWModelParameters>(
```
mimoo, line 68: I highly suggest adding doc comments even for non-public functions. For example, I'm bit confused as what this function does here.

dlog/commitment/src/commitment.rs
```diff
@@ -382,6 +382,17 @@ pub trait CommitmentCurve: AffineCurve {
         crate::combine::window_combine(g1, g2, Self::ScalarField::one(), x2)
     }
 
+    // Combine where x1 = one
```
mimoo, line 385: change to `///`

mimoo:
thanks for the pointers!

PR #270, opened 2022-01-20
Big cleaning PR

mimoo:
quick PR to do some clean up:

* delete any plonkv1 crates (`dlog/plonk` and `circuits/plonk`)
* move all tests contained in `dlog/tests` to their respective crates
* make the kimchi-circuits crate a module of kimchi
* make kimchi a top-level crate
* rename commitment to poly-commitment
* remove bazel files
* get rid of the `nolookup` folder that was a legacy thing

for some reasons now the tests are very fast o.o I suspect that moving the tests from their own crates to modules of their respective crate is where the gain came from


jspada (approved):
lgtm!

PR #273, opened 2022-01-24
First version of the Cairo runner

querolita:
Cairo works in two phases. First, it executes the program received as input/ in agreement with the verifier, in order to generate the witness of the computation. This part is called the Cairo runner. This witness is basically the full memory content after execution. The full memory is composed by: the compiled input program, a set of hints (I need to receive clarification from the Cairo guys for this part), cells that are allocated throughout execution, and finally the public outputs of the computation. The second part of Cairo is the one providing constraints that need to be satisfied by the witness. Starkware passes these constraints to a Stark prover, but we will be performing this step using a Kimchi-styled proof system. We refer to this second part as Turshi (an Egyptian pickled veggie, thanks @mimoo). 

This pull request is for a first version of the Cairo runner. It contains three main modules: 
- `word` : defines a Cairo word, the native data that is handled by the Cairo runner. In this phase, this is represented as a u64 integer and a flag for negative numbers (useful when one gets negative offsets). For reference, Cairo instructions are represented as 64-bit words. In a future version, we will use field elements here, but for debugging purposes it was more intuitive to stick to u64 for now. From a `CairoWord`, one can access useful methods for decomposing the Cairo instruction: finding register offsets and flags. It comes with a simple test to decode offsets, and another one for decoding instruction flags.
- `bytecode`: refers to the stack of bytecode instructions defining the input program (initially), and ends up adding more entries to the vector containing values computed throughout execution. A `CairoBytecode` can be read at a given position, can be assigned values at a given position, can be resized (enlarged) accordingly to fit new cells, and can be visualized. the same cell can be written many times in execution. But in the proving stage (Turshi), the memory will have to satisfy a write-once behavior. This is not a problem at this point, as obtaining the witness and proving it are two (related but) different tasks. `CairoBytecode` is a struct that contains the length of the compiled program (initial shared input with the verifier), and more importantly, a vector of Option<CairoWord>. The reason that this is an optional type, is that we needed a way to differ between cells assigned with a zero value and non-written cells. If these are None, then they were not still initialized. This becomes more important in the execution. Comes with a unit test.
- `program`:  it contains some structs that are useful throughout execution.
- - `CairoRegisters`: stores program counter, allocation pointer and frame pointer.
- - `CairoVariables`: stores auxiliary variables: content and address of destination, first operand and second operand, result of operation, and size of instruction (instructions with immediate values such as assignments have 2 words). For `dst`, `op0` and `op1` I chose an Option<u128> to represent the possibility that cells were unoccupied, and "negative" u64 elements (as opposed as if I had chosen u64 for the type directly).
- - `CairoStep`: defines a step of the computation. It contains a mutable reference to a `CairoBytecode` (the memory), current and next registers, and auxiliary variables. We have methods to compute each of the necessary values of computation for this struct. These are an "adapted" version (execution-wise, rather than checking-wise) of the state transition written in page 33 of the Cairo whitepaper. Comes with a unit test.
- - `CairoProgram`: defines a full program with a mutable reference to a `CairoBytecode` and the initial values of the registers. We have methods to execute a full program, that is formed by consecutive `CairoStep`s. It comes with two tests, one for a basic program and another one for a program including imports, outputs and builtins. There's still one issue pending from Cairo's response to understand one part of the non-deterministic execution of the Cairo runner.


mimoo:
Can you rebase or merge master into your branch :o? Also should I start reviewing this?

querolita:
Thanks for pointing out @mimoo! I think master is merged now. I have included some constants for readability of the state transition function. I will be adding some suggestions from Joseph as well, but feel free to take a look into it when you can 👍

querolita:
@imeckler sent me a pointer to what seems to be the field that the Cairo team is using [here](https://github.com/starkware-libs/cairo-lang/blob/cf8266fd5d1ff66962579ff7967ac5cdcf699f77/src/starkware/crypto/starkware/crypto/signature/nothing_up_my_sleeve_gen.py#L35)
It seems to be `FIELD_PRIME = 2 ** 251 + 17 * 2 ** 192 + 1` that is written in hexadecimal as `max = 0x800000000000011000000000000000000000000000000000000000000000001`
I have been checking this against the [Cairo playground](https://www.cairo-lang.org/playground/) to learn what is the largest available value it can represent, and it seems to be the case (that number, the modulus, goes back again to zero). Negative numbers also follow the same field (e.g., the number `max-1` becomes -1).
This clarifies that when we represent ours with Pallas, whose sizes are larger, there shouldn't be any issues.

mimoo:
I don't think master is merge, Github seems to be pointing a lot of conflicting files

kimchi/cairo/0001-cairo-gates-and-constraints.md
```diff
@@ -0,0 +1,256 @@
+## Summary
```
mimoo, line 1: can you create a separate PR for that RFC (like https://github.com/o1-labs/proof-systems/pull/250/files)

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.publen
+    }
+
+    /// Get size of the full memory including dummy 0th entry
+    pub fn size(&self) -> u64 {
+        (self.vect.len()) as u64
+    }
+
+    /// Enlarges memory with enough additional None slots if necessary before writing or reading
+    fn enlarge(&mut self, index: u64) {
```
mimoo, line 38: nit: usually this is called "resize", but if cairo calls this enlarge let's use that :o

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+        // if you want to access an index of the memory but its size is less or equal than this
+        if self.size() <= index {
+            // you will need to extend the vector with enough spaces (taking into account that
+            // vectors start by index 0 and size starts in 1)
+            let additional = index - self.size() + 1;
+            self.vect.reserve(additional.try_into().unwrap());
```
mimoo, line 44: `unwrap()` -> `except()`

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Get size of the full memory including dummy 0th entry
+    pub fn size(&self) -> u64 {
+        (self.vect.len()) as u64
+    }
+
+    /// Enlarges memory with enough additional None slots if necessary before writing or reading
+    fn enlarge(&mut self, index: u64) {
+        // if you want to access an index of the memory but its size is less or equal than this
+        if self.size() <= index {
```
mimoo, line 40: you probably want `<` and not `<=`

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+            // you will need to extend the vector with enough spaces (taking into account that
+            // vectors start by index 0 and size starts in 1)
+            let additional = index - self.size() + 1;
+            self.vect.reserve(additional.try_into().unwrap());
+            for _ in 0..additional {
+                // Consider CairoBytecode having Option<CairoWord> so one can have None here
+                self.vect.push(None);
+            }
```
mimoo, line 48: similar:

```rust
self.vect.extend(iter::repeat(None).take(additional));
```

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.enlarge(index);
+        self.vect[index as usize] = Some(CairoWord::new(elem));
+    }
+
+    /// Read element in memory address
+    /// Because of how assignments work in Cairo (called assert-equal), they
+    /// behave in two ways: either check two variables are equal or assign
+    /// the value of one variable to the address of the other one so that
+    /// their content will be the same. This means you may first read to
+    /// addressses of memory that were still not instantiated, and you need
```
mimoo, line 63: typo: addressses

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// to enlarge the vector before reading (with None values).
+    pub fn read(&mut self, index: u64) -> Option<i128> {
+        self.enlarge(index);
+        if self.vect[index as usize].is_some() {
+            Some(self.vect[index as usize].unwrap().to_i128())
```
mimoo, line 68: always avoid `unwrap()`. Here write:

```rust
if let Some(x) = self.vect[index as usize] { ... }
```

or better:

```rust
self.vect[index as usize].map(|x| x.to_i128())
```

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+                // Consider CairoBytecode having Option<CairoWord> so one can have None here
+                self.vect.push(None);
+            }
+        }
+    }
+
+    /// Write u64 element in memory address
+    pub fn write(&mut self, index: u64, elem: i128) {
+        self.enlarge(index);
+        self.vect[index as usize] = Some(CairoWord::new(elem));
```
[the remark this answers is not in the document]
mimoo: maybe `index` should be renamed to `address` or `addr`?

kimchi/cairo/src/runner/definitions.rs
```diff
@@ -0,0 +1,72 @@
+/// Defining some constants for easier readability of the steps
+/// When they refer to single bit flagsets, only one constant is needed
```
mimoo, line 2: these should be `//!` comments

kimchi/cairo/src/runner/definitions.rs
```diff
@@ -0,0 +1,72 @@
...
+    PCREL = 2,
+    PCJNZ = 4,
+    APZ2 = 0,
+    APADD = 1,
+    APONE = 2,
+    OPCJMPINC = 0,
+    OPCCALL = 1,
+    OPCRET = 2,
+    OPCAEQ = 4,
+}*/
```
mimoo, line 25: delete?

mimoo (commented):
The code looks well structured so this is great! Here's a few notes:

* I think the RFC should be a separate PR, so that we can comment on it and keep working on it while this PR can get merged.
* Turshi shouldn't be part of this PR ideally, so we can focus on a single library at the moment
* minor comment: I feel like the distinction between the modules that defines the structures and logic around instructions, and the modules that define the state transition logic and the runner, could be made clearer.

PS: can you also update the main `README.md` and add a `README.md` in your folder?

kimchi/cairo/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod runner;
```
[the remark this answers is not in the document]
querolita: Thanks for the suggestion! I will be adding it :)

In this case I am calling it the cairo runner as this is the name they use in their paper. It basically refers to the piece of code that executes (runs) the cairo program and obtains the memory instance that will be used for the proof generation. So it is supposed to contain the instruction-related functions, and the "state transition logic" as you say, but from an execution perspective, rather than constraints-checking. Maybe I could state this point clearer, as I have changed the modules names very often.

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+/// This data structure stores the memory of the program
+pub struct CairoBytecode {
+    /// length of the public memory
+    publen: u64,
+    /// full memory vector, None if non initialized
+    pub vect: Vec<Option<CairoWord>>,
+}
+
+impl CairoBytecode {
+    /// Create a new memory structure from a vector of u64
```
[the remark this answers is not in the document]
querolita: Oh good catch! That comment is an old one, I should definitely update references to u64 elements, as it is no longer that easy. Basically, Cairo instructions they are 64-bit elements and explained as such in the video tutorial I shared. But the truth is the runner cannot distinguish between "real" instructions, and other types of cells such as immediate values and addresses. Actually, immediate values can be negative, and for this reason I switched to i128 instead (so as to be able to both represent negative values and the largest u64 word). But then I could check that the Cairo language itself operates with field elements directly. So I need to switch to this even more general format. This means memory will be field elements, and only some of them can be interpreted as "64-bit" instructions. So I will have to create some converters and the index trait for accessing vector positions that are indexed by field elements directly, not only just usizes.

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /// Get size of the public memory
+    pub fn public(&self) -> u64 {
+        self.publen
+    }
+
+    /// Get size of the full memory including dummy 0th entry
+    pub fn size(&self) -> u64 {
```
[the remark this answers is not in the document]
querolita: So the reason that it is returning u64 is that the code is working for writing and reading from memory data that fits into 64 bits. But it is in any case something that is going to change (aka the previous comment), as I am switching to field compatibility directly. Nonetheless, given the uses that this function may have, this one could return a usize, instead of u64 or field elements I think.

kimchi/cairo/src/runner/bytecode.rs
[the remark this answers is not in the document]
querolita: Oh you mean that it will panic if index is larger than usize? Meaning for instance, in a machine with 32-bit native words and casting a u64 into a usize directly? Or do you mean if the index is not yet allocated and one tries to access it? If it is the latter, the idea is that the `enlarge` (aka resize, as suggested in another comment) allocates enough memory to store enough cells so that accessing the index address shouldn't be a problem. And good suggestion for the `addr` name, I am changing it :)

kimchi/cairo/src/runner/bytecode.rs
querolita: Thanks!

kimchi/cairo/src/runner/word.rs
```diff
@@ -0,0 +1,314 @@
...
+// TODO(querolita):
+// - Enlarge native word type to u256
+
+// Number of Cairo flags
+pub const NUM_FLAGS: usize = 16;
+
+/// A Cairo instruction for simulation
+#[derive(Clone, Copy)]
+pub struct CairoWord {
+    /// 64bit word
```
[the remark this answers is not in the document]
querolita: Oh thanks, better with the dash yes. 
So, you're indeed right. At the beginning I was coding separate structures and functions depending on whether they were related to the runner (pre-computation step) or the actual proof generation, as the runner was working for native rust types instead of fields elements, unlike turshi, that should work for fields. Now that the code looks quite stable in terms of functionality, and the fact that I realised Cairo will work for fields even at the runner level, I will be "merging" or "renaming" some data structures that were somehow duplicates of each other: e.g. CairoBytecode vs CairoMemory, CairoWord vs CairoInstruction. Now, for the former I do think it will become CairoMemory. But the latter is a bit more tricky. Meaning, not all Cairo words (seen as cells in memory) actually correspond to Cairo instructions (seen as 64-bit data with a certain structure). But this is not very relevant, what's important here is that they will be merged to work directly over field elements.

kimchi/cairo/src/runner/word.rs
```diff
@@ -0,0 +1,314 @@
+/*
+    The Cairo Language works natively for field elements in the finite field with
+    modulus 0x800000000000011000000000000000000000000000000000000000000000001
+    This is the hexadecimal value for 2 ^ 251 + 17 * 2 ^ 192 + 1
+    Our Pallas curves have 256 bits, so Cairo native instructions will fit.
+*/
```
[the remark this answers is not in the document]
querolita: Thank you!

kimchi/cairo/src/turshi/constraint.rs
```diff
@@ -0,0 +1,519 @@
...
+        } else {
+            unimplemented!(); // invalid instruction
+        }
+
+        // convert to Fp all of it
+
+        constraint();
+    }
+}
+*/
```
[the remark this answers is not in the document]
querolita: Oh, that file is indeed important. It will contain (and is containing right now in a more advanced version of that file) the actual Kimchi constraints representing the cairo gate (what we called turshi). But you're right that in this PR we should focus only on the runner and leave all of the other contents for a different PR. The reason that this file was commented out was to ignore it, but I can see this is not the cleanest way to proceed: better have different PRs for each large module.

kimchi/cairo/src/turshi/instruction.rs
```diff
@@ -0,0 +1,396 @@
+/*****************************************************************************************************************
+
+This source file implements Cairo instruction primitive.
+
+*****************************************************************************************************************/
```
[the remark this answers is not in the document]
querolita: thanks!

kimchi/cairo/src/turshi/instruction.rs
```diff
@@ -0,0 +1,396 @@
...
+// Affine curve point type
+//pub use Pallas::Affine as CurvePoint;
+// Base field element type
+//pub type BaseField = <CurvePoint as AffineCurve>::BaseField;
+// Scalar field element type
+//pub type ScalarField = <CurvePoint as AffineCurve>::ScalarField;
+
+use crate::runner::word::CairoWord;
+
+const NUM_FLAGS: usize = 16;
```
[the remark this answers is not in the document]
querolita: Oh I missed this one! Thanks, as you can see in this file, there is a lot of "duplicate" code with respect to the word.rs file. I am merging CairoWord and CairoInstruction to be the same thing, basically.

kimchi/cairo/src/turshi/instruction.rs
```diff
@@ -0,0 +1,396 @@
...
+/// A Cairo instruction
+pub struct CairoInstruction<F: FftField> {
+    /// Cairo word for execution
+    pub word: CairoWord,
+    /// Corresponding field element,
+    pub elem: F,
+    /// Vector of 16 flags as field elements
+    pub flags: Vec<F>,
+    /// Vector of 3 offsets as field elements
+    pub offsets: Vec<F>,
```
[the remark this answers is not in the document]
querolita: Agree +100. I know I have too many public elements in the code, and I still didn't figure out what would be the minimal access to the library from outside to decide which fields/methods should be made public. For ease of coding within the cairo folder I made them public, but I know I will be more selective when it comes to privateness of some parts of the code. Also, some fields are just elements such a usize or things like that, but when the struct has more complex data such as pointers to vectors with a given lifetime, I am not sure if I would like to have getters and setters that would give copies of those vectors, as I prefer them to use the "original" data (hence, the lifetime) all the time. An example of this is the CairoProgram, which starts with an initial instantiation of the memory, and with each consecutive step that's made, that memory keeps getting updated. But I may be missing some good practices about how to proceed when one wants this behaviour. Would you do anything smarter in this case and/or getter and setters could work similarly without ending up with clones of the vector every time it's used?

kimchi/cairo/src/turshi/mod.rs
```diff
@@ -0,0 +1,4 @@
+//pub mod constraint;
+//pub mod instruction;
+//pub mod machine;
+//pub mod memory;
```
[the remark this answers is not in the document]
querolita: Yeah, I was commenting out everything related to turshi so as to keep this PR be a runner-only thing. But yeah, it looks like this is not the good way to do this. I am going to update this PR to only contain the runner folder (also the RFC going to another PR), and then have a different one for turshi stuff.

kimchi/cairo/src/lib.rs
```diff
@@ -0,0 +1,2 @@
...
+pub mod turshi;
```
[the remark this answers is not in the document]
querolita: Yes, I will do that 👍

kimchi/cairo/src/runner/bytecode.rs
querolita: Actually Cairo never mentions this method about resizing memory cells before accessing them. I realised we needed that as I went through some real programs and observed their behaviour regarding memory use. So basically, the memory needs no enlarging/resizing at proving time, as at that point the memory is seen as a write-once stack, but at the precomputation stage (runner) it is indeed necessary as one's building the stack iteratively.

kimchi/cairo/src/runner/bytecode.rs
querolita: Mmm I do think I need `<=` here for a "weird" reason, let me explain. At first my vectors (memory, basically) started with the 0 index, as usual when programming. But then I realised that the tests one could extract from the Cairo playground started with a memory address of 1. That means, that every time I was checking a program, I needed to manually "adapt" the memory content to be one less value for the addresses that contained stack addresses inside. And this is a pain when you do it more than twice. So I just gave up and inserted a dummy "zero" entry at the beginning of the memory vector, so that useful entries started on position 1 (thus the second) as in the playground. So for example: say I have a program with 3 instructions. Then they will go from position 0 (with the dummy), then 1, 2 and 3. So one would have a 4-elements sized memory with only 3 "real" cells. Now suppose you want to write in the 4th (fake) position, meaning the fifth. Then you need to enlarge it by one. And since `self.size() = 4` and `index = 4`, then I need to include the `<=`. But when I get a "normal" test-suite that won't use the playground outputs, then I will be able to remove the equal and have a normal behaviour.

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Read element in memory address
+    /// Because of how assignments work in Cairo (called assert-equal), they
+    /// behave in two ways: either check two variables are equal or assign
+    /// the value of one variable to the address of the other one so that
+    /// their content will be the same. This means you may first read to
+    /// addressses of memory that were still not instantiated, and you need
+    /// to enlarge the vector before reading (with None values).
+    pub fn read(&mut self, index: u64) -> Option<i128> {
+        self.enlarge(index);
```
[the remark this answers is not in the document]
querolita: Right, it is a bit weird at the beginning. But the reason for this is their nondeterministic execution model. This basically means that you need to execute your program with correct values that will satisfy the constraints at the proof system level. And one of these constraints is assert-equals. They mean exactly that. Sometimes they are assertions (is this value the same as this other value), and other times it is an assignment (store in this value the value of the other). So in any case, when you reach an assert-equal constraint, what it does is to check that two things have the same contents. No matter if they came from an assertion or an assignment. More concretely, what is being checked is that `dst = res`. So for this to be true, there are two possible ways: either in `dst_addr` we stored `res`, or the value of `dst` was loaded into the address of `res`, which is `op1_addr`. Then, the fact that I was modelling the memory to be an `Option<CairoWord>` allows us to distinguish between a `None` entry or an entry with an immediate value of `0`. So if `dst` is `None` you know it is `res` that you need to copy there. But if it is the other way round, you do the opposite. In both cases, the constraint of the assert equal should be satisfied.

kimchi/cairo/src/runner/definitions.rs
querolita: thanks!

kimchi/cairo/src/runner/definitions.rs
querolita: Oh yes! This was an attempt to represent those constants in a more elegant way using enums as Joseph suggested me the other day. But I probably didn't imagine the same type of solution as he was, and indeed the commented lines do not work (as rust does not allow repeated values in an enum). So by now, I will keep those as constants, as their only goal is to increase readability when defining the step behaviour, as opposed to just writing the integers (which you may not remember what they correspond to).

kimchi/cairo/src/runner/bytecode.rs
querolita: Oh I see, so the `map` function would return an `Option<CairoWord>` meaning if the content of the address is a `None`, it will return a `None`... It looks way neater this way, it avoids using the conditional explicitly, thanks for the training! Nonetheless, what's the reason that `unwrap()` should be avoided? Perhaps the fact that it can panic if executed on a `None` element, and thus too verbose and non-elegant code?

mimoo: yeah! So in general you want to write code that can't panic, especially if it's library code. It's better to return an Error or an Option and let the caller decide what they want to do with that. If it doesn't make sense to return an Error/Option and let the caller decide, you can make the panic more explicit in several ways:

* use `expect()`, so that you can document why you think panicking is ok
* change the name of the function to reflect that. In OCaml functions that panic are appended with `_exn`, but I don't think there's a convention in Rust. Knowing that, I'm not sure if it makes sense to do that.
* document that the function can panic, this is what the rust stdlib does

querolita: Thanks for the tips, I will be using them from now on 👍

kimchi/cairo/src/runner/bytecode.rs
querolita: Thank you, I had never seen this clean layout before :)

kimchi/cairo/src/runner/bytecode.rs
```diff
@@ -0,0 +1,106 @@
...
+        (self.vect.len()) as u64
+    }
+
+    /// Enlarges memory with enough additional None slots if necessary before writing or reading
+    fn enlarge(&mut self, index: u64) {
+        // if you want to access an index of the memory but its size is less or equal than this
+        if self.size() <= index {
+            // you will need to extend the vector with enough spaces (taking into account that
+            // vectors start by index 0 and size starts in 1)
+            let additional = index - self.size() + 1;
```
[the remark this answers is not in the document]
querolita: For this, I think it makes sense to stick to the second option you're suggesting. Otherwise I would be moving the task of checking space to the `read` and `write` methods instead, and I prefer to have it in the same `resize` function. However, it may make sense to call it `check_or_resize` or something like that...

kimchi/src/circuits/cairo/runner/memory.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.resize(addr.to_u64());
+        if self[addr].is_some() {
+            Some(self[addr].unwrap().get_word())
+        } else {
+            None
+        }
+    }
+
+    /// Visualize content of memory excluding the 0th dummy entry
+    pub fn view(&mut self) {
```
mrmr1993, line 93: Does it make sense to implement `Display`/`Debug` here instead?

kimchi/src/circuits/cairo/runner/program.rs
```diff
@@ -0,0 +1,466 @@
...
+        } // no more values than 0 and 1 because op0_reg is one bit
+        self.vars.op0 = self.mem.read(self.vars.op0_addr);
+    }
+
+    /// This function computes the second operand address and content and the instruction size
+    pub fn set_op1(&mut self) {
+        if self.instr().op1_src() == OP1_DBL {
+            self.vars.size = F::one(); // double indexing
+            self.vars.op1_addr = self.vars.op0.unwrap() + self.instr().off_op1(); // should be positive for address
+            self.vars.op1 = self.mem.read(self.vars.op1_addr);
```
mrmr1993, line 149: Factor this out of all the branches?

kimchi/src/circuits/cairo/runner/program.rs
```diff
@@ -0,0 +1,466 @@
...
+            // execute current step and increase time counter
+            step.execute();
+            //i += 1;
+            match step.next {
+                None => end = true, // if find no next registers, end
+                _ => {
+                    // if there are next registers
+                    end = false;
+                    regs = step.next.unwrap();
+                    if curr_ap <= regs.pc {
```
mrmr1993, line 336: Since `ap` can be controlled by the program, it seems like this may not be the right termination condition (for example, you could write to some memory and then jump to it to execute it). Should we check if the memory at `pc` is `None` instead?

kimchi/src/circuits/cairo/runner/memory.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Read element in memory address
+    /// Because of how assignments work in Cairo (called assert-equal), they
+    /// behave in two ways: either check two variables are equal or assign
+    /// the value of one variable to the address of the other one so that
+    /// their content will be the same. This means you may first read to
+    /// addresses of memory that were still not instantiated, and you need
+    /// to enlarge the vector before reading (with None values).
+    pub fn read(&mut self, addr: F) -> Option<F> {
+        self.resize(addr.to_u64());
+        if self[addr].is_some() {
```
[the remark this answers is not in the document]
querolita: Oh! At first I thought you were suggesting me to use an `if let` block such as:
`if let Some(word) = self[addr] { Some(word.get_word()) }`
`else { None }`
But then I saw what you meant, that the `map()` function already returns `None` if it cannot map to that expression!
That is very neat, thanks for sharing :)

kimchi/src/circuits/cairo/runner/word.rs
```diff
@@ -0,0 +1,265 @@
...
+        CairoWord { word }
+    }
+
+    /// Returns the content of the word as a field element
+    pub fn get_word(&self) -> F {
+        self.word
+    }
+
+    /// Returns the destination offset in biased representation
+    pub fn off_dst(&self) -> F {
```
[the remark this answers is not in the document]
querolita: Oh, do you mean that I move all of the decomposition methods to a trait (i.e. `Decomposition`) for `CairoWord<F>`, or as a trait for field elements?

kimchi/src/circuits/cairo/runner/program.rs
```diff
@@ -0,0 +1,466 @@
+//! This module represents a Cairo execution step
```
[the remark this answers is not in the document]
querolita: Oh sure, basically I mixed `step` and `program` together and forgot to change the doc comment accordingly. Nonetheless, now I wonder if it makes sense to call it `runner`, as it involves all of the running logic.

kimchi/src/circuits/cairo/runner/program.rs
```diff
@@ -0,0 +1,466 @@
...
+        Self::new()
+    }
+}
+
+/// A data structure to store a current step of Cairo computation
+pub struct CairoStep<'a, F: PrimeField> {
+    // current step of computation
+    //step: u64,
+    /// current word of the program
+    pub mem: &'a mut CairoMemory<F>,
```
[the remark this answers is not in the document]
querolita: I chose a mutable reference with a given lifetime because I didn't want to copy or clone the whole structure of `CairoMemory` every time it was being called in a different function. In particular, since each step is given by the status of the memory at that time and a series of variables, and then the next step takes the status of memory and builds on top of it, I thought the program could just be "passing" the same object for every step. So mainly, the reason is potential efficiency and saving some space.

kimchi/src/circuits/cairo/runner/memory.rs
```diff
@@ -0,0 +1,137 @@
...
+        // idx should not be too big, this should be safe)
+        let addr: u64 = idx.to_u64();
+        &self.data[addr as usize]
+    }
+}
+
+impl<F: PrimeField> IndexMut<F> for CairoMemory<F> {
+    fn index_mut(&mut self, idx: F) -> &mut Self::Output {
+        let addr: u64 = idx.to_u64();
+        &mut self.data[addr as usize]
```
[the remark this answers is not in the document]
querolita: Oh this is a very good question indeed... I hadn't think of that, since I was calling to `resize()` at the beginning of `write()` and `read()` methods. But at the end, these two are connected to the `Index` trait, so I could try to move this to `index()` and keep a clearer separation of concerns 👍

kimchi/src/circuits/cairo/runner/memory.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub data: Vec<Option<CairoWord<F>>>,
+}
+
+impl<F: PrimeField> Index<F> for CairoMemory<F> {
+    type Output = Option<CairoWord<F>>;
+    fn index(&self, idx: F) -> &Self::Output {
+        // Safely convert idx from F to usize (since this is a memory address
+        // idx should not be too big, this should be safe)
+        let addr: u64 = idx.to_u64();
+        &self.data[addr as usize]
```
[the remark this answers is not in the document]
querolita: So it seems like I can call `resize` inside `index_mut()` to avoid having to call `resize()` inside `write()`, but I cannot avoid calling `resize()` in `read()`. Since `write()` is assigning a value to an address of memory, it looks like it is using the `IndexMut` trait beneath. Unlike `read()`, which is only "looking into" an address of memory, and thus seems to be calling `Index` instead. And since in `Index` the `self` is immutable, I cannot call `resize()` there.

kimchi/src/circuits/cairo/runner/memory.rs
```diff
@@ -0,0 +1,137 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// Visualize content of memory excluding the 0th dummy entry
+    pub fn view(&mut self) {
+        for i in 1..self.size() {
+            if self.read(F::from(i)).is_some() {
+                println!("{}: 0x{:?}", i, self.read(F::from(i)).unwrap().to_hex());
```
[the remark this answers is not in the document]
querolita: Oh sure, I'll do that! In any case, I think it is more intuitive if I don't use `to_hex()` here, but a little endian version of it. Otherwise things like `0x8000u16` get printed as `0x0080`.

kimchi/src/circuits/cairo/runner/program.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// This function returns the current word instruction being executed
+    pub fn instr(&mut self) -> CairoWord<F> {
+        CairoWord::new(self.mem.read(self.curr.pc).unwrap())
+    }
+
+    /// This function computes the first operand address
+    pub fn set_op0(&mut self) {
+        if self.instr().op0_reg() == OP0_AP {
+            // reads first word from allocated memory
+            self.vars.op0_addr = self.curr.ap + self.instr().off_op0();
```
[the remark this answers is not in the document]
querolita: I guess it all depends on what assumptions we are making on the kind of input we would be executing the runner with. The input should be a program that was compiled trustfully, and thus should not generate instructions accessing invalid addresses of memory. Meaning, we should not be executing the Cairo runner on field elements that do not correspond to real bytecode computations. So if this is the case, then I think we don't need to worry about over/underflows. Now, if we wanted this piece of code to be "safe enough", then we should decide what should be a reasonable limit for the memory size for instance. We are already doing this implicitly, as we are casting the address to a `u64`. But I guess this size is already too large and would cause memory overflows in any case. Even `u32` seems too large, but then `u16` may be too small.... Any thoughts on this?

kimchi/src/circuits/cairo/runner/memory.rs
querolita: Oh I was only using this `view` thing for some unit tests. But I can turn it into a trait instead 👍

kimchi/src/circuits/cairo/runner/program.rs
querolita: Ah well yes, I can put this aside (but taking into account the `else` branch, that corresponds to invalid instructions). But here again, if we were assuming that we are only executing the runner for valid bytecodes, then this should never arise.

kimchi/cairo/src/runner/bytecode.rs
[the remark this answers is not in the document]
mimoo: > Oh you mean that it will panic if index is larger than usize

no I meant if `index` is larger than the size of `vect`

kimchi/src/circuits/cairo/runner/program.rs
querolita: Ok yeah, this is a tricky one and you got it :D Short answer is: yeah, it makes sense that the program finishes when the program counter points to a `None` address, meaning that there are no more instructions. But there are two problems with this: 
1. Instructions are given at compilation time. And in compilation time you still don't know how long the execution will take, so you cannot predict where there will be an address that is empty. 
2. Recall that the compiled program and all of the variables created in execution will share the same memory struct. If you knew in compilation time where there will be an empty cell, you would be leaking information about how long the computation takes. But this is input (witness) dependent, so you cannot really tell in compilation (and should not).
3. A correctness condition for the memory is that it has no holes. So you cannot really leave empty cells in the middle. In issue [294](https://github.com/o1-labs/proof-systems/issues/294) I give more intuition about the shape of the memory cells. 

So long answer would be: the last return of the program moves the program counter to `current+1`, and that address should be the first "empty" cell at the beginning of the execution (the first empty cell after compilation). And in reality, I have noticed that real Cairo executions tend to write in that cell the address where outputs are written. Thus, it's a cell that is written afterwards, some kind of retrospective. And the final `ap` points to that same address, where the outputs are written. So that's why I found that condition to be a candidate halting condition. 

Nonetheless, and as explained in the issue, I still need to carefully analyse their repo to make sure what's their mechanism to fill these cells and address halting.

kimchi/cairo/src/runner/bytecode.rs
mimoo: yeah I can never remember how to spell that word...

cairo/src/helper.rs
```diff
@@ -0,0 +1,99 @@
+//! This module inlcudes some field helpers that are useful for Cairo
+
+use ark_ff::FftField;
+use o1_utils::FieldHelpers;
+
+//(TODO move to utils inside FieldHelpers)
+
+/// Field element helpers for Cairo
+pub trait CairoFieldHelpers<F: FftField> {
```
mimoo, line 9: can you remove the trait bound here on `F`? Not sure if that makes sense

cairo/src/helper.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Return field element as byte, if it fits. Otherwise returns least significant byte
+    fn ls_byte(self) -> u8;
```
mimoo, line 11: better to spell this out `least_significant_byte` or use common acronym `lsb`

cairo/src/word.rs
```diff
@@ -0,0 +1,429 @@
+//! The Cairo language works natively for field elements in the finite field with
+//! modulus 0x800000000000011000000000000000000000000000000000000000000000001
+//! This is the hexadecimal value for 2 ^ 251 + 17 * 2 ^ 192 + 1
+//! Our Pallas curves have 256 bits, so Cairo native instructions will fit.
```
mimoo, line 4: I thought the pallas curves had 255 bits :o

cairo/src/word.rs
```diff
@@ -0,0 +1,429 @@
...
+
+use crate::definitions::*;
+use crate::helper::CairoFieldHelpers;
+use ark_ff::FftField;
+use o1_utils::field_helpers::FieldHelpers;
+
+/// A Cairo instruction for simulation
```
mimoo, line 11: if this is an instruction, why is it called a word?

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
+//! This module represents a run of a Cairo program as a consecution of execution
```
mimoo, line 1: -> as a "series" of execution steps?

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+
+/// A structure to store program counter, allocation pointer and frame pointer
+#[derive(Clone, Copy)]
+pub struct CairoPointers<F: FftField> {
+    /// Program counter: points to address in memory
+    pub pc: F,
+    /// Allocation pointer: points to first free space in memory
+    pub ap: F,
+    /// Frame pointer: points to the beginning of the stack in memory (for arguments)
+    pub fp: F,
```
mimoo, line 17: should these really be public? My understanding is that you should not set them directly, and they should be set only through the state transition

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    /// This function computes the first operand address
+    pub fn set_op0(&mut self) {
+        let reg = {
+            match self.instr().op0_reg() {
+                OP0_AP => self.curr.ap, // reads first word from allocated memory
+                _ => self.curr.fp,      // reads first word from input stack
+            } // no more values than 0 and 1 because op0_reg is one bit
+        };
```
mimoo, line 121:
```rust
let reg = match self.instr().op0_reg() {
...
};
```

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+        self.set_op1();
+        self.set_res();
+        self.set_dst();
+        // If the Option<> thing is not a guarantee for continuation of the program, we may be removing this
+        let next_pc = self.next_pc();
+        let (next_ap, next_fp) = self.next_apfp();
+        self.next = Some(CairoPointers::new(
+            next_pc.unwrap(),
+            next_ap.unwrap(),
+            next_fp.unwrap(),
```
mimoo, line 105: unwrap -> expect

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+        }
+    }
+
+    /// Executes a Cairo step from the current registers
+    pub fn execute(&mut self) {
+        // This order is important in order to allocate the memory in time
+        self.set_op0();
+        self.set_op1();
+        self.set_res();
+        self.set_dst();
```
mimoo, line 98: I think it'd be much nicer if the code matched the paper as closely as possible so that we can easily audit it. It also looks like CairoVariables is unnecessary here? It's just to keep track of variables between these functions right? Why not just keep track of these variables within the logic of execute?

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+        from starkware.cairo.common.serialize import serialize_word
+        func main{output_ptr : felt*}():
+            tempvar x = 10
+            tempvar y = x + x
+            tempvar z = y * y + x
+            serialize_word(x)
+            serialize_word(y)
+            serialize_word(z)
+            return ()
+        end
```
mimoo, line 412: if possible, include a link to the UI from where I can easily recompile this code and obtain the instructions below (for example if I want to modify the test)

mimoo (approved):
Left you a bunch of small comments. Feel free to go over them in another PRs (or ignore them if you find them irrelevant).

kimchi/cairo/src/turshi/instruction.rs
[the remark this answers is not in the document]
mimoo: Getters/setters doesn’t add clones :o you can return references. Note that If you implement both getter and setter it’s like having a `pub` on the field.

kimchi/src/circuits/cairo/runner/program.rs
```diff
@@ -0,0 +1,466 @@
...
+            unimplemented!(); // invalid instruction
+        }
+        (next_ap, next_fp)
+    }
+}
+
+/// This struct stores the needed information to run a program
+pub struct CairoProgram<'a, F: PrimeField> {
+    /// full execution memory
+    mem: &'a mut CairoMemory<F>,
```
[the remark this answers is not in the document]
mimoo: Yeah I’d think this is better as a mut ref

cairo/src/helper.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn to_u64(self) -> u64;
+
+    /// Return a field element in hexadecimal in little endian
+    fn to_hex_le(self) -> String;
+
+    /// Return a vector of field elements from a vector of i128
+    fn vec_to_field(vec: Vec<i128>) -> Vec<F>;
+}
+
+impl<F: FftField> CairoFieldHelpers<F> for F {
```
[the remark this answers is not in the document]
querolita: Oh sure, our conversation! I will check if I can indeed switch to Field instead

cairo/src/helper.rs
querolita: Ah, well I thought `lsb` would stand for least significant bit instead of byte. In any case, I can switch to the verbose format if that makes more sense.

cairo/src/helper.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn chunk_u16(self, pos: usize) -> F;
+
+    /// Return first 64 bits of the field element
+    fn to_u64(self) -> u64;
+
+    /// Return a field element in hexadecimal in little endian
+    fn to_hex_le(self) -> String;
+
+    /// Return a vector of field elements from a vector of i128
+    fn vec_to_field(vec: Vec<i128>) -> Vec<F>;
```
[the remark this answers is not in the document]
querolita: Mmm fair enough! This was only used in one unit test, so I played down this documentation.

cairo/src/definitions.rs
```diff
@@ -0,0 +1,60 @@
+//! Defining some constants for easier readability of the steps
```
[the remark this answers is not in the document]
querolita: Oh I like this!

cairo/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![warn(missing_docs)]
+//! This module contains the code that executes a compiled Cairo program and generates the memory
+/// The Cairo runner includes code to execute a bytecode compiled Cairo program,
+/// and obtain a memory instantiation after the execution. It uses some code to
+/// represent Cairo instructions and their decomposition, together with their logic
+/// which is represented as steps of computation making up the full program.
+pub mod definitions;
```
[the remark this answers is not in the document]
querolita: Oh let me check... You're right, it will simply attach the `//!` inside `definitions.rs` after the `///` comment. I can unify these things.

cairo/src/word.rs
querolita: Oh well they do, I meant that they were large enough to fit the Cairo modulus. I'll write it more precisely.

cairo/src/word.rs
```diff
@@ -0,0 +1,429 @@
+//! The Cairo language works natively for field elements in the finite field with
+//! modulus 0x800000000000011000000000000000000000000000000000000000000000001
```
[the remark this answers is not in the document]
querolita: This is the size of the finite field that they use in Cairo. It comes from the operation: 2 ^ 251 + 17 * 2 ^ 192 + 1. 

This means that this number `0x800000000000011000000000000000000000000000000000000000000000001` would become zero in a real Cairo execution, using the Cairo implementation. In our case however, it would actually correspond to the full amount `0x800000000000011000000000000000000000000000000000000000000000001`, as it fits inside the Pallas fields.

cairo/src/word.rs
```diff
@@ -0,0 +1,429 @@
...
+use crate::helper::CairoFieldHelpers;
+use ark_ff::FftField;
+use o1_utils::field_helpers::FieldHelpers;
+
+/// A Cairo instruction for simulation
+#[derive(Clone, Copy)]
+pub struct CairoWord<F: FftField> {
+    /// field element. Instructions will fit in 64 bits
+    word: F,
+}
```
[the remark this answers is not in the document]
querolita: Oooh! Ok thanks! So this is like what you did with the `ChaCha0` struct and the `PhantomData` thing! Cool, thanks! Then I must assume that its content (the actual word) will be publicly accessible?

cairo/src/word.rs
querolita: I guess this comes from the previous version of the code where I had different entities for CairoWord and CairoInstruction. But I switched to only having CairoWords. Basically, all cairo instructions are cairo words, but not all cairo words are instructions (where instructions stands for 64-bit words with the flag and offsets structure). I will unify notation so it's clearer.

cairo/src/runner.rs
querolita: 👍  or series of consecutive execution steps, I think that's what I wanted to say and my English failed

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+//! steps, each of which define the execution logic of Cairo instructions
+
+use crate::definitions::*;
+use crate::memory::CairoMemory;
+use crate::word::{CairoWord, Decomposition};
+use ark_ff::FftField;
+
+/// A structure to store program counter, allocation pointer and frame pointer
+#[derive(Clone, Copy)]
+pub struct CairoPointers<F: FftField> {
```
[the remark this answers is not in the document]
querolita: Mmmm I like `CairoContext` as a substitute for `CairoVariables`. Here for the pointers... I use `CairoState` for the Cairo gate type that refers to instruction related checks... You know what? Then I'll use `CairoState` for `CairoPointers` and `CairoInstruction` for the `GateType`. Cool!

cairo/src/runner.rs
querolita: Absolutely right. I should have a getter for these, as I really only extract them. One shouldn't be able to tamper them.

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+    op0_addr: F,
+    /// Second operand address
+    op1_addr: F,
+    /// Size of the instruction
+    size: F,
+}
+
+impl<F: FftField> CairoVariables<F> {
+    /// This function creates an instance of a default CairoVariables struct
+    pub fn new() -> CairoVariables<F> {
```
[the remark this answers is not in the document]
querolita: Ah! Very neat 😄

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+    }
+}
+impl<F: FftField> Default for CairoVariables<F> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A data structure to store a current step of Cairo computation
+pub struct CairoStep<'a, F: FftField> {
```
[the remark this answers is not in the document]
querolita: Ok! I thought these two must go together.

cairo/src/helper.rs
querolita: I can try! I thought they had to be specified both in the trait and the impl, but I guess it's more flexible if you only specify the generic at the impl level... Cool feature.

cairo/src/runner.rs
querolita: ah! you told me last time and I fixed this in some files, but then there are some unwraps left here and there 🤕

cairo/src/runner.rs
querolita: Oh, precisely that ordering of `op0, op1, res, dst` is the one in the paper (not sure if that's what you mean). 

Regarding the variables, do you mean we should keep those variables inside the impl of CairoStep instead of a field inside the struct itself? Answering your question, these variables only live within the step, and they do not connect to other parts of the following step, other than the logics itself. If you instead mean we could pass them as inputs/outputs of the individual functions that decompose the execution? Since there are so many of these, I thought one would get very long functions calls by doing it that way. But I am not sure this is what you meant.

cairo/src/runner.rs
```diff
@@ -0,0 +1,472 @@
...
+            // right part is addition
+            } else if self.instr().res_log() == RES_MUL {
+                self.vars.res = Some(self.vars.op0.unwrap() * self.vars.op1.unwrap());
+            // right part is multiplication
+            } else {
+                panic!("Invalid instruction");
+            }
+        } else {
+            // multiple bits take value 1
+            panic!("Invalid instruction");
```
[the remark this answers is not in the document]
querolita: Mmm ok, that will require some changes, but it can be a good next step.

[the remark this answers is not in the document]
querolita: Actually the paper has a mixture of conditionals and match statements. But I can try to follow the paper style more literally so that I am using the exact same statement at each point.

cairo/src/runner.rs
querolita: Oh sure I can :)

cairo/src/helper.rs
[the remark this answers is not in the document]
querolita: For this I need to switch to Field also in o1-utils::field_helper, but it's fine. I will ask Joseph before I do so.

cairo/src/runner.rs
querolita: Actually I think I don't even need a getter, as I am only using them within the same module

cairo/src/runner.rs
querolita: And the same for other functions of the module, true

cairo/src/runner.rs
mimoo: When I was reading the paper, I would have imagined the state transition to be implemented as a single function, not as a number of functions. If all these functions were inlined you wouldn't need the Variables if I understand correctly.

cairo/src/word.rs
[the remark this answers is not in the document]
mimoo: no you'd still have to write `pub struct CairoWord<F>(pub F)` for it to be public

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ff7b7674aad084f00c4c40da8e438848

cairo/src/word.rs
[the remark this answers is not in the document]
mimoo: maybe include a link to some source stating that?

cairo/src/lib.rs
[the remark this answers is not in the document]
mimoo: ok good to know, I was wondering :)

cairo/src/helper.rs
mimoo: Maybe lsb only stands for bit, not sure :D I feel like I've seen it being used for both but I might be wrong. In any case the return value (u8 instead of bool) and documentation would be enough to indicate that it's a byte

querolita: Oh good point... I honestly prefer `lsb` so I will be changing it in a future PR

cairo/src/word.rs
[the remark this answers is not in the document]
querolita: Awesome, I will :)

cairo/src/word.rs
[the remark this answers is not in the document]
querolita: Oh nice! I see!

cairo/src/runner.rs
querolita: Oh, in the paper it is indeed just one function... I preferred to have it as different functions because it helped me to better distinguish each of the sections of the transition function. But then sure, if I want to avoid having a Variables struct then I will need to pass these values as arguments/return. I can do it for the next version of the runner if you think it looks neater that way 👍

PR #276, opened 2022-02-07
Use Expr for generic gate

mrmr1993:
This PR switches the generic gate to use the `Expr` construction.

The test in the generic polynomial file has also been deleted, due to its heavy use of the original expanded version of the code. There is a remaining end-to-end test in `kimchi/src/tests/generic.rs` that I was able to use to test this version.


kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -4,276 +4,20 @@ This source file implements generic constraint polynomials.
...
-        let f = cs.gnrc_lnrz(&w_zeta, generic_zeta).interpolate();
-        let f_zeta = f.evaluate(&zeta);
-
-        // check that f(z) = t(z) * Z_H(z)
-        let z_h_zeta = cs.domain.d1.evaluate_vanishing_polynomial(zeta);
-        assert!(f_zeta == t_zeta * z_h_zeta);
-    }
+use ark_ff::FftField;
+
+pub fn constraint<F: FftField>() -> E<F> {
```
querolita, line 14: will this need to be adapted to the new double-generic gate? Meaning, to include the six wires l1, r1, o1, l2, r2, o2

kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -4,276 +4,20 @@ This source file implements generic constraint polynomials.
...
+    let v = |c| E::cell(c, CurrOrNext::Curr);
+    let w = |i| v(Column::Witness(i));
+    v(Column::Index(GateType::Generic)) // Selector
+        * (w(0) * v(Column::Coefficient(0)) // Left input
```
querolita, line 18: nit: would you like `L_COEFF`, `R_COEFF`, `O_COEFF` here?

querolita (commented):
really succinct and elegant, but I guess it will require changes for the double generic gate?

kimchi/src/circuits/polynomials/generic.rs
mimoo: Yeah, I think that’s fine but the generic gate is a special case in the verifier circuit in mina so it doesn’t sound straight forward. I’m not sure how much that optimization was doing for us though in the grand scheme of things.

kimchi/src/circuits/polynomials/generic.rs
jspada: i would

PR #277, opened 2022-02-07
Jspada/field helpers to utils

### LINKED ISSUE #242: Move signer FieldHelpers to utils crate.
This is generally useful, not just for signer.

jspada:
This PR moves the field related helpers to the `utils` crate.  This was an earlier code review suggestion that we decided to do later, but it would be helpful and save rewriting some code if I had them available in `o1_utils` now.

Closes #242


mrmr1993 (approved):
Nice!

querolita (approved):
It's nice having this trait inside the utils crate :)

PR #285, opened 2022-02-08
[gate][poseidon] refactor to make it more readable

mimoo:
This PR should improve the readability of the poseidon gate

### DIFF
--- a/kimchi/src/circuits/polynomials/poseidon.rs
+++ b/kimchi/src/circuits/polynomials/poseidon.rs
@@ -1,72 +1,83 @@
-/*****************************************************************************************************************
-
-This source file implements Posedon constraint polynomials.
-
-*****************************************************************************************************************/
+//! This source file implements Posedon constraint polynomials.
 
 use crate::circuits::expr::{Cache, Column, ConstantExpr, E};
 use crate::circuits::gate::{CurrOrNext, GateType};
 use crate::circuits::gates::poseidon::*;
 use ark_ff::{FftField, SquareRootField};
 use oracle::poseidon::{PlonkSpongeConstants15W, SpongeConstants};
+use std::ops::RangeInclusive;
 use CurrOrNext::*;
 
-/// An equation of the form `(curr | next)[i] = round(curr[j])`
-pub struct RoundEquation {
-    pub source: usize,
-    pub target: (CurrOrNext, usize),
+/// All the information needed to construct a round in the Poseidon custom gate.
+pub struct RoundSpec {
+    /// the columns that contain the input
+    pub input_cols: RangeInclusive<usize>,
+    /// the row that contain the output
+    pub output_row: CurrOrNext,
+    /// the columns that contain the output
+    pub output_cols: RangeInclusive<usize>,
 }
 
-pub const ROUND_EQUATIONS: [RoundEquation; ROUNDS_PER_ROW] = [
-    RoundEquation {
-        source: 0,
-        target: (Curr, 1),
+/// Specifies in which columns the input of a row is stored,
+/// as well as which row and columns the output the permutation is stored.
+/// A Poseidon gates performs 5 rounds, with the last 5 round outputing
+/// its result on the next row. Each round acts on a state of 3 columns.
+/// The layout is also shuffled, so that the 4th round of the gate is stored
+/// right after the first round. This is to let the permutation access it,
+/// in case the final output lands on this state, instead of the next row's.
+pub const ROUND_EQUATIONS: [RoundSpec; ROUNDS_PER_ROW] = [
+    RoundSpec {
+        input_cols: 0..=2,
+        output_row: Curr,
+        output_cols: 6..=8,
     },
-    RoundEquation {
-        source: 1,
-        target: (Curr, 2),
+    RoundSpec {
+        input_cols: 6..=8,
+        output_row: Curr,
+        output_cols: 9..=11,
     },
-    RoundEquation {
-        source: 2,
-        target: (Curr, 3),
+    RoundSpec {
+        input_cols: 9..=11,
+        output_row: Curr,
+        output_cols: 12..=14,
     },
-    RoundEquation {
-        source: 3,
-        target: (Curr, 4),
+    RoundSpec {
+        input_cols: 12..=14,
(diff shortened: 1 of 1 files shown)


kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -76,30 +87,28 @@ pub fn constraint<F: FftField + SquareRootField>() -> E<F> {
...
+            let res =
+                E::cell(Column::Witness(input_col), Curr).pow(PlonkSpongeConstants15W::SPONGE_BOX);
+            sboxed.push(cache.cache(res));
+        }
 
-        res.extend(round_to_cols(target_round).enumerate().map(|(j, col)| {
-            let rc = E::cell(Column::Coefficient(idx), Curr);
+        for (state_i, output_col) in eq.output_cols.clone().enumerate() {
+            // round constant
+            let mut output = E::cell(Column::Coefficient(round * 3 + state_i), Curr);
```
mrmr1993, line 101: If feels a bit strange to hard-code 3 here, but have the ranges be defined separately. I know that they match now, but this will be confusing if we change the number again.

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -1,73 +1,84 @@
...
+    pub output_cols: RangeInclusive<usize>,
 }
 
-pub const ROUND_EQUATIONS: [RoundEquation; ROUNDS_PER_ROW] = [
-    RoundEquation {
-        source: 0,
-        target: (Curr, 1),
+/// Specifies in which columns the input of a row is stored,
+/// as well as which row and columns the output the permutation is stored.
+/// A Poseidon gates performs 5 rounds, with the last 5 round outputing
```
mrmr1993, line 23: `s/gates/gate`

kimchi/src/circuits/polynomials/poseidon.rs
mrmr1993: Perhaps a `STATE_SIZE` constant instead of a literal 3 here?

kimchi/src/circuits/polynomials/poseidon.rs
mrmr1993: `s/outputing/outputting`

mrmr1993: `s/last 5 round/last round`

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -76,30 +87,28 @@ pub fn constraint<F: FftField + SquareRootField>() -> E<F> {
...
-        let sboxed: Vec<_> = round_to_cols(source)
-            .map(|i| {
-                cache.cache(
-                    E::cell(Column::Witness(i), Curr).pow(PlonkSpongeConstants15W::SPONGE_BOX),
-                )
-            })
-            .collect();
+    for (round, eq) in ROUND_EQUATIONS.iter().enumerate() {
+        // sbox
+        let mut sboxed = vec![];
```
mrmr1993, line 92: The size of this vector is known in advance, use `with_capacity` to avoid allocations with `push`.

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -76,30 +87,28 @@ pub fn constraint<F: FftField + SquareRootField>() -> E<F> {
...
+        for input_col in eq.input_cols.clone() {
```
mrmr1993, line 93: I find this mutability significantly less readable than the original.

mrmr1993 (changes requested):
I don't find this more readable. Left a few nits, but I don't think I would be happy to approve this unless it's only the doc-comments.

kimchi/src/circuits/polynomials/poseidon.rs
querolita: For this part, I am also inclined to think that the original approach seems more direct (...) _continue in line 102_

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -1,73 +1,84 @@
...
+/// Specifies in which columns the input of a row is stored,
+/// as well as which row and columns the output the permutation is stored.
+/// A Poseidon gates performs 5 rounds, with the last 5 round outputing
+/// its result on the next row. Each round acts on a state of 3 columns.
+/// The layout is also shuffled, so that the 4th round of the gate is stored
+/// right after the first round. This is to let the permutation access it,
+/// in case the final output lands on this state, instead of the next row's.
+pub const ROUND_EQUATIONS: [RoundSpec; ROUNDS_PER_ROW] = [
+    RoundSpec {
+        input_cols: 0..=2,
```
querolita, line 30: Do you think it could make sense to predefine some ranges to assign these - otherwise hardcoded - fields? I am referring to the pattern/cycle for input and output columns: 0-2 -> 6-8 -> 9-11 -> 12-14 -> 3-5 -> 0-2

querolita (commented):
I am fine with making any part of the codebase more readable and accessible. I think this PR made a good progress in this direction, particularly when defining the `RoundEquation`s in a more informative manner. But I would keep the spongebox part as in the original version, while accepting the changes for the `mds`.

kimchi/src/circuits/polynomials/poseidon.rs
mimoo: my opinion is that the imperative way is more idiomatic in Rust and more readable (unless you come from a functional language). Also it most of the time takes less lines of code.

kimchi/src/circuits/polynomials/poseidon.rs
mimoo: I'm not sure I understand the question exactly. We could pre-compute these ranges based on a `STATE_SIZE`, but we're then losing the clean layout written in this form. One idea here was to remove the clever calculation and explicitly write out the layout.

mrmr1993:
Noting here in case I forget: we discussed in our call that it would probably be cleaner to have the outer loop that maps over the constraints become a function, and then call that function with each of the inputs, rather than using that const list.

PR #286, opened 2022-02-08
[gate][chacha] refactor to distinguish the different subgates

mimoo:
I perform several changes to the chacha gate:

* separate the different chacha gates into different functions (the previous code wasn't making it clear that we're dealing with 4 gates)
* closures -> functions, this will facilitate changes when we use the alpha iterator later
* move functions for testing in a `testing` module

### DIFF
--- a/kimchi/src/circuits/polynomials/chacha.rs
+++ b/kimchi/src/circuits/polynomials/chacha.rs
@@ -142,22 +142,24 @@
 //! And we'll check that y' is the sum of the shifted nybbles.
 //!
 
-use crate::circuits::expr::{Column, ConstantExpr as C, E};
-use crate::circuits::gate::{CurrOrNext, GateType};
+use crate::circuits::{
+    expr::{index, witness, ConstantExpr as C, E},
+    gate::{CurrOrNext, GateType},
+};
 use ark_ff::{FftField, Field, Zero};
 use CurrOrNext::*;
 
-// The lookup table for 4-bit xor.
-// Note that it is constructed so that (0, 0, 0) is the last position in the table.
-//
-// This is because tables are extended to the full size of a column (essentially)
-// by padding them with their final value. And, having the value (0, 0, 0) here means
-// that when we commit to this table and use the dummy value in the `lookup_sorted`
-// columns, those entries that have the dummy value of
-//
-// 0 = 0 + joint_combiner * 0 + joint_combiner^2 * 0
-//
-// will translate into a scalar multiplication by 0, which is free.
+/// The lookup table for 4-bit xor.
+/// Note that it is constructed so that (0, 0, 0) is the last position in the table.
+///
+/// This is because tables are extended to the full size of a column (essentially)
+/// by padding them with their final value. And, having the value (0, 0, 0) here means
+/// that when we commit to this table and use the dummy value in the `lookup_sorted`
+/// columns, those entries that have the dummy value of
+///
+/// 0 = 0 + joint_combiner * 0 + joint_combiner^2 * 0
+///
+/// will translate into a scalar multiplication by 0, which is free.
 pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
     let mut res = vec![vec![]; 3];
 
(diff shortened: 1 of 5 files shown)


kimchi/src/index.rs
```diff
@@ -172,8 +172,12 @@ pub fn constraints_expr<F: FftField + SquareRootField>(
             )
         }
     };
+
     if chacha {
-        expr + chacha::constraint(super::range::CHACHA.start)
+        let expr = expr + chacha::constraint_chacha0(super::range::CHACHA.start);
```
mrmr1993, line 177: Why split these? It was more readable before, and it would be easy to create a function that calls them all.

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -164,203 +166,53 @@ pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
...
-    for _ in 0..10 {
-        for [a, b, c, d] in CHACHA20_QRS {
-            qr(a, b, c, d);
-        }
-    }
-
-    rows
-}
+/// Implementation details
+struct Impl<F>(PhantomData<F>);
```
mrmr1993, line 170: This doesn't appear to be doing anything. Delete?

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -164,203 +166,53 @@ pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
...
-    // like y^x' or x+z
-    let chunks_over_2_rows = |col_offset: usize| -> Vec<_> {
+    /// 8-nybble sequences that are laid out as 4 nybbles per row over the two row,
+    /// like y^x' or x+z
+    fn chunks_over_2_rows(col_offset: usize) -> Vec<E<F>> {
         (0..8)
             .map(|i| {
                 let r = if i < 4 { Curr } else { Next };
-                w(col_offset + (i % 4), r)
+                Self::w(col_offset + (i % 4), r)
```
mrmr1993, line 186: `Self::w` is the same number of characters as `witness`.. Why don't you move `w` to `Expr`, call it `witness`, and use it throughout?

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -164,203 +166,53 @@ pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
...
-    };
-    for _ in 0..10 {
-        for [a, b, c, d] in CHACHA20_QRS {
-            qr(a, b, c, d);
-        }
+impl<F> Impl<F>
+where
+    F: FftField,
+{
+    fn w(i: usize, r: CurrOrNext) -> E<F> {
```
mrmr1993, line 176: This is not a function of the 'chacha implementation', probably should not live here.

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -371,63 +223,233 @@ pub fn constraint<F: FftField>(alpha0: usize) -> E<F> {
...
-        let low_bits = chunks_over_2_rows(5);
-        let yprime = w(0, Curr);
+    }
+}
 
-        let one_half = F::from(2u64).inverse().unwrap();
+/// a += b; d ^= a; d <<<= 16 (=4*4)
+pub fn constraint_chacha0<F: FftField>(alpha0: usize) -> E<F> {
+    let index = |g: GateType| E::cell(Column::Index(g), Curr);
+    index(GateType::ChaCha0) * Impl::line(alpha0, 4)
```
mrmr1993, line 243: Duplicated code here isn't ideal. It would be preferable to have
```rust
pub fn constraint_shift_by_nybbles<F: FftField>(alpha0: usize, gate_type: GateType) -> E<F> {
    let line_number =
        match gate_type {
            GateType::ChaCha0 => 4,
            GateType::ChaCha1 => 3,
            GateType::ChaCha2 => 2,
            _ => panic!("The gate type {} is not a nybble-shifting ChaCha gate!", gate_type)
        };
    let index = |g: GateType| E::cell(Column::Index(g), Curr);
    index(gate_type) * Impl::line(alpha0, line_number)
}
```

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -164,203 +166,53 @@ pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
...
-}
-
-// TODO: Assert somehow that we're using the right number of alphas.
-pub fn constraint<F: FftField>(alpha0: usize) -> E<F> {
-    let w = |i: usize, r: CurrOrNext| E::<F>::cell(Column::Witness(i), r);
 
-    // 8-nybble sequences that are laid out as 4 nybbles per row over the two row,
-    // like y^x' or x+z
-    let chunks_over_2_rows = |col_offset: usize| -> Vec<_> {
+    /// 8-nybble sequences that are laid out as 4 nybbles per row over the two row,
```
mrmr1993, line 180: Is this a doc-comment on a private function? :thinking:

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -164,203 +166,53 @@ pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
...
             })
-    };
+    }
 
-    // Constraints for the line L(x, x', y, y', z, k), where k = 4 * nybble_rotation
-    let line = |nybble_rotation: usize| -> E<F> {
-        let y_xor_xprime_nybbles = chunks_over_2_rows(3);
-        let x_plus_z_nybbles = chunks_over_2_rows(7);
-        let y_nybbles = chunks_over_2_rows(11);
+    /// Constraints for the line L(x, x', y, y', z, k), where k = 4 * nybble_rotation
```
mrmr1993, line 203: Private function doc-comment?

mrmr1993 (changes requested):
I feel like this refactor makes the code worse. I am opposed to merging this.

kimchi/src/index.rs
mimoo: the idea was that they are separate gates, so let's make that explicit

kimchi/src/circuits/polynomials/chacha.rs
mimoo: yeah, you can still export private function comments if you want with a flag, so that's useful :o

kimchi/src/circuits/polynomials/chacha.rs
mimoo: that was a hack to avoid having to declare the trait bounds on every functions, but yeah I agree it's ugly...

kimchi/src/circuits/polynomials/chacha.rs
mimoo: The idea is to make this more explicit, so I think it's better to maintain a 3 lines of duplicated code

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -180,270 +182,282 @@ pub fn xor_table<F: Field>() -> Vec<Vec<F>> {
...
+    (0..8)
+        .map(|i| {
+            let r = if i < 4 { Curr } else { Next };
+            witness(col_offset + (i % 4), r)
+        })
+        .collect()
+}
+
+/// Creates a constraint to enforce that b is either 0 or 1.
+fn boolean<F: Field>(b: &E<F>) -> E<F> {
```
mrmr1993, line 201: I would be tempted to move this to `expr` as well, since it's only related to chacha in the sense that its gate uses it, but we use it in several other gates.

mimoo: ok, implementing this in another PR ;)

mimoo: https://github.com/o1-labs/proof-systems/pull/290

mrmr1993: I don't like that that PR comes *after* this one, since it leads to 'approving' a change that I don't approve of. In future, I would prefer if the equivalent of #290 was against master and the equivalent of this PR was built on top, by merging the new PR's branch and fixing any changes.

mrmr1993 (approved):

PR #287, opened 2022-02-09
Move table_id from SingleLookup to JointLookup

mrmr1993:
This PR moves the lookup `table_id` from `SingleLookup` to `JointLookup`.

Since we never want to do a 'joint lookup' from multiple tables -- the values we look up will all come from the same table, combined in the same row -- it makes sense for it to live on the `JointLookup` instead.

This is a small part of the work towards implementing #250.


jspada (approved):
lgtm!

PR #288, opened 2022-02-09
prover/verifier cleaning

mimoo:
*  inline `combine_evaluations`
* change some variable names to improve readability
* add tests for each gate within the prover

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -12,7 +12,7 @@ use crate::circuits::{
 use crate::plonk_sponge::FrSponge;
 use ark_ec::AffineCurve;
 use ark_ff::UniformRand;
-use ark_ff::{FftField, Field, One, PrimeField, Zero};
+use ark_ff::{Field, One, PrimeField, Zero};
 use ark_poly::{
     univariate::DensePolynomial, Evaluations, Polynomial, Radix2EvaluationDomain as D, UVPolynomial,
 };
@@ -20,6 +20,7 @@ use array_init::array_init;
 use commitment_dlog::commitment::{
     b_poly_coefficients, CommitmentCurve, CommitmentField, OpeningProof, PolyComm,
 };
+use itertools::Itertools;
 use lookup::CombinedEntry;
 use o1_utils::ExtendedDensePolynomial;
 use oracle::{rndoracle::ProofError, sponge::ScalarChallenge, FqSponge};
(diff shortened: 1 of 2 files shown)


kimchi/src/prover.rs
```diff
@@ -424,79 +395,175 @@ where
...
-                drop(chacha);
-                t4
+
+                if cfg!(test) {
+                    let (_, res) = chacha
+                        .clone()
+                        .interpolate()
+                        .divide_by_vanishing_poly(index.cs.domain.d1)
+                        .unwrap();
+                    assert!(res.is_zero());
```
mimoo, line 520: (btw this is the test that made me change the panic into a TODO elsewhere)

kimchi/src/prover.rs
```diff
@@ -424,79 +395,175 @@ where
...
-        // divide contributions with vanishing polynomial
-        let (mut t, res) = (&(&t4.interpolate() + &t8.interpolate()) + &public_poly)
-            .divide_by_vanishing_poly(index.cs.domain.d1)
-            .map_or(Err(ProofError::PolyDivision), Ok)?;
-        if !res.is_zero() {
-            return Err(ProofError::PolyDivision);
-        }
+                let constraints =
+                    lookup::constraints(&lcs.dummy_lookup_values[0], index.cs.domain.d1);
+                for (constraint, alpha_pow) in constraints.into_iter().zip_eq(pows) {
```
mrmr1993, line 534: I'm confused by `zip_eq` here; isn't the alpha iterator infinite?

kimchi/src/prover.rs
```diff
@@ -424,79 +395,175 @@ where
...
-                lookup::constraints(&lcs.dummy_lookup_values[0], index.cs.domain.d1)
-                    .iter()
-                    .map(|e| e.evaluations(&env))
-                    .collect(),
-            ),
-        };
+            // lookup
+            if let Some(lcs) = index.cs.lookup_constraint_system.as_ref() {
+                let mut alpha_pow = alphas[alphas.len() - 1];
+                let pows = (0..).map(|_| {
```
mrmr1993, line 527: It makes more sense (to me, at least) to do this in the loop below rather than constructing an iterator.

o1-david: are you commenting on the power of alpha thing or the stuff below? For the powers of alpha it doesn't matter too much as I'm going to remove this in the alpha iterator PR

mimoo: note that I'm going to remove this code in the alpha iterator PR

mimoo: also this code is not correct, the `(0..)` is going to time out. I guess we don't have tests for lookup?

kimchi/src/prover.rs
mimoo: good catch, this is an artifact of the alpha iterator PR (since I spill out the correct number of powers of alpha, we can use a `zip_eq`). But I'm confused how the tests are passing with this.

kimchi/src/prover.rs
mimoo: which prompted me to check wth was happening, and https://github.com/o1-labs/proof-systems/pull/289

thanks for catching these :D

mrmr1993 (approved):

Issue #271, opened 2022-01-21
gates should implement an interface


mimoo:
* every gate should implement the same clear interface
* every gate should also return a vector of constraints, and we should have some logic that always combine these constraints (so we can't forget to combine them)

cc @jspada @querolita

mimoo:
Currently the prover code looks like a lot of repeat of this:

```rust
            let emulscalar8 =
                endomul_scalar::constraint(range::ENDOMUL_SCALAR.start).evaluations(&env);
            t8 += &emulscalar8;

            if cfg!(test) {
                let (_, res) = emulscalar8
                    .clone()
                    .interpolate()
                    .divide_by_vanishing_poly(index.cs.domain.d1)
                    .unwrap();
                assert!(res.is_zero());
            }
```

it'd be really awesome if we can just have something like that:

```rust
for gate in gates {
   let gate_eval = gate.evaluations(&env);
   test_gate!(gate_eval);
   evaluation += gate_eval;
}
```

querolita:
Actually I thought we were thinking of some trait for gates to have a common set of functions such as `create`, `verify`, `gadget` or similar.

mimoo:
I have a PR out to add an interface for gates: https://github.com/o1-labs/proof-systems/pull/295

I don't think we can create an interface for create, as different gates take different inputs. Although... saying that I think the following interface would work:

```rust
pub trait Gate {
   type Inputs;
   fn create(input: Inputs, ...) -> ...
}
```

Issue #249, opened 2021-12-09
benchmarking


mimoo:
We can run benchmark in CI by applying the `benchmark` label to a PR. It is not perfect because it runs in the Github action host. To bypass that, we use [iai](https://github.com/bheisler/iai) to measure the number of CPU cycles, which should work well in CI. It's not a really nice interface though, and it doesn't less us easily isolate what we want to bench. Ideally, CI should tell us for every single PR if we're regressing performance-wise. Here's a list of things that we can work on to gradually improve the situation:

- [ ] run benchmarks on bare metal (AWS or something) for benchmarks
- [ ] run benchmarks on both master, and the latest PR, and post a comment with a comparison
- [ ] I think we can still keep iai, but if we do then try to clean up the cycles (take the cycles with setup only, and then without setup, and subtract. For verifier time, since it's so small, do a LOT of verifications to reach a larger number of cycles that can be compared more easily?)
- [ ] run benchmarks on every PR
- [ ] run benchmark with the same SRS size we use for Mina, and a similar circuit size
- [ ] run benchmarks for both the prover time, and the verification

PS: this GHA looks cool: https://github.com/marketplace/actions/continuous-benchmark btw

PR #295, opened 2022-02-11
Implementing an Argument interface

mimoo:
This PR adds a new Argument interface to represent the three types of arguments we have: gates, (p)lookup, permutation.

```rust
/// A constraint type represents a polynomial that will be part of the final equation f (the circuit equation)
pub enum ArgumentType {
    /// Gates in the PLONK constraint system.
    /// As gates are mutually exclusive (a single gate is set per row),
    /// we can reuse the same powers of alpha across gates.
    Gate(GateType),
    /// The permutation argument
    Permutation,
    /// The lookup argument
    Lookup,
}

/// The interface for a minimal argument implementation.
pub trait Argument {
    /// The field being used for this argument.
    type Field: FftField;

    /// The type of constraints that this will produce.
    /// This is important to enforce that we don't combine the constraints
    /// with powers of alpha that collide with other mutually inclusive arguments.
    const ARGUMENT_TYPE: ArgumentType;

    /// The number of constraints created by the argument.
    const CONSTRAINTS: usize;

    /// Returns the set of constraints required to prove this argument.
    fn constraints(&self) -> Vec<E<Self::Field>>;

    /// Returns constraints safely combined via the passed combinator.
    fn combined_constraints(&self, alphas: &Alphas<Self::Field>) -> E<Self::Field> {
        let constraints = self.constraints();
        assert!(constraints.len() == Self::CONSTRAINTS);
        let alphas = alphas.get_exponents(Self::ARGUMENT_TYPE, Self::CONSTRAINTS);
        let combined_constraints = E::combine_constraints(alphas, constraints);

        // An optional gate type, if used to define a gate.
        // This is used to filter the gate, to avoid applying it on the entire domain.
        if let ArgumentType::Gate(gate_type) = Self::ARGUMENT_TYPE {
            index(gate_type) * combined_constraints
        } else {
            combined_constraints
        }
    }
}
```

Following this, the alpha iterator can be heavily simplified. The rest of this PR applies the changes to all the arguments.

This should close https://github.com/o1-labs/proof-systems/issues/271


kimchi/src/circuits/argument.rs
```diff
@@ -0,0 +1,23 @@
...
+
+/// The interface for a minimal argument implementation
+pub trait Argument {
+    type Field: FftField;
+
+    /// The number of constraints created by the argument
+    const CONSTRAINTS: usize;
+
+    /// Returns constraints safely combined via the passed combinator
+    fn constraint(&self, combinator: &Alphas<Self::Field>) -> Expr<ConstantExpr<Self::Field>>;
```
mrmr1993, line 22: Would it make more sense to return an `[Expr<ConstantExpr<Self::Field>>; Self::CONSTRAINTS]`? It seems like `constraint<T: Argument>(&self) -> Expr<..>` could be written once to combine them, rather than doing it multiple times times

kimchi/src/circuits/argument.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use ark_ff::FftField;
+
+use crate::{
+    alphas::Alphas,
+    circuits::expr::{ConstantExpr, Expr},
+};
+
+/// The interface for a minimal argument implementation
+pub trait Argument {
```
mrmr1993, line 15: Perhaps it's not the intention here, but a `Gate` interface could also expose the selector, so we could combine that automatically

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -254,50 +248,122 @@ fn line<F: Field>(nybble_rotation: usize) -> Vec<E<F>> {
...
+
+impl<F> Argument for ChaCha0<F>
+where
+    F: FftField,
+{
+    type Field = F;
+    const CONSTRAINTS: usize = 5;
+
+    fn constraint(&self, alphas: &Alphas<F>) -> E<F> {
+        // a += b; d ^= a; d <<<= 16 (=4*4)
```
mrmr1993, line 263: Pity to repeat this for each

kimchi/src/index.rs
```diff
@@ -163,36 +172,35 @@ pub fn constraints_expr<F: FftField + SquareRootField>(
...
-    let alphas = powers_of_alpha.register(ConstraintType::Gate, largest_constraint_num);
+    let highest_constraints = VarbaseMul::<F>::CONSTRAINTS;
+    powers_of_alpha.register(ConstraintType::Gate, highest_constraints);
 
-    let mut expr = poseidon::constraint(alphas.clone().take(poseidon::CONSTRAINTS));
-    expr += varbasemul::constraint(alphas.clone().take(varbasemul::CONSTRAINTS));
-    expr += complete_add::constraint(alphas.clone().take(complete_add::CONSTRAINTS));
-    expr += endosclmul::constraint(alphas.clone().take(endosclmul::CONSTRAINTS));
-    expr += endomul_scalar::constraint(alphas.clone().take(endomul_scalar::CONSTRAINTS));
+    let mut expr = Poseidon::default().constraint(&powers_of_alpha);
```
mrmr1993, line 183: It might be cleaner to have a function that returns an `impl trait Argument` than do the default dance on a dummy object. Wdyt?

kimchi/src/circuits/argument.rs
mimoo: image

looks like const generics don't quite work yet, that's a bummer but we can fix that at some point in the future :o

kimchi/src/alphas.rs
```diff
@@ -38,120 +39,114 @@ use std::{
...
+            // the zero gate is not used, so we default to it
+            ArgumentType::Gate(GateType::Zero)
+        } else {
+            ty
+        };
 
-        Alphas {
-            alphas,
-            mapping: powers.mapping.clone(),
+        if self.mapping.insert(ty, range).is_some() {
```
querolita, line 77: I am a bit confused about the functionality of this check, but perhaps a short comment could solve it. In particular, it's about how the ranges increase, or when this `register()` method should be invoked. Before the new alphas merged, we used to have ranges that grew consecutively together with the `GateType`, up to around power 79 or so for `ChaChaFinal`. Then, after the alphas PR, I was thinking we were sticking to 21 powers or so (the largest number of constraints, in any available gate). And as long as the selector polynomials were mutually exclusive, that was fine. 
But here, it seems to suggest that one could potentially have "duplicate" entries of `GateTypes` with different ranges, right? I am assuming this won't ever be happening in the code, as this `register()` method will only be called once per `GateType`. But I would like to double check with you that this is indeed the case. If so, perhaps it is safer if the mapping checked only the `ty` part, regardless the range? Or is there a reason why the check is made on the full tuple `(ty, range)`?

kimchi/src/circuits/argument.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// The number of constraints created by the argument.
+    const CONSTRAINTS: usize;
+
+    /// Returns the set of constraints required to prove this argument.
+    // TODO: return a [_; Self::CONSTRAINTS] once generic consts are stable
+    fn constraints(&self) -> Vec<E<Self::Field>>;
+
+    /// Returns constraints safely combined via the passed combinator.
+    fn combined_constraints(&self, alphas: &Alphas<Self::Field>) -> E<Self::Field> {
```
mrmr1993, line 44: As above, so below

kimchi/src/circuits/argument.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// This is important to enforce that we don't combine the constraints
+    /// with powers of alpha that collide with other mutually inclusive arguments.
+    const ARGUMENT_TYPE: ArgumentType;
+
+    /// The number of constraints created by the argument.
+    const CONSTRAINTS: usize;
+
+    /// Returns the set of constraints required to prove this argument.
+    // TODO: return a [_; Self::CONSTRAINTS] once generic consts are stable
+    fn constraints(&self) -> Vec<E<Self::Field>>;
```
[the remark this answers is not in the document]
mimoo: I need `self` for the plookup argument

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -518,10 +567,12 @@ mod tests {
             h.insert(Column::Index(GateType::Generic));
             h
         };
-        let mut expr = constraint_chacha0(0..5);
-        expr += constraint_chacha1(0..5);
-        expr += constraint_chacha2(0..5);
-        expr += constraint_chacha_final(0..9);
+        let mut alphas = Alphas::<F>::default();
+        alphas.register(ArgumentType::Gate(GateType::ChaChaFinal), 9);
```
querolita, line 571: perhaps a counter for this 9? I bet this was suggested before, not so important I guess

querolita (commented):
Great ideas to handle alphas and essentially gates better

kimchi/src/circuits/argument.rs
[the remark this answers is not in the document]
mimoo: thinking a bit more about that, we could potentially augment the interface like this:

```rust
trait Argument {
  type Input;
  fn constraints(input: Input) -> ...
}
```

what do you think about experimenting with these in another PRs though? Same for the other comment about hiding the ugliness :o

kimchi/src/index.rs
mimoo: yeah in general I agree, but I think you also end up making things a bit too magical without really reducing the complexity. There's a trade off.

kimchi/src/alphas.rs
```diff
@@ -38,120 +39,114 @@ use std::{
...
 /// This type can be used to create a mapping between powers of alpha and constraint types.
-/// See [Builder::default] to create one,
+/// See [Alphas::default] to create one,
 /// and [Builder::register] to register a new mapping.
 /// Once you know the alpha value, you can convert this type to a [Alphas].
 #[derive(Debug, Default, Serialize, Deserialize, Clone)]
-pub struct Builder {
+pub struct Alphas<F> {
     /// The next power of alpha to use
     /// the end result will be [1, alpha^next_power)
```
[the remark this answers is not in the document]
mimoo: true!

kimchi/src/alphas.rs
```diff
@@ -38,120 +39,114 @@ use std::{
...
+    /// If set to [Some], you can't register new constraints.
+    alphas: Option<Vec<F>>,
 }
 
-impl Builder {
-    /// Registers a new [ConstraintType],
+impl<F: Field> Alphas<F> {
+    /// Registers a new [ArgumentType],
     /// associating it with a number `powers` of powers of alpha.
-    /// The function returns an iterator of powers that haven't been used yet.
```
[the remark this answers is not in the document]
mimoo: good idea

kimchi/src/alphas.rs
mimoo: If I understand your question correctly, you're misreading this line: 

* this inserts `range` under the mapping for `ty` (`Gate`, `Permutation`, or `Lookup`). 
* Importantly, if this is a `Gate(_)`, the range will always be stored under `Gate(Zero)` due to the line above (so you CAN'T have disjoint ranges for different gates). 
* The result of inserting is either `None` (all good, we never registered that argument before), or the previous value `Some(_)` in which case we panic because there's already a range registered under that argument.

kimchi/src/alphas.rs
```diff
@@ -38,120 +39,114 @@ use std::{
...
         }
     }
 
+    /// Instantiates the ranges with an actual field element `alpha`.
+    /// Once you call this function, you cannot register new constraints via [register].
+    pub fn instantiate(&mut self, alpha: F) {
+        let mut last_power = F::one();
+        let mut alphas = Vec::with_capacity(self.next_power);
+        alphas.push(F::one());
+        for _ in 0..(self.next_power - 1) {
```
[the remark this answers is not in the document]
mimoo: I feel like whenever you read this kind of lines your brain needs to active anyway, we can never write ranges like that in an obvious way (or at least I never found a way to do it). Your suggestion takes less room though so I'll adopt it :D

kimchi/src/alphas.rs
querolita: Oooh I see! I think I was misunderstanding the possible values that `ty` could take! In particular, I had forgotten that it was only `Gate`, `Permutation` and `Lookup`, and I was assuming it took any `GateType` such as `Poseidon` or `ChaCha1`. Thanks, this clarifies everything.

kimchi/src/alphas.rs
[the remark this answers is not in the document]
querolita: Yeah, I guess one needs to pay attention to the details anyway 😄

mimoo:
compared with the bench on master: https://github.com/o1-labs/proof-systems/pull/303#issuecomment-1041110002 looks good :D it'd be nicer if we could have a bench that gives you the comparison with master directly though (https://github.com/o1-labs/proof-systems/issues/249)

kimchi/src/circuits/argument.rs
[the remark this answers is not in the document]
mrmr1993: This PR is an experiment itself. I'd prefer if we limited ourselves to just the gates for this PR, since they can be done cleanly. Perhaps there's then a nice abstraction for the non-local constraints, but it need not be the same one IMO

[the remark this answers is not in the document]
mimoo: I'm going to argue that we should get this in as this is much cleaner for the permutation and the lookup as well. It shows exactly where we create the constraints and it allows us to use the `Alphas` quite easily with it as well.

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -254,50 +245,105 @@ fn line<F: Field>(nybble_rotation: usize) -> Vec<E<F>> {
 // Gates
 //
 
-/// a += b; d ^= a; d <<<= 16 (=4*4)
-pub fn constraint_chacha0<F: FftField>(alphas: impl Iterator<Item = usize>) -> E<F> {
-    index(GateType::ChaCha0) * E::combine_constraints(alphas, line(4))
+/// Implementation of the ChaCha0 gate
+#[derive(Default)]
+pub struct ChaCha0<F>(PhantomData<F>);
```
querolita, line 250: Oh so you're redefining a `ChaCha0` that is a struct (and the same for each of the `GateTypes`). This explains what I was confused about in our meeting: meaning, I once tried to implement a trait for an enum variant (e.g. GateType::ChaCha0), and rust didn't seem to support that directly. So the reason that you can do this now is ChaCha0 here is a struct. Would it make sense to have some data inside it, instead of just PhantomData? I mean, the corresponding GateType, so that later on when you're defining the `const ARGUMENT_TYPE` you could just access the content of the struct? Well... I guess it is not so much better, it will perhaps take more rows than simply writing manually GateType::ChaCha0 in the very single place that you need it...

mimoo: I'd say it wouldn't make sense to store data that is known at compile-time. The `PhantomData` is a trick to make sure we can use the generic field `F`.

kimchi/src/circuits/argument.rs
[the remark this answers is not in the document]
querolita: I agree with David on the cleaner aspect of this approach. I am not saying this is not subject to further improvements, but so far I like the idea of having this `ArgumentType` for the three possible variants. We can think of ways to hide some ugliness in a future PR. And the same goes for a slightly different treatment of the non local constraints. But accepting this PR only for the gates would become a bit tricky, as we would be using quite different approaches for the other types of arguments, which I think could make the code a bit less readable.

kimchi/src/circuits/polynomials/chacha.rs
querolita: Ok thanks :)

kimchi/src/circuits/argument.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// The permutation argument
+    Permutation,
+    /// The lookup argument
+    Lookup,
+}
+
+/// The interface for a minimal argument implementation.
+pub trait Argument {
+    /// The field being used for this argument.
+    type Field: FftField;
```
mrmr1993, line 29: Is there a reason to do this vs. an `Argument<F: FftField>` trait bound? It currently feels redundant, given that we do an `impl<F: FftField> Argument for Foo` everywhere.

mrmr1993 (approved):
Nice! Left a nit.

kimchi/src/circuits/argument.rs
mimoo: That's what I did first, but I couldn't make it work for some reason. I admit that I don't remember what was the issue, but it'd be nice to make it work.

PR #299, opened 2022-02-15
Move LookupConstraintSystem creation to its own function

mrmr1993:
This PR adds a `LookupConstraintSystem::create` function, which abstracts out the logic for creating a `LookupConstraintSystem`. This will allow us to call this function as part of generating the linearisation when we want to implement plookup in pickles.

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -319,6 +319,60 @@ pub enum GateError {
     Custom { row: usize, err: String },
 }
 
+impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
+    pub fn create(
+        gates: &Vec<CircuitGate<F>>,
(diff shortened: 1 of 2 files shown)


kimchi/src/circuits/constraints.rs
```diff
@@ -319,6 +319,60 @@ pub enum GateError {
...
+                        table_poly.push(poly);
+                        table_eval.push(eval);
+                    }
+                    lookup_tables_polys.push(table_poly);
+                    lookup_tables8.push(table_eval);
+                }
+
+                // generate the look up selector polynomials
+                let lookup_selectors = lookup_info.selector_polynomials(domain, gates);
+                Some(LookupConstraintSystem {
```
mimoo, line 362: nit: you can return `Self`

mimoo (approved):

PR #300, opened 2022-02-15
Automatically infer and expose used tables from used gates

mrmr1993:
This PR
* tracks the table associated with each gate that uses lookups
* generates the table when used, automatically inserting it into the list of tables considered when creating the proof
* removes the explicit use of `xor_table` in the chacha test

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -331,9 +331,13 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
             Some(lookup_used) => {
                 let d1_size = domain.d1.size();
 
+                let (lookup_selectors, gate_lookup_tables) =
+                    lookup_info.selector_polynomials_and_tables(domain, gates);
+
                 // get the last entry in each column of each table
-                let dummy_lookup_values: Vec<Vec<F>> = lookup_tables
+                let dummy_lookup_values: Vec<Vec<F>> = gate_lookup_tables
                     .iter()
+                    .chain(lookup_tables.iter())
                     .map(|table| table.iter().map(|col| col[col.len() - 1]).collect())
                     .collect();
 
@@ -341,7 +345,11 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
                 let mut lookup_tables_polys: Vec<Vec<DP<F>>> = vec![];
                 let mut lookup_tables8: Vec<Vec<E<F, D<F>>>> = vec![];
 
-                for (table, dummies) in lookup_tables.into_iter().zip(&dummy_lookup_values) {
+                for (table, dummies) in gate_lookup_tables
+                    .into_iter()
+                    .chain(lookup_tables.into_iter())
+                    .zip(&dummy_lookup_values)
+                {
                     let mut table_poly = vec![];
                     let mut table_eval = vec![];
                     for (mut col, dummy) in table.into_iter().zip(dummies) {
@@ -358,7 +366,6 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
                 }
 
                 // generate the look up selector polynomials
-                let lookup_selectors = lookup_info.selector_polynomials(domain, gates);
                 Some(LookupConstraintSystem {
                     lookup_selectors,
                     dummy_lookup_values,
--- a/kimchi/src/circuits/gate.rs
+++ b/kimchi/src/circuits/gate.rs
@@ -159,6 +159,9 @@ pub struct LookupInfo<F> {
     /// A map from the kind of gate (and whether it is the current row or next row) to the lookup
     /// constraint (given as an index into `kinds`) that should be applied there, if any.
     pub kinds_map: HashMap<(GateType, CurrOrNext), usize>,
+    /// A map from the kind of gate (and whether it is the current row or next row) to the lookup
+    /// table that is used by the gate, if any.
+    pub kinds_tables: HashMap<(GateType, CurrOrNext), GateLookupTable>,
     /// The maximum length of an element of `kinds`. This can be computed from `kinds`.
     pub max_per_row: usize,
     /// The maximum joint size of any joint lookup in a constraint in `kinds`. This can be computed from `kinds`.
@@ -179,11 +182,17 @@ pub enum LookupsUsed {
     Joint,
 }
 
+/// Enumerates the different 'fixed' lookup tables used by individual gates
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub enum GateLookupTable {
+    Xor,
+}
+
 impl<F: FftField> LookupInfo<F> {
     /// Create the default lookup configuration.
     pub fn create() -> Self {
-        let (kinds, kinds_set): (Vec<_>, Vec<_>) =
-            GateType::lookup_kinds::<F>().into_iter().unzip();
+        let (kinds, locations_with_tables): (Vec<_>, Vec<_>) = GateType::lookup_kinds::<F>();
+        let (kinds_map, kinds_tables) = GateType::lookup_kinds_map::<F>(locations_with_tables);
         let max_per_row = max_lookups_per_row(&kinds);
         LookupInfo {
             max_joint_size: kinds.iter().fold(0, |acc0, v| {
@@ -191,7 +200,8 @@ impl<F: FftField> LookupInfo<F> {
                     .fold(acc0, |acc, j| std::cmp::max(acc, j.entry.len()))
             }),
 
-            kinds_map: GateType::lookup_kinds_map::<F>(kinds_set),
+            kinds_map,
+            kinds_tables,
             kinds: kinds,
             max_per_row,
             empty: vec![],
@@ -275,7 +285,10 @@ impl GateType {
     /// See circuits/kimchi/src/polynomials/chacha.rs for an explanation of
     /// how these work.
     #[allow(clippy::type_complexity)]
-    pub fn lookup_kinds<F: Field>() -> Vec<(Vec<JointLookup<F>>, HashSet<(GateType, CurrOrNext)>)> {
+    pub fn lookup_kinds<F: Field>() -> (
+        Vec<Vec<JointLookup<F>>>,
+        Vec<(HashSet<(GateType, CurrOrNext)>, Option<GateLookupTable>)>,
+    ) {
         let curr_row = |column| LocalPosition {
             row: CurrOrNext::Curr,
             column,
@@ -336,26 +349,51 @@ impl GateType {
             chacha_final_where.insert((ChaChaFinal, *r));
         }
 
-        vec![
-            (chacha_pattern, chacha_where),
-            (chacha_final_pattern, chacha_final_where),
-        ]
+        let lookups = [
+            (chacha_pattern, chacha_where, Some(GateLookupTable::Xor)),
+            (
+                chacha_final_pattern,
+                chacha_final_where,
+                Some(GateLookupTable::Xor),
+            ),
+        ];
+
+        // Convert from an array of tuples to a tuple of vectors
+        {
+            let mut patterns = Vec::with_capacity(lookups.len());
+            let mut locations_with_tables = Vec::with_capacity(lookups.len());
+            for (pattern, location, table) in lookups {
+                patterns.push(pattern);
+                locations_with_tables.push((location, table));
+            }
+            (patterns, locations_with_tables)
+        }
     }
 
     pub fn lookup_kinds_map<F: Field>(
-        lookup_kinds: Vec<HashSet<(GateType, CurrOrNext)>>,
-    ) -> HashMap<(GateType, CurrOrNext), usize> {
-        let mut res = HashMap::new();
-        for (i, locs) in lookup_kinds.into_iter().enumerate() {
-            for (g, r) in locs {
-                if let std::collections::hash_map::Entry::Vacant(e) = res.entry((g, r)) {
+        locations_with_tables: Vec<(HashSet<(GateType, CurrOrNext)>, Option<GateLookupTable>)>,
+    ) -> (
+        HashMap<(GateType, CurrOrNext), usize>,
+        HashMap<(GateType, CurrOrNext), GateLookupTable>,
+    ) {
+        let mut index_map = HashMap::with_capacity(locations_with_tables.len());
+        let mut table_map = HashMap::with_capacity(locations_with_tables.len());
+        for (i, (locs, table_kind)) in locations_with_tables.into_iter().enumerate() {
+            for location in locs {
+                if let std::collections::hash_map::Entry::Vacant(e) = index_map.entry(location) {
                     e.insert(i);
                 } else {
                     panic!("Multiple lookup patterns asserted on same row.")
                 }
+                if let Some(table_kind) = table_kind {
+                    if let std::collections::hash_map::Entry::Vacant(e) = table_map.entry(location)
+                    {
+                        e.insert(table_kind);
+                    }
+                }
             }
         }
-        res
+        (index_map, table_map)
     }
 }
 
--- a/kimchi/src/circuits/gate.rs
+++ b/kimchi/src/circuits/gate.rs
@@ -208,7 +208,7 @@ impl<F: FftField> LookupInfo<F> {
 
             kinds_map,
             kinds_tables,
-            kinds: kinds,
+            kinds,
             max_per_row,
             empty: vec![],
         }
@@ -235,6 +235,7 @@ impl<F: FftField> LookupInfo<F> {
 
     /// Each entry in `kinds` has a corresponding selector polynomial that controls whether that
     /// lookup kind should be enforced at a given row. This computes those selector polynomials.
+    #[allow(clippy::type_complexity)]
     pub fn selector_polynomials_and_tables(
         &self,
         domain: &EvaluationDomains<F>,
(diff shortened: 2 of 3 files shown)


kimchi/src/circuits/gate.rs
```diff
@@ -235,6 +235,7 @@ impl<F: FftField> LookupInfo<F> {
 
     /// Each entry in `kinds` has a corresponding selector polynomial that controls whether that
     /// lookup kind should be enforced at a given row. This computes those selector polynomials.
+    #[allow(clippy::type_complexity)]
```
mimoo, line 238: Is there a way to make these returned types clearer? I think Clippy is right to warn here.

kimchi/src/circuits/gate.rs
```diff
@@ -189,16 +182,17 @@ pub enum LookupsUsed {
 impl<F: FftField> LookupInfo<F> {
     /// Create the default lookup configuration.
     pub fn create() -> Self {
-        let kinds = lookup_kinds::<F>();
+        let (kinds, kinds_set): (Vec<_>, Vec<_>) =
```
mimoo: The type signature seems unecessary

kimchi/src/circuits/gate.rs
```diff
@@ -189,16 +182,17 @@ pub enum LookupsUsed {
...
         LookupInfo {
             max_joint_size: kinds.iter().fold(0, |acc0, v| {
                 v.iter()
                     .fold(acc0, |acc, j| std::cmp::max(acc, j.entry.len()))
             }),
 
-            kinds_map: GateType::lookup_kinds_map::<F>(),
-            kinds,
+            kinds_map: GateType::lookup_kinds_map::<F>(kinds_set),
+            kinds: kinds,
```
mimoo: `kinds,` is enough

kimchi/src/circuits/gate.rs
```diff
@@ -380,24 +398,31 @@ impl GateType {
...
+                locations_with_tables.push(GatesLookupSpec {
+                    gate_positions: locations,
+                    gate_lookup_table: table,
+                });
             }
             (patterns, locations_with_tables)
         }
     }
 
     #[allow(clippy::type_complexity)]
```
mimoo, line 412: You forgot to remove this :)

kimchi/src/circuits/gate.rs
```diff
@@ -342,25 +385,62 @@ impl GateType {
...
     }
 
-    pub fn lookup_kinds_map<F: Field>() -> HashMap<(GateType, CurrOrNext), usize> {
-        let mut res = HashMap::new();
-        let lookup_kinds = Self::lookup_kinds::<F>();
-        for (i, (_, locs)) in lookup_kinds.into_iter().enumerate() {
-            for (g, r) in locs {
-                if let std::collections::hash_map::Entry::Vacant(e) = res.entry((g, r)) {
+    #[allow(clippy::type_complexity)]
+    pub fn lookup_kinds_map<F: Field>(
```
mimoo, line 413: What does function do? Could use some doc

kimchi/src/circuits/gate.rs
```diff
@@ -273,6 +298,24 @@ impl<F: FftField> LookupInfo<F> {
     }
 }
 
+/// Specifies the relative position of gates and the fixed lookup table (if applicable) that a
```
mimoo, line 301: No biggy, but I always try to line break after punctuation and not in the middle of a sentence as it looks weird otherwise

kimchi/src/circuits/gate.rs
```diff
@@ -273,6 +298,24 @@ impl<F: FftField> LookupInfo<F> {
...
+/// given lookup configuration should apply to.
+pub struct GatesLookupSpec {
+    /// The set of positions relative to an active gate where a lookup configuration applies.
+    pub gate_positions: HashSet<(GateType, CurrOrNext)>,
+    /// The fixed lookup table that should be used for these lookups, if applicable.
+    pub gate_lookup_table: Option<GateLookupTable>,
+}
```
mimoo, line 308: What are cases when there is no fixed lookup table associated?

kimchi/src/circuits/gate.rs
```diff
@@ -273,6 +298,24 @@ impl<F: FftField> LookupInfo<F> {
...
+
+/// Specifies mapping from positions defined relative to gates into lookup data.
+pub struct GatesLookupMaps {
+    /// Enumerates the selector that should be active for a particular gate-relative position.
+    pub gate_selector_map: HashMap<(GateType, CurrOrNext), usize>,
```
mimoo, line 313: Should that usize be an enum?

kimchi/src/circuits/gate.rs
```diff
@@ -273,6 +298,24 @@ impl<F: FftField> LookupInfo<F> {
...
+    /// Enumerates the fixed tables that should be used for lookups in a particular gate-relative
+    /// position.
```
mimoo, line 315: No need to break that comment for a single word

kimchi/src/circuits/gate.rs
```diff
@@ -186,18 +182,35 @@ pub enum LookupsUsed {
     Joint,
 }
 
+/// Enumerates the different 'fixed' lookup tables used by individual gates
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum GateLookupTable {
+    Xor,
+}
+
+pub fn get_table<F: FftField>(table_name: GateLookupTable) -> Vec<Vec<F>> {
```
mimoo, line 191: Missing doc on this and the previous enum

kimchi/src/circuits/gate.rs
```diff
@@ -342,25 +385,62 @@ impl GateType {
...
+            let mut patterns = Vec::with_capacity(lookups.len());
+            let mut locations_with_tables = Vec::with_capacity(lookups.len());
+            for (pattern, locations, table) in lookups {
+                patterns.push(pattern);
+                locations_with_tables.push(GatesLookupSpec {
+                    gate_positions: locations,
+                    gate_lookup_table: table,
+                });
+            }
+            (patterns, locations_with_tables)
```
mimoo, line 408: There should be a type for patterns, it’s too chaotic imo to have all these tuples. 

```rust
Struct Pattern {
  /// what lookups to perform in a row
  Lookups: Vec<JointLookup>
  /// the gate that has these lookups
  Locations: HashSet<(GateType,  …)>
  /// the table used in the lookups
  Table
}
```

mimoo: I’m also wondering if this is the best way to specify the layout of all these lookups. This might be unrelated to your PR, I still find it extremely hard to follow these values and how they are loosely related

mimoo: Imo we should build some sort of abstractions that gate implementors can then use to construct fixed tables, and use table lookups as part of their gates. If the abstraction is clean, it’ll be easier to build on top of that in the future. Wdyt?

kimchi/src/circuits/gate.rs
```diff
@@ -342,25 +385,62 @@ impl GateType {
...
+        for (
+            i,
+            GatesLookupSpec {
+                gate_positions: locs,
+                gate_lookup_table: table_kind,
+            },
+        ) in locations_with_tables.into_iter().enumerate()
+        {
+            for location in locs {
+                if let std::collections::hash_map::Entry::Vacant(e) = index_map.entry(location) {
```
mimoo, line 427: Import Entry instead

kimchi/src/circuits/gate.rs
```diff
@@ -342,25 +385,62 @@ impl GateType {
...
+                    if let std::collections::hash_map::Entry::Vacant(e) = table_map.entry(location)
+                    {
+                        e.insert(table_kind);
+                    }
+                }
             }
         }
-        res
+        GatesLookupMaps {
+            gate_selector_map: index_map,
```
mimoo, line 441: Index_map should be called gate_selector_map throughout, same for table_map

kimchi/src/circuits/gate.rs
```diff
@@ -342,25 +385,62 @@ impl GateType {
...
+            for location in locs {
+                if let std::collections::hash_map::Entry::Vacant(e) = index_map.entry(location) {
                     e.insert(i);
                 } else {
                     panic!("Multiple lookup patterns asserted on same row.")
                 }
+                if let Some(table_kind) = table_kind {
+                    if let std::collections::hash_map::Entry::Vacant(e) = table_map.entry(location)
+                    {
+                        e.insert(table_kind);
```
mimoo, line 435: I feel like “kind” is used differently in the lookup code, but maybe it’s me misunderstanding the code. It’s also called table_id elsewhere

kimchi/src/circuits/gate.rs
```diff
@@ -225,35 +238,47 @@ impl<F: FftField> LookupInfo<F> {
...
+        let mut gate_tables = HashSet::new();
 
         // TODO: is take(n) useful here? I don't see why we need this
         for (i, gate) in gates.iter().enumerate().take(n) {
             let typ = gate.typ;
 
-            if let Some(v) = self.kinds_map.get(&(typ, CurrOrNext::Curr)) {
-                res[*v][i] = F::one();
+            if let Some(selector_index) = self.kinds_map.get(&(typ, CurrOrNext::Curr)) {
+                selector_values[*selector_index][i] = F::one();
```
mimoo, line 256: I’m not a fan of this selector index, but confused enough about the flow that I wouldn’t be able to suggest a better way

kimchi/src/circuits/gate.rs
```diff
@@ -225,35 +238,47 @@ impl<F: FftField> LookupInfo<F> {
...
     /// Each entry in `kinds` has a corresponding selector polynomial that controls whether that
     /// lookup kind should be enforced at a given row. This computes those selector polynomials.
-    pub fn selector_polynomials(
+    #[allow(clippy::type_complexity)]
+    pub fn selector_polynomials_and_tables(
         &self,
         domain: &EvaluationDomains<F>,
         gates: &[CircuitGate<F>],
-    ) -> Vec<E<F, D<F>>> {
+    ) -> (Vec<E<F, D<F>>>, Vec<Vec<Vec<F>>>) {
```
mimoo, line 246: If I understand correctly, this vec of vec of vec is simply a list of all tables before being combined. We should have a type for a table (and a type for a combined table)

kimchi/src/circuits/gate.rs
```diff
@@ -186,18 +182,35 @@ pub enum LookupsUsed {
...
+    match table_name {
+        GateLookupTable::Xor => crate::circuits::polynomials::chacha::xor_table(),
+    }
+}
+
 impl<F: FftField> LookupInfo<F> {
     /// Create the default lookup configuration.
     pub fn create() -> Self {
-        let kinds = lookup_kinds::<F>();
+        let (kinds, locations_with_tables): (Vec<_>, Vec<_>) = GateType::lookup_kinds::<F>();
```
mimoo, line 200: Unecessary type sig

mimoo (approved):
I spent several hours looking at the code and I’m still very confused about the flow. I think the abstractions are leaking, and I’m really worried that if we continue to build on the current implementation it’ll become harder and harder to reason about the logic. That being said these are mostly comments unrelated to the PR and I this looks like a step in the right direction to me.

kimchi/src/circuits/constraints.rs
```diff
@@ -331,17 +331,25 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
                 let d1_size = domain.d1.size();
 
+                let (lookup_selectors, gate_lookup_tables) =
+                    lookup_info.selector_polynomials_and_tables(domain, gates);
+
                 // get the last entry in each column of each table
-                let dummy_lookup_values: Vec<Vec<F>> = lookup_tables
+                let dummy_lookup_values: Vec<Vec<F>> = gate_lookup_tables
                     .iter()
+                    .chain(lookup_tables.iter())
```
mimoo, line 340: So these are non-fixed lookup tables?

mimoo: in which case, I would suggest making that argument an `Option`, and also does that work currently or is that a step in the direction of making it work?

kimchi/src/circuits/gate.rs
mrmr1993: When you're using a runtime or side-loaded lookup table. So none yet, but this is future-proofing.

kimchi/src/circuits/gate.rs
mrmr1993: I don't think so, since the number and order of the selectors changes depending on the gates that are active in a circuit.

kimchi/src/circuits/gate.rs
mrmr1993: My ultimate plan is for them to go away entirely. IMO we should just be using the gate selectors (added together, under the assumption that they're mutually exclusive) rather than creating a dedicated selector.

kimchi/src/circuits/gate.rs
mrmr1993: `table_id` is a distinct, but related, concept. I don't think we should confuse the two here.

kimchi/src/circuits/gate.rs
mrmr1993: We'd still need somewhere for the map to live if we want to use it in a closure. Might as well store it here directly.

kimchi/src/circuits/gate.rs
mrmr1993: I'm already of the opinion that the comments in rust are too long (they're a nuisance to read on a phone, for example), I'd prefer not to make this line even longer.

kimchi/src/circuits/gate.rs
mrmr1993: As above, the line is already too long IMO

kimchi/src/circuits/gate.rs
mrmr1993: The name already says what it does; I don't know what comment I could add that wouldn't be redundant.

kimchi/src/circuits/gate.rs
mimoo: I was pointing to that also:

```rust
pub struct LookupInfo<F> {
    /// A single lookup constraint is a vector of lookup constraints to be applied at a row.
    /// This is a vector of all the kinds of lookup constraints in this configuration.
    pub kinds: Vec<Vec<JointLookup<F>>>,
```

which seems to be a different "kind"

kimchi/src/circuits/gate.rs
mrmr1993: > What do we really want to get here?
> 
> * Kimchi can have gates that can perform lookups, oddly gates actually don’t encode these, they are encoded outside of the gate itself.

This is kinda inverted; some gates 'require' lookups in order to behave correctly, but the gates themselves do not perform them. I agree that it would be better if this lived alongside the gates.

> * Then a gate can perform lookups on two rows at a time, which is opaque from the outside (and might lead to UB if the next row does something weird)

Again, the gate isn't doing it. There's an assertion in this file that ensures that we don't trigger 2 lookup configurations on the same line.

> * A gate can also perform several lookups, but they all have to be for the same table. Oddly we don’t always carry this information

They don't have to be from the same table. That they currently are simplified things for this PR, but nothing about the lookup's setup cares about the particular table.

> * A lookup is either a single lookup (you want to check for the existence of a special value) or for a joint lookup (your table has more than one column). Oddly, a jointlookup is built on top of a single lookup

We never use single lookup in isolation, and once we add `table_id` properly, we probably never will. We should rename them to something that reflects that 'single' is a linear combination of witness fields to form one value in the lookup, and that 'joint' is the actual entry in the table, comprising one or more 'single' lookups and the table ID thereof.

> * A single lookup, and thus a joint lookup, can be formed from a linear combination of witness columns (not used currently)

I take it you haven't looked at `chacha_final_pattern`...

> Imo we should build some sort of abstractions that gate implementors can then use to construct fixed tables, and use table lookups as part of their gates. If the abstraction is clean, it’ll be easier to build on top of that in the future. Wdyt? I’m scared that if we continue to build on top of the current abstractions the logic will become harder and harder to work on.

I'm going to try and improve things as I work through implementing 'extended lookup tables', but my priority is getting that done rather than iterating on the existing one.

kimchi/src/circuits/gate.rs
mimoo: every public function should be documented, please add one here

kimchi/src/circuits/gate.rs
mimoo: These are really meant to be read on the generated docs though

kimchi/src/circuits/gate.rs
mimoo: that sounds like ideology :P

kimchi/src/circuits/gate.rs
mimoo: as above, public functions need to be documented, even if redundant

kimchi/src/circuits/constraints.rs
mrmr1993: IMO an option is redundant when we're passing a slice/vector, since `None` should be equivalent to the empty slice (and indeed `Vec::with_capacity(0)` is non-allocating, so it's a drop-in replacement).

kimchi/src/circuits/gate.rs
```diff
@@ -188,7 +188,9 @@ pub enum GateLookupTable {
     Xor,
 }
 
-pub fn get_table<F: FftField>(table_name: GateLookupTable) -> Vec<Vec<F>> {
+pub type LookupTable<F> = Vec<Vec<F>>;
```
mimoo, line 193: missing doc :P

PR #303, opened 2022-02-16
[easy] remove dead code

mimoo:
it seems like we were using these functions (ToBytes, FromBytes) and traits (num-derive) before we moved to serde for serialization. So removing this code now.


mrmr1993:
Can you make a corresponding mina PR, to confirm that this code is dead?

mimoo:
will do after I can get the double generic gate in

jspada (commented):
Assuming it's not used from `mina`, then good to remove.

joseandro:
@mimoo what's the latest on this?

mimoo:
keeping it around until I have time to do it. BTW we already have a bot checking on stale PRs/issues ;)

PR #310, opened 2022-02-17
[WIP][argument] use argument trait for lookup

mimoo:
originally part of https://github.com/o1-labs/proof-systems/pull/295, but now part of its own PR: it implements the `Argument` trait for lookup. Since lookup needs access to some dynamic data to build the constraints, we change the API slightly to allow that.

### DIFF
--- a/kimchi/src/circuits/argument.rs
+++ b/kimchi/src/circuits/argument.rs
@@ -38,11 +38,11 @@ pub trait Argument {
 
     /// Returns the set of constraints required to prove this argument.
     // TODO: return a [_; Self::CONSTRAINTS] once generic consts are stable
-    fn constraints() -> Vec<E<Self::Field>>;
+    fn constraints(&self) -> Vec<E<Self::Field>>;
 
     /// Returns constraints safely combined via the passed combinator.
-    fn combined_constraints(alphas: &Alphas<Self::Field>) -> E<Self::Field> {
-        let constraints = Self::constraints();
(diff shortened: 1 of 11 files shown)


mrmr1993 (commented):
I don't think it makes sense to use the same abstraction for lookups as it does for gates, since there's quite a lot of divergence in the details. I think that a 'bundle of constraints' is too nebulous an idea to create meaningful traits from.

I would personally prefer if the `Argument` trait became a `Gate` trait, and then we add a new trait `GlobalConstraint` that applies to every row. To my mind, there are 3 kinds of `GlobalConstraint`, those being `Gates` that combines all of the gate constraints, `Permutation` that computes the permutation argument across all of the rows, and `Lookup` that computes the lookup aggregation and applies the lookups.

For example, I could imagine something like
```rust
trait GlobalConstraint<F: FftField> {
  fn selector() -> Expr<F>; // Returns `VanishesOnLast4Rows` for lookup, permutation; returns `1` for the combined gates
  fn constraints(index: ProverIndex<F>) -> Expr<F>;
}
```
with some extra machinery to make sure each is using a *different* set of alphas (whereas the set for each gate is overlapping). Wdyt?

mimoo:
It definitely made sense to me for the permutation and gates, as they do apply to the whole domain (it's just that gates are filtered). But I need to spend more times in lookup land to see if it does there as well. Once I fill in the blanks in that diagram I'll have a better idea on how to respond :D

Screen Shot 2022-02-25 at 2 07 27 PM

PR #314, opened 2022-02-18
Make FieldHelpers work with Field instead of FftField

### LINKED ISSUE #313: Make FieldHelpers work on more general Field instead of FftField

jspada:
Closes #313

### DIFF
--- a/utils/src/field_helpers.rs
+++ b/utils/src/field_helpers.rs
@@ -1,7 +1,7 @@
-use ark_ff::FftField;
+use ark_ff::Field;
 
 /// Field element helpers
-pub trait FieldHelpers<F: FftField> {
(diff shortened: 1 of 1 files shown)


utils/src/field_helpers.rs
```diff
@@ -1,7 +1,7 @@
-use ark_ff::FftField;
+use ark_ff::Field;
 
 /// Field element helpers
-pub trait FieldHelpers<F: FftField> {
+pub trait FieldHelpers<F: Field> {
```
querolita, line 4: Perhaps here use `FieldHelpers<F>` , without specifying the concrete generic type, and have it only in the implementation.

querolita (approved):

PR #319, opened 2022-02-21
Update field helpers and signer to use thiserror

### LINKED ISSUE #318: Update field helpers and signer to use thiserror
The field helpers and signer should use `thiserror`.  This will also resolve a lifetime issue with the field helpers interface.  Additionally, chainsafe needs an implementation of `domain_prefix_to_bytes(prefix: &str)` so we ought to expose this for them.

jspada:
* Resolve FieldHelpers::from_bytes(...) lifetime issue
    * Make domain_prefix_to_bytes(...) public for CS
    * Add size_in_bytes() field helper

Closes #318


querolita (approved):

signer/src/lib.rs
```diff
@@ -49,6 +51,18 @@ impl From<NetworkId> for u8 {
     }
 }
 
+/// Transform domain prefix string to bytes
+pub fn domain_prefix_to_bytes<F: PrimeField>(prefix: &str) -> Vec<u8> {
```
hanabi1224, line 55: any chance to use `&[u8]` type for `prefix`?  We define all our prefix strings like `b'foo'` which is `&u[8]`

hanabi1224: This is actually what we already have [here](https://github.com/ChainSafe/mina-rs/blob/d78d78f701ae4177f4dc13043166fe3690c39ecf/crypto/src/hash/prefixes.rs#L11), how about exposing `domain_prefix_to_field`?

utils/src/field_helpers.rs
```diff
@@ -19,19 +30,27 @@ pub trait FieldHelpers<F> {
 
     /// Serialize to bits
     fn to_bits(self) -> Vec<bool>;
+
+    /// Field size in bytes
+    fn size_in_bytes() -> usize
+    where
+        F: PrimeField,
+    {
+        ((F::size_in_bits() as f64) / 8.0).ceil() as usize
```
hanabi1224, line 39: How about avoiding f64 conversion by doing
```
let bits = F::size_in_bits();
match bits % 8 {
    0 => bits / 8,
    _ => bits / 8 + 1,
}
```

signer/src/lib.rs
jspada: Hi! Thanks for your feedback.

* Wrt `&[u8]` prefix -> bytes - This doesn't make sense because the concept is to transform a string into bytes, rather than bytes into bytes.  IOW, your "prefix strings" should really be strings.

* Wrt exposing `domain_prefix_to_field` - it doesn't make sense because, again from our library's perspective, there's really no such thing.  We have two concepts
  * prefix string -> bytes
  * bytes -> field
  
  If you want this function, you should make your own helper by using `domain_prefix_to_bytes` and `FieldHelpers::from_bytes`.   By using these two interfaces we will be sure your code does the correct things and doesn't require many updates if we change any low-level crypto things.
  
* Wrt Edit - I think that would be a unguided optimisation for something that is highly unlikely to become a bottleneck

hanabi1224: Thanks, regarding bullet point 2, we have all prefix strings represented as const &[u8] at compile-time, but without being padded into field bit size, so what we really need is `bytes -> field` with padding, we can either implement in [our lib](https://github.com/ChainSafe/mina-rs/pull/154/files#diff-d0426229102065c0759aa57dae8dd11982a0fe33c38376e270127bc44f67a865R13) or make it another function in `FieldHelpers` trait, like [this proposal](https://github.com/o1-labs/proof-systems/pull/312/files#diff-ac9b891bb54a8652cdb3091e9686b7b43ea7f7079b85f90fa2b42d52d280e9baR35)

jspada: Our bytes -> field helper semantics are selected to match those of arkworks, which should not have mina-specific domain string padding logic as part of it.  (Think of this as another layer.)

However, the `domain_prefix_to_bytes(...)` function provided is part of the (mina-specific) signer crate and, thus, has the mina-related padding logic.

This separation of concerns is important.

Re strings vs bytes; if your downstream code wants to avoid all strings that's one thing, but it doesn't make sense to push this requirement on upstream code.

I think the best way forward is for you to either to represent your prefix strings as strings OR convert them to strings before calling `domain_prefix_to_bytes(...)`.  Either of these solutions avoids rewriting existing proof-systems code is fine with respect to our 3rd party implementation requirements.

utils/src/field_helpers.rs
jspada: This makes sense - I'll swap it to something like `bit_count / 8 + (bit_count % 8 != 0) as usize`

signer/src/lib.rs
hanabi1224: > has the mina-related padding logic.

Does 'mina-related padding logic' refer to `format!("{:*<MAX_DOMAIN_STRING_LEN$}", prefix)` or `bytes.resize(F::size_in_bytes(), 0)` or both?  The latter is only padding/trimming unknown-sized byte array to F::size_in_bytes(), would it fit into generic FieldHelpers trait?

jspada: Yeah, I would say both are mina-specific things-- `from_bytes` does only what arkworks does and `domain_prefix_to_bytes(...)` does the mina-specfic stuff.

I will just move the field conversion step (i.e. call to from_bytes) into `domain_prefix_to_bytes(...)` and rename to `field_from_domain_prefix(...)`.  This won't require any changes to `from_bytes` and should give you what you want.

hanabi1224: Thank you so much!

jspada: The Hashable interface changes should be in another PR, so this one will just update to thiserror and expose`field_from_domain_prefix` for now.  We can then discuss the `Hashable` trait and open another PR for that.

mrmr1993 (approved):
Nice!

PR #321, opened 2022-02-22
Introduce a CryptoHash trait to easily hash types

mimoo:
this introduces the CryptoHash trait, which you can implement for a type to get access to a `digest()` function:

```rust
#[derive(Serialize)]
struct A {
    thing: u8,
}
impl CryptoDigest for A {
    const PREFIX: &'static [u8; 15] = b"kimchi-circuit0";
}
println!("digest: {:?}", a.digest());
```

The `PREFIX` is a domain-separation string that is simply prepended to the real input.

Benefit of this approach:

* less error-prone since we generate the boilerplate by implementing the trait
* uses a domain-separation string by default

It would be nice if:

* we could just use a macro
    ```rust
    #[derive(Serialize, CryptoDigest("kimchi-circuit")]
    struct A { 
    ```
* we could make sure all types that implement CryptoDigest use different prefixes
* the hash state with the prefix is pre-computed (for example using [sha2-const](https://github.com/saleemrashid/sha2-const))

### DIFF
--- a/utils/src/hasher.rs
+++ b/utils/src/hasher.rs
@@ -1,0 +1,56 @@
+//! This module includes the [CryptoDigest] trait,
+//! which provides a generic interface for hashing.
+//!
+//! To use it, simply implement [CryptoDigest] for your type:
+//!
+//! ```
+//! use o1_utils::hasher::CryptoDigest;
+//! use serde::Serialize;
+//!
+//! #[derive(Serialize)]
+//! struct A {
+//!     thing: u8,
+//! }
+//!
+//! impl CryptoDigest for A {
+//!     const PREFIX: &'static [u8; 15] = b"kimchi-circuit0";
+//! }
+//!
+//! let a = A { thing: 1 };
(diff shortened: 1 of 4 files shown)


jspada:
Do you think this can be integrated with the signer crate somehow so that I could use it to wrap around poseidon and then use it from there?

mimoo:
IIUC there's two different "hashable", one that's for SHA-256 and one that's for Poseidon. They are pretty different since the one for SHA-256 relies on messagepack for encoding AND implements both that encoding and the hashing in the same function (`digest`), whereas the one for poseidon is split between an encoding function and a hashing function right? I'm not sure how you would reconcile these two.

utils/src/hasher.rs
```diff
@@ -0,0 +1,60 @@
+//! This module includes the [CryptoDigest] trait,
```
mrmr1993, line 1: Nit: `includes` is probably not the right word here. `exposes`/`provides`/`defines` might be better suited

mrmr1993 (approved):

PR #323, opened 2022-02-25
Relax Hashable trait constraint to use Clone instead of Copy

hanabi1224:
This PR relaxes `Hashable` trait constraint to use `Clone` instead of `Copy`. 

Background: most of the data structures under mina-rs are not deriving the `Copy` trait, for better visibility of performance overhead and to avoid implicit copy-on-move,  so it's hard for us to implement this trait when it requires `Copy`.  Please let me know your feedback and concerns, thanks!

P.S. In this PR, I did not make changes to any structs that implements `Hashable` in the `signer` crate (e.g. `Transaction`) so they are still deriving `Copy`.


mimoo (approved):
looks good!

mimoo:
looks like circleci tests are blocking. If you want you can remove the `.circleci` folder as part of this PR, since we don't need circleCI anymore (the checks in the github actions are actually better). cc @mrmr1993 we'd need to enforce the github action tests in the protected branch master though.

jspada (approved):

hanabi1224:
Sure, `.circleci` folder removed

mimoo:
humm... not sure why it still shows the circleci tests. cc @mrmr1993 since I can't investigate this myself

mrmr1993:
The current CI for github actions doesn't run `cargo test`. We want the CircleCI tests to run those, at least until we change the github one to do so.

mimoo:
Wow... Good catch. I was convinced it did. @hanabi1224 can you add `cargo run test` to the `.github/workflows/rust.yml`?

thanks for catching that @mrmr1993

.github/workflows/rust.yml
```diff
@@ -15,25 +15,34 @@ jobs:
...
 
       - name: Set up cargo/rust
-        run: |
-          rustup component add rustfmt
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt, clippy
+          override: true
```
mimoo, line 23: <3

hanabi1224:
@mimoo I've enabled tests in GitHub CI. Plz note that I changed build&test to release mode because tests are very slow in debug mode

@jspada Plz let me know if I still need to restore the `.circleci` folder in this case.

.github/workflows/rust.yml
```diff
@@ -15,25 +15,34 @@ jobs:
...
 
-      - name: build
+      # `cargo check` compiles the packages without performing the final step of code generation,
+      # which is faster than running `cargo build`.
+      # Use release build to speed up tests
+      - name: build & test
         run: |
-          cargo build
+          cargo check --release
+          cargo test --release
```
mimoo, line 36: can you remove `--release` for tests? As it disable some checks like integer overflows

.github/workflows/rust.yml
```diff
@@ -15,25 +15,34 @@ jobs:
...
-        run: |
-          cargo fmt -- --check
+        run: cargo fmt -- --check
 
       - name: Run cargo clippy
-        run: |
-          rustup component add clippy
-          cargo clippy -- -D warnings
+        # Need clippy for `--features ocaml_types`?
+        run: cargo clippy --release -- -D warnings
```
mimoo, line 48: does it really make sense to run cargo clippy in release?

.github/workflows/rust.yml
mimoo: just read your message above, that's a shame :( I guess let's leave the release then

hanabi1224: Is integer overflow checks the only thing to be concerned?  I believe it can be enabled via [rustc flags](https://doc.rust-lang.org/cargo/reference/profiles.html#overflow-checks)

mimoo: I'm actually not sure what is getting disabled in release mode. But that could be a good compromise!

.github/workflows/rust.yml
hanabi1224: Removed

.github/workflows/rust.yml
hanabi1224: > could be a good compromise!

I will add the `RUSTFLAGS` environment variable then : )

The tests take minutes in debug mode, and I see many warnings like 
```
test bench::tests::test_bench has been running for over 60 seconds
test tests::chacha::chacha_prover has been running for over 60 seconds
```
BTW, previous `cargo test` in the circleci was also under release mode

.github/workflows/rust.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+env:
+  RUSTFLAGS: -Coverflow-checks=y
```
mimoo, line 10: sorry but last nit :D I'm wondering if this makes `cargo check` and `cargo clippy` slowlier since this env will be applied to all of them instead of just the the tests. Perhaps `RUSTFLAGS="-Coverflow-checks=y" cargo test --release` is better?

hanabi1224: Ideally, they share the build cache, so using the same flags should save the overall time. (But the incremental build is not enabled by default in recent rust releases)

.github/workflows/rust.yml
hanabi1224: > not sure what is getting disabled

According to the [doc](https://doc.rust-lang.org/cargo/reference/profiles.html#release)
```
debug-assertions = false
overflow-checks = false
```

.github/workflows/rust.yml
hanabi1224: I just compared the build logs, the performance impact of build time is negligible, I guess it's OK but plz let me know if u still like applying the flags to cargo test only.

```
Finished release [optimized] target(s) in 1m 31s
```

```
Finished release [optimized] target(s) in 1m 34s
```

hanabi1224: To further speed up the job, I commented out `cargo test --release` because it's covered by `cargo test --release --features ocaml_types`

.github/workflows/rust.yml
```diff
@@ -36,10 +36,16 @@ jobs:
...
+          # This is covered by the next step,
+          # commented out to make the job faster
+          #
+          # cargo test --release
 
       - name: build & test ocaml types
         run: |
+          # Cannot use `--all-features` here
+          # Becauses `3`, `3w` and `5w` features under `oracle` crate are orthogonal,
+          # Should this be changed?
```
mimoo, line 48: if you find a way to fix this it'd be nice!

hanabi1224: I think it's only because the tests are orthogonal, not the features, I can make a fix in a separate PR if u like.

mimoo: that'd be awesome

.github/workflows/rust.yml
mimoo: ah, it'd be good to have debug assertions as well!

hanabi1224: Added `-Cdebug-assertions=y`, also enabled incremental build by setting `CARGO_INCREMENTAL` to `true`

.github/workflows/rust.yml
hanabi1224: Fixed by https://github.com/o1-labs/proof-systems/pull/329

jspada: @mimoo - We still need `3w` (i.e. `PlonkSpongeConstantsBasic` config) in addition to `5w,` since mainnet transactions payment and delegation use this configuration.

.github/workflows/rust.yml
```diff
@@ -15,25 +21,40 @@ jobs:
...
-        run: sudo apt update && sudo apt install ocaml
+        run: |
+          sudo apt update
+          sudo apt install -y ocaml
 
-      - name: build
+      # `cargo check` compiles the packages without performing the final step of code generation,
+      # which is faster than running `cargo build`.
+      # Use release build to speed up tests
+      - name: build & test
```
mrmr1993, line 39: Can you rename this to `check build` or similar?

.github/workflows/rust.yml
```diff
@@ -15,25 +21,40 @@ jobs:
...
         run: |
-          cargo build
+          cargo check --release
+          # This is covered by the next step,
```
mrmr1993, line 42: Delete this comment

.github/workflows/rust.yml
```diff
@@ -15,25 +21,40 @@ jobs:
...
 
-      - name: build ocaml types
+      - name: build & test ocaml types
         run: |
-          cargo build --features ocaml_types
+          # Cannot use `--all-features` here
+          # Becauses `3`, `3w` and `5w` features under `oracle` crate are orthogonal,
+          # Should this be changed?
+          cargo check --release --features ocaml_types
+          cargo test --release --features ocaml_types
```
mrmr1993, line 53: Move this to its own step

mrmr1993 (commented):
I don't love that this PR has been hijacked by a complete CI refactor, but I guess we're already here now. Can you delete `.circleci`? I'll change the backend to use these new tests once this is ready to merge.

Sprinkled a few other nits, for flavour.

.github/workflows/rust.yml
hanabi1224: Sure, or I can create a separate PR for this, as suggested by Joseph

hanabi1224: @mrmr1993  Here u go https://github.com/o1-labs/proof-systems/pull/332  I will resolve ur comments in the new PR and revert all CI related changes in this one,.

.github/workflows/rust.yml
hanabi1224: Fixed https://github.com/o1-labs/proof-systems/pull/332/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR40

.github/workflows/rust.yml
hanabi1224: Fixed

PR #328, opened 2022-03-01
[specs][kimchi] specify poseidon + double generic gates

mimoo:
Specifying two gates in this PR. Rest incoming.

### DIFF
--- a/kimchi/src/circuits/polynomials/generic.rs
+++ b/kimchi/src/circuits/polynomials/generic.rs
@@ -3,13 +3,15 @@
 //~ The double generic gate contains two generic gates.
 //~
 //~ A generic gate is simply the 2-fan in gate specified in the
-//~ vanilla PLONK protocol that allows us to do:
+//~ vanilla PLONK protocol that allows us to do operations like:
 //~
 //~ * addition of two registers (into an output register)
 //~ * or multiplication of two registers
 //~ * equality of a register with a constant
 //~
-//~ In each cases, registers can also be scaled by a constant.
+//~ More generally, the generic gate controls the coefficients $c_i$ in the equation:
+//~
+//~ $$c_0 \cdot l + c_1 \cdot r + c_2 \cdot o + c_3 \cdot (l \times r) + c_4$$
 //~
 //~ The layout of the gate is the following:
 //~
--- a/kimchi/src/circuits/polynomials/generic.rs
+++ b/kimchi/src/circuits/polynomials/generic.rs
@@ -1,43 +1,35 @@
-//! This module implements the double generic gate,
-//! which contains two generic gates.
-//!
-//! A generic gate is simply the 2-fan in gate specified in the
-//! vanilla PLONK protocol that allows us to do:
-//!
-//! * addition of two registers (into an output register)
-//! * or multiplication of two registers
-//! * equality of a register with a constant
-//!
-//! In each cases, registers can also be scaled by a constant.
-//!
-//! The layout of the gate is the following:
-//!
-//! |  0 |  1 |  2 |  3 |  4 |  5 | 6 | 7 | 8 | 9 | 10 | 11 | 11 | 12 | 13 | 14 |
-//! |:--:|:--:|:--:|:--:|:--:|:--:|:-:|:-:|:-:|:-:|:--:|:--:|:--:|:--:|:--:|:--:|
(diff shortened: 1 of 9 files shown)


kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -10,28 +10,34 @@ use std::marker::PhantomData;
...
     RoundEquation {
         source: 0,
         target: (Curr, 1),
     },
+    //~ * the second round targets the first row
     RoundEquation {
         source: 1,
         target: (Curr, 2),
     },
+    //~ * the second round targets the first row
```
mrmr1993, line 30: You probably intended to write `third` here.

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -10,28 +10,34 @@ use std::marker::PhantomData;
...
     RoundEquation {
         source: 2,
         target: (Curr, 3),
     },
+    //~ * the second round targets the first row
```
mrmr1993, line 35: `forth`

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -10,28 +10,34 @@ use std::marker::PhantomData;
...
-pub struct RoundEquation {
+struct RoundEquation {
     pub source: usize,
     pub target: (CurrOrNext, usize),
 }
 
-pub const ROUND_EQUATIONS: [RoundEquation; ROUNDS_PER_ROW] = [
+//~ A poseidon gate records 5 rounds of the poseidon permutation:
+const ROUND_EQUATIONS: [RoundEquation; ROUNDS_PER_ROW] = [
+    //~ * the first round targets the first row
```
mrmr1993, line 20: This comment could easily say `The first four rounds target the first row`, rather than copy-pasting below. It's probably also relevant to mention the layout beyond just the row.

mimoo: Actually I was just toying around on how to specify poseidon, I think it’s going to be hard due to how it’s implemented, but ignore most of that for now

kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -1,43 +1,35 @@
...
+//~ The double generic gate contains two generic gates.
+//~
+//~ A generic gate is simply the 2-fan in gate specified in the
+//~ vanilla PLONK protocol that allows us to do:
+//~
+//~ * addition of two registers (into an output register)
+//~ * or multiplication of two registers
+//~ * equality of a register with a constant
+//~
+//~ In each cases, registers can also be scaled by a constant.
```
mrmr1993, line 12: This feels a little under-specified, since in practice we use it for more than the options above. For example,
* we constrain `x = 1 or -1` with `x*x-1 = 0`,
  - we use this for the 'sign' for signed n-bit integers
* we treat an `x = 1 or -1` as if it were a boolean mask with `(x+1)*y/2 = z`, which expands to `0 * x + 2^{-1} * y + 2^{-1} * x*y + (-1) * z + 0 = 0`
  - this allows us to compute `if x > 0 then x else 0` for a signed n-bit integer `x`.

Really, this gives us a way to define a gate that satisfies a linear function `f(x, y, x*y, z) = 0`. I don't know if there's a more elegant way to state this though; perhaps these additional examples are sufficient.

mrmr1993 (approved):
Really nice, but left a small nit on the generic gate description.

kimchi/src/circuits/polynomials/generic.rs
mimoo: yeah that's true, perhaps these use-cases could be added to the `GenericGateSpec` as well

mrmr1993: I don't think we need explicit constructors for them, but a constructor like
```rust
    FullGeneric {
        /// Optional coefficient that can be multiplied with the left operand.
        left_coeff: Option<F>,
        /// Optional coefficient that can be multiplied with the right operand.
        right_coeff: Option<F>,
        /// Optional coefficient that can be multiplied with the multiplication result.
        mul_coeff: Option<F>,
        /// Optional coefficient that can be multiplied with the output.
        output_coeff: Option<F>,
        /// Optional constant to add to the expression.
        constant_term: Option<F>,
    },
```
might be nice for completeness.

PR #329, opened 2022-03-01
Remove orthogonal features "3", "3w" and "5w"

hanabi1224:
As follow-up of [this comment](https://github.com/o1-labs/proof-systems/pull/323#discussion_r816414141)

- Remove orthogonal features "3", "3w" and "5w" under `oracle` crate to allow `cargo build/test --all-features`
- Remove `export_test_vectors` only deps from dependency tree when the crate is used as lib
- Make  "3", "3w" and "5w" the 2nd positional parameter of the `export_test_vectors` tool

### DIFF
--- a/oracle/Cargo.toml
+++ b/oracle/Cargo.toml
@@ -25,22 +25,22 @@ ocaml-gen = { path = "../ocaml/ocaml-gen", optional = true }
 syn = { version = "1.0.76", optional = true }
 
 # for export_test_vectors
-num-bigint = "0.4.0"
-serde_json = "1.0"
-hex = "0.4"
-ark-serialize = "0.3.0"
+# maybe better to move the tool into a separate crate
+# to avoid polluting the dependency tree of this lib
+num-bigint = { version = "0.4.0", optional = true }
+serde_json = { version = "1.0", optional = true }
+hex = { version = "0.4", optional = true }
+ark-serialize = { version = "0.3.0", optional = true }
 
 [dev-dependencies]
 hex = "0.4"
+serde_json = "1.0"
+ark-serialize = "0.3.0"
 
 [features]
+default=[]
 ocaml_types = [ "ocaml", "ocaml-gen", "syn" ]
-
-# for test vectors
-default = ["3w"]
-3w = [ ]
-5w = [ ]
-3 = [ ]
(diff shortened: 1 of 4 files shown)


jspada:
The 3w oracle configuration isn't orthogonal as it's used in mainnet.  We first should update the oracle unit tests to the current supported list.

hanabi1224:
@jspada These 3 features being removed are only used in the `export_test_vectors` CLI tool, no code logic or unit tests are changed in this PR. With the changes, the `export_test_vectors` tool has the same functionalities as before, it only changes the way how these features are specified. (via CLI positional argument now)

mrmr1993 (commented):
LGTM, but the github action will need to be updated for the new tests you added in #332.

.circleci/config.yml
```diff
@@ -1,21 +0,0 @@
-version: 2.1
```
mrmr1993, line 1: This merge commit doesn't seem to be a merge

oracle/Cargo.toml
```diff
@@ -25,21 +25,23 @@ ocaml-gen = { path = "../ocaml/ocaml-gen", optional = true }
 syn = { version = "1.0.76", optional = true }
 
 # for export_test_vectors
-num-bigint = "0.4.0"
-serde_json = "1.0"
-hex = "0.4"
-ark-serialize = "0.3.0"
+# maybe better to move the tool into a separate crate
```
mrmr1993, line 28: Agreed, can you open an issue for this?

.circleci/config.yml
hanabi1224: Sorry that was a bad merge, now all conflicts are resolved and github CI is updated to use `cargo test --all-features`, plz consider square merging this PR since it has many unuseful merge commits, thx!

mrmr1993 (approved):
Looks great, thanks!

.circleci/config.yml
mrmr1993: No worries :slightly_smiling_face:

We've found that squash merging caused us real headache (at least, in the mina repo) when resolving conflicts between longer-lived branches. Our current flow is to `git merge` for PR merges, and to bring branches up-to-date. This also allows us to build 'PR chains' of branches of dependent changes.

History ends up a bit messier, but it's a small price to pay compared to the amount of time we used to spend handling merge conflicts and branch drift!

PR #336, opened 2022-03-05
[kimchi][poseidon] refactor test

mimoo:


kimchi/src/circuits/gates/poseidon.rs
```diff
@@ -181,3 +184,52 @@ impl<F: FftField> CircuitGate<F> {
         })
     }
 }
+
+/// `generate_witness(row, params, witness_cols, input)` uses a sponge initialized with
+/// `params` to generate a witness for starting at row `row` in `witness_cols`,
+/// and with input `input`.
+pub fn generate_witness<F: Field>(
```
querolita, line 191: I like this generation of the witness being part of the gate itself, rather than part of the tests only. Are you planning to extend this refactoring to other gates as well, even for the ones whose witness generation isn't as complex?

querolita (approved):
This PR makes gates tests more readable plus it provides a function to generate the Poseidon witness that can be called from the tests more succinctly.

kimchi/src/circuits/gates/poseidon.rs
o1-david: imo it'd be really great if that was the case, but I haven't thought too much about it. For it to be really useful it'd have to be something that can be used by the OCaml code. [Right now the OCaml code assembles the execution trace/witness](https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml#L941), but it'd be nice if it could just call a function with the right parameters. But for that to work we'd have to create an execution trace/witness_cols struct

mimoo: imo it'd be really great if that was the case, but I haven't thought too much about it. For it to be really useful it'd have to be something that can be used by the OCaml code. [Right now the OCaml code assembles the execution trace/witness](https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml#L941), but it'd be nice if it could just call a function with the right parameters. But for that to work we'd have to create an execution trace/witness_cols struct

querolita: Oh I see... Actually I was not sure where the witness was being generated, so it is in OCaml I see... In any case, it seems like this is being a thing in Cairo and ForeignMul, but these are examples of quite complex witness generations.

mimoo: yeah which is not great IMO, because if someone wants to use this library they'll have no idea how to do that :) I'm actually writing something that could be a start atm

PR #343, opened 2022-03-07
[WIP] Introduce a witness type

mimoo:
This PR introduces a `Witness` type. WIP

### DIFF
--- a/kimchi/src/bench.rs
+++ b/kimchi/src/bench.rs
@@ -1,8 +1,6 @@
 use crate::{
     circuits::{
-        gate::CircuitGate,
-        polynomials::generic::GenericGateSpec,
-        wires::{Wire, COLUMNS},
+        gate::CircuitGate, polynomials::generic::GenericGateSpec, wires::Wire, witness::COLUMNS,
     },
     index::{testing::new_index_for_test, Index, VerifierIndex},
     prover::ProverProof,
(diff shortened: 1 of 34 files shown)


mrmr1993 (changes requested):
I don't think this PR makes the code any clearer or easy to maintain, and I'm pretty put off by the direct abstraction leaks in `get_cols_mut`, `set`, and `inner` (so effectively the whole API).

I would strongly prefer to see this PR closed out, or reduced to only the 'visu' changes.

mimoo:
It's kind of a half PR, as it doesn't implement the full idea (or apply it to the whole codebase). I think a complete abstraction around `Witness` would allow us to easily contain all the witness functions in one place (creation/manipulation of witness, perhaps we can keep track of what row we're on there as well, zk-ifying the rows). 

I think the biggest argument for an abstraction is this: a user of the library shouldn't have to figure out that they have to create an array of `COLUMNS` rows of the right height to create a proof. Perhaps it doesn't make as much sense right now as we don't have the functionality to generate the witness anyway, but I think it's worthy to explore at some point.

mimoo:
probably would be a good idea if we could do `witness.col(5).row(3) = Cell { col, row }`

querolita:
Ok, makes sense @mimoo

PR #345, opened 2022-03-08
Fix 32-bit arch expr::pow by using the correct number of bits

mrmr1993:
This PR is the minimal possible fix for the poseidon failure in the WASM version of the code.

This was caused by the assumption that `usize` is always 64-bit -- where it's 32-bit under WASM -- and is a symptom of misusing `usize` for unsigned integers that aren't related to memory accesses.


mitschabaude:
Omg how awesome is this

PR #350, opened 2022-03-08
Switch to correct arithmetic sponge params

### LINKED ISSUE #349: Kimchi is using legacy arithmetic sponge params but should be on kimchi version

jspada:
* Switch away from legacy params to kimchi params
   * Rename params to be easily recognisable
   * Resample round consts and MDS to be unique (from new name)
   * Rename SpongeConstants to be more recognisable and pair with param names
   * Legacy: f{p,q}_legacy and PlonkSpongeConstantsLegacy
   * Kimchi: f{p,q}_kimchi and PlonkSpongeConstantsKimchi
   * Signer crate stay on legacy, everything else to kimchi
   
   Closes #349


mrmr1993 (approved):

oracle/src/pasta/fp_kimchi.rs
```diff
@@ -0,0 +1,828 @@
+use crate::poseidon::ArithmeticSpongeParams;
```
mimoo, line 1: so IIUC these parameters were missing before? How was anything between ocaml and rust working then?

mrmr1993: They previously had a different name, but they were present before. I believe @jspada is planning to create a Mina PR and a snarky PR with corresponding changes before merging this, so that everything keeps in sync

mimoo: I'm confused :o it looks like we were using the now {fp, fq}_legacy, whereas after this PR we use {fp, fq}_kimchi which are brand new in this repo

jspada: They were present under another name, but unused in proof-systems, which as far and Matthew and I know was a mistake.

This PR adds them back and gives them better names so they cannot be confused again, here nor anywhere else.  It also lines up the names of the params with the sponge constant names, so that it's more obvious which combos are valid. I'll be opening up mina and snarky PRs corresponding to this also.

mimoo:
> They were present under another name

why are fp_kimchi and fq_kimchi brand new files then :D ? if they were present what were the filenames before? I don't see them in the PR

jspada:
> > They were present under another name
> 
> why are fp_kimchi and fq_kimchi brand new files then :D ? if they were present what were the filenames before? I don't see them in the PR

This is because they have been resampled.  The way the sage script works is that the name, e.g. "3" or "kimchi" is hashed into things to assure unique parameters are generated.

kimchi/src/circuits/expr.rs
```diff
@@ -1984,11 +1984,11 @@ where
         let mut res = String::new();
         for (k, v) in env.into_iter() {
             let rhs = v.ocaml_str();
-            let cached = format!("let {} = {rhs} in", k.var_name());
+            let cached = format!("let {} = {rhs} in ", k.var_name());
```
mimoo, line 1987: I think this little change will actually annoy the formatter on the ocaml side :o

Issue #346, opened 2022-03-08
Don't use `usize` for anything except memory accesses


mrmr1993:
We should never use `usize` for anything except memory accesses (ie. array indexing and lengths), and we should always prefer fixed-size integers unless we are directly managing memory.

This was the root cause of the poseidon breakage in SnarkyJS+WASM (#345): we were using encoding integers as `usize` and then treating them as if `usize = u64`. We will continue to create these kinds of issues until we fix this.

mimoo:
I've noticed a lot of `usize` in the codebase that don't really make sense, even to interpret rows of a circuit it doesn't make sense to use a `usize` imo.

mimoo:
BTW, the problematic conversions:

1. `usize as u32`, you can have a truncation.
2. `u64 as u32`, you can have a truncation.
3. `u64 as usize`, same as the previous one when running on a machine where `usize = u32`.

I think the first two are unlikely to be seen in our codebase, but the third one is definitely happening in a number of places. The third one is only an issue when we're using values that are larger than 32 bits, which might be unlikely anyway. Still, we might want to use `usize::try_from(u64).expect("32-bit conversion issue")` in places where we do this. (Or create a shorter helper to do this.)

mrmr1993:
We probably want `u32` nearly everywhere, `usize` for memory, and `u64` pretty much just for the encodings that use it (e.g. limbs of `BigInt256`/`Fp`/`Fq`).

There's a slight wrinkle in dealing with OCaml, since the int type uses `usize` (where the top bit is never set); however, porting this code to js_of_ocaml+WASM will also cause a problem, because JS can only faithfully represent a u32 in the number type. In these cases we should probably `TryInto::try_into::<u32>(ocaml_int).map_err(|_| { Error(ocaml::failwith "function_name: Integer argument uses more than 32 bits")})?`. A macro for this might be nice.

mimoo:
Yeah I think that's a good idea, switching everything to `u32`

> There's a slight wrinkle in dealing with OCaml, since the int type uses usize (where the top bit is never set);

That's really annoying that ocaml-rs let's you make that mistake. I brought that up here https://github.com/zshipko/ocaml-rs/issues/71 but I don't think it's sensible that we have to remember these things when writing bindings

mrmr1993:
> I brought that up here https://github.com/zshipko/ocaml-rs/issues/71 but I don't think it's sensible that we have to remember these things when writing bindings

[...]

PR #351, opened 2022-03-08
usize -> u64 or u32

mimoo:
helps https://github.com/o1-labs/proof-systems/issues/346

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -169,7 +169,7 @@ pub enum ConstantExpr<F> {
     EndoCoefficient,
     Mds { row: usize, col: usize },
     Literal(F),
-    Pow(Box<ConstantExpr<F>>, usize),
+    Pow(Box<ConstantExpr<F>>, u64),
     // TODO: I think having separate Add, Sub, Mul constructors is faster than
     // having a BinOp constructor :(
     Add(Box<ConstantExpr<F>>, Box<ConstantExpr<F>>),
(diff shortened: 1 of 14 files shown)


kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -554,7 +554,7 @@ mod tests {
                 h.insert(Column::Witness(i));
             }
             for i in 0..(lookup_info.max_per_row + 1) {
-                h.insert(Column::LookupSorted(i));
+                h.insert(Column::LookupSorted(i as usize));
```
mrmr1993, line 557: Nit: `LookupSorted` should also become `u32` (although not necessarily in this PR)

kimchi/src/circuits/expr.rs
```diff
@@ -169,7 +169,7 @@ pub enum ConstantExpr<F> {
     EndoCoefficient,
     Mds { row: usize, col: usize },
     Literal(F),
-    Pow(Box<ConstantExpr<F>>, usize),
+    Pow(Box<ConstantExpr<F>>, u64),
```
mrmr1993, line 172: I don't think we ever use a large `Pow`, it might be safer to move to `u32`

mrmr1993 (approved):
Great, thank you for this!

kimchi/src/circuits/expr.rs
mimoo: I kept `u64` as this is what arkworks expect when we use `pow` there, but I don't see a problem moving this to `u32` either

kimchi/src/circuits/polynomials/chacha.rs
mimoo: I tried that but this led to another rabbit hole so made a minimal PR

PR #353, opened 2022-03-09
[easy][utils][dense_polynomial] improve doc + code

mimoo:
as the title says

### DIFF
--- a/utils/src/dense_polynomial.rs
+++ b/utils/src/dense_polynomial.rs
@@ -22,10 +22,10 @@ pub trait ExtendedDensePolynomial<F: Field> {
     /// This function evaluates polynomial in chunks.
     fn eval(&self, elm: F, size: usize) -> Vec<F>;
 
-    /// Multiplies the chunks of a polynomial with powers of zeta^n to make it of degree n.
+    /// Multiplies the chunks of a polynomial with powers of zeta^n to make it of degree n-1.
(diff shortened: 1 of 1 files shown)


utils/src/dense_polynomial.rs
```diff
@@ -22,7 +22,10 @@ pub trait ExtendedDensePolynomial<F: Field> {
     /// This function evaluates polynomial in chunks.
     fn eval(&self, elm: F, size: usize) -> Vec<F>;
 
-    /// Multiplies the chunks of a polynomial with powers of zeta^n
+    /// Multiplies the chunks of a polynomial with powers of zeta^n to make it of degree n.
```
querolita, line 25: I am not sure I understand the reasoning. 
If `f(x) = f0(x) + x^n * f1(x) + x^{2n} * f2(x)` and all three `f0, f1, f2` have degree n-1, then I understand that `f` has degree `n-1+2n = 3n-1`.

querolita (commented):
I have one comment, once solved I will get back to the PR and accept

utils/src/dense_polynomial.rs
mimoo: that's correct, that's why we do:

*`x^n` -> `z^n`
* `z^2n` -> `z^2n`

mimoo: probable the comment here should say "of degree n-1"

mrmr1993 (approved):

PR #354, opened 2022-03-09
[specs][kimchi] overview + prover

mimoo:
Adding an overview, and a prover spec. There's still a lot of elements that are placeholders (like lookup, or endomorphism trick) or that rely on missing specifications (like poly commitments, or poseidon).


book/specifications/kimchi/template.md
```diff
@@ -1,16 +1,87 @@
...
+|  row  |    0    |   1   |   2   |   3   |    4     |   5   |    6     |
+| :---: | :-----: | :---: | :---: | :---: | :------: | :---: | :------: |
+|   0   |   0,0   |  0,1  |  0,2  |  0,3  | **80,6** |  0,5  |   0,6    |
+|  ...  |         |       |       |       |          |       |          |
+|  80   |  80,0   | 80,1  | 80,2  | 80,3  |   80,4   | 80,5  | **90,0** |
+|  ...  |         |       |       |       |          |       |          |
+|  90   | **0,4** | 90,1  | 90,2  | 90,3  |   90,4   | 90,5  |   90,6   |
+
+**Wiring (Permutation) trace**. You can think of the permutation trace as an extra register that is used to enforce the wiring specified in the wiring table. 
+It is a single column that applies on all the rows as well, which the prover computes as part of a proof.
```
jspada, line 58: Could you explain more about how this is represented and integrated, maybe by giving the structure of the polynomial or formula as it relates to this?

book/specifications/kimchi/template.md
```diff
@@ -1,16 +1,87 @@
 # Kimchi
```
jspada, line 1: Awesome work!

jspada (approved):

book/src/specs/kimchi.md
```diff
@@ -1,16 +1,87 @@
...
+
+|  row  |   0   |   1   |   2   |   3   |   4   |   5   |   6   |   7   |   8   |   9   |  10   |  11   |  12   |  13   |  14   |
+| :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
+|   0   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |
+
+
+**Wiring (or Permutation)**. For gates to take output of other gates, we use a wiring table to wire registers together. 
+It is defined at every row, but only for the first $7$ registers. 
+Each cell specifies a `(row, column)` tuple that it should be wired to. 
+Note that if three or more registered are wired together, they must form a cycle.
```
querolita, line 46: registers

kimchi/src/prover.rs
```diff
@@ -102,7 +102,11 @@ where
             index.cs.verify(&witness).expect("incorrect witness");
         }
 
-        // ensure we have room for the zero-knowledge rows
+        //~ 1. Ensure we have room in the witness for the zero-knowledge rows.
```
querolita, line 105: great point

kimchi/src/prover.rs
```diff
@@ -126,18 +131,28 @@ where
...
         .interpolate();
 
-        // commit to the wire values
+        //~ 5. Commit (non-hidding) to the negated public input polynomial. **TODO: seems unecessary**
+        let public_comm = index.srs.commit_non_hiding(&public_poly, None);
+
+        //~ 6. Absorb the public polynomial with the Fq-Sponge. **TODO: seems unecessary**
+        fq_sponge.absorb_g(&public_comm.unshifted);
+
+        //~ 7. Commit to the witness columns by creating `COLUMNS` hidding commitments.
```
querolita, line 153: hiding

kimchi/src/prover.rs
```diff
@@ -148,7 +163,13 @@ where
...
         });
 
-        // compute witness polynomials
+        //~ 8. Absorb the witness commitments with the Fq-Sponge.
+        w_comm
+            .iter()
+            .for_each(|c| fq_sponge.absorb_g(&c.0.unshifted));
+
+        //~ 9. Compute the witness polynomials by interpolating each `COLUMNS` of the witness.
+        //~    TODO: why not do this first, and the commit? Why commit from evaluation directly?
```
querolita, line 172: then

kimchi/src/prover.rs
```diff
@@ -405,10 +433,14 @@ where
             }
         };
 
-        //
-        // compute quotient polynomial
-        //
-
+        //~ 23. Compute the quotient polynomial (the $t$ in $f = Z_H \cdot t$).
```
querolita, line 436: awesome summary

kimchi/src/prover.rs
```diff
@@ -650,123 +688,191 @@ where
...
-        let chunked_evals_zeta_omega = ProofEvaluations::<Vec<Fr<G>>> {
-            s: array_init(|i| {
-                index.cs.sigmam[0..PERMUTS - 1][i].eval(zeta_omega, index.max_poly_size)
-            }),
-            w: array_init(|i| witness_poly[i].eval(zeta_omega, index.max_poly_size)),
-            z: z_poly.eval(zeta_omega, index.max_poly_size),
-            lookup: lookup_evals(zeta_omega),
-            generic_selector: index.cs.genericm.eval(zeta_omega, index.max_poly_size),
-            poseidon_selector: index.cs.psm.eval(zeta_omega, index.max_poly_size),
+        //~ 29. Chunk evaluate the following polynomials at both $\zeta$ and $\zeta \omega$:
```
querolita, line 691: do we ever give the intuition that $\zeta$ refers to a row $i$ and $\zeta\omega$ to row $i+1$? if so, where?

kimchi/src/prover.rs
```diff
@@ -650,123 +688,191 @@ where
...
             ]
         };
+
+        //~ 37. Absorb all the polynomial evaluations in $\zeta$ and $\zeta\omega$:
+        //~     - the public polynomial
+        //~     - z
+        //~     - generic selector
+        //~     - poseidon selector
+        //~     - the 15 register/witness
+        //~     - the 6 sigmas evaluations
```
querolita, line 848: 6 or 7? also, continuing on another comment of the review, is there a sigma for each I/O register?

kimchi/src/prover.rs
```diff
@@ -776,29 +882,20 @@ where
...
             .collect::<Vec<_>>();
+
+        //~ 44. Then, include:
+        //~     - the negated public polynomial (TODO: why?)
+        //~     - the ft polynomial
+        //~     - the permutation aggregation polynomial z polynomial
+        //~     - the generic selector
+        //~     - the poseidon selector
+        //~     - the 15 registers/witness columns
+        //~     - the 6 s
```
querolita, line 898: again I ask if 6s or 7 (because 7 I/O registers)

querolita (approved):
Great specification of kimchi and even more importantly, the prover.

book/specifications/kimchi/template.md
mimoo: This should be specified in the permutation section. But to answer the question quickly, it is supposed to be this polynomial (without the division by the vanishing polynomial, and with 7 columns instead of 3, and [with some part removed that the verifier can compute themselves](https://o1-labs.github.io/mina-book/crypto/plonk/final_check.html))

Screen Shot 2022-03-11 at 11 48 42 AM

taken from https://o1-labs.github.io/mina-book/crypto/plonk/zkpm.html#zero-knowledge-for-the-permutation-polynomial

mimoo: @querolita the sigma permutation polynomials is what I call "wiring (or permutation)" right above that. Probably a good idea to also call it sigmas.

book/specifications/kimchi/template.md
```diff
@@ -1,16 +1,87 @@
...
-This document specifies the kimchi variant of PLONK.
+* This document specifies *kimchi*, a zero-knowledge proof system that's a variant of PLONK.
+* This document does not specify how circuits are created, or executed. Only how to convert a circuit and its execution into a proof.
 
 ## Overview
 
-TKTK
+There's three main algorithms to kimchi:
+
+* [Setup](#constraint-system-creation): takes a circuit and produces a prover index, and a verifier index.
```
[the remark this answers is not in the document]
mimoo: > what's the name of the other part of the setup, where some elements of the SRS would be generated for up to a certain bound? 

so we don't have that, as I pointed out in another comment. I think it'd be great to have though, especially as mina reuses the same SRS for all circuits, and @imeckler PR on the trimmedSRS was going in that direction (where an index only uses part of the SRS, if the circuit is not the maximum size)

book/specifications/kimchi/template.md
```diff
@@ -1,16 +1,87 @@
...
+**Gates**. A circuit is described by a series of gates, that we list in a table. 
+The columns of the tables list the gates, while the rows are the length of the circuit. 
+For each row, only a single gate can take a value $1$ while all other gates take the value $0$.
+
+|  row  | Generic | Poseidon | CompleteAdd | VarBaseMul | EndoMul | EndoMulScalar | ChaCha0 | ChaCha1 | ChaCha2 | ChaChaFinal |
+| :---: | :-----: | :------: | :---------: | :--------: | :-----: | :-----------: | :-----: | :-----: | :-----: | :---------: |
+|   0   |    1    |    0     |      0      |     0      |    0    |       0       |    0    |    0    |    0    |      0      |
+|   1   |    0    |    1     |      0      |     0      |    0    |       0       |    0    |    0    |    0    |      0      |
+
+**Coefficients**. The coefficient table has 15 columns, and is used to tweak the gates.
```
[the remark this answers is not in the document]
mimoo: > Is the number of columns of the coefficients table necessarily linked to the number of columns of the registers?

Indirectly. Right it is linked to the Poseidon gate, which needs 15 round constants to be specified somewhere (and different poseidon gate might have different round constants, so it's not fixed). In turn, the poseidon manages to do 15 transitions because we have 15 registers/columns.

book/specifications/kimchi/template.md
```diff
@@ -1,16 +1,87 @@
...
+|  row  |   0   |   1   |   2   |   3   |   4   |   5   |   6   |   7   |   8   |   9   |  10   |  11   |  12   |  13   |  14   |
+| :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
+|   0   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |   /   |
+
+
+**Wiring (or Permutation)**. For gates to take output of other gates, we use a wiring table to wire registers together. 
+It is defined at every row, but only for the first $7$ registers. 
+Each cell specifies a `(row, column)` tuple that it should be wired to. 
+Note that if three or more registered are wired together, they must form a cycle. 
+For example, if register `(0, 4)` is wired to both registers `(80, 6)` and `(90, 0)` then you would have the following table:
```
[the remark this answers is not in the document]
mimoo: yeah, so your example wouldn't work because you can't convert that to the permutation argument

book/specifications/kimchi/template.md
querolita: Oooooh I see ok ok... I was thinking of the above polynomial as a separate thing, but never as a possible additional column in the trace.
Regarding the sigmas, thanks for clarifying. These two things are quite interconnected and the way you have split it in two makes it easier.

book/specifications/kimchi/template.md
```diff
@@ -94,14 +166,75 @@ TODO: use expr to define the index columns?
...
 
+Originally, kimchi is based on an interactive protocol that was transformed into a non-interactive one using the [Fiat-Shamir](https://o1-labs.github.io/mina-book/crypto/plonk/fiat_shamir.html) transform.
+For this reason, it can be useful to visualize the high-level interactive protocol before the transformation:
+
+
+```mermaid
+sequenceDiagram
+    participant Prover
+    participant Verifier
+    Prover->>Verifier: public & witness commitment
```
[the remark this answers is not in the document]
mimoo: public input yeah, but I think we should remove this from the protocol if we can. This means that we would have to modify the verifier circuit though... cc @imeckler

kimchi/src/prover.rs
mimoo: We don't. I'd be a great addition though, not sure where we could add that in the spec :o if you want to give it a shot be my guest!

kimchi/src/prover.rs
mimoo: 6 actually, because one is not evaluated and kept as a commitment

Issue #356, opened 2022-03-10
Create test of verifier with public inputs


mimoo:
I don't believe there's any test that include public inputs for the verify function

mimoo:
https://github.com/o1-labs/proof-systems/pull/360

PR #360, opened 2022-03-10
[tests] create a test for public inputs

mimoo:
will write a PR description in 3, 2, 1, ...

closes https://github.com/o1-labs/proof-systems/issues/356

### DIFF
--- a/kimchi/src/circuits/polynomials/generic.rs
+++ b/kimchi/src/circuits/polynomials/generic.rs
@@ -313,6 +313,7 @@ pub mod testing {
             &self,
             row: usize,
             witness: &[Vec<F>; COLUMNS],
+            public: &[F],
         ) -> Result<(), String> {
             // assignments
             let this: [F; COLUMNS] = array_init(|i| witness[i][row]);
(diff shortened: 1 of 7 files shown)


kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -330,9 +331,14 @@ pub mod testing {
...
                     * this[register_offset + 1];
+                let public = if coeffs_offset == 0 {
+                    public.get(row).cloned().unwrap_or(F::zero())
+                } else {
+                    F::zero()
+                };
                 ensure_eq!(
                     zero,
-                    sum + mul + self.coeffs[coeffs_offset + 4],
+                    sum + mul + self.coeffs[coeffs_offset + 4] - public,
```
mrmr1993, line 341: I don't think this is right: the public input is already accounted for as part of the witness, and the gates for the public input should already be at the beginning of the witness. The old version was already correct.

mrmr1993 (changes requested):
Only the updates to the test framework are right here; the rest (changes to `verify` functions) should be reverted

mimoo:
sorry didn't see you were reviewing, the tests still don't pass so I'm looking for what's wrong

kimchi/src/circuits/polynomials/generic.rs
mimoo: I'm not sure I understand. If you don't remove the public element, which is in the witness, then the sum is not going to be equal to zero.

For the first `public` rows the equation is going to be `generic_selector( alpha^gen1 * (1 * w[0])) = w[0]`

mimoo: indeed the tests pass now, so that must be that I'm right :P

mrmr1993: Ah, indeed. My mistake!

mrmr1993 (approved):

PR #361, opened 2022-03-10
[verifier] reuse pre-computed lagrange bases we already have

mimoo:
before:

* we expected the verifier to pass enough lagrange bases that we have public inputs
* verify was a batch verify function

after:

* we rely on the pre-computed lagrange bases we already have
* verify -> batch_verify

### DIFF
--- a/kimchi/src/verifier.rs
+++ b/kimchi/src/verifier.rs
@@ -390,9 +390,9 @@ where
 ///     index: VerifierIndex
 ///     RETURN: verification status
 #[allow(clippy::type_complexity)]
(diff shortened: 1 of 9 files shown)


kimchi/src/verifier.rs
```diff
@@ -383,301 +383,313 @@ where
             combined_inner_product,
```
mimoo, line 383: make sure to check 
Screen Shot 2022-03-10 at 10 22 12 AM

imeckler (approved):

PR #363, opened 2022-03-10
Include `table_id` in the calculation of lookup terms

mrmr1993:
This PR modifies the calculation for lookup terms to include the `table_id` for the lookup.

Currently, the code only allows for 1 table, and the table that we use has `table_id= 0`, so the term for the `table_id` is erased. Once there is support for multiple tables, this will become actually useful.


kimchi/src/prover.rs
```diff
@@ -711,17 +747,31 @@ where
...
+                                    .collect()
+                            });
+                        match lcs.table_ids.as_ref() {
+                            None => base_table,
+                            Some(table_ids) => {
+                                let table_combiner =
+                                    joint_combiner.pow([lcs.configuration.max_joint_size as u64]);
+                                base_table
+                                    .into_iter()
+                                    .zip(table_ids.eval(e, index.max_poly_size))
```
mimoo, line 769: btw this has now changed https://github.com/o1-labs/proof-systems/pull/467/

```rust
to_chunked_polynomial(index.max_poly_size).evaluate_chunks(e)
```

that being said, do we really want to handle split polynomials this way for lookup tables (and table_ids)? As we will need to cleanly ignore specific rows (the zk rows). More generally I feel like it'd be better to just deal with evaluations instead of chunked evaluations and treat the quotient polynomial (I believe the only exception) as an exception with a special type or something

kimchi/src/verifier.rs
```diff
@@ -539,6 +539,13 @@ where
                                 scalars.push(scalar * j);
                                 commitments.push(t);
                             }
+                            if let Some(table_ids) = lindex.table_ids.as_ref() {
+                                scalars.push(
+                                    (scalar * constants.joint_combiner)
+                                        .pow([lindex.max_joint_size as u64]),
```
mimoo, line 545: shouldn't this be 

```rust
scalar * (constants.joint_combiner.pow(...))
```

kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -157,12 +157,32 @@ fn single_lookup<F: FftField>(s: &SingleLookup<F>) -> E<F> {
...
+fn joint_lookup<F: FftField>(j: &JointLookup<F>, max_joint_size: u32) -> E<F> {
+    // The domain-separation term, used to ensure that a lookup against a given table cannot
+    // retrieve a value for some other table.
+    let table_id_contribution = {
+        let table_id: F = {
+            if j.table_id >= 0 {
+                F::from(j.table_id as u32)
+            } else {
+                -F::from(-j.table_id as u32)
+            }
```
mimoo, line 169: use the `i32_to_field()` function

mimoo (approved):
LGTM besides the few comments

kimchi/src/circuits/gate.rs
```diff
@@ -55,6 +55,24 @@ pub struct SingleLookup<F> {
...
+
+// The domain-separation term calculated from the given `table_id`, used to ensure that a lookup
+// against a given table cannot retrieve a value for some other table.
+pub fn table_id_contribution<F: Field>(joint_combiner: F, table_id: F, max_joint_size: u32) -> F {
+    // Here, we use `joint_combiner^max_joint_size` rather than incrementing the powers of the
+    // `joint_combiner` in the table value calculation below. This ensures that we can avoid
+    // using higher powers of the `joint_combiner` when we have only one table with a
+    // `table_id` of 0.
+    joint_combiner.pow([max_joint_size as u64]) * table_id
+}
```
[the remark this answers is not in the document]
mrmr1993: Yea, I was planning to do that when I fix merge conflicts

kimchi/src/circuits/gate.rs
```diff
@@ -87,21 +108,26 @@ impl<F: Field> SingleLookup<F> {
 /// A spec for checking that the given vector belongs to a vector-valued lookup table.
 #[derive(Clone, Serialize, Deserialize)]
 pub struct JointLookup<F> {
-    pub table_id: usize,
+    pub table_id: i32,
```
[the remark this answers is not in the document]
mrmr1993: They're not strictly necessary, but for ease of use I was planning to have negative IDs for 'runtime' tables and positive ones for the 'fixed' tables. We currently hard-code the table ID for XOR, this was mostly a workaround so that we don't have to account for that / any other fixed tables that we add.

I'll add a comment to this effect.

kimchi/src/verifier.rs
mrmr1993: Good catch!

mimoo: scalar is probably `1` anyway here right?

kimchi/src/prover.rs
```diff
@@ -711,17 +747,31 @@ where
...
+                            .lookup_table
+                            .iter()
+                            .map(|p| p.eval(e, index.max_poly_size))
+                            .rev()
+                            .fold(vec![Fr::<G>::zero()], |acc, x| {
+                                acc.into_iter()
+                                    .zip(x.iter())
+                                    .map(|(acc, x)| acc * joint_combiner + x)
+                                    .collect()
+                            });
```
mimoo, line 761: btw I think it'd be nice to move all the similar (join columns with joint_combiner code) to a single place behind some abstraction. Could open an issue about that

kimchi/src/verifier.rs
mrmr1993: Must be, because my tests passed. Still, better to fix it for future :)

utils/src/field_helpers.rs
```diff
@@ -86,6 +87,14 @@ impl<F: Field> FieldHelpers<F> for F {
     }
 }
 
+pub fn i32_to_field<F: From<u64> + Neg<Output = F>>(i: i32) -> F {
```
querolita, line 90: Very useful! I have some of these for Cairo. Nonetheless, I think you can do it with intermediate `u32` instead. Meaning, `i32` has one leading bit for positive/negative, and then 31 bits of actual digits [-2147483648..2147483647], but `u32` has 32 whole bits for natural numbers [0..4294967295]. Nonetheless, I guess that the impact on efficiency is minimal.

PR #366, opened 2022-03-11
[verifier] clean up

mimoo:
list of changes:

* remove `combined_inner_product` as it was unused!
* extract the per-proof logic in the verifier to a `to_batch` (could use a better name)
* `to_batch` constructs a new type `BatchEvaluationProof` instead of a tuple (could use a better name as well)
* `BatchEvaluationProof` uses a new type `Evaluations` instead of a tuple (could use a better name as well)

my focus was on extracting the per-proof logic so that it's easier to specify. As there were tuples of tuples of vecs of vecs everywhere it was a bit hard to do, hence the new types created to track these. This improves the verify code mostly, but I think the poly-commitment could still use some love (hopefully will happen when we spec that part out).


kimchi/src/verifier.rs
```diff
@@ -43,7 +43,7 @@ where
     /// the computed powers of alpha
     pub all_alphas: Alphas<Fr<G>>,
     /// public polynomial evaluations
-    pub p_eval: [Vec<Fr<G>>; 2],
+    pub p_eval: Vec<Vec<Fr<G>>>,
```
mrmr1993, line 46: What motivates this relaxation of types?

mimoo: that was a fast review o_O basically we later use the polycommitment verification API with vectors of vectors of evaluations. I'm guessing this is because the API is compatible with any number of evaluations (not just 2)

querolita (approved):

kimchi/src/verifier.rs
```diff
@@ -380,11 +338,301 @@ where
             polys,
             zeta1,
             ft_eval0,
-            combined_inner_product,
         }
     }
 }
 
+fn to_batch<'a, G, EFqSponge, EFrSponge>(
```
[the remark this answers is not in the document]
mimoo: it'd probably be a good idea if we explain the difference between aggregating (on the prover side) and batching (on the verifier side). I've seen a lot of confusions between these terms

PR #367, opened 2022-03-11
[easy][bench] reflect our verifier circuit size in the bench

mimoo:


kimchi/src/bench.rs
```diff
@@ -25,8 +25,9 @@ type SpongeParams = PlonkSpongeConstantsKimchi;
 type BaseSponge = DefaultFqSponge<VestaParameters, SpongeParams>;
 type ScalarSponge = DefaultFrSponge<Fp, SpongeParams>;
 
-/// the circuit size. This influences the size of the SRS
-pub const CIRCUIT_SIZE: usize = (1 << 14) + 1; // SRS will be 2^15
+/// The circuit size. This influences the size of the SRS.
+/// At the time of this writing our verifier circuits have 27164 & 18054 gates.
+pub const CIRCUIT_SIZE: usize = 27164;
```
mrmr1993, line 30: If we want to be able to meaningfully compare over time, this value shouldn't be regularly updated. Why don't we go for `2^15-4` instead, so that we're using the same size of SRS, but changes to the verifier circuit won't make it look like we changed the performance of the code here.

mimoo: I don't think we need to update this value over time, it's just that it's nice to have a bench that shows you the time that a proof takes in production (since I would expect the size of our proofs to remain near this size for the foreseeing future).

I think what you're thinking about is a different kind of bench, where we'd be able to see timings for different SRS sizes. That'd be really awesome to have as well (and post in our README), but for a different PR :o

mimoo: similar to this btw: https://github.com/novifinancial/winterfell#performance

PR #368, opened 2022-03-11
[specs][kimchi] index and constraint system parts

mimoo:
To make it easier to spec the constraint system and index setup:

* index -> prover_index and verifier_index
* index::linearization -> linearization

This also adds some spec comment in order to fill the spec out. More to come.

### DIFF
--- a/book/specifications/kimchi/template.md
+++ b/book/specifications/kimchi/template.md
@@ -88,9 +88,17 @@ TODO: use expr to define the index columns?
 
 {sections.constraint_system}
 
-## Prover and Verifier Index Creation
+## Prover Index
 
-{sections.indexes}
+{sections.prover_index}
+
+## Verifier Index
+
+The verifier index is essentially a number of pre-computations containing:
+
+* the (non-hidding) commitments of all the required polynomials
+
+{sections.verifier_index}
 
 ## Proof Data Structure
 
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -188,10 +188,33 @@ fifth round:
 
 ## Constraint System Creation
 
+1. +3 on gates.len() here to ensure that we have room for the zero-knowledge entries of the permutation polynomial
+   see https://minaprotocol.com/blog/a-more-efficient-approach-to-zero-knowledge-for-plonk
+2. pad the rows: add zero gates to reach the domain size
 
 
 ## Prover and Verifier Index Creation
 
+
+### The prover Index
+
+
+#### Linearization
+
+
+#### Prover Index Creation
+
+1. compute the linearization
+2. set `max_quot_size` to the degree of the quotient polynomial,
+   which is obtained by looking at the highest monomial in the sum
+    $$\sum_{i=0}^{PERMUTS} (w_i(x) + \beta k_i x + \gamma)$$
+   where the $w_i(x)$ are of degree the size of the domain.
+
+### The verifier Index
+
+The verifier index is essentially a number of pre-computations containing:
+
+* the (non-hidding) commitments of all the required polynomials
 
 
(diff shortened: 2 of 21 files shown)


jspada (approved):
just moving things around.  LGTM!

book/specifications/kimchi/template.md
```diff
@@ -88,9 +88,17 @@ TODO: use expr to define the index columns?
...
+## Prover Index
 
-{sections.indexes}
+{sections.prover_index}
+
+## Verifier Index
+
+The verifier index is essentially a number of pre-computations containing:
+
+* the (non-hidding) commitments of all the required polynomials
```
querolita, line 99: are these polynomials, partly, each of the gate constraints of the specific circuit? Meaning, the specification of the circuit, which is public information (thus non-hiding commitments being used).
Typo -> non-hiding, only 1 d I think

book/src/specs/kimchi.md
```diff
@@ -188,9 +188,28 @@ fifth round:
 
 ## Constraint System Creation
 
+1. +3 on gates.len() here to ensure that we have room for the zero-knowledge entries of the permutation polynomial
+   see https://minaprotocol.com/blog/a-more-efficient-approach-to-zero-knowledge-for-plonk
+2. pad the rows: add zero gates to reach the domain size
 
 
-## Prover and Verifier Index Creation
+## Prover Index
```
querolita, line 196: If I understand correctly, this index is dependent on the circuit itself? Like, this prover index is made effective only after the "general" SRS has been created? ( I refer to this question in the other PR as well, making a distinction between Setup and Index in papers)

book/src/specs/kimchi.md
```diff
@@ -188,9 +188,28 @@ fifth round:
...
+   which is obtained by looking at the highest monomial in the sum
+    $$\sum_{i=0}^{PERMUTS} (w_i(x) + \beta k_i x + \gamma)$$
+   where the $w_i(x)$ are of degree the size of the domain.
+
+
+## Verifier Index
+
+The verifier index is essentially a number of pre-computations containing:
+
+* the (non-hidding) commitments of all the required polynomials
```
querolita, line 212: non hiding

kimchi/src/linearization.rs
```diff
@@ -0,0 +1,106 @@
+//! This module implements the linearization.
```
querolita, line 1: I like this to have a separate file

querolita (approved):
I like this modular approach

book/src/specs/kimchi.md
mimoo: > If I understand correctly, this index is dependent on the circuit itself?

Yeah you basically compile the circuit to these two indexes

book/specifications/kimchi/template.md
querolita: Oh! so only those polynomials that will tell the verifier what gate appears in each row I see... The mutually exclusive polynomials you mean.

Well I have one question. I understand that the verifier checks that `quot(X) * van(X) = poly(x)`, where `poly` is the set of all constraints, permutations etc. Then, how does it check that `poly(X)` indeed corresponds to "the shape that it should"? Will it "rebuild" the equation checks using the committed witness together with the set of constraints? (but perhaps this approach takes linear time in the number of rows, right?) I don't know if I am making myself clear.

mimoo: that's what I call the polynomial dance between the prover and the verifier. The verifier doesn't want to just get poly (or f) from the prover, the verifier recreates an f from the constraints of each gate, the witness, the permutation polynomial, etc. Oh, and they do that hidden in the exponent, so as a commitment, not as a polynomial directly.

PR #371, opened 2022-03-11
[specs] address feedback

mimoo:
addresses feedback from @querolita on specs

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -148,7 +148,7 @@ where
         )
         .interpolate();
 
-        //~ 5. Commit (non-hidding) to the negated public input polynomial. **TODO: seems unecessary**
+        //~ 5. Commit (non-hiding) to the negated public input polynomial. **TODO: seems unecessary**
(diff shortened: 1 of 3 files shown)


kimchi/src/prover.rs
```diff
@@ -173,7 +173,7 @@ where
             .for_each(|c| fq_sponge.absorb_g(&c.0.unshifted));
 
         //~ 9. Compute the witness polynomials by interpolating each `COLUMNS` of the witness.
-        //~    TODO: why not do this first, and the commit? Why commit from evaluation directly?
+        //~    TODO: why not do this first, and then commit? Why commit from evaluation directly?
```
mrmr1993, line 176: * we need to modify the domain to `d8`, so we find the polynomial that evaluates to the witness on `d1`, then evaluate it on `d8`
* the verifier only gets the evaluations, so we need to commit based on the evaluations in order for the verifier to generate a matching sponge state.

mrmr1993 (approved):

PR #376, opened 2022-03-12
Include `table_id` in the calculation of constraints

mrmr1993:
This PR builds upon #363, calculating the `table_id` column when needed, tracking a table ID for the dummy value, and adding the `table_id` to the constraints.

That this works correctly can be tested by changing `XOR_TABLE_ID` from `0` to `1` and running `cargo test`. You can also verify that this has added the correct constraints by replacing one use of `XOR_TABLE_ID` with `1` and the others with `0`, and noting that the chacha test fails as expected.


kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -261,6 +261,9 @@ pub struct LookupConfiguration<F: FftField> {
     /// particular row, so that we can treat each row uniformly as having the same number of
     /// lookups.
     pub dummy_lookup_value: Vec<F>,
+
+    /// The table ID to combine with the placeholder lookup value.
+    pub dummy_lookup_table_id: i32,
```
mimoo, line 266: why not store as `F` directly?

kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -517,6 +520,13 @@ pub fn sorted<
             sorted[i].push(end_val);
         }
 
+        // Duplicate the final sorted value, to fix the off-by-one in the last lookup row.
+        // This is caused by the snakification: all other sorted columns have the value from the
+        // next column added to their end, but the final sorted column has no subsequent column to
+        // pull this value from.
+        let final_sorted_col = &mut sorted[max_lookups_per_row];
+        final_sorted_col.push(final_sorted_col[final_sorted_col.len() - 1].clone());
```
mimoo, line 528: I'm struggling to understand what is happening here. 

* we're making the assumption that final_sorted_col is the last column, but that's only true if the tables are taking a single column (and are not divided over several columns)
* we're adding one more value in that last column, but that means the last column will be of size `n + 1 > n` now, and I imagine the last value will be dismissed when interpolated in the domain of size `n`

kimchi/src/circuits/gate.rs
```diff
@@ -217,7 +220,11 @@ pub enum GateLookupTable {
     Xor,
 }
 
-pub type LookupTable<F> = Vec<Vec<F>>;
+/// A table of values that can be used for a lookup, along with the ID for the table.
+pub struct LookupTable<F> {
+    pub id: i32,
+    pub data: Vec<Vec<F>>,
+}
```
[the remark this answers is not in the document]
mrmr1993: I agree, it would be better to manage this automatically wherever we can. Although I haven't built anything in this direction, that was part of the motivation for 'negative' table IDs, so that snarky and other library users can allocate runtime tables without concern over how we might change this automatic allocation.

kimchi/src/circuits/polynomials/lookup.rs
mrmr1993: I have an issue for combining this with the table value as a `JointLookup`. This PR predates the API change, so it was easier at the time to use an `i32`; I'll update this after merging.

kimchi/src/circuits/polynomials/lookup.rs
mrmr1993: This is because we reverse the columns: this was an edge case before, because the last column was previously not reversed. It actually doesn't matter for this change, but I discovered this in testing with previous PRs, because we push a value to the end of every other column. Off-by-1 errors abound :)

mimoo (approved):

PR #381, opened 2022-03-12
Support multiple lookup tables

mrmr1993:
This PR builds upon #363. This
* adds support for multiple tables, by concatenating them into one combined table, and accumulating the corresponding table IDs in a column
* fixes the dummy lookup value as 0, for efficiency
* adds non-panic error handling to `LookupConstraintSystem::create`
* adds domain separation tests, using the chacha gate.


mrmr1993:
> can I start reviewing this one btw :o ?

It depends on #376, which in turn depends on #363, so it's probably best to review those first.

kimchi/src/circuits/constraints.rs
```diff
@@ -343,35 +355,74 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
+                        lookup_table.extend((0..table_len).map(|_| F::zero()))
+                    }
+                }
+
+                // Note: we use `>=` here to leave space for the dummy value.
+                if lookup_table[0].len() >= max_num_entries {
+                    return Err(LookupError::LookupTableTooLong {
+                        length: lookup_table[0].len(),
+                        maximum_allowed: max_num_entries - 1,
+                    });
```
mimoo, line 399: I'm confused with this one as well. max_num_entries is the size of d1 minus the rows where the grand product argument is not applied. This would be the zk_rows as well as the row before that (needed to check that the final value of the grand product is 1). It looks like max_num_entries is calculated this way, so why not `>` instead of `>=`?

kimchi/src/circuits/constraints.rs
```diff
@@ -343,35 +355,74 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
-                    .data
-                    .iter()
-                    .map(|col| col[col.len() - 1])
-                    .collect();
-                let dummy_lookup_table_id = lookup_table.id;
+                // Pad up to the end of the table with the dummy value.
+                lookup_table
+                    .iter_mut()
+                    .for_each(|col| col.extend((col.len()..max_num_entries).map(|_| F::zero())));
+                table_ids.extend((table_ids.len()..max_num_entries).map(|_| F::zero()));
```
mimoo, line 411: same here with iter::repeat

kimchi/src/circuits/constraints.rs
```diff
@@ -343,35 +355,74 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
-                let dummy_lookup_value: Vec<F> = lookup_table
-                    .data
-                    .iter()
-                    .map(|col| col[col.len() - 1])
-                    .collect();
-                let dummy_lookup_table_id = lookup_table.id;
+                // Pad up to the end of the table with the dummy value.
+                lookup_table
+                    .iter_mut()
+                    .for_each(|col| col.extend((col.len()..max_num_entries).map(|_| F::zero())));
```
mimoo, line 410: par_iter_mut?

also, why use `F::zero()` here? This is creating an assumption that the dummy value for the `table_id:0` is always going to be `0` right? It'd be better not to rely on this assumption here :o

kimchi/src/circuits/constraints.rs
```diff
@@ -343,35 +355,74 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
+                table_ids.extend((table_ids.len()..max_num_entries).map(|_| F::zero()));
 
                 // pre-compute polynomial and evaluation form for the look up tables
                 let mut lookup_table_polys: Vec<DP<F>> = vec![];
                 let mut lookup_table8: Vec<E<F, D<F>>> = vec![];
-                for (mut col, dummy) in lookup_table.data.into_iter().zip(&dummy_lookup_value) {
+                for col in lookup_table.into_iter() {
                     // pad each column to the size of the domain
```
mimoo, line 417: remove comment

kimchi/src/circuits/constraints.rs
```diff
@@ -593,7 +644,10 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
         // ------
 
         let lookup_constraint_system =
-            LookupConstraintSystem::create(&gates, lookup_tables, &domain);
+            match LookupConstraintSystem::create(&gates, lookup_tables, &domain) {
+                Ok(res) => res,
+                Err(_) => None?,
+            };
```
mimoo, line 650: note that there's a PR that will make this function return a Result instead of an Option: https://github.com/o1-labs/proof-systems/pull/490

for now, you can use

```rust
let lookup_constraint_system = LookupConstraintSystem::create(&gates, lookup_tables, &domain).ok()?;
```

mimoo (approved):
left you some comments :o but I'm confused as to why the side-loaded API is used to emulate the multiple fixed table feature.

kimchi/src/circuits/constraints.rs
```diff
@@ -343,35 +355,74 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
+                        std::cmp::max(max_width, data.len())
+                    });
+
+                let max_num_entries = d1_size - (ZK_ROWS as usize) - 1;
+
+                let mut lookup_table = vec![Vec::with_capacity(d1_size); max_table_width];
+                let mut table_ids: Vec<F> = Vec::with_capacity(d1_size);
+                let mut non_zero_table_id = false;
+                for table in lookup_tables.iter() {
+                    let table_len = table.data[0].len();
```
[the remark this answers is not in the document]
mimoo: it is checked later!

kimchi/src/tests/chacha.rs
```diff
@@ -98,3 +99,159 @@ fn chacha_prover() {
...
+    };
+
+    // One real witness..
+    push_rows(false);
+    // .. and one fake witness.
+    push_rows(true);
+
+    let lookup_tables = vec![
+        LookupTable {
+            id: 0,
```
mimoo, line 204: this will collide with the XOR lookup table no? shouldn't we panic when we detect tables with the same id?

kimchi/src/tests/chacha.rs
```diff
@@ -98,3 +99,159 @@ fn chacha_prover() {
...
+            data: vec![vec![Fp::from(0); 10]],
+        },
+        LookupTable {
+            id: table_id,
+            data: fakes,
+        },
+    ];
+
+    // create the index
+    let index = new_index_for_test_with_lookups(gates, PUBLIC, lookup_tables);
```
mimoo, line 214: I'm not sure I understand how this could even work, since the verifier will not use these tables when creating the linearized commitment of the lookup grand product argument

mimoo: ok actually nevermind, the verifier index will copy the big lookup table from the prover index

kimchi/src/circuits/constraints.rs
```diff
@@ -343,35 +355,74 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
                     .chain(lookup_tables.into_iter())
                     .collect();
 
-                if lookup_tables.len() > 1 {
-                    panic!("Multiple lookup tables are currently not supported");
+                // Get the max width of all lookup tables
```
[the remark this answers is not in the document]
mimoo: ok I think I get how this is working. But these are neither side-loaded tables, nor fixed tables right?

kimchi/src/circuits/constraints.rs
mrmr1993: It's hard-coded as zero above.

kimchi/src/circuits/constraints.rs
mrmr1993: The dummy is now hard-coded as 0 to make the constraints cheaper. We use that final position to insert the zero dummy value, in case it isn't already present.

kimchi/src/tests/chacha.rs
mrmr1993: We could do, but that would leave us unable to write these tests.

kimchi/src/circuits/constraints.rs
mimoo: I think it'd be a good idea to add a debug assert somewhere to make sure this assumption doesn't change

kimchi/src/tests/chacha.rs
mimoo: perhaps we could find another way to write tests that would make the API less prone to misusage

PR #382, opened 2022-03-12
rename Index to ProverIndex

### LINKED ISSUE #379: rename Index -> ProverIndex
we have a VerifierIndex, so we should have a ProverIndex (which is called Index atm)

relevant code: https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/prover_index.rs#L29

Trivo25:
refactors `Index` to `ProverIndex` as mentioned in #379
`cargo test`, `cargo check` and `cargo build` all run successfully

closes #379

### DIFF
--- a/kimchi/src/prover_index.rs
+++ b/kimchi/src/prover_index.rs
@@ -19,7 +19,7 @@ type Fr<G> = <G as AffineCurve>::ScalarField;
 type Fq<G> = <G as AffineCurve>::BaseField;
 
 //~
-//~ ### The prover ProverIndex
+//~ ### The prover Index
 //~
 
 /// The index used by the prover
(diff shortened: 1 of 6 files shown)


mimoo:
Awesome thanks a lot for the PR :)

kimchi/src/prover_index.rs
```diff
@@ -19,14 +19,13 @@ type Fr<G> = <G as AffineCurve>::ScalarField;
 type Fq<G> = <G as AffineCurve>::BaseField;
 
 //~
-//~ ### The prover Index
```
mimoo, line 22: can you actually revert that? This is what is causing the test issue (as these comments `//~` are special comments that generate the spec

Trivo25: Oh thats interesting, good to know those type of comments actually have a usage!

mimoo: yeah, they get extract to create this specification here which I'm working on at the moment: https://o1-labs.github.io/proof-systems/specs/kimchi.html

mimoo (approved):
Awesome, thanks a lot :)

Issue #390, opened 2022-03-12
[oracle] split parameters for sponge vs permutation


mimoo:
parameters for the permutation should be distinct from parameters for the sponge:

```rust
impl SpongeConstants for PlonkSpongeConstantsLegacy {
    const ROUNDS_FULL: usize = 63; // permutation
    const ROUNDS_PARTIAL: usize = 0; // permutation
    const HALF_ROUNDS_FULL: usize = 0; // permutation
    const SPONGE_CAPACITY: usize = 1; // sponge
    const SPONGE_WIDTH: usize = 3; // sponge
    const SPONGE_RATE: usize = 2; // sponge
    const SPONGE_BOX: u32 = 5; // permutation
    const FULL_MDS: bool = true; // permutation
    const INITIAL_ARK: bool = true; // permutation
}
```

perhaps split `SpongeConstants` into `PermutationParams` and `SpongeParams`?

EDIT: we're tabling this for now, we might revisit later, but now the PR is about changing to something like this:

```rust
trait SpongeConstants {
    const SPONGE_CAPACITY: usize = 1;
    const SPONGE_WIDTH: usize = 3;
    const SPONGE_RATE: usize = 2;

    const PERM_ROUNDS_FULL: usize = 63;
    const PERM_ROUNDS_PARTIAL: usize = 0;
    const PERM_HALF_ROUNDS_FULL: usize = 0;
    const PERM_SBOX: u32 = 5;
    const PERM_FULL_MDS: bool = true;
    const PERM_INITIAL_ARK: bool = true;
}
```

frisitano:
PR up.  I wonder if we also want to change `PlonkSpongeConstantsLegacy` and `PlonkSpongeConstantsKimchi` to `PlonkSpongeParamsLegacy` and `PlonkSpongeParamsKimchi` for consistency.

mrmr1993:
> parameters for the permutation should be distinct from parameters for the sponge:
> ...
> perhaps split `SpongeConstants` into `PermutationParams` and `SpongeParams`?

@mimoo I think this issue will serve to make the code worse rather than better, and would prefer to see this and the relevant PR closed without further action.

At a meta level: please stop opening issues like this, demanding eefact

mimoo:
[...]

PR #392, opened 2022-03-13
Split params into permutation and sponge

frisitano:
This PR splits `SpongeConstants` into `PermutationParams` and `SpongeParams`.

closes: #390

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ We assume that you already have:
 
 Then, you can create an URS for your circuit in the following way:
 
-```ignore
+```rust,ignore
 use kimchi::circuits::constraints;
 use mina_curves::pasta::{fp::Fp, vesta::{Affine, VestaParameters}, pallas::Affine as Other};
 use oracle::{


oracle/src/poseidon.rs
```diff
@@ -34,13 +40,16 @@ impl SpongeConstants for PlonkSpongeConstantsLegacy {
...
-    const ROUNDS_FULL: usize = 55;
-    const ROUNDS_PARTIAL: usize = 0;
-    const HALF_ROUNDS_FULL: usize = 0;
+impl SpongeParams for PlonkSpongeConstantsKimchi {
     const SPONGE_CAPACITY: usize = 1;
     const SPONGE_WIDTH: usize = 3;
     const SPONGE_RATE: usize = 2;
+}
+
+impl PermutationParams for PlonkSpongeConstantsKimchi {
```
mimoo, line 49: it feels a bit weird to me that both `SpongeParams` and `PermutationParams` are implemented on `PlonkSpongeConstantsKimchi`. I would think that you should be able to:

* instantiate a specific permutation via some parameters
* instantiate a sponge using sponge parameters AND a specific permutation. 

Does that make sense? I'm not too sure how this can be cleanly done. Perhaps, `ArithmeticSponge` should also take a `P: Permutation` with:

```rust
trait Permutation {
    const ROUNDS_FULL: usize;
    const ROUNDS_PARTIAL: usize;
    const HALF_ROUNDS_FULL: usize;
    const SPONGE_BOX: u32;
    const FULL_MDS: bool;
    const INITIAL_ARK: bool;
 
    fn poseidon_block_cipher(params: &ArithmeticSpongeParams<F>, state: &mut Vec<F>) {
      // default implementation in the trait
    }
}

struct PermutationKimchi;

impl Permutation for PermutationKimchi {
    const ROUNDS_FULL: usize = 55;
    const ROUNDS_PARTIAL: usize = 0;
    const HALF_ROUNDS_FULL: usize = 0;
    const SPONGE_BOX: u32 = 7;
    const FULL_MDS: bool = true;
    const INITIAL_ARK: bool = false;
}
```

mimoo (commented):
Nice! Thanks for this first PR :)

I think it's going in the right direction, I left you a few comments. Let me know what you think.

oracle/src/poseidon.rs
mrmr1993: @mimoo I'm in the process of replying on your issue, but this proposal in particular seems to make the code more complex for no benefit.

mimoo: I agree for the current version, I'm still curious if we can land on something that looks good (even if we don't merge it quickly, to make sure there's no other issues in snarkyjs & co). 

@frisitano, I don't want to make you lose your time here, with the possibility of this not being landed (we've had issues with complex changes in oracle in the past). I'm still interested to see where this could go, but happy to suggest another task if you want to do something that'll be merged for sure :)

PR #399, opened 2022-03-13
Add a lookup gate kind

mrmr1993:
This PR builds upon #381, defining a lookup gate (currently pinned to perform lookups on the table with ID 1).

This gate has no selector or constraints associated with it, except for the lookups defined against it. The layout is

| 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 |
|---|---|---|---|---|---|---|---|---|---|----|----|----|----|----|
| `table_id` | `i_1` | `x_1` | `i_2` | `x_2` | `i_3` | `x_3` |

where `(i_1, x_1)`, `(i_2, x_2)`, and `(i_3, x_3)` are the pairs of values that make up 3 entries read from the table. The intention is that this will mostly be used with a runtime lookup table, where the `i` corresponds to the index into that table and the `x` corresponds to the value at that index.


kimchi/src/circuits/gate.rs
```diff
@@ -422,13 +423,36 @@ impl GateType {
...
 
+        let lookup_gate_pattern = (0..3)
+            .map(|i| {
+                // each row represents an XOR operation
+                // where l XOR r = o
+                //
+                // 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
+                // - i v - - - - - - - -  -  -  -  -
+                // - - - i v - - - - - -  -  -  -  -
+                // - - - - - i v - - - -  -  -  -  -
```
mimoo, line 434: you can reuse this table if you want

```
| r0 |  r1  |  r2  |  r3  |  r4  |  r5  |  r6  |
|:--:|:----:|:----:|:----:|:----:|:----:|:----:|
|    | idx1 | val1 | idx2 | val2 | idx3 | val3 |
```

kimchi/src/circuits/gate.rs
```diff
@@ -422,13 +423,36 @@ impl GateType {
             chacha_final_where.insert((ChaChaFinal, *r));
         }
 
+        let lookup_gate_pattern = (0..3)
+            .map(|i| {
+                // each row represents an XOR operation
+                // where l XOR r = o
```
mimoo, line 429: unrelated to XOR no?

kimchi/src/circuits/gate.rs
```diff
@@ -422,13 +423,36 @@ impl GateType {
...
+                // - i v - - - - - - - -  -  -  -  -
+                // - - - i v - - - - - -  -  -  -  -
+                // - - - - - i v - - - -  -  -  -  -
+                let index = curr_row(2 * i + 1);
+                let value = curr_row(2 * i + 2);
+                let l = |loc: LocalPosition| SingleLookup {
+                    value: vec![(F::one(), loc)],
+                };
+                JointLookup {
+                    table_id: 1,
```
mimoo, line 441: a few things:

* use a hardcoded const like `XOR_TABLE_ID`
* why is table_id set to `1`? If it is supposed to represent the runtime table?

mimoo (commented):
left a few comments

kimchi/src/circuits/gate.rs
mrmr1993: The next PR in the chain (#400) does it properly, using the first position in the witness. This is really a temporary placeholder until then, since those changes are orthogonal. I chose 1 because it makes it easier to write a reasonable test without requiring the changes in #400.

mimoo (approved):

PR #400, opened 2022-03-13
Specify lookup table as part of lookup gate witness

mrmr1993:
This PR tweaks the behaviour of table ID in lookups. With this, the `Lookup` gate can use different table IDs, by setting the desired table ID specified as the first witness value.

Note: the same table ID is used for all 3 lookups in the gate. It would be nice to use more, but
* we've already used all of the permutation-constrained cells in the witness
* it's best to avoid evaluating the coefficient polynomials, if we're able to
* it's best to avoid adding a new polynomial, if we're able to (technically worse than evaluating the coefficients).


mimoo:
> it's best to avoid evaluating the coefficient polynomials, if we're able to

just to check my understanding: 

* as part of the protocol the prover evaluates the grand product argument constraint
* so the verifier needs to produce its commitment to verify that evaluation
* the commitment must be linearized to avoid multiplication of commitments
* there's f, t, and s used in this equation. s and t are evaluated as part of the proof, but f is derived using lookup selector commitments which we don't evaluate

Screen Shot 2022-04-04 at 2 48 08 PM

so if we were to use the coefficients vectors to specify the table id, we'd need to choose between evaluating those or evaluating the lookup selectors

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -179,44 +179,65 @@ impl<F: Copy> SingleLookup<F> {
...
             table_id_combiner,
             self.entry.iter(),
-            i32_to_field(self.table_id),
+            &self.table_id,
         )
     }
 }
 
-impl<F: Copy> JointLookup<SingleLookup<F>> {
+impl<F: Copy> JointLookup<SingleLookup<F>, LookupTableID> {
```
mimoo, line 222: can't do `impl<F> JointLookupSpec<F> { ... }` ?

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -179,44 +179,65 @@ impl<F: Copy> SingleLookup<F> {
...
     /// Reduce linear combinations in the lookup entries to a single value, resolving local
     /// positions using the given function.
-    pub fn reduce<K, G: Fn(LocalPosition) -> K>(&self, eval: &G) -> JointLookup<K>
+    pub fn reduce<K, G: Fn(LocalPosition) -> K>(&self, eval: &G) -> JointLookupValue<K>
```
mimoo, line 225: I was really confused as to why this was taking a closure instead of writing the code without a closure, looking at this again I'm confused again

Also why can't `K` be the `F` here?

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -179,44 +179,65 @@ impl<F: Copy> SingleLookup<F> {
...
+        K: Neg<Output = K>,
+        K: From<u64>,
     {
+        let table_id = match self.table_id {
+            LookupTableID::Constant(table_id) => i32_to_field(table_id),
+            LookupTableID::WitnessColumn(column) => eval(LocalPosition {
+                row: CurrOrNext::Curr,
+                column,
+            }),
+        };
```
mimoo, line 238: maybe this should be implemented on your `TableId` type instead?

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -425,8 +427,8 @@ pub struct LookupConfiguration<F: FftField> {
     /// This is used to pad the lookups to `max_lookups_per_row` when fewer lookups are used in a
     /// particular row, so that we can treat each row uniformly as having the same number of
     /// lookups.
-    #[serde_as(as = "JointLookup<o1_utils::serialization::SerdeAs>")]
-    pub dummy_lookup: JointLookup<F>,
+    #[serde_as(as = "JointLookupValue<o1_utils::serialization::SerdeAs>")]
+    pub dummy_lookup: JointLookupValue<F>,
```
mimoo, line 431: the name still confused me today. If it's joint, then why is it still a vec :)

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -474,14 +476,14 @@ pub fn constraints<F: FftField>(configuration: &LookupConfiguration<F>, d1: D<F>
 
         // combine the columns of the dummy lookup row
         let dummy_lookup = {
-            let expr_dummy: JointLookup<ConstantExpr<F>> = JointLookup {
+            let expr_dummy: JointLookupValue<ConstantExpr<F>> = JointLookup {
```
mimoo, line 479: it'd be nice to create a type for that as well :D `JointLookupExpr`

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -474,14 +476,14 @@ pub fn constraints<F: FftField>(configuration: &LookupConfiguration<F>, d1: D<F>
...
                 entry: configuration
                     .dummy_lookup
                     .entry
                     .iter()
                     .map(|x| ConstantExpr::Literal(*x))
                     .collect(),
-                table_id: configuration.dummy_lookup.table_id,
+                table_id: ConstantExpr::Literal(configuration.dummy_lookup.table_id),
```
mimoo, line 486: this does feel unefficient though, what if your circuit doesn't make use of the XOR table, it sounds like you'd still have to include it for the dummy queries to work

kimchi/src/circuits/constraints.rs
```diff
@@ -413,7 +413,7 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
                 // table 0.
                 let dummy_lookup = JointLookup {
                     entry: vec![],
-                    table_id: 0,
+                    table_id: F::zero(),
```
mimoo, line 416: the inconvenience of havint a `table_id: F` at this point is that we can't specify these things with an enum or a hardcoded `XOR_TABLE_ID` or something

kimchi/src/circuits/serialization_helper.rs
```diff
@@ -1,4 +1,4 @@
-use crate::circuits::lookup::lookups::JointLookup;
+use crate::circuits::lookup::lookups::{JointLookup, JointLookupValue};
```
mimoo, line 1: this file should live inside lookup/

mimoo (approved):

kimchi/src/circuits/lookup/tables.rs
```diff
@@ -123,7 +123,7 @@ pub fn combine_table_entry<'a, F, I>(
     joint_combiner: &F,
     table_id_combiner: &F,
     v: I,
-    table_id: F,
+    table_id: &F,
```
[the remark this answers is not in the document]
mrmr1993: The generic allows this to also use `expr` types, so it's better not to pass them by value.

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -179,44 +179,65 @@ impl<F: Copy> SingleLookup<F> {
     }
 }
 
+/// The table ID associated with a particular lookup
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub enum LookupTableID {
+    /// Look up the value from the given fixed table ID
+    Constant(i32),
+    /// Look up the value in the table with ID given by the value in the witness column
+    WitnessColumn(usize),
```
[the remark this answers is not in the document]
mrmr1993: This is true of all current and planned lookups, although it's easy to relax if we need it.

kimchi/src/circuits/lookup/lookups.rs
mrmr1993: `K` is a vector space over `F`, which allows us to mix expr types with field constants using the same code.

kimchi/src/circuits/lookup/constraints.rs
mrmr1993: I suppose we should standardise on e.g. 'joint' meaning that it represents multiple values, and 'joined' meaning those multiple values have been combined using the joint combiner.

kimchi/src/circuits/lookup/constraints.rs
mrmr1993: The choices are currently:
* use a table with ID 0 that has an all-zero entry (e.g. XOR)
* don't use a table with ID 0, and only the dummy is in that table
We can make it more flexible -- the dummy value is free -- but because we hard-code the table ID for the fixed tables at the moment, I thought this assumption was fine.

kimchi/src/circuits/constraints.rs
mrmr1993: As above: this uses the unwritten assumption that table 0 is either empty or has an all-zero entry

mrmr1993: Perhaps that assumption ought to be written

kimchi/src/circuits/lookup/lookups.rs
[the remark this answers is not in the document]
mimoo: yeah, I was suggesting that it might be worth commenting that there :o

kimchi/src/circuits/constraints.rs
mimoo: or better: asserted

kimchi/src/circuits/lookup/constraints.rs
mimoo: I just run into this issue because my circuit was too small to host the current table. I think we should have a Zero table that's just a single column 0, and always use that table when we use lookups. The only downside I see is that if you only use the XOR table, then it will have id = 1 and will be a bit harder to combine

PR #401, opened 2022-03-13
[specs][kimchi] adding data structures for prover & verifier index

mimoo:


mrmr1993 (approved):
Looks good, although it might be cleaner if it could strip the annotations for the spec

mimoo:
Yeah I've been thinking about that actually, I need to change the templating on cargo-spec and I think we could maybe substitute things more easily with the new templating

PR #404, opened 2022-03-14
feat: non recursive proofs

iammadab:


mimoo:
Hey @iammadab, thanks for the PR! Is this ready for review? It says it's still a draft.

iammadab:
Technically ready for review but missing tests to demonstrate the new functionality (not sure if this is necessary tho)

kimchi/src/prover.rs
```diff
@@ -85,11 +85,26 @@ impl<G: CommitmentCurve> ProverProof<G>
...
     ///     witness: computation witness
     ///     index: ProverIndex
     ///     RETURN: prover's zk-proof
     pub fn create<EFqSponge: Clone + FqSponge<Fq<G>, G, Fr<G>>, EFrSponge: FrSponge<Fr<G>>>(
+        groupmap: &G::Map,
+        witness: [Vec<Fr<G>>; COLUMNS],
+        index: &ProverIndex<G>,
+    ) -> Result<Self> {
+        Self::create_recursive::<EFqSponge, EFrSponge>(groupmap, witness, index, Vec::new())
+    }
```
mimoo, line 98: I think you'll have to change the way `create_recursive` work, can you modify a test (for example `tests/generic.rs`) to use the non-recursive create? Let's see if this passes tests :o

iammadab: changed tests/generic.rs verify_proof function to use non-recursive create, (commented out the prev argument), test still passes which is weird, it's as tho prev not being present has no effect.

mimoo: looked at the code, looks like the recursion stuff doesn't do anything if the vector is empty! Can you add the test as a commit here, LGTM otherwise :)

mimoo:
for completion, the prover side iterates over the prev_challenges to add them to the aggregated evaluation proof

```rust
        let polys = prev_challenges
            .iter()
            .map(|(chals, comm)| {
                (
                    DensePolynomial::from_coefficients_vec(b_poly_coefficients(chals)),
                    comm.unshifted.len(),
                )
            })
            .collect::<Vec<_>>();
```

and the verifier side adds the commitments and evaluations to the aggregated verification

```rust
        let polys: Vec<(PolyComm<G>, _)> = self
            .prev_challenges
            .iter()
            .zip(self.prev_chal_evals(index, &ep, &powers_of_eval_points_for_chunks))
            .map(|(c, e)| (c.1.clone(), e))
            .collect();
```

kimchi/src/prover.rs
```diff
@@ -85,11 +85,26 @@ impl<G: CommitmentCurve> ProverProof<G>
 where
     G::BaseField: PrimeField,
 {
-    /// This function constructs prover's zk-proof from the witness & the ProverIndex against SRS instance
+    /// This function constructs prover's non-recursive zk-proof from the witness & the ProverIndex against SRS instance
     ///     witness: computation witness
     ///     index: ProverIndex
     ///     RETURN: prover's zk-proof
```
mimoo, line 91: I'd be nice to have a nicer doc comment (that matches how the rust stdlib is documented for example, instead of listing the arguments like that). Also, I suggest not mentioning non-recursion here, and only mentioning recursion in the doc of the other function

iammadab:
Noticed that all tests except generic and poseidon, do not make use of prev_challanges, refactored their proof creation to make use of the non-recursive variant.

mimoo:
btw, the README could use some refactoring to show the non-recursive API (or both API) if you want to update this in another PR :o

Issue #202, opened 2021-10-28
[kimchi] move gate code into polynomials


mimoo:
We currently have two folders containing related logic for each gates we have:

* a gate folder
* a polynomial folder

I believe there's been a trend to move the gate code into the polynomial code, just so that it lives in the same file. We should finish that trend and get rid of the gate folder.

frisitano:
I will take this

PR #412, opened 2022-03-15
Purge gate folder

frisitano:
This PR migrates code from `kimchi/src/circuits/gates/` to the corresponding file in `kimichi/src/circuits/polynomials/`.  The empty `kimchi/src/circuits/gates/` folder has then been purged. 

closes: #202


frisitano:
Please do not merge just yet.  After reading https://www.cryptologie.net/article/553/the-code-is-the-specification-introducing-cargo-spec/ I want to investigate if any of the changes I have made have broken the spec and if so I will fix.

frisitano:
just checked and it doesn't look like this will impact the spec.

mimoo:
CI would actually have yelled at you if you changed something in the spec :) so don't worry. Thanks a lot for this PR!

mimoo:
sorry, one more thing. In other polynomials we've always had the gate construction code first, and then the constraints after, but in your PR you added the gate stuff at the end of the file instead of at the beginning. Would you mind changing that and moving things to the beginning of the file instead? Just so that we're consistent.

PR #419, opened 2022-03-15
[CI] fix ocaml setup

mimoo:
hopefully this will fix issues we see in CI


mimoo:
this greatly increases the time CI takes, and it seems like we don't have the ocaml error anymore

PR #429, opened 2022-03-16
Turshi 15 wires

querolita:
This code aims at providing gate constraints for Cairo programs in the Kimchi 15-wire format. It contains a witness creation function that builds a full Cairo gadget (a set of Cairo gates) from a set of steps of the Cairo runner. It is equipped with the following gates:

- `CairoInstruction`: takes care of the instruction logics. It accesses 2 rows (`Next` is a `Zero` gate).
- `CairoTransition`: takes care of the correct transition between steps of computation. It accesses 2 rows (`Next` is a `CairoTransition` gate or `Zero` for the last one). 
- `CairoClaim`: takes care of the claimed initial and final registers of the computation. It only accesses the `Curr` row. 

A Cairo gadget for an N-instruction `CairoProgram` will look like this:
N  (`CairoInstruction` | `Zero`)
N-1 ( `CairoTransition`)
1 ( `Zero`)
1  (`CairoClaim`)

The copy constraints are not still imposed, nor the memory checks. I am adding these to a later version of turshi that uses more than 15 wires. I realised that I need to perform too much wiring otherwise, and if I switch to a larger witness length I can reduce the number of Cairo gates significantly. Plus, I found that this 15-wire version can be of independent interest, so I thought it may be worth having a PR for itself.

Nonetheless, this PR modifies the number of powers of alpha to be 28 instead of 21. This number will be larger when I merge the gates into a single one, so I am not sure we want to do this "now".


mimoo:
is this ready for review :o ? also would you mind meeting up to talk through this one?

querolita:
This version is outdated, I will create a new one that puts everything together

PR #405, opened 2022-03-14
Split out non-table-specific parts of `LookupConstraintSystem`

mrmr1993:
This PR builds upon #400, splitting out a `LookupConfig` structure from the rest of the `LookupConstraintSystem`. `LookupConfig` doesn't depend on the contents of the lookup tables, and so can be shared across multiple different constraint systems.

In particular, this is useful for generating a linearisation in the OCaml bindings, since we can specify the 'shape' of lookup tables in the verifier, but we can't be confident of the table's contents.

PR #430, opened 2022-03-16
Split out non-table-specific parts of `LookupConstraintSystem` (redux)

mrmr1993:
This is PR #405 cherry-picked onto master. To quote from there:

> This PR builds upon https://github.com/o1-labs/proof-systems/pull/400, splitting out a LookupConfig structure from the rest of the LookupConstraintSystem. LookupConfig doesn't depend on the contents of the lookup tables, and so can be shared across multiple different constraint systems.

> In particular, this is useful for generating a linearisation in the OCaml bindings, since we can specify the 'shape' of lookup tables in the verifier, but we can't be confident of the table's contents.

### DIFF
--- a/kimchi/src/circuits/polynomials/lookup.rs
+++ b/kimchi/src/circuits/polynomials/lookup.rs
@@ -236,6 +236,10 @@ pub struct LookupConfiguration<F: FftField> {
     pub max_joint_size: u32,
 
     #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    /// A placeholder value that is known to appear in the lookup table.
+    /// This is used to pad the lookups to `max_lookups_per_row` when fewer lookups are used in a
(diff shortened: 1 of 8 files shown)


kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -217,6 +222,23 @@ pub fn zk_patch<R: Rng + ?Sized, F: FftField>(
...
+    /// The kind of lookups used
+    pub lookup_used: LookupsUsed,
+
+    /// The maximum number of lookups per row
+    pub max_lookups_per_row: usize,
+    /// The maximum number of elements in a vector lookup
+    pub max_joint_size: u32,
+
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub dummy_lookup_value: Vec<F>,
```
mimoo, line 239: doc comment!

mimoo (approved):
LGTM!

PR #433, opened 2022-03-18
Use a single implementation for `combine_table_entry`

mrmr1993:
This PR generalises `JointLookup` so that it can handle `SingleLookup<F>`s, `F`s, or `Expr<ConstantExpr<F>>`s as arguments, and relaxes the constraints on `combine_table_entry`.

This allows us to
* store the 'dummy' lookup value as a `JointLookup` in the constraint system, rather than unpacking and re-packing it
  - not done here; this will be in a follow-up PR
* unify the 3 different implementations of `combine_table_entry` into one
  - happily, the 'fold' implementation in `combine_table_entry` minimises the number of multiplications compared to the others, so this is also an efficiency gain
* have a single point to change when we add `table_id`s

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -1925,7 +1925,7 @@ impl<F: Field> Mul<F> for Expr<ConstantExpr<F>> {
     type Output = Expr<ConstantExpr<F>>;
 
     fn mul(self, y: F) -> Self::Output {
-        Expr::Constant(ConstantExpr::Literal(F::from(y))) * self
+        Expr::Constant(ConstantExpr::Literal(y)) * self
     }
 }
 
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -1921,6 +1921,14 @@ impl<F: Field> From<u64> for Expr<ConstantExpr<F>> {
     }
 }
 
+impl<F: Field> Mul<F> for Expr<ConstantExpr<F>> {
+    type Output = Expr<ConstantExpr<F>>;
+
+    fn mul(self, y: F) -> Self::Output {
+        Expr::Constant(ConstantExpr::Literal(F::from(y))) * self
+    }
+}
+
 //
 // Display
 //
--- a/kimchi/src/circuits/gate.rs
+++ b/kimchi/src/circuits/gate.rs
@@ -3,6 +3,7 @@
 use crate::circuits::{constraints::ConstraintSystem, domains::EvaluationDomains, wires::*};
 use ark_ff::bytes::ToBytes;
 use ark_ff::{FftField, Field};
+use ark_ff::{One, Zero};
 use ark_poly::{Evaluations as E, Radix2EvaluationDomain as D};
 use num_traits::cast::ToPrimitive;
 use o1_utils::hasher::CryptoDigest;
@@ -10,6 +11,7 @@ use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use std::collections::{hash_map::Entry, HashMap, HashSet};
 use std::io::{Result as IoResult, Write};
+use std::ops::{AddAssign, Mul, MulAssign};
 
 type Evaluations<Field> = E<Field, D<Field>>;
 
@@ -75,31 +77,45 @@ pub fn combine_table_entry<'a, F: Field, I: DoubleEndedIterator<Item = &'a F>>(
     v.rev().fold(F::zero(), |acc, x| joint_combiner * acc + x)
 }
 
-impl<F: Field> SingleLookup<F> {
+impl<F: Copy> SingleLookup<F> {
     /// Evaluate the linear combination specifying the lookup value to a field element.
-    pub fn evaluate<G: Fn(LocalPosition) -> F>(&self, eval: G) -> F {
+    pub fn evaluate<K: Zero + Mul<F, Output = K>, G: Fn(LocalPosition) -> K>(&self, eval: G) -> K {
         self.value
             .iter()
-            .fold(F::zero(), |acc, (c, p)| acc + *c * eval(*p))
+            .fold(K::zero(), |acc, (c, p)| acc + eval(*p) * *c)
     }
 }
 
 /// A spec for checking that the given vector belongs to a vector-valued lookup table.
 #[derive(Clone, Serialize, Deserialize)]
-pub struct JointLookup<F> {
-    pub table_id: usize,
-    pub entry: Vec<SingleLookup<F>>,
+pub struct JointLookup<SingleLookup> {
+    pub table_id: i32,
+    pub entry: Vec<SingleLookup>,
 }
 
-impl<F: Field> JointLookup<F> {
+pub type JointLookupSpec<F> = JointLookup<SingleLookup<F>>;
+
+impl<F: Copy> JointLookup<SingleLookup<F>> {
+    pub fn reduce<K: Zero + Mul<F, Output = K>, G: Fn(LocalPosition) -> K>(
+        &self,
+        eval: &G,
+    ) -> JointLookup<K> {
+        JointLookup {
+            table_id: self.table_id,
+            entry: self.entry.iter().map(|s| s.evaluate(eval)).collect(),
+        }
+    }
+}
+
+impl<F: Zero + One + AddAssign + MulAssign + Clone> JointLookup<F> {
     // TODO: Support multiple tables
     /// Evaluate the combined value of a joint-lookup.
-    pub fn evaluate<G: Fn(LocalPosition) -> F>(&self, joint_combiner: F, eval: &G) -> F {
+    pub fn evaluate(&self, joint_combiner: F) -> F {
         let mut res = F::zero();
         let mut c = F::one();
         for s in self.entry.iter() {
-            res += c * s.evaluate(eval);
-            c *= joint_combiner;
+            res += c.clone() * s.clone();
+            c *= joint_combiner.clone();
         }
         res
     }
@@ -158,7 +174,7 @@ pub enum GateType {
 pub struct LookupInfo<F> {
     /// A single lookup constraint is a vector of lookup constraints to be applied at a row.
     /// This is a vector of all the kinds of lookup constraints in this configuration.
-    pub kinds: Vec<Vec<JointLookup<F>>>,
+    pub kinds: Vec<Vec<JointLookupSpec<F>>>,
     /// A map from the kind of gate (and whether it is the current row or next row) to the lookup
     /// constraint (given as an index into `kinds`) that should be applied there, if any.
     pub kinds_map: HashMap<(GateType, CurrOrNext), usize>,
@@ -170,10 +186,10 @@ pub struct LookupInfo<F> {
     /// The maximum joint size of any joint lookup in a constraint in `kinds`. This can be computed from `kinds`.
     pub max_joint_size: u32,
     /// An empty vector.
-    empty: Vec<JointLookup<F>>,
+    empty: Vec<JointLookupSpec<F>>,
 }
 
-fn max_lookups_per_row<F>(kinds: &[Vec<JointLookup<F>>]) -> usize {
+fn max_lookups_per_row<F>(kinds: &[Vec<JointLookupSpec<F>>]) -> usize {
     kinds.iter().fold(0, |acc, x| std::cmp::max(x.len(), acc))
 }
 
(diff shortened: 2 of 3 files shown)


kimchi/src/circuits/expr.rs
```diff
@@ -1921,6 +1921,14 @@ impl<F: Field> From<u64> for Expr<ConstantExpr<F>> {
     }
 }
 
+impl<F: Field> Mul<F> for Expr<ConstantExpr<F>> {
+    type Output = Expr<ConstantExpr<F>>;
+
+    fn mul(self, y: F) -> Self::Output {
+        Expr::Constant(ConstantExpr::Literal(F::from(y))) * self
```
mimoo, line 1928: you might want to either remove the `F::from` or have `F: Into<Field>`

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
+    'a,
+    F: 'a + Zero + One + Add + Mul + Clone,
+    I: DoubleEndedIterator<Item = &'a F>,
+>(
     joint_combiner: F,
     v: I,
 ) -> F {
-    v.rev().fold(F::zero(), |acc, x| joint_combiner * acc + x)
+    v.rev()
+        .fold(F::zero(), |acc, x| joint_combiner.clone() * acc + x.clone())
```
mimoo, line 81: if you add Copy instead of Clone you can remove the clone here (Field is Copy)

mrmr1993: `F` is copy but `Expr` isn't (and we use `JointCombiner` rather than a field element there).

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
 /// analogously using `joint_combiner`.
 ///
 /// This function computes that combined value.
-pub fn combine_table_entry<'a, F: Field, I: DoubleEndedIterator<Item = &'a F>>(
+pub fn combine_table_entry<
+    'a,
+    F: 'a + Zero + One + Add + Mul + Clone,
```
mimoo, line 74: I'm wondering why `F: Field + Zero` here is not enough?

Also I'm wondering how I'm supposed to read `F: 'a`? And why it doesn't work without that

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
-    v.rev().fold(F::zero(), |acc, x| joint_combiner * acc + x)
+    v.rev()
+        .fold(F::zero(), |acc, x| joint_combiner.clone() * acc + x.clone())
 }
 
-impl<F: Field> SingleLookup<F> {
+impl<F: Copy> SingleLookup<F> {
     /// Evaluate the linear combination specifying the lookup value to a field element.
-    pub fn evaluate<G: Fn(LocalPosition) -> F>(&self, eval: G) -> F {
+    pub fn evaluate<K: Zero + Add + Mul<F, Output = K>, G: Fn(LocalPosition) -> K>(
```
mimoo, line 86: in general it's nicer to move these bounds to a `where` clause at the end of the function signature

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
     joint_combiner: F,
     v: I,
 ) -> F {
-    v.rev().fold(F::zero(), |acc, x| joint_combiner * acc + x)
+    v.rev()
+        .fold(F::zero(), |acc, x| joint_combiner.clone() * acc + x.clone())
 }
 
-impl<F: Field> SingleLookup<F> {
+impl<F: Copy> SingleLookup<F> {
```
mimoo, line 84: is there a reason behind that relaxation and the creation of K? Why not just take everything as a `Field`?

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
 }
 
 /// A spec for checking that the given vector belongs to a vector-valued lookup table.
 #[derive(Clone, Serialize, Deserialize)]
-pub struct JointLookup<F> {
-    pub table_id: usize,
-    pub entry: Vec<SingleLookup<F>>,
+pub struct JointLookup<SingleLookup> {
+    pub table_id: i32,
+    pub entry: Vec<SingleLookup>,
```
mimoo, line 100: why not just make this `Vec<SingleLookup<F>>` and the struct generic in `F`?

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
-        for s in self.entry.iter() {
-            res += c * s.evaluate(eval);
-            c *= joint_combiner;
+    pub fn evaluate(&self, joint_combiner: F) -> F {
+        combine_table_entry(joint_combiner, self.entry.iter())
+    }
+}
+
+impl<F: Copy> JointLookup<SingleLookup<F>> {
+    pub fn reduce<K: Zero + Mul<F, Output = K>, G: Fn(LocalPosition) -> K>(
```
mimoo, line 114: same here

kimchi/src/circuits/gate.rs
mrmr1993: Because `reduce` erases `SingleLookup` and returns a concrete value/`Expr` for the lookup

kimchi/src/circuits/gate.rs
mrmr1993: `K` can be `Expr`; that was the point of this PR.

kimchi/src/circuits/gate.rs
mimoo: ah I see. My tolerance for generics is usually pretty low ^^ I'm wondering how we can make this clearer. Maybe by changing `F` with `F_or_Expr` or something. Food for thought

mimoo: and then creating an actual trait `F_or_Expr : Zero + One + ...` and implement it on what we need

kimchi/src/circuits/gate.rs
mimoo: *facepalm* I did get straight into the code without reading the description

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
+        &self,
+        eval: &G,
+    ) -> JointLookup<K> {
+        JointLookup {
```
mimoo, line 118: can you Self in place of JointLookup here. I'm not sure it's objectively better actually... but I've been conditioned to do this

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
+            table_id: self.table_id,
+            entry: self.entry.iter().map(|s| s.evaluate(eval)).collect(),
```
mimoo, line 120: I think this should work too:

```rust
let entry = self.entry.iter().map(|s| s.evaluate(eval)).collect();
Self { entry, ..self }
```

kimchi/src/circuits/gate.rs
mimoo: yeah I see that now. So it's like we have multiple types. Do you think it'd be a good idea to define them somewhere and use them? Something like that:

```rust
type EvaluatedJointLookup<F> = JointLookup<F>;
type ExprSingleLookup<F> = SingleLookup<ConstantExpr<F>>;
type ExprJointLookup<F> = JointLookup<SingleLookup<ConstantExpr<F>>>;
```

might be a stupid idea

mrmr1993: We never use any of those types directly, not sure it's worth it.

kimchi/src/circuits/gate.rs
mrmr1993: `F` may be a type that is defined within a function; if the scope of `'a` is broader than the function, then the data outlives the type. Thus `F: 'a`.

kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -661,16 +635,26 @@ pub fn constraints<F: FftField>(configuration: &LookupConfiguration<F>, d1: D<F>
             .collect()
     };
 
+    let eval = |pos: LocalPosition| {
+        E::Cell(Variable {
+            col: Column::Witness(pos.column),
+            row: pos.row,
+        })
+    };
```
mimoo, line 643: we already have these helpers in expr.rs

kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -661,16 +635,26 @@ pub fn constraints<F: FftField>(configuration: &LookupConfiguration<F>, d1: D<F>
...
-    let f_term = |spec: &Vec<_>| {
+    let f_term = |spec: &Vec<JointLookupSpec<_>>| {
         assert!(spec.len() <= lookup_info.max_per_row);
         let padding = complements_with_beta_term[lookup_info.max_per_row - spec.len()].clone();
 
         spec.iter()
-            .map(|j| E::Constant(ConstantExpr::Gamma) + joint_lookup(j))
+            .map(|j| {
+                E::Constant(ConstantExpr::Gamma)
+                    + j.evaluate(E::constant(ConstantExpr::JointCombiner), &eval)
```
mimoo, line 656: I'm confused as to how you can pass a reference to a closure to evaluate, which doesn't seem to expect a closure there

mimoo: I think rust-analyzer is getting confused as to which `evaluate()` function is called o_O

mimoo: https://github.com/rust-analyzer/rust-analyzer/issues/11759

kimchi/src/circuits/gate.rs
```diff
@@ -68,40 +69,67 @@ pub struct SingleLookup<F> {
...
+
+    pub fn evaluate<
+        K: Zero + One + Add + Mul + Mul<F, Output = K> + Clone,
+        G: Fn(LocalPosition) -> K,
+    >(
+        &self,
+        joint_combiner: K,
+        eval: &G,
+    ) -> K {
+        self.reduce(eval).evaluate(joint_combiner)
```
mimoo, line 132: honestly I find this function really hard to parse.

* we have a `JointLookup<SingleLookup<F>>` for some `F: Zero + One + Add + Mul + Clone`
* we want a `K` for some `K: Zero + One + Add + Mul + Mul<F, Output = K> + Clone`
* in the mean time, we have some eval function that takes a `LocalPosition` and return a `K`

it's just too generic to really grasp what is really going on IMO. Is there a way to make this clearer?

mimoo: interestingly my IDE is also working against myself, pointing me to the wrong reduce function

mimoo: I'm wondering if we can just have code duplication for the different instantiations of these functions, this would make the code much clearer imo and we wouldn't need to pass a closure around (IIUC)

kimchi/src/circuits/gate.rs
mrmr1993: Types mismatch, `self: JointLookup<SingleLookup<F>>`, return value is `JointLookup<F>`.

kimchi/src/circuits/polynomials/lookup.rs
mrmr1993: Commented there; presumably it's due to name reuse.

kimchi/src/circuits/gate.rs
mrmr1993: I don't think I can, because the return type is different from `Self`..

kimchi/src/circuits/polynomials/lookup.rs
```diff
@@ -142,29 +139,6 @@ use CurrOrNext::*;
...
-                })
-        })
-        .fold(E::zero(), |acc, e| acc + e)
-}
-
-fn joint_lookup<F: FftField>(j: &JointLookup<F>) -> E<F> {
-    j.entry
-        .iter()
-        .enumerate()
-        .map(|(i, s)| E::constant(ConstantExpr::JointCombiner.pow(i as u64)) * single_lookup(s))
```
mimoo, line 164: so if I understand correctly this is replaced by the logic in `combine_table_entry`

mimoo (approved):
Approved, but I find the generic implementation really hard to follow, I'm even wondering how I'll ever be able to change this code without keeping 10 different low-level details in my head.

kimchi/src/circuits/polynomials/lookup.rs
mrmr1993: Exactly, yes

Issue #386, opened 2022-03-12
move ceil_log2 to utils


mimoo:
move [ceil_log2](https://github.com/o1-labs/marlin/blob/4ea3d287e777b5c9d6dc49a82083631412ccbd63/poly-commitment/src/commitment.rs#L323) out of poly-commitment, and into the utils crate, as it's a more general function

prateekm33:
I can pick this one up @mimoo

prateekm33:
is there a specific file that we want to move it to in the `utils` crate or create a new file under that dir?

mrmr1993:
None of the existing files seem like a good fit, but either in the toplevel `lib.rs` or a new file (e.g. `math.rs`) is sound good to me.

PR #436, opened 2022-03-18
move ceil_log2 into o1_utils::math package

Trivo25:
moves `ceil_log2` into the package `o1_utils::math` under `utils/src/math.rs`

see https://github.com/o1-labs/proof-systems/issues/386


mimoo (approved):
LGTM, thanks for the PR!

PR #439, opened 2022-03-18
move ceil_log2 tests into math module

Trivo25:
This is a follow-up to https://github.com/o1-labs/proof-systems/pull/436 which moves the corresponding test case of `ceil_log2` into the newly created `math` module in order to keep the code and structure clean (forgot it with last PR)


mrmr1993 (approved):
Great, thanks!

Issue #394, opened 2022-03-13
Factor out parts of `ConstraintSystem` that don't depend on the constraint system


mrmr1993:
There are some computations that we store in `ConstraintSystem` that depend only on the domain. By moving these to their own sub-structure, and by creating them in a function that depends on the domain alone, we can avoid recomputing them for each constraint system, and share them between different circuits over the same domain.

Fields that don't depend on the constraint system:
* [`zkpm`](https://github.com/o1-labs/proof-systems/blob/cc5ef9e0a4108e9ca623588d2187219cac090a01/kimchi/src/circuits/constraints.rs#L80)
* polynomials [in the section labelled `Constant polynomials`](https://github.com/o1-labs/proof-systems/blob/cc5ef9e0a4108e9ca623588d2187219cac090a01/kimchi/src/circuits/constraints.rs#L141-L164)

These probably make sense to move to their own structure with a dedicated `create` function, and could even be moved to their own module.

mimoo:
Note that someone is looking into this TODO: https://github.com/o1-labs/proof-systems/blob/cc5ef9e0a4108e9ca623588d2187219cac090a01/kimchi/src/circuits/constraints.rs#L563

which might mean that some of the unused constants will disappear.

mimoo:
btw, somewhat related: https://github.com/o1-labs/proof-systems/issues/375

chee-chyuan:
happy to look into this

mimoo:
Btw I'm recommending a different route for now: keep the fields in the struct, but lazily evaluate them instead of deserializing them. IMO it's a good first step. 

We should create another issue to make parameters based on domain + field sharable. Perhaps we could create a struct wrapping Field with more parameters that we can share. I think this is something that needs be thought of while designing the trimmedSRS

mimoo:
Thinking more about this, it sounds like using an enum for tracking either a reference or a value (which is what we used to do for the SRS) is a good path. `Cow` seems to provide exactly this with a bunch of handy traits (although we don't need the mutability part):

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0187f7d388e31109228ae2e0a7edf22d

mimoo:
btw, cf https://www.reddit.com/r/rust/comments/tjlh18/a_struct_field_thats_either_a_reference_or_value/ a number of people recommend to use Arc or Rc so that you don't have to deal with lifetimes. I think that might be a better approach.

PR #441, opened 2022-03-19
Refactor constraint system (zk polynomial and constant polynomial

chee-chyuan:
issue https://github.com/o1-labs/proof-systems/issues/394

Prelim work in an attempt to to refactor out zk polynomial and constant polynomial from the `ConstraintSystem` struct.
Setting it as draft as I am not sure if im doing the right thing. happy to scrap if redo if i have done something wrongly.

Also how do we write tests for these?


mimoo:
Hey! Thanks for taking a stab at it. I suggest that `ConstantPolynomial` should:

* have the `zkpm` field in it
* be renamed to something more descriptive. For example, `ProverPrecomputations`.
* be stored as a field `precomputations` within the `ConstraintSystem`.
* perhaps also the module should be renamed `prover_precomputations`

The way you implemented this, the pre-computations are not really pre-computations as they will be re-computed every time someone creates a proof. Instead, you could use the once_cell crate to declare a global HashMap that stores these pre-computation indexed per-domain. If the `ProverPrecomputations` doesn't exist for a specific domain, create it and store it in the hashmap the first time (so that subsequent use just fetch the data from there).

Does that make sense?

kimchi/src/circuits/constant_polynomial.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub l04: E<F, D<F>>,
+    /// 0-th Lagrange evaluated over domain.d8
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub l08: E<F, D<F>>,
+    /// zero evaluated over domain.d8
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub zero4: E<F, D<F>>,
+    /// zero evaluated over domain.d8
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub zero8: E<F, D<F>>,
```
mimoo, line 27: I think zero4 and zero8 can be deleted, I don't see them used (at least in this PR, not sure if tests pass)

kimchi/src/circuits/constant_polynomial.rs
```diff
@@ -0,0 +1,71 @@
...
+
+impl<F: FftField> ConstantPolynomial<F> {
+    pub fn create(domain: EvaluationDomains<F>) -> Option<Self> {
+        let l1 = DP::from_coefficients_slice(&[F::zero(), F::one()])
+            .evaluate_over_domain_by_ref(domain.d8);
+        // TODO: These are all unnecessary. Remove
+        let l04 =
+            E::<F, D<F>>::from_vec_and_domain(vec![F::one(); domain.d4.size as usize], domain.d4);
+        let l08 =
+            E::<F, D<F>>::from_vec_and_domain(vec![F::one(); domain.d8.size as usize], domain.d8);
```
mimoo, line 53: I really dislike this naming scheme btw :D 

I think it'd be a good idea to rename these fields:

* `l1` -> `constant_x_d1`
*  `l04` -> `constant_one_d4`
* `l08` -> `constant_one_d8`

couldn't find better names, but at least I can remember what there are after a few times seeing these names

kimchi/src/circuits/constraints.rs
```diff
@@ -291,25 +247,6 @@ pub fn vanishes_on_last_4_rows<F: FftField>(domain: D<F>) -> DP<F> {
     &(&(&x - &c(w1)) * &(&x - &c(w2))) * &(&(&x - &c(w3)) * &(&x - &c(w4)))
 }
```
mimoo, line 248: these are functions that are only related to the permutation, so I think they should be moved to permutation.rs

kimchi/src/circuits/constant_polynomial.rs
chee-chyuan: i think you are right. planning to look into that next

kimchi/src/circuits/constraints.rs
chee-chyuan: just so im following we are removing `vanishes_on_last_4_rows` to permutation.rs ya?

chee-chyuan:
> Hey! Thanks for taking a stab at it. I suggest that `ConstantPolynomial` should:
> 
> * have the `zkpm` field in it
> * be renamed to something more descriptive. For example, `ProverPrecomputations`.
> * be stored as a field `precomputations` within the `ConstraintSystem`.
> * perhaps also the module should be renamed `prover_precomputations`
> 
> The way you implemented this, the pre-computations are not really pre-computations as they will be re-computed every time someone creates a proof. Instead, you could use the once_cell crate to declare a global HashMap that stores these pre-computation indexed per-domain. If the `ProverPrecomputations` doesn't exist for a specific domain, create it and store it in the hashmap the first time (so that subsequent use just fetch the data from there).
> 
> Does that make sense?

let me wrap by head around this. i have to confess that im a rust newb

kimchi/src/circuits/constraints.rs
mimoo: and `eval_vanishes_on_last_4_rows` as well. Although it seems to be outside the scope of the first issue, I was just commenting as I noticed these :)

chee-chyuan: no worries. just wanted to be sure not to break things :p

chee-chyuan:
hey @mimoo  just updated this. lmk

mrmr1993:
> The way you implemented this, the pre-computations are not really pre-computations as they will be re-computed every time someone creates a proof. Instead, you could use the once_cell crate to declare a global HashMap that stores these pre-computation indexed per-domain. If the `ProverPrecomputations` doesn't exist for a specific domain, create it and store it in the hashmap the first time (so that subsequent use just fetch the data from there).

@mimoo I'm not a fan of this `HashMap` approach, since it effectively represents a memory leak. I would prefer if we take the `ProverPrecomputations` as an argument (could be optional, where we generate it if it's `None`), and caching is the responsibility of the library user.

kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -207,7 +207,8 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
 
             // alpha
             let alpha_pow = {
-                let mut res = self.l04.clone();
+                let precomputation = self.get_precomputation();
+                let mut res = precomputation.constant_1_d4.clone();
```
mimoo, line 211: it'd be nice not to have to call `get_precomputation` every time. I think it would be nicer if we could do this:

```rust
let mut res = self.precomputations().constant_1_d4()
```

kimchi/src/circuits/zk_polynomial.rs
```diff
@@ -0,0 +1,59 @@
+use crate::circuits::constraints::ZK_ROWS;
```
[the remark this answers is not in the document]
mimoo: and move the logic here to `permutation.rs` as well

mimoo:
> @mimoo I'm not a fan of this HashMap approach, since it effectively represents a memory leak. I would prefer if we take the ProverPrecomputations as an argument (could be optional, where we generate it if it's None), and caching is the responsibility of the library user.

@mrmr1993 yeah we ended up talking about this privately so pasting the formatted convo here for context:

I think there are two paths. Either use `once_cell` inside the `ConstraintSystem` struct, and initialize it with the create function when the field is accessed through a function.
In the future we can think of creating configuration structs that can be shared across many constraint systems.
`#[serde(skip)]` is totally what you should do! As in: we don’t want to serialize or deserialize this field, we just want to compute it and cache it when we need it. 
And yeah you will not share this between constraint systems with the same domain / parameters, but I don’t think we have a clean way of doing this atm anyway. In the future we can brainstorm about a `FieldParameter` structure that’s a wrapper around `Fp` or `Fq + data` like this that we can cache

mimoo:
@chee-chyuan 

This is great and going in the right direction! I left you a few more comments

mrmr1993:
> I'm not sure what would be the cleaner path to be honest, but it'd be nice to be able to have a more transparent approach for the user (versus something that they have to manually pass).

@mimoo generally application developers should be the ones managing memory, not libraries. I think what we're learning here was that the old set-up -- generating the precomputations when creating the constraint system -- was the best ergonomics already. We could offer a more-complex function that allows them to be passed in, which we can use in Mina, and advanced users can discover if they have issues with memory/disk space usage.

> * be renamed to something more descriptive. For example, `ProverPrecomputations`.

@chee-chyuan I know David explicitly suggested this name, but everything in the prover index and constraint system is a 'prover precomputation'. It would be better to choose a name like `DomainConstantEvaluations` that actually reflects the contents.

mimoo:
note that we talked privately again about this, I believe @chee-chyuan is investigating using Cow/Arc/Borrow to provide this.

chee-chyuan:
Note: am closing this pr favor of a new branch so that i can avoid resolving the conflicts

chee-chyuan:
refer to https://github.com/o1-labs/proof-systems/pull/517 instead

Issue #426, opened 2022-03-16
Fr<G> and Fq<G>


mimoo:
We use type aliases `Fr<G>` and `Fq<G>` in many places in the codebase. Every time we do this, we're re-declaring these.

A few things about that:

- [ ] `Fr` and `Fq` are not good names. Especially as Fq represents the base field of Vesta. Instead they should be named something like `ScalarField` and `BaseField`.
- [ ] We should avoid re-declaring these throughout the codebase. It would be good to declare these useful aliases within the utils crate, somewhere, as a prologue perhaps

if you don't know the prologue pattern, it's usually about defining something like this:

```rust
pub mod prologue {
  pub use super::{ScalarField, BaseField};
}
```

and then using it as 

```rust
use some_module::prologue::*;
```

- [ ] In some cases, it would be nice to be explicit about the use of Fr and Fq. For example, in `prover.rs`:

```diff
- let alpha_chal = ScalarChallenge(fq_sponge.challenge());
+ let alpha_chal: BaseField<F> = ScalarChallenge(fq_sponge.challenge());

//~ 18. Derive $\alpha$ from $\alpha'$ using the endomorphism (TODO: details)
- let alpha = alpha_chal.to_field(&index.srs.endo_r);
+ let alpha: ScalarField<F> = alpha_chal.to_field(&index.srs.endo_r);
```

frisitano:
I will take this.

frisitano:
What is the advantage of using the prologue pattern for this?  I'm guessing it's to do with future extensibility.  You can easily add to the prologue and the new types will be immediately available every the prologue is used?

PR #443, opened 2022-03-19
Field aliases

frisitano:
This PR renames and migrates commonly used field types.  The types have been renamed from `Fr` and `Fq` to `ScalarField` and `BaseField`.  They have been migrated to `o1_utils::types::fields::*` where they can be used across crates.  I have added explicit type declaration for:
```
//~ 18. Derive $\alpha$ from $\alpha'$ using the endomorphism (TODO: details)
- let alpha = alpha_chal.to_field(&index.srs.endo_r);
+ let alpha: ScalarField<F> = alpha_chal.to_field(&index.srs.endo_r);
```
but received a compilation error when applying the following change so it has not been included:
```
- let alpha_chal = ScalarChallenge(fq_sponge.challenge());
+ let alpha_chal: BaseField<F> = ScalarChallenge(fq_sponge.challenge());
```

It should be noted that this PR also includes changes to the kimchi book (`kimchi.md`)

closes: #426

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -39,15 +39,9 @@ use commitment_dlog::{
     evaluation_proof::OpeningProof,
 };
 use itertools::Itertools;
-use o1_utils::ExtendedDensePolynomial;
+use o1_utils::{types::fields::*, ExtendedDensePolynomial};
 use oracle::{sponge::ScalarChallenge, FqSponge};
 use std::collections::HashMap;
-
-/// Alias to refer to the scalar field of a curve.
-type Fr<G> = <G as AffineCurve>::ScalarField;
-
-/// Alias to refer to the base field of a curve.
-type Fq<G> = <G as AffineCurve>::BaseField;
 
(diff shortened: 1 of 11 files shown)


kimchi/src/prover.rs
```diff
@@ -398,7 +395,7 @@ where
         let alpha_chal = ScalarChallenge(fq_sponge.challenge());
```
mimoo, line 395: it'd be nice to give the same treatment to the Fq-Sponge, Fr-Sponge, fq_sponge, fr_sponge at some point :)

mimoo (approved):
reviewed everything, looks like an awesome change to me! Makes the codebase much clearer! Thanks a lot for that

mimoo:
note that you cannot do

```diff
- let alpha_chal = ScalarChallenge(fq_sponge.challenge());
+ let alpha_chal: BaseField<F> = ScalarChallenge(fq_sponge.challenge());
```

because `alpha_chal` here is of type `ScalarChallenge<BaseField<G>>`

PR #448, opened 2022-03-22
[commitment] re-write combined_inner_product

mimoo:
work in progress

- [x] rewrite the combined_inner_product in a clearer and safer way
- [x] ensure that it matches the previous function behavior
- [ ] replace the previous function with this one

### DIFF
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -21,7 +21,8 @@ use ark_poly::{
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use core::ops::{Add, Sub};
 use groupmap::{BWParameters, GroupMap};
-use o1_utils::math;
+use itertools::{izip, multizip};
+use o1_utils::math::{self, powers_of};
 use o1_utils::types::fields::*;
 use o1_utils::ExtendedDensePolynomial as _;
 use oracle::{sponge::ScalarChallenge, FqSponge};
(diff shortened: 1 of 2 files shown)


poly-commitment/src/commitment.rs
```diff
@@ -424,6 +425,58 @@ pub fn combined_inner_product<G: CommitmentCurve>(
...
+/// for combining the different evaluations as well as the different polynomials.
+/// In addition, the function expects a list of evaluations for each polynomial
+/// of the form `([f(x1), f(x2)], None)`
+/// where `None` means that the polynomial's degree is not upperbounded.
+/// You must provide the SRS's length for degree upperbounds to work.
+///
+/// **Warning**: the function will crash if the number of evaluations of
+/// each polynomial doesn't match the given evaluation points or
+/// if the degree upperbound is greater than the SRS length.
+#[allow(clippy::type_complexity)]
```
mimoo, line 442: I need to remove this clippy

mimoo: Perhaps i should make a video explaining this change

mimoo:
@rot256 is proposing something even simpler: https://gist.github.com/rot256/d0b86b406e45b8dda2baccc2154ebf4d

utils/src/math.rs
```diff
@@ -17,6 +19,18 @@ pub fn ceil_log2(d: usize) -> usize {
     ceil_log2
 }
 
+/// Returns `num` powers of `x`, starting with and including 1.
+/// For example, `(x, 4)` gives you `[1, 2, 4, 8]`
+pub fn powers_of<F: Field>(x: F, num: usize) -> Vec<F> {
+    let mut res = Vec::with_capacity(num);
+    let mut last = F::one();
```
mrmr1993, line 26: Nit: `curr`? It's the current value rather than the 'last' one!

poly-commitment/src/commitment.rs
```diff
@@ -424,6 +425,58 @@ pub fn combined_inner_product<G: CommitmentCurve>(
...
+pub fn combined_inner_product2<F: Field>(
+    evaluation_points: &[F],
+    poly_combinator: F,
+    eval_combinator: F,
+    polynomials_evals: &[(Vec<F>, Option<usize>)],
```
mrmr1993, line 447: Nit: slice rather than `Vec`?

PR #450, opened 2022-03-22
add assert to test

Trivo25:
unit test case for `ceil_log2` wasn't being asserted, only printed. Add `assert!(res == *expected)` to it


mimoo (approved):
ah strange! thanks for the fix

Issue #437, opened 2022-03-18
factor our honer's polynomial evaluation


mimoo:
We use honer all over the place when we compute a polynomial 

```
f(x) = c0 + x c1 + x^2 c2 + x^3 c3
``` 

as 

```
x ( x (x (c3) + c2) + c1) + c0 
```

For example:

```rust
some_iterator.fold(F::zero(), |acc, x| joint_combiner.clone() * acc + x.clone())
```

it'd be good to factor this out as a helper in o1-utils like that:

```rust
fn evaluate_poly<F>(coefficients: Iter<Item=F>, zero: F, x: F) -> F where F: Add {
  coefficients.fold(zero, |acc, coeff| x * acc + coeff)
}
```

or 

```rust
fn evaluate_poly<F>(polynomial: DensePolynomial, zero: F, x:F) -> F where F: Add {
  polynomial.coeffs.fold(zero, |acc, coeff| x * acc + coeff)
}
```

I'm not sure exactly where it is used in the codebase, but I've seen plenty of implementations written in different ways

mimoo:
cc @Trivo25 that could be nice to add to the module you created :)

Trivo25:
will do! Was looking for some functions to add to the math lib already as well ;)

mimoo:
This issue should be about using this instead now:

```rust
DensePolynomial::<F>::from_coefficients_slice(&coeffs).evaluate(x);
```

duguorong009:
@mimoo @mrmr1993 
Can you please check the PR #876 which I created for this issue?
I could not find any other horner implementation in the repo, except for `utils::DensePolynomial`.
All other evaluations are using either of `DensePolynomial::from_coefficients_slice()` or `DensePolynomial::from_coefficients_vec()` of `arkworks::ark_poly` crate.

mimoo:
thanks @duguorong009 !

PR #451, opened 2022-03-22
move horner schema to utils

Trivo25:
adds horner schema evaluation of polynomials to `o1_utils::math` plus tests - i went with the iterator-variant instead of passing a `DensePolynomial` directly to make it more "generic"

Not sure if its necessary to make Field-specific tests or if primitive integer tests are enough, i added them nonetheless 

i did not replace any implementations yet

would like to hear some thoughts about that

#437

### DIFF
--- a/utils/src/math.rs
+++ b/utils/src/math.rs
@@ -29,25 +29,38 @@ where
 
 #[cfg(test)]
 mod tests {
+
     use super::*;
+    use mina_curves::pasta::Fp;
 
     #[test]
     fn test_eval() {
         /* [
             ([..coeffs..], (x, expected_y))
         ] */
-        let tests = [
-            (vec![5, 0, 7, 9], (2, 63)),
-            (vec![0, 0, 0, 3], (2, 3)),
-            (vec![16, 91, 23, 111], (32, 618_319)),
-            (vec![8, 0, 2, 0], (16, 32_800)),
-            (vec![3, 4, 1, -5], (3, 115)),
-            (vec![3, 4, 1, -5], (-6, -515)),
+        let test_set = [
+            ([5, 0, 7, 9], (2, 63)),
+            ([0, 0, 0, 3], (2, 3)),
+            ([16, 91, 23, 111], (32, 618_319)),
+            ([8, 0, 2, 0], (16, 32_800)),
+            ([0, 0, 0, 0], (11111, 0)),
         ];
 
-        for (coeffes, (x, expected)) in tests {
+        // runs the test_set as primitive i32 types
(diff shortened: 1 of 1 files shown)


utils/src/math.rs
```diff
@@ -17,9 +19,45 @@ pub fn ceil_log2(d: usize) -> usize {
...
 }
 
+/// Evaluates a polynomial's `coefficients` at a given point `x`, using [Horner's scheme](https://en.wikipedia.org/wiki/Horner%27s_method)
+pub fn evaluate_polynomial<F, I>(coefficients: I, zero: F, x: F) -> F
+where
+    I: Iterator<Item = F>,
+    F: Mul<F, Output = F> + Add<F, Output = F> + Copy,
+{
+    coefficients.fold(zero, |acc, coeff| x * acc + coeff)
+}
```
mimoo, line 29: I think this is not going to work as you're going through the coeffs in the wrong order no?
Also looking at this, I realize that we have a very similar function in `ExtendedDensePolynomial::eval_polynomial`, perhaps we can just use that?

Trivo25: > coeffs in the wrong order 

interesting, because I am not sure in what order the coeffs usually would be (especially in the `DensePolynomial` implementation) - I assumed they would be listed "from left to right", so `[2, 6, 2, 1]` would correspond to `2x³ + 6x² + 2x + 1`, I guess in this case they are handled the other way around natively? I was confused at first when I saw that the coeffs were reversed in the already existing implementations

> I realize that we have a very similar function in ExtendedDensePolynomial::eval_polynomial, perhaps we can just use that?

I did notice the implementation of `eval_polynomial` as well, assuming most calculations are done using the `ExtendedDensePolynomial` trait, the additional implementation could just be ignored

mimoo: yeah they are handled from left to right as far as I remember

mimoo: Do you want to try replacing occurrences of this pattern with the `eval_polynomial` function? Sorry, didn't realize it was already there ^^

mimoo: also I realize that we might just be able to do this:

```rust
DensePolynomial::<F>::from_coefficients_slice(&coeffs).evaluate(x);
```

I bet arkworks `evaluate` function is implemented like this (we should verify though)

Trivo25: will look into it 👍

Trivo25: @mimoo ye, arkwork's implementation of `evaluate` also uses horner's scheme, but more efficiently in parallel - so probably better to use that. I will go through the code and see if it can be replaced somewhere

Issue #453, opened 2022-03-23
Make changes of domain in Evaluations type clearer


mimoo:
ATM I believe we rely on three different domains `d1, d4, d8` where the size of `d8` is 8 times the size of `d1`, and the size of `d4` is 4 times the size of `d1`.

Because of the way domains are generated in our multiplicative group, we have a special relationship between these groups. From `domain.rs`:

```rust
// ensure the relationship between the three domains in case the library's behavior changes
assert_eq!(d2.group_gen.square(), d1.group_gen);
assert_eq!(d4.group_gen.square(), d2.group_gen);
assert_eq!(d8.group_gen.square(), d4.group_gen);
```

This means that we can easily switch from evaluations in different domains. We use this in different ways:

**1. to switch between fields**.

**2. to (binary) operate on evaluations on different domains**. For example:

```rust
mul.evals
                .par_iter_mut()
                .enumerate()
                .for_each(|(i, eval)| *eval *= mul_selector_d8[2 * i]);
```

**3. To retrieve evaluations like we're in d1**. For example, the following code in `prover.rs` converts the `Evaluations` from domain `d8` to `d1`:

```rust
let row = lookup_table8.iter().map(|e| &e.evals[8 * i]);
```

We have these all over the place, and it is sometimes really hard to parse them (especially in `expr.rs` I find). It would be nice to have an abstraction so that one could do something like this for example

```rust
let thing_d1 = domains.to_d1(thing_d8);
```

or 

```rust
domains.d2_times_d4(thing_d2, thing_d4);
```

I'm not sure what would be the cleanest way to achieve this.

mimoo:
I've toyed with it here: https://github.com/o1-labs/proof-systems/pull/455

Not clear if this is a good way. It'd be much cooler if we could just do something like this:

```rust
let thing_d1: D1 = (thing_d2 * thing_d4).into();
```

but we'd need to introduce wrapper types

```rust
pub struct E1<F>(Evaluations<F>);

trait SafeEvaluations {
  const SIZE: usize;
  fn from_evals(evals: &Evaluations) -> Self;
}

impl<F> E1<F> {
  fn from_evals(evals: Evaluations) -> Self {
    assert!(Self::SIZE == evals.domain().size());
    Self(evals);
  }
}
```

mrmr1993:
[...]

PR #455, opened 2022-03-23
[WIP][kimchi][domains] implement conversion between domains helpers

mimoo:
https://github.com/o1-labs/proof-systems/issues/453

looking for feedback on this approach

### DIFF
--- a/kimchi/src/circuits/domains.rs
+++ b/kimchi/src/circuits/domains.rs
@@ -1,5 +1,5 @@
 use ark_ff::FftField;
-use ark_poly::{EvaluationDomain, Radix2EvaluationDomain as Domain};
+use ark_poly::{EvaluationDomain, Evaluations, Radix2EvaluationDomain as Domain};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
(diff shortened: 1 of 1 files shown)


kimchi/src/circuits/domains.rs
```diff
@@ -40,13 +42,139 @@ impl<F: FftField> EvaluationDomains<F> {
...
+        };
+
+        let bigger_size = bigger.domain().size();
+        let smaller_size = smaller.domain().size();
+
+        let step = match bigger_size {
+            x if x == smaller_size => 1,
+            x if x == smaller_size * 2 => 2,
+            x if x == smaller_size * 4 => 4,
+            x if x == smaller_size * 8 => 8,
```
mimoo, line 154: one idea to make this safer is to check here what we enforce in [create]:

```rust
x if x == smaller_size * 2 => { assert_eq!(smaller.group_gen.square(), bigger.group_gen); 2 }
```

mimoo: another way to deal with this is to pass a reference to `domains: &EvaluationDomains<F>` and enforce that the domain of both the `lhs` and `rhs` is among the ones defined in our `domains`.

Issue #294, opened 2022-02-11
automate nondeterminism (hint addresses)


querolita:
The current version of our Cairo runner still requires some additional "manual help" to be able to fully behave as in the Cairo playground (that so far, I have been introducing manually from the outputs of the playground for given tests). To give a bit of context, the runner should only receive as parameters the compiled Cairo program given in bytecode format (what is represented as `CairoMemory<F>` and initial values for the pointers `pc` and `ap`. This is the data that a `CairoRunner` receives. Once a Cairo program is executed, the content of the `CairoMemory` should contain the following data:

- First addresses: the compiled bytecode program
- "Hint" addresses: some "hints", addresses to auxiliary variables needed to interpret the program. 
- Middle addresses: all of the temporary values obtained in execution, kind of like the "RAM". 
- Final addresses: output/return values

Right now, the first, middle and final addresses are implemented. What is lacking is the hints addresses. The reason of this is that we still didn't address one part of the nondeterministic computational model. In particular, the "hints" corresponds to addresses of memory that can only be obtained once the full program is executed. Then, they will include addresses such as where are outputs located (pointer to final addresses, only can be known after executing the whole program), or addresses that are larger than the final `ap` (which is the address that is read after the return command, and seems to correspond to the halting condition). Thus,the reason that these addresses cannot be given as public inputs (mostly, later on in Turshi) is that knowing these addresses can filter information about how long the computation took, and thus breaking zero knowledge (as some execution branches could have depended on the value of some witnesses, for example). 

I asked the cairo engineers in Discord and this is what I got:

_QUESTION_

> I have two questions regarding Cairo, I hope this is the right place to ask them. A little bit of context: I have been running some examples on the playground and I realised that right after the memory addresses containing the compiled bytecode for the program, there will be a few (2 or 3) cells containing what I think are addresses to "output" locations or the first address that is empty. Thus, when the return in the main updates the program counter, it gets to read those positions. The questions are:
> - How are these positions computed? Meaning, how does the runner know where it will store outputs /ending position before running the program. It looks a bit like the snake that bites its tail. Is this related to the nondeterministic computation model?
> - What is the halting condition? If the "final" next_pc indeed points to unallocated positions, perhaps the halting condition is: execute until current allocation pointer <= next_pc ... But in the paper I also found an example with an infinite loop at the end that tried to even the number of steps. 

_ANSWER 1_ 

> FeedTheFed | StarkWare — 
[...]

Issue #283, opened 2022-02-07
permutation constraints for memory


querolita:
The current version of the Cairo constraints do not yet address the memory checks: singularity (write once) and continuity (no holes). Even with a larger number of wires, I am not sure these can be added to the constraint system without adding more rows. This is because the memory checks are performed on 4 pairs of vectors, so these 8 values at each step may be too much for the copy constraints that only take the first 7 columns.

querolita:
UPDATE: constraints are written in a correctness fashion: the argument is complete but still not sound. I need to add randomness on the constraints from the `CairoInitial` and `CairoMemory` gates to include zeta and alpha.

Issue #282, opened 2022-02-07
custom wire length variant of the constraints


querolita:
With current 15-wire Kimchi, the Cairo constraints need at least 3 types of `CircuitGate`s taking at the end about `3n+1` rows, where `n` is the number of steps of computation. With a more generous number of wires we could keep the number of rows at `n+1`.

PR #457, opened 2022-03-23
Turshi15

querolita:
This PR provides a first functioning version of all the Cairo gate constraints. It still doesn't have the wiring between I/O registers and public input, but it should be easily addable. It needs about 5 rows per Cairo instruction in a 15-wire kimchi. The delta between this PR and the one for the Cairo runner are the following: 
- Added two fields to `CairoProgram`: one gives the list of instructions of the execution, the other one gives the ordered list of memory accesses throughout execution.
- `turshi.rs`: includes 
  - 7 Cairo gates definitions; 
  -  a witness generator from an execution of the runner; 
  -  a gadget generation from a witness, 
  -  constraints for the operands, operations, registers, instruction checks, registers updates, and memory singularity, continuity and self-permutation-ness 
  - a checking function that replicates the constraints expressions in field format
  - a verification function that evaluates the constraints directly
  - one simple check using the same program as the runner's tests
- some changes needed to be added to prover, expr, verifier, gate, wires,  index, for it to work. But further modifications will have to be carried out

These are the following types of gates (more or less, in gadget order). 
- `CairoInitial`: takes care of the first memory chunk and the initial claim about the registers.
- `CairoMemory`: takes care of the i-th memory chunk, connecting the partial permutation ratio with previous ones.
- `CairoInstruction`: takes care of the logics of the Cairo instruction.
- `CairoFlags`: performs registers update checks using the values of the flags
- `CairoTransition`: performs the check of a correct transition between consecutive instructions
- `CairoAuxiliary`: does nothing, it's just a row that holds 3 values. It would be great to merge this one somewhere else, or make it a `Zero` gate instead
- `CairoClaim`: takes care of the final registers claim and the final self-permutation check of the memory

Some remarks/TODOs for future versions:
- (Issue [283](https://github.com/o1-labs/proof-systems/issues/283)) The permutation argument that is embedded on the memory checks is correct, but not sound. Meaning, I still have to add a randomness parameters (called zeta and alpha) to make it as in the whitepaper.
- I have to write the wiring of the Cairo rows. In the comments I wrote some sort of table of the current layout and it explains what columns should be copied and from where. But this is still pending.
- I want to create full tests that involve both the prover and the verifier, not only these "gates checks"
- (Issue [282](https://github.com/o1-labs/proof-systems/issues/282))Think about a different number of columns and how much it would benefit Cairo and other complex gates. In this approach, if `n` is the number of instructions, the full gadget for the program takes `5n-1` rows.
- (Issue [294](https://github.com/o1-labs/proof-systems/issues/294))The runner still needs to address the hints independently: meaning, run the program `backwards` and figure out the intermediate 2 or 3 values of the memory that need to be set for this nondeterministic feature of the memory


querolita:
I am closing this PR to split it into smaller chunks: one for the instruction logics and another one for the memory logics that may be subject to change

PR #460, opened 2022-03-25
Re-export Hash type of the Hasher trait

hanabi1224:
This PR re-exports the Hash type (`Fp`) of the `Hasher<H: Hashable>` trait, so that the caller doesn't need to explicitly depend on the `mina-curves` crate to use this Hash type. 

@jspada plz take a look and let me know your feedback, thx!

### DIFF
--- a/hasher/src/lib.rs
+++ b/hasher/src/lib.rs
@@ -5,7 +5,7 @@ pub mod poseidon;
 pub mod roinput;
 
 use ark_ff::PrimeField;
-use mina_curves::pasta::Fp;
+pub use mina_curves::pasta::Fp as Hash;
 use o1_utils::FieldHelpers;
 pub use roinput::ROInput;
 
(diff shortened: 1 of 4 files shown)


hasher/src/lib.rs
```diff
@@ -5,7 +5,7 @@ pub mod poseidon;
 pub mod roinput;
 
 use ark_ff::PrimeField;
-use mina_curves::pasta::Fp;
+pub use mina_curves::pasta::Fp as Hash;
```
mimoo, line 8: I don't really like this naming, because `Fp` is not a `Hash`. I'm guessing we're doing this to make it more user-friendly? But will there be a cost down the line of not being explicit about which field you're in? This seems like a fairly low-level library. Perhaps `FieldDigest` would be a better name?

Also, it'd be nice to separate the `pub use`s  from the `use`.

jspada: This is a good point @mimoo.  Though I like the spirit of this change making things more clear, it may be best to leave it as `Fp`.  In some places Mina does a change of field after obtaining the `Fp` element hash output.  For security reasons we must make sure it's very clear whenever changing field happens, so the safest option is probably to leave it `Fp`.

hanabi1224: Fixed

jspada: I think it's better to leave `Fp` in one place and not re-export it here.  Since this PR doesn't really introduce anything else, I propose just to close it.

jspada: Change of heart -> we'll take it.

jspada (approved):
LGTM!

Issue #432, opened 2022-03-17
need better abstractions to represent chunked polynomials, chunked evaluations, chunked commitments, etc.


mimoo:
I believe we mostly used chunked SOMETHING for the quotient polynomial, and the f polynomial, in their different forms (coefficient form, evaluation form, or commitment form).

The code is quite messy in parts where we handle the chunked polynomials IMO. This issue is mostly about brainstorming what could be a good abstraction for these to make the code clearer.

For example, this is part of the prover:

```rust
        let chunked_evals = {
            let chunked_evals_zeta = ProofEvaluations::<Vec<Fr<G>>> {
                s: array_init(|i| {
                    index.cs.sigmam[0..PERMUTS - 1][i].eval(zeta, index.max_poly_size)
                }),
                w: array_init(|i| witness_poly[i].eval(zeta, index.max_poly_size)),
                z: z_poly.eval(zeta, index.max_poly_size),
                lookup: lookup_evals(zeta),
                generic_selector: index.cs.genericm.eval(zeta, index.max_poly_size),
                poseidon_selector: index.cs.psm.eval(zeta, index.max_poly_size),
            };
            let chunked_evals_zeta_omega = ProofEvaluations::<Vec<Fr<G>>> {
                s: array_init(|i| {
                    index.cs.sigmam[0..PERMUTS - 1][i].eval(zeta_omega, index.max_poly_size)
                }),
                w: array_init(|i| witness_poly[i].eval(zeta_omega, index.max_poly_size)),
                z: z_poly.eval(zeta_omega, index.max_poly_size),
                lookup: lookup_evals(zeta_omega),
                generic_selector: index.cs.genericm.eval(zeta_omega, index.max_poly_size),
                poseidon_selector: index.cs.psm.eval(zeta_omega, index.max_poly_size),
            };

            [chunked_evals_zeta, chunked_evals_zeta_omega]
        };
```

mimoo:
also this `eval` function should really be called `linearization`

mimoo:
perhaps something like this:

```
struct ChunkedPolynomial<F>(Vec<DensePolynomial<F>>);

impl ChunkedPolynomial {
  fn linearize(&self, chunk_size: usize, x: F) -> DensePolynomial<F>
}
```

with a function to convert a polynomial into a chunked polynomial?

mimoo:
[...]

PR #467, opened 2022-03-27
refactor ChunkedPolynomials

xiangxiecrypto:
This PR (related to issue #432):
- add a new module  in `utils/chunked_polynomial.rs` with an abstract struct `ChunkedPolynomials`.
- the new `ChunkedPolynomials` struct contains two method: `linearize` and `compress_polynomial`. `linearize` evaluates all the chunked polynomials in the same point. `compress_polynomial` compresses all the chunked polynomials into one with `zeta_n`.
- refactor the original struct `ChunkedPolynomial` with `ScaledChunkedPolynomials` based on the above abstract. This struct contains `(Vec[F], Vec[P])`.
-  add a function `to_chunked_polynomials` in trait `ExtendedDensePolynomial`, which allows to transfer polynomials into multiple chunks.
- modify accordingly in the `prover.rs` file.

### DIFF
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -16,7 +16,6 @@ use ark_ff::{
 };
 use ark_poly::{
     univariate::DensePolynomial, EvaluationDomain, Evaluations, Radix2EvaluationDomain as D,
-    UVPolynomial,
 };
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use core::ops::{Add, Sub};
@@ -422,65 +421,6 @@ pub fn combined_inner_product<G: CommitmentCurve>(
         }
     }
     res
-}
-
-enum OptShiftedPolynomial<P> {
-    Unshifted(P),
-    Shifted(P, usize),
-}
-
-/// A formal sum of the form
-/// `s_0 * p_0 + ... s_n * p_n`
-/// where each `s_i` is a scalar and each `p_i` is an optionally shifted polynomial.
-pub struct ChunkedPolynomial<F, P>(Vec<(F, OptShiftedPolynomial<P>)>);
-
-impl<F, P> Default for ChunkedPolynomial<F, P> {
-    fn default() -> ChunkedPolynomial<F, P> {
-        ChunkedPolynomial(vec![])
-    }
-}
-
-impl<F, P> ChunkedPolynomial<F, P> {
-    pub fn add_unshifted(&mut self, scale: F, p: P) {
-        self.0.push((scale, OptShiftedPolynomial::Unshifted(p)))
-    }
-
-    pub fn add_shifted(&mut self, scale: F, shift: usize, p: P) {
-        self.0
-            .push((scale, OptShiftedPolynomial::Shifted(p, shift)))
-    }
-}
-
-impl<'a, F: Field> ChunkedPolynomial<F, &'a [F]> {
-    pub fn to_dense_polynomial(&self) -> DensePolynomial<F> {
-        let mut res = DensePolynomial::<F>::zero();
-
-        let scaled: Vec<_> = self
-            .0
-            .par_iter()
-            .map(|(scale, segment)| {
-                let scale = *scale;
-                match segment {
-                    OptShiftedPolynomial::Unshifted(segment) => {
-                        let v = segment.par_iter().map(|x| scale * *x).collect();
-                        DensePolynomial::from_coefficients_vec(v)
-                    }
-                    OptShiftedPolynomial::Shifted(segment, shift) => {
-                        let mut v: Vec<_> = segment.par_iter().map(|x| scale * *x).collect();
-                        let mut res = vec![F::zero(); *shift];
-                        res.append(&mut v);
-                        DensePolynomial::from_coefficients_vec(res)
-                    }
-                }
-            })
-            .collect();
-
-        for p in scaled {
-            res += &p;
-        }
-
-        res
-    }
 }
 
 /// Contains the evaluation of a polynomial commitment at a set of points.
@@ -901,7 +841,7 @@ mod tests {
     use super::*;
 
     use crate::srs::SRS;
-    use ark_poly::Polynomial;
+    use ark_poly::{Polynomial, UVPolynomial};
     use array_init::array_init;
     use mina_curves::pasta::{fp::Fp, vesta::Affine as VestaG};
     use oracle::constants::PlonkSpongeConstantsKimchi as SC;
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -908,8 +908,12 @@ mod tests {
 
         // evaluate the polynomials at these two points
         let poly1_chunked_evals = vec![
-            poly1.to_chunked_polynomial(srs.g.len()).evaluate_chunks(elm[0]),
-            poly1.to_chunked_polynomial(srs.g.len()).evaluate_chunks(elm[1]),
+            poly1
+                .to_chunked_polynomial(srs.g.len())
+                .evaluate_chunks(elm[0]),
+            poly1
+                .to_chunked_polynomial(srs.g.len())
+                .evaluate_chunks(elm[1]),
         ];
 
         fn sum(c: &[Fp]) -> Fp {
@@ -920,8 +924,12 @@ mod tests {
         assert_eq!(sum(&poly1_chunked_evals[1]), poly1.evaluate(&elm[1]));
 
         let poly2_chunked_evals = vec![
-            poly2.to_chunked_polynomial(srs.g.len()).evaluate_chunks(elm[0]),
-            poly2.to_chunked_polynomial(srs.g.len()).evaluate_chunks(elm[1]),
+            poly2
+                .to_chunked_polynomial(srs.g.len())
+                .evaluate_chunks(elm[0]),
+            poly2
+                .to_chunked_polynomial(srs.g.len())
+                .evaluate_chunks(elm[1]),
         ];
 
         assert_eq!(sum(&poly2_chunked_evals[0]), poly2.evaluate(&elm[0]));
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -908,8 +908,8 @@ mod tests {
 
         // evaluate the polynomials at these two points
         let poly1_chunked_evals = vec![
-            poly1.eval(elm[0], srs.g.len()),
-            poly1.eval(elm[1], srs.g.len()),
+            poly1.to_chunked_polynomials(srs.g.len()).linearize(elm[0]),
+            poly1.to_chunked_polynomials(srs.g.len()).linearize(elm[1]),
         ];
 
         fn sum(c: &[Fp]) -> Fp {
@@ -920,8 +920,8 @@ mod tests {
         assert_eq!(sum(&poly1_chunked_evals[1]), poly1.evaluate(&elm[1]));
 
         let poly2_chunked_evals = vec![
-            poly2.eval(elm[0], srs.g.len()),
-            poly2.eval(elm[1], srs.g.len()),
+            poly2.to_chunked_polynomials(srs.g.len()).linearize(elm[0]),
+            poly2.to_chunked_polynomials(srs.g.len()).linearize(elm[1]),
         ];
 
         assert_eq!(sum(&poly2_chunked_evals[0]), poly2.evaluate(&elm[0]));
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -908,8 +908,8 @@ mod tests {
 
         // evaluate the polynomials at these two points
         let poly1_chunked_evals = vec![
-            poly1.to_chunked_polynomials(srs.g.len()).linearize(elm[0]),
-            poly1.to_chunked_polynomials(srs.g.len()).linearize(elm[1]),
+            poly1.to_chunked_polynomial(srs.g.len()).evaluate_chunks(elm[0]),
+            poly1.to_chunked_polynomial(srs.g.len()).evaluate_chunks(elm[1]),
         ];
 
         fn sum(c: &[Fp]) -> Fp {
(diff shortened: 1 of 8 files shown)


utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    pub fn add_shifted(&mut self, scale: F, shift: usize, p: P) {
+        self.scale.push(scale);
+        self.chunked_polynomials
+            .add_chunk(OptShiftedPolynomial::Shifted(p, shift))
+    }
+}
+
+impl<'a, F: Field> ScaledChunkedPolynomials<F, &'a [F]> {
+    /// check length?
```
querolita, line 101: if this is a TODO, change to normal comment // 
and if it is indeed a TODO, is there any test to check the length? it seems like an important check to make sure dense polynomials will be correctly formed

utils/src/dense_polynomial.rs
```diff
@@ -19,14 +21,7 @@ pub trait ExtendedDensePolynomial<F: Field> {
...
 
-    /// This function evaluates polynomial in chunks.
-    fn eval(&self, elm: F, size: usize) -> Vec<F>;
-
-    /// Multiplies the chunks of a polynomial with powers of zeta^n to make it of degree n-1.
-    /// For example, if a polynomial can be written `f = f0 + x^n f1 + x^2n f2`
-    /// (where f0, f1, f2 are of degree n-1), then this function returns the new semi-evaluated
-    /// `f'(x) = f0(x) + zeta^n f1(x) + zeta^2n f2(x)`.
-    fn chunk_polynomial(&self, zeta_n: F, n: usize) -> Self;
+    fn to_chunked_polynomials(&self, size: usize) -> ChunkedPolynomials<DensePolynomial<F>>;
```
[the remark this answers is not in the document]
mimoo: `to_chunked_polynomial` singular no?

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        while coeffs.last().map_or(false, |c| c.is_zero()) {
+            coeffs.pop();
+        }
+
+        DensePolynomial { coeffs }
+    }
+}
+enum OptShiftedPolynomial<P> {
```
mimoo, line 61: nit: missing space between the two structs

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+    Unshifted(P),
+    Shifted(P, usize),
+}
+
+/// A formal sum of the form
+/// `s_0 * p_0 + ... s_n * p_n`
+/// where each `s_i` is a scalar and each `p_i` is an optionally shifted polynomial.
```
mimoo, line 68: this comment seems unrelated to the struct below

utils/src/chunked_polynomial.rs
mimoo: also some doc would be useful

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+
+///pub struct ChunkedPolynomial<F, P>(Vec<(F, OptShiftedPolynomial<P>)>);
+
+pub struct ScaledChunkedPolynomials<F, P> {
+    scale: Vec<F>,
+    chunked_polynomials: ChunkedPolynomials<OptShiftedPolynomial<P>>,
+}
+
+impl<F, P> Default for ScaledChunkedPolynomials<F, P> {
+    fn default() -> ScaledChunkedPolynomials<F, P> {
```
mimoo, line 78: derive Default on the struct instead

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+
+use rayon::prelude::*;
+
+#[derive(Clone)]
+pub struct ChunkedPolynomials<P> {
+    pub polys: Vec<P>,
+    pub degree: usize,
+}
+
+impl<P> Default for ChunkedPolynomials<P> {
```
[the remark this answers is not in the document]
mimoo: also, should this really have a default implementation? It sounds to me like you would always create this struct from a `DensePolynomial`

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
+use ark_ff::{Field, Zero};
+use ark_poly::polynomial::{univariate::DensePolynomial, Polynomial, UVPolynomial};
+
+use rayon::prelude::*;
+
+#[derive(Clone)]
+pub struct ChunkedPolynomials<P> {
+    pub polys: Vec<P>,
+    pub degree: usize,
```
mimoo, line 9: both the struct and the fields need documentation. Here `degree` sounds like it could be renamed to `chunk_degree` or something

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+/// A formal sum of the form
+/// `s_0 * p_0 + ... s_n * p_n`
+/// where each `s_i` is a scalar and each `p_i` is an optionally shifted polynomial.
+
+///pub struct ChunkedPolynomial<F, P>(Vec<(F, OptShiftedPolynomial<P>)>);
+
+pub struct ScaledChunkedPolynomials<F, P> {
+    scale: Vec<F>,
+    chunked_polynomials: ChunkedPolynomials<OptShiftedPolynomial<P>>,
+}
```
mimoo, line 75: I'm a bit confused what this struct is about. (I know you didn't add it.) Same comment as previously, I think this should be `Polynomial` singular. What puzzles me is why we have several shifted polynomials, when we usually only care about the last one (the shift being interesting for upperbounding the degree of a poly)

utils/src/lib.rs
```diff
@@ -1,6 +1,8 @@
+pub mod chunked_polynomial;
 pub mod dense_polynomial;
 pub mod evaluations;
 pub mod field_helpers;
+
```
mimoo, line 5: remove the space

mimoo (changes requested):
I'm mid-review, making sure that this doesn't get merged as I think it still needs some work.

That being said, I really like the direction so far :) I think this is going to greatly improve the codebase.

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+impl<P> ChunkedPolynomials<P> {
+    pub fn add_chunk(&mut self, p: P) {
+        self.polys.push(p)
+    }
+}
+
+impl<F: Field> ChunkedPolynomials<DensePolynomial<F>> {
```
mimoo, line 27: I'm not a fan of this generic parameter. I think it's make more sense to define `ChunkedPolynomial` as

```rust
pub struct ChunkedPolynomial<F> {
    pub polys: Vec<DensePolynomial<F>>,
    pub degree: usize,
}
```

as this is what we're going to use it with mostly.

poly-commitment/src/commitment.rs
```diff
@@ -424,65 +423,6 @@ pub fn combined_inner_product<G: CommitmentCurve>(
...
-            .collect();
-
-        for p in scaled {
-            res += &p;
-        }
-
-        res
-    }
-}
-
```
mimoo, line 485: I actually think we should try not to touch this part of the code (poly-commitment) yet. As this logic seems a bit unnecessary to me at the moment. I think this PR could be more useful if targeted to the kimchi crate to begin with.

utils/src/dense_polynomial.rs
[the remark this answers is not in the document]
querolita: Oh way better indeed

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// This function evaluates polynomial in chunks.
+    pub fn linearize(&self, elm: F) -> Vec<F> {
+        let mut res: Vec<F> = vec![];
+        for poly in self.polys.iter() {
+            let eval = poly.evaluate(&elm);
+            res.push(eval);
+        }
+        res
+    }
```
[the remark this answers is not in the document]
querolita: good point

utils/src/chunked_polynomial.rs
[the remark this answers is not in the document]
xiangxiecrypto: > also, should this really have a default implementation? It sounds to me like you would always create this struct from a `DensePolynomial`

This is for `ScaledChunkedPolynomial` , the original `ChunkedPolynomial` used in opening proof.

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+            for (coeff, poly_coeff) in coeffs.iter_mut().zip(&poly.coeffs) {
+                *coeff += scale * poly_coeff;
+            }
+
+            scale *= zeta_n;
+        }
+
+        while coeffs.last().map_or(false, |c| c.is_zero()) {
+            coeffs.pop();
+        }
```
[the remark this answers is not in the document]
xiangxiecrypto: > I feel like this could be rewritten in a clearer way. Perhaps something like this:
> 
> ```rust
> let coeffs = coeffs.rev().skip_while(|c| c.is_zero()).rev().collect();
> ```

well, I copy the codes here... Yes, the new one is better i think.

utils/src/chunked_polynomial.rs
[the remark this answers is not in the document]
xiangxiecrypto: > I wouldn't call this function `linearize`. Linearizing a poly to me is:
> 
> given a poly `f(x) = f0(x) + x^n f1(x) + x^2n f2(x) + ...` and a point `z`, return f(x) = f0(x) + z^n f1(x) + z^2n f2(x) + ...`
> 
> where all the `f_i` are of degree `n-1`
> 
> this function is more like `evaluate_chunks` to me

Ok, this means I have to rename `linearize` and `compress_polynomials`

utils/src/chunked_polynomial.rs
xiangxiecrypto: > I'm not a fan of this generic parameter. I think it's make more sense to define `ChunkedPolynomial` as
> 
> ```rust
> pub struct ChunkedPolynomial<F> {
>     pub polys: Vec<DensePolynomial<F>>,
>     pub degree: usize,
> }
> ```
> 
> as this is what we're going to use it with mostly.

I thought about this before, the reason I use more generic parameters is to be compatible with the definition of `ScaledChunkedPolynomials`. Perhaps, we could refactor it.

poly-commitment/src/commitment.rs
xiangxiecrypto: > I actually think we should try not to touch this part of the code (poly-commitment) yet. As this logic seems a bit unnecessary to me at the moment. I think this PR could be more useful if targeted to the kimchi crate to begin with. Would you mind moving the poly-commitment stuff in this PR to another PR/branch for now?

Well, if we refactor `ChunkedPolynomial` into a new module, it is necessary that we move this part into the new file.

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
+use ark_ff::{Field, Zero};
+use ark_poly::polynomial::{univariate::DensePolynomial, Polynomial, UVPolynomial};
+
+use rayon::prelude::*;
+
+#[derive(Clone)]
+pub struct ChunkedPolynomials<P> {
```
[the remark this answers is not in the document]
xiangxiecrypto: > wouldn't `ChunkedPolynomial` singular be better here? Since it's a polynomial that's chunked

Yes, that is better.

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,164 @@
...
+
+            scale *= zeta_n;
+        }
+
+        while coeffs.last().map_or(false, |c| c.is_zero()) {
+            coeffs.pop();
+        }
+
+        DensePolynomial { coeffs }
+    }
```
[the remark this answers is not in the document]
xiangxiecrypto: > this is the `linearize` function to me :o could be used here for example: https://github.com/o1-labs/marlin/blob/37b7ad53f968c1368ce61ca791aedafe3c06fc33/kimchi/src/prover.rs#L839

Hmmm, maybe I misunderstand the meaning in Issue #432 :), will change that.

utils/src/chunked_polynomial.rs
xiangxiecrypto: > I'm a bit confused what this struct is about. (I know you didn't add it.) Same comment as previously, I think this should be `Polynomial` singular. What puzzles me is why we have several shifted polynomials, when we usually only care about the last one (the shift being interesting for upperbounding the degree of a poly)

This struct is used in commitment, where one has to bound the degree in the original implementation.

utils/src/chunked_polynomial.rs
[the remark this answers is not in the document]
xiangxiecrypto: > I feel like this could be rewritten in a clearer way. Perhaps something like this:
> 
> ```rust
> let coeffs = coeffs.rev().skip_while(|c| c.is_zero()).rev().collect();
> ```

This is not the functionality we want. Here we remove the leading zeros, while the code above remove all the zero coefficient.
Actually, I just copy the codes in the original implementation. :)

xiangxiecrypto:
Update:
- Use less generic definition of `ChunkedPolynomial` with field `Vec<DensePolynomial<F>>`.
- Add documentations and use suitable names for pub structs and functions.
- Move `ScaledChunkedPolynomial` into `evaluation_proof.rs`, the struct and functions are private.

utils/src/chunked_polynomial.rs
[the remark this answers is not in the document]
mimoo: I think they're all equivalent :o https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e7dda821e6964961bdb6d350929edaac

> Actually, I just copy the codes in the original implementation. :)

yeah sorry about that ^^

utils/src/chunked_polynomial.rs
```diff
@@ -0,0 +1,74 @@
...
+        let res = (one + zeta)
+            * (one + zeta_n + zeta_n * zeta.square() + zeta_n * zeta.square() * zeta.square());
+
+        // 1 + x + x^2 + x^3 + x^4 + x^5 + x^6 + x^7 = (1+x) + x^2 (1+x) + x^4 (1+x) + x^6 (1+x)
+        let coeffs = [one, one, one, one, one, one, one, one];
+        let f = DensePolynomial::from_coefficients_slice(&coeffs);
+
+        let eval = f.to_chunked_polynomial(2).linearize(zeta_n).evaluate(&zeta);
+
+        assert!(eval == res);
```
mimoo, line 72: this should be equal to `f.evaluate(&zeta)` as well

poly-commitment/src/evaluation_proof.rs
```diff
@@ -12,6 +12,60 @@ use rand_core::{CryptoRng, RngCore};
...
+            .collect();
+
+        for p in scaled {
+            res += &p;
+        }
+
+        res
+    }
+}
+
```
mimoo, line 68: I see, ScaledChunkedCommitment is only used here. I think I have some idea on how to get rid of it in the future

xiangxiecrypto:
> 

Let me think about it.

Thanks for the reviews @mimoo @querolita :)

PR #478, opened 2022-03-30
Fixup ocaml-gen to emit snake-case type parameters

mrmr1993:
This PR tweaks ocaml-gen to use snake-case for type parameters. This makes it far easier to use OCaml derivers etc., which are currently broken by the capitalisation that we use for variables.

Example output:
```ocaml
  type nonrec 'caml_f scalar_challenge = { inner : 'caml_f } [@@boxed]
  type nonrec 'caml_f random_oracles =
    { joint_combiner : 'caml_f scalar_challenge * 'caml_f
    ; beta : 'caml_f
    ; gamma : 'caml_f
    ; alpha_chal : 'caml_f scalar_challenge
    ; alpha : 'caml_f
    ; zeta : 'caml_f
    ; v : 'caml_f
    ; u : 'caml_f
    ; zeta_chal : 'caml_f scalar_challenge
    ; v_chal : 'caml_f scalar_challenge
    ; u_chal : 'caml_f scalar_challenge
    }
```

### DIFF
--- a/ocaml/ocaml-gen/derive/src/lib.rs
+++ b/ocaml/ocaml-gen/derive/src/lib.rs
@@ -174,10 +174,9 @@ pub fn derive_ocaml_enum(item: TokenStream) -> TokenStream {
         .params
         .iter()
         .filter_map(|g| match g {
-            GenericParam::Type(t) => Some(&t.ident),
+            GenericParam::Type(t) => Some(t.ident.to_string().to_case(Case::Snake)),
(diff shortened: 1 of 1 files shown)


mimoo (approved):
nice, note that we should try to keep https://github.com/o1-labs/ocaml-gen up to date as well so we can remove ocaml-gen from this repo down the line. I'm not sure how far back it is though

PR #479, opened 2022-03-30
[mina-hasher] Absorb domain string on update

hanabi1224:
This PR fixes a bug that the `domain_string` is never properly consumed in the example Merkle tree scenario described in [readme](https://github.com/o1-labs/proof-systems/tree/master/hasher)

changes:
- Changes the behavior of `update` function in Poseidon hasher, to absorb the domain string of the input when available. 
- Adds a test case for Merkle tree scenario to make sure the domain string is properly consumed.

```rust
use ark_ff::Zero;
use mina_hasher::{create_legacy, Fp, Hashable, Hasher, ROInput};

#[derive(Clone)]
struct ExampleMerkleNode {
    height: u64,
    left: Fp,
    right: Fp,
}

impl Hashable for ExampleMerkleNode {
    type D = ();

    fn to_roinput(&self) -> ROInput {
        let mut roi = ROInput::new();

        roi.append_field(self.left);
        roi.append_field(self.right);

        roi
    }

    fn domain_string(this: Option<&Self>, _: Self::D) -> Option<String> {
        match this {
            None => format!("Unused").into(),
            // This is never called.
            Some(x) => format!("ExampleMerkleNode{:03}", x.height).into(),
        }
    }
}

// Used like this
let mut hasher = create_legacy::<ExampleMerkleNode>(());
let node = ExampleMerkleNode {
    height: 3,
    left: Fp::zero(),
    right: Fp::zero(),
};
let out = hasher.hash(&node);
// Or like this..
let out = hasher.update(&node).digest();
```

### DIFF
--- a/hasher/src/lib.rs
+++ b/hasher/src/lib.rs
@@ -80,9 +80,7 @@ pub trait Hashable: Clone {
     /// be enforced by the developer implementing the traits (see [`Hashable`] for
     ///  more details). The domain string may be parameterized by the contents of
     /// `this` and/or the generic `domain_param` argument.
-    ///
-    /// **Note:** You should always return `Some(String)`. A `None` return value
-    /// is only used for testing.
+    /// `this` is always None on `init`, and is always Some on `update`
     fn domain_string(this: Option<&Self>, domain_param: Self::D) -> Option<String>;
 }
 
--- a/hasher/src/poseidon.rs
+++ b/hasher/src/poseidon.rs
@@ -24,6 +24,7 @@ pub struct Poseidon<SC: SpongeConstants, H: Hashable> {
     sponge: ArithmeticSponge<Fp, SC>,
     sponge_state: SpongeState,
     state: Vec<Fp>,
+    domain_param: H::D,
     phantom: PhantomData<H>,
 }
 
@@ -33,6 +34,7 @@ impl<SC: SpongeConstants, H: Hashable> Poseidon<SC, H> {
             sponge: ArithmeticSponge::<Fp, SC>::new(sponge_params),
             sponge_state: SpongeState::Absorbed(0),
             state: vec![],
+            domain_param: domain_param,
             phantom: PhantomData,
         };
 
@@ -68,8 +70,7 @@ where
         self.sponge.reset();
 
         if let Some(domain_string) = H::domain_string(None, domain_param) {
-            self.sponge
-                .absorb(&[domain_prefix_to_field::<Fp>(domain_string)]);
+            self.sponge.absorb(&[domain_prefix_to_field(domain_string)]);
             self.sponge.squeeze();
         }
 
@@ -81,6 +82,9 @@ where
     }
 
     fn update(&mut self, input: &H) -> &mut dyn Hasher<H> {
+        if let Some(domain_string) = H::domain_string(Some(input), self.domain_param) {
+            self.sponge.absorb(&[domain_prefix_to_field(domain_string)]);
+        }
         self.sponge.absorb(&input.to_roinput().to_fields());
 
         self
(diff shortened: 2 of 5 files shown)


hasher/src/lib.rs
```diff
@@ -80,9 +80,7 @@ pub trait Hashable: Clone {
     /// be enforced by the developer implementing the traits (see [`Hashable`] for
     ///  more details). The domain string may be parameterized by the contents of
     /// `this` and/or the generic `domain_param` argument.
-    ///
-    /// **Note:** You should always return `Some(String)`. A `None` return value
-    /// is only used for testing.
+    /// `this` is always None on `init`, and is always Some on `update`
```
jspada, line 83: This is a good additional comment.

hasher/src/lib.rs
```diff
@@ -80,9 +80,7 @@ pub trait Hashable: Clone {
     /// be enforced by the developer implementing the traits (see [`Hashable`] for
     ///  more details). The domain string may be parameterized by the contents of
     /// `this` and/or the generic `domain_param` argument.
-    ///
-    /// **Note:** You should always return `Some(String)`. A `None` return value
```
jspada, line 84: This comment should remain, as returning None should not happen.  The `this` however can be None.

hasher/src/poseidon.rs
```diff
@@ -81,6 +82,9 @@ where
     }
 
     fn update(&mut self, input: &H) -> &mut dyn Hasher<H> {
+        if let Some(domain_string) = H::domain_string(Some(input), self.domain_param) {
```
jspada, line 85: Setting the domain string on update should definitely not be happening.

hasher/tests/merkle.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[test]
+fn ensure_domain_string_is_invoked() {
+    let node = TestMerkleNode {
+        height: 3,
+        left: Fp::zero(),
+        right: Fp::zero(),
+    };
+    for i in 0..10 {
+        let mut hasher = create_legacy(());
```
jspada, line 46: Yeah, it's not intended to be used like this.  As the examples show, you're force to init the hasher with something for domain separation.  For example, start with 0.

Then if you're forced to do some "weird" hashing, such as using the height as domain separation before each hash, then you need to use the right interface, which is not the `update` interface.

Instead you should use one of the one-shot interfaces, such as

```rust
let data = Example { height: 82, left: Fp::zero(), right: Fp::zero() };
let out = hasher.hash(&data);

// Or if you define domain_string without using `this`... like this
let out = hasher.init(data.height).update(&data).digest();
let out = hasher.init_and_hash(data.height, &data);
```

The update interface is for incremental stuff, so the pattern would have to be

```
init // sets domain separation
update
update
...
update
digest

init // sets another domain separation
update
update
...
update
digest
```

Update must definitely not be setting domain separation each time!

jspada (changes requested):

hasher/tests/merkle.rs
hanabi1224: `init` works except it resets the initial states, and in my case that prevents proper pooling, should there be an `init_lite` fn?

jspada: It's not necessary to get what you want.  You can always call `init` to set the domain separation to whatever you want, whenever you want (e.g. after popping from the pool).

For example,

```
1. Setup merkle_hasher_pool with hashers domain separation initialised to whatever height (e.g. choose `0`)
2. hasher <- merkle_hasher_pool // grab hasher from pool
3. hasher.init(height) // set domain separation for height that you like
4. out <- hasher.hash(merkle_node)
5. put hasher back in the pool
```

hanabi1224: That works thanks!

jspada:
no bug

PR #482, opened 2022-03-30
Remove superfluous and confusing 'this' argument from Hashable::domai…

### LINKED ISSUE #481: simplify Hashable interface by removing unneeded domain_string this parameter
It turns out we don't need `Option<&Self>` as a parameter for domain string creation since Merkle hashing could be done without this.  Thus, we should remove it from the Hashable interface to make things simpler and clearer.   Also, I don't think the concept was good and should be removed on those grounds alone.

jspada:
Closes #481

### DIFF
--- a/hasher/README.md
+++ b/hasher/README.md
@@ -29,7 +29,7 @@ impl Hashable for Example {
         roi
     }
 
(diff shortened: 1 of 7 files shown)


querolita (approved):
Neat!

PR #484, opened 2022-03-31
Add derives to compressed pub key

willemolding:
It would be great if you could add derivations of `Debug` `PartialEq` and `Eq` to the public API for `CompressedPubKey`. This makes it consistent with `PubKey` and makes it much easier to integrate into our internal types which use the above traits for tests.

I've also included a helper for `from_address`, this is just a nice-to-have and not required.


jspada (approved):
Nice enhancement.  Thanks!

PR #485, opened 2022-03-31
specify endomul_scalar gate

querolita:
This is a specification of the `endomul_scalar` gate. 

It includes the previous documentation comments as part of the spec now, together with a bunch of new content to explain more verbosely the resulting constraints.


mimoo (approved):
fantastic again!!!

PR #486, opened 2022-03-31
[mina-hasher] Change append_hashable to take borrow param

hanabi1224:
This PR changes the signature of `append_hashable` in `mina_hasher::Hashable` from `fn append_hashable(&mut self, input: impl Hashable)` to `fn append_hashable(&mut self, input: &impl Hashable)`

When it takes an owned value, it forces an unnecessary `clone` or `copy` at caller, like in the original doc example, `roi.append_hashable(self.a1);` works because `struct A` derives `Copy` and an implict `copy` happens here.

### DIFF
--- a/hasher/src/roinput.rs
+++ b/hasher/src/roinput.rs
@@ -73,7 +73,7 @@ impl ROInput {
     }
 
     /// Append a `Hashable` input
-    pub fn append_hashable(&mut self, input: impl Hashable) -> &mut Self {
+    pub fn append_hashable(&mut self, input: &impl Hashable) -> &mut Self {
         self.append_roinput(input.to_roinput());
         self
     }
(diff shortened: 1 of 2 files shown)


hasher/src/roinput.rs
```diff
@@ -73,7 +73,7 @@ impl ROInput {
     }
 
     /// Append a `Hashable` input
-    pub fn append_hashable(&mut self, input: impl Hashable) -> &mut Self {
+    pub fn append_hashable(&mut self, input: &impl Hashable) -> &mut Self {
```
jspada, line 76: Great PR.  I should have done this.  Can you please also make `append_roinput` take roi as a reference parameter?

jspada (commented):
Great PR.  Can you also make

hasher/src/roinput.rs
jspada: Or you can leave it.  Up to you.

PR #490, opened 2022-03-31
rename to ProverError

### LINKED ISSUE #454: Return Result instead of Option
- [ ] `ProofError` should be renamed to `ProverError`
- [ ] `ConstraintSystem::create` and `EvaluationDomains::create` should return `Result<_, ProverError>` instead of an `Option`. Or perhaps, they should simply panic?

Trivo25:
renames `ProofError` to `ProverError` and returns a `Result<_, ProverError>` instead of an `Option`

I am currently using `ProverError::Prover("some error")` as error, should there maybe be another error type in the `ProverError` enum that specifies the exact error? 

closes #454


kimchi/src/circuits/domains.rs
```diff
@@ -21,24 +23,34 @@ impl<F: FftField> EvaluationDomains<F> {
...
-        let n = Domain::<F>::compute_size_of_domain(n)?;
+    pub fn create(n: usize) -> Result<Self, ProverError> {
+        let n = Domain::<F>::compute_size_of_domain(n)
+            .ok_or(ProverError::Prover("couldn't compute size of domain"))
+            .unwrap();
 
-        let d1 = Domain::<F>::new(n)?;
+        let d1 = Domain::<F>::new(n)
+            .ok_or(ProverError::Prover("could not construct domain"))
+            .unwrap();
```
mimoo, line 33: since the function returns a `Result`, return an error instead of unwrapping here (using `?`)

kimchi/src/circuits/domains.rs
```diff
@@ -21,24 +23,34 @@ impl<F: FftField> EvaluationDomains<F> {
...
-        let d8 = Domain::<F>::new(8 * n)?;
+        let d2 = Domain::<F>::new(2 * n)
+            .ok_or(ProverError::Prover("could not construct domain"))
+            .unwrap();
+        let d4 = Domain::<F>::new(4 * n)
+            .ok_or(ProverError::Prover("could not construct domain"))
+            .unwrap();
+        let d8 = Domain::<F>::new(8 * n)
+            .ok_or(ProverError::Prover("could not construct domain"))
+            .unwrap();
```
mimoo, line 46: same here

kimchi/src/circuits/constraints.rs
```diff
@@ -385,15 +388,19 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
...
         // for some reason we need more than 1 gate for the circuit to work, see TODO below
         assert!(gates.len() > 1);
 
         //~ 2. Create a domain for the circuit. That is,
         //~    compute the smallest subgroup of the field that
         //~    has order greater or equal to `n + ZK_ROWS` elements.
-        let domain = EvaluationDomains::<F>::create(gates.len() + ZK_ROWS as usize)?;
+        let domain = match EvaluationDomains::<F>::create(gates.len() + ZK_ROWS as usize) {
+            Ok(val) => val,
+            Err(e) => return Err(e),
```
mimoo, line 401: use `?` instead of a `match`

kimchi/src/circuits/domains.rs
mimoo: I would suggest errors like "construction of domain d2 did not work as intended"

kimchi/src/circuits/domains.rs
```diff
@@ -21,24 +23,34 @@ impl<F: FftField> EvaluationDomains<F> {
     /// and `d8` (of size `8n`). If generator of `d8` is `g`, the generator
     /// of `d4` is `g^2`, the generator of `d2` is `g^4`, and the generator of `d1` is `g^8`.
     // TODO(mimoo): should we instead panic/return an error if any of these return None?
-    pub fn create(n: usize) -> Option<Self> {
-        let n = Domain::<F>::compute_size_of_domain(n)?;
+    pub fn create(n: usize) -> Result<Self, ProverError> {
+        let n = Domain::<F>::compute_size_of_domain(n)
+            .ok_or(ProverError::Prover("couldn't compute size of domain"))
+            .unwrap();
```
mimoo, line 29: "could not compute size of domain {n}"

mimoo (commented):
Thanks for the PR :) going in the right direction, left you a few comments

kimchi/src/circuits/domains.rs
mimoo: yeah I do this in places with ProverError or VerifierError. See for example https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/error.rs#L15 (you should check the documentation for thiserror also)

mimoo: perhaps these errors should be a `ProverError::IndexGeneration` or something like that

kimchi/src/circuits/domains.rs
mimoo: the `ok_or` turn your Option into an Result though

kimchi/src/circuits/constraints.rs
```diff
@@ -654,7 +658,9 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
         // ------
 
         let lookup_constraint_system =
-            LookupConstraintSystem::create(&gates, lookup_tables, &domain).ok()?;
+            LookupConstraintSystem::create(&gates, lookup_tables, &domain)
```
Trivo25, line 661: @mimoo  `LookupConstraintSystem::create()` returns a somewhat weird combination of `Result<Option<Self>, LookupError>` - not sure if it would be better to return a `Result<LookupConstraintSystem, LookupError>` instead

mimoo: yeah that would be better

mimoo: actually, sorry, that looks normal to me (albeit a bit weird). Since we don't know at this point if lookups will be used. The function analyzes the circuit and returns `None` if no gates are used that make use of lookups.

mimoo: so here just doing `ok()?` is not going to work because you have two different error types. Not sure what's the best way forward:

1. create a `SetupError` error type?
2. create an enum `ProverError::Lookup(LookupError)` and use `map_err` to convert to that
3. something else

I would personally go for 1. and convert the error of `LookupConstraintSystem::create` to such an error type. Others might have some opinions (cc @mrmr1993 @querolita @jspada)

Trivo25: > actually, sorry, that looks normal to me (albeit a bit weird). Since we don't know at this point if lookups will be used. The function analyzes the circuit and returns `None` if no gates are used that make use of lookups.

ye thats right, I saw that after I was looking into as well

mimoo: btw I just hit that thing, and it's going to be really useful to return an error here instead of returning `None` (which doesn't tell you where the error happened)

Trivo25: > btw I just hit that unwrap. It's going to be really useful to return an error here instead of returning None (which doesn't tell you where the error happened)

 so you think its worth it re-writing the `Result<Option<Self>, LookupError>` from `LookupConstraintSystem::create()` to just return a normal `Result<LookupConstraintSystem, LookupError>` instead of the Option `None` then? Can do it that way if that's how you want to deal with the nested `Result<Option, _>`

mimoo: oh no, the `Option` is important to keep, I meant more generally the constraint system (not the lookup constraint system) doesn't return an error and we end up unwrapping an option if we reach that case (which doesn't tell you why it didn't work)

chee-chyuan:
hi @Trivo25 just wondering the progress on this. will be relying on this pr for a change that i need to make

Trivo25:
> hi @Trivo25 just wondering the progress on this. will be relying on this pr for a change that i need to make

Oh sorry, thats awkward.. things been busy and I completely forgot about it lol I will finish it asap! Thanks for the reminder

Trivo25:
cc @mimoo finally got around it! adjusted the `LookupConstraintSystem` to return an error instead of panicking and brought everything up to date with master so the CS etc now returns an `Result<Option, Error>` instead of only the option

kimchi/src/circuits/constraints.rs
```diff
@@ -583,9 +587,11 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
         //
         // Lookup
         // ------
-
         let lookup_constraint_system =
-            LookupConstraintSystem::create(&gates, lookup_tables, &domain).ok()?;
+            match LookupConstraintSystem::create(&gates, lookup_tables, &domain) {
+                Ok(ok) => ok,
+                Err(_) => return Err(ProverError::SetupError),
+            };
```
mimoo, line 594: can you rewrite this as a one-liner as well as preserve the error when you bubble it up? Something like

```rust
let lookup_constraint_sytem = LookupConstraintSystem::create(&gates, lookup_tables, &domain).map_err(|e| ProverError::SetupError(e))?;
```
```

File: kimchi/src/circuits/constraints.rs [hunk 7]
```diff
@@ -403,7 +406,7 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
         fr_sponge_params: ArithmeticSpongeParams<F>,
         public: usize,
         precomputations: Option<Arc<DomainConstantEvaluations<F>>>,
-    ) -> Option<Self> {
+    ) -> Result<Option<Self>, ProverError> {
```
mimoo, line 409: why not `Result<Self, ProverError>`? Is there a usecase for `Ok(None)` here?

kimchi/src/circuits/domains.rs
```diff
@@ -21,24 +23,33 @@ impl<F: FftField> EvaluationDomains<F> {
...
     /// of `d4` is `g^2`, the generator of `d2` is `g^4`, and the generator of `d1` is `g^8`.
     // TODO(mimoo): should we instead panic/return an error if any of these return None?
-    pub fn create(n: usize) -> Option<Self> {
-        let n = Domain::<F>::compute_size_of_domain(n)?;
+    pub fn create(n: usize) -> Result<Self, ProverError> {
+        let n = Domain::<F>::compute_size_of_domain(n)
+            .ok_or(ProverError::Prover("could not compute size of domain n"))?;
 
-        let d1 = Domain::<F>::new(n)?;
+        let d1 = Domain::<F>::new(n).ok_or(ProverError::Prover(
```
mimoo, line 30: I don't think these should be `ProverError` as we're doing setup stuff here. Maybe worth creating a `SetupError`?

mimoo (commented):
Hey! Nice :) I think it looks great except a few things I left you some comments on

kimchi/src/circuits/constraints.rs
Trivo25: true that, seems like the Option is not needed at all in that case

kimchi/src/circuits/domains.rs
```diff
@@ -22,25 +22,25 @@ impl<F: FftField> EvaluationDomains<F> {
...
     /// and `d8` (of size `8n`). If generator of `d8` is `g`, the generator
     /// of `d4` is `g^2`, the generator of `d2` is `g^4`, and the generator of `d1` is `g^8`.
-    // TODO(mimoo): should we instead panic/return an error if any of these return None?
-    pub fn create(n: usize) -> Result<Self, ProverError> {
-        let n = Domain::<F>::compute_size_of_domain(n)
-            .ok_or(ProverError::Prover("could not compute size of domain n"))?;
+    pub fn create(n: usize) -> Result<Self, SetupError> {
+        let n = Domain::<F>::compute_size_of_domain(n).ok_or(SetupError::DomainCreation(
+            "could not compute size of domain",
+        ))?;
```
mimoo, line 28: nit: it'd be nice to have the value of `n` within the error

mimoo: (same for the others)

kimchi/src/circuits/constraints.rs
```diff
@@ -248,11 +251,13 @@ pub enum GateError {
...
 
 /// Represents an error found when computing the lookup constraint system
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum LookupError {
     /// One of the lookup tables has columns of different lengths
+    #[error("Inconsisten table length")]
     InconsistentTableLength,
     /// The combined lookup table is larger than allowed by the domain size
+    #[error("Lookup table too long")]
```
mimoo, line 260: nit: would have been nice to also display the values `length` and `maximum_allowed` in the error message

kimchi/src/circuits/constraints.rs
```diff
@@ -412,6 +417,7 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
         //~    compute the smallest subgroup of the field that
         //~    has order greater or equal to `n + ZK_ROWS` elements.
         let domain = EvaluationDomains::<F>::create(gates.len() + ZK_ROWS as usize)?;
+
         assert!(domain.d1.size > ZK_ROWS);
```
mimoo, line 421: since we're returning an error, we could now return an error here instead of an assert

kimchi/src/error.rs
```diff
@@ -34,3 +34,13 @@ pub enum VerifyError {
     #[error("lookup used in circuit, but proof is missing lookup commitments")]
     LookupCommitmentMissing,
 }
+
+/// Errors that can arise when preparing the setup
+#[derive(Error, Debug, Clone)]
+pub enum SetupError {
+    #[error("the domain could not be constructed: {0}")]
```
mimoo, line 41: this error message doesn't seem relevant

kimchi/src/prover_index.rs
```diff
@@ -107,9 +107,10 @@ pub mod testing {
         lookup_tables: Vec<LookupTable<Fp>>,
     ) -> ProverIndex<Affine> {
         let fp_sponge_params = oracle::pasta::fp_kimchi::params();
+
+        // not sure if theres a smarter way instead of the double unwrap, but should be fine in the test
```
mimoo, line 111: delete

mimoo (approved):
This is great! Thank-you so much for pushing this to completion :) I'll merge this one, as the comments I made are minor now and can be addressed in another PR

kimchi/src/error.rs
```diff
@@ -34,3 +34,13 @@ pub enum VerifyError {
...
+    ConstraintSystem(String),
```
mimoo, line 42: I suggest using `LookupError(LookupError)`

PR #492, opened 2022-04-02
Jspada/commitments combined and improve

jspada:
This combines the two book sections entitled "Commitments" into one comprehensive section.


jspada:
> give $\mathsf{commit}(x, r)$ a name like $c$ or something short

Oh, I was trying to be consistent with Izaak's use of `commit`.  Also, I think he uses `c = commit(...)` below for shorthand already.  Is this sufficient or still want it changed?

jspada:
> It is **not** a one way function, but is considered to be so given the number of hardness assumptions relying on discrete logs. So in practice, it looks like a OWF and we use it like so, but they are not.

Awesome- thanks! Updating..

utils/src/field_helpers.rs
```diff
@@ -1,4 +1,4 @@
-use ark_ff::{Field, PrimeField};
```
mimoo, line 1: unrelated?

book/src/plonk/commitments.md
```diff
@@ -1,60 +0,0 @@
-# Commitments
-
-## What's a commitment
-
-A commitment is a cryptographic primitive that allows you to lock "something" (perhaps a number) without revealing it (yet). This will produce a **commitment**. Later, someone can ask you to reveal the "something". After you do, then can verify that it is linked to the commitment by **opening** it.
-
-<img src="../../img/commitment.png" width="300px">
```
mimoo, line 7: I thought that was a helpful diagram o.o

book/src/plonk/commitments.md
```diff
@@ -1,60 +0,0 @@
...
-
-Commitments are usually defined with these two properties:
-
-* **hiding**: it does not reveal what is commited
-* **binding**: you can't open a commitment to a different value (think second pre-image)
-
-## A simple example
```
mimoo, line 14: why remove this section?

book/src/fundamentals/zkbook_commitment.md
```diff
@@ -25,40 +25,133 @@ $$
...
+
+On the other hand, Pedersen commitments are **unconditionally hiding**, as there is no way (even with a magic computer) to reveal what $x$ is without knowing $r$.
+Lack of perfect binding is the reason why most of the "proofs" we will see later in this book are not referred to as proofs, but instead are referred to as
+**arguments** of knowledge (although we may care little about this distinction). Just remember that you need perfect binding to be called a proof.
+
+> Interestingly, it is impossible to have a commitment scheme that has both perfect hiding and perfect binding.
+
+To recap, in cryptography the following distinctions are important
+
+- **Perfect.** The property that an algorithm is statistically sound without hardness assumptions, also known as unconditional or statistical soundness.
```
mimoo, line 92: cc @imeckler that's what we were talking about the other day. perfect = statistical

utils/src/field_helpers.rs
jspada: good spot.

book/src/plonk/commitments.md
jspada: They don't load for me.  I can put it back in or, if you like, I can convert if there's a new way in latex/md?

utils/src/field_helpers.rs
jspada: Fixed!

book/src/plonk/commitments.md
mimoo: I was commenting on the simple example section, which talks about SHA-256 :o

book/src/plonk/commitments.md
mimoo: commitment

ah, I might have messed up some images when I moved the book to this repo, it was this diagram

book/src/plonk/commitments.md
jspada: Ah, because Izzy already had one there.

jspada: Under "Simple commitments". Pretty comprehensive and clean explanation with hash function.

book/src/plonk/commitments.md
jspada: Cool - I'll put it back in the simple section.  Does it look the way you want now (the one above)?

book/src/plonk/commitments.md
mimoo: ah indeed! Nice :)

querolita:
> Oh, I was trying to be consistent with Izaak's use of commit. Also, I think he uses c = commit(...) below for shorthand already. Is this sufficient or still want it changed?

It's fine that you call the commitment commit(x,r) instead of just c, but then the open algorithm should receive as input the commitment commit(x,r) ( you open a commitment, optionally receive as input the message x, and in any case output the randomness r, called the opening)

book/src/fundamentals/zkbook_commitment.md
```diff
@@ -36,7 +36,7 @@ If the hash function is one-way, then this commitment is also hiding.
 ## Algebraic and homomorphic commitments
 
 Instead of a cryptographic hash function, we can use elliptic curve cryptography to construct a commitment scheme.  Here elliptic curve scalar
-multiplication is used as the one way function.  Suppose we have
+multiplication is used like a one way function based on some hardness assumptions.  Suppose we have
```
mimoo, line 39: maybe based on the hardness assumption of the ECDLOG?

mimoo (approved):
LGTM

querolita:
> open(x, r, c) right?

well, either verify(x,r,c) and ( open(x,c) -> r or open(c) -> x,r )

querolita (approved):
Well of course, I accept the PR. I left a few minor comments here and there, but it's fine :D

book/src/fundamentals/zkbook_commitment.md
querolita: yes exactly

jspada:
> Also, the fact that they are algebraic does not necessarily give you homomorphism "for free". Only _some_ of them, like Pedersen, do. This is just a nonsense example, but if the commitment consisted on performing the operation x^2 , and you had com(a) and com(b), the com(a+b) is not simply com(a) + com(b), you would also need the term 2_a_b, because (a+b)^2 = a^2 + b^2 + 2ab

Good point -- I've updated text to make it clear by removing the word "consequently"

book/src/fundamentals/zkbook_commitment.md
jspada: Nice -- I'll move this specific up from the section below.

book/src/plonk/commitments.md
jspada: I've made a latex version that support scaling and fixes the issue with the above diagram ("open" box should be "verify")

PR #493, opened 2022-04-03
Turshi instruction logic

querolita:
This PR provides a functioning version of the Cairo instruction logics gate constraints. It still doesn't have the wiring between I/O registers and public input, but it should be easily addable. This PR does not contain the memory logics, which has been moved to a different PR for readability and ease of review. In particular, some parts of the memory logics are subject to changes for lookup tables, so that part is less stable. It needs about 4 rows per Cairo instruction in a 15-wire kimchi. The delta between this PR and the one for the Cairo runner are the following: 
- Added one field to `CairoProgram` that gives the list of instructions of the execution.
- `turshi.rs`: includes 
  - 4 Cairo gates definitions; 
  -  a witness generator from an execution of the runner; 
  -  a gadget generation from a witness, 
  -  constraints for the operands, operations, registers, instruction checks, registers updates
  - a checking function that replicates the constraints expressions in field format
  - a verification function that evaluates the constraints directly
  - one simple check using the same program as the runner's tests
- some changes needed to be added to prover, expr, verifier, gate, wires,  index, for it to work. But further modifications will have to be carried out

These are the following types of gates (more or less, in gadget order). 
- `CairoClaim`: takes care of the initial and final registers claim
- `CairoInstruction`: takes care of the logics of the Cairo instruction.
- `CairoFlags`: performs registers update checks using the values of the flags
- `CairoTransition`: performs the check of a correct transition between consecutive instructions
- `Zero`: does nothing, it's just a row that holds 3 values. It would be great to merge this one somewhere else, but it doesn't seem to fit. So it is a `Zero` gate for now (used to be `CairoAuxiliary`, but I think it can be `Zero` like in ChaCha).

Some remarks/TODOs for future versions:
- I have to write the wiring of the Cairo rows. In the comments I wrote some sort of table of the current layout and it explains what columns should be copied and from where. But this is still pending.
- I want to create full tests that involve both the prover and the verifier, not only these "gates checks"
- (Issue [282](https://github.com/o1-labs/proof-systems/issues/282))Think about a different number of columns and how much it would benefit Cairo and other complex gates. In this approach, if `n` is the number of instructions, the full gadget for the program takes `4n-1` rows.
- (Issue [294](https://github.com/o1-labs/proof-systems/issues/294))The runner still needs to address the hints independently: meaning, run the program `backwards` and figure out the intermediate 2 or 3 values of the memory that need to be set for this nondeterministic feature of the memory


cairo/src/runner.rs
```diff
@@ -310,21 +513,32 @@ impl<'a, F: Field> CairoProgram<'a, F> {
...
         prog
     }
 
     /// Outputs the total number of steps of the execution carried out by the runner
-    pub fn get_steps(&self) -> F {
+    pub fn steps(&self) -> F {
         self.steps
     }
 
     /// Outputs the final value of the pointers after the execution carried out by the runner
```
jspada, line 527: Fix comment.  Should be "the initial values of the ..."

jspada:
Awesome work-- Looks great!

jspada (approved):
Approving,  but please see comments/feedback and wait for others to review also before merging.

cairo/src/word.rs
```diff
@@ -28,10 +28,26 @@ impl<F: Field> CairoWord<F> {
     pub fn word(&self) -> F {
         self.0
     }
+
+    /*
```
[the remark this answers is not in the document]
querolita: Sure!

cairo/src/helper.rs
```diff
@@ -1,30 +1,30 @@
...
-use o1_utils::FieldHelpers;
+use o1_utils::{field_helpers::i128_to_field, FieldHelpers};
 
 //(TODO move to utils inside FieldHelpers)
 
 /// Field element helpers for Cairo
 pub trait CairoFieldHelpers<F> {
     /// Return field element as byte, if it fits. Otherwise returns least significant byte
-    fn least_significant_byte(self) -> u8;
+    fn lsb(self) -> u8;
```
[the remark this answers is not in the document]
querolita: :)

cairo/src/word.rs
```diff
@@ -209,53 +214,49 @@ impl<F: Field> Decomposition<F> for CairoWord<F> {
...
-        self.f_op0_fp().least_significant_byte()
+        self.f_op0_fp().lsb()
     }
 
     fn op1_src(&self) -> u8 {
         // op1_src = 4*fOP1_AP + 2*fOP1_FP + fOP1_VAL
-        2 * (2 * self.f_op1_ap().least_significant_byte()
-            + self.f_op1_fp().least_significant_byte())
-            + self.f_op1_val().least_significant_byte()
+        2 * (2 * self.f_op1_ap().lsb() + self.f_op1_fp().lsb()) + self.f_op1_val().lsb()
```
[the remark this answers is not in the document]
querolita: it's already there, isn't it? this one:
`// opcode = 4*fOPC_AEQ + 2*fOPC_RET + fOPC_CALL`

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -592,6 +592,7 @@ mod tests {
                 aggreg: F::rand(rng),
                 table: F::rand(rng),
             }),
+            cairo_selector: [F::zero(); crate::circuits::polynomials::turshi::CIRCUIT_GATE_COUNT],
```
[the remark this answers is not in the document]
querolita: sure, like above

kimchi/src/tests/turshi.rs
```diff
@@ -0,0 +1,178 @@
...
+    //view_witness(&witness);
+
+    // Create the Cairo circuit
+    let num = prog.trace().len();
+    let circuit = CircuitGate::<F>::create_cairo_gadget(0, num);
+
+    // Verify each gate
+    let mut row = 0;
+    for gate in circuit {
+        let res = gate.ensure_cairo_gate(row, &witness);
```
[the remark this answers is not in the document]
querolita: all of the `assert_eq` are handled within `ensure_cairo_gate` and if there's an error, `res` will be an error as well

cairo/src/helper.rs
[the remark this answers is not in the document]
mimoo: `of_lsb` is really an ocaml thing. in Rust just I've seen `lsb`, or `to_lsb` (or `into_lsb` if you're consuming `self` and `self` is not `Copy`)

cairo/src/helper.rs
```diff
@@ -43,22 +43,14 @@ impl<F: Field> CairoFieldHelpers<F> for F {
...
     }
 
     fn vec_to_field(vec: &[i128]) -> Vec<F> {
-        vec.iter()
-            .map(|i| {
-                if *i < 0 {
-                    -F::from((-(*i)) as u64)
-                } else {
-                    F::from((*i) as u64)
-                }
```
mimoo, line 59: note that this function exists here: https://github.com/o1-labs/proof-systems/blob/master/utils/src/field_helpers.rs#L90

cairo/src/runner.rs
```diff
@@ -3,9 +3,12 @@
 
 use crate::flags::*;
 use crate::memory::CairoMemory;
-use crate::word::{CairoWord, Decomposition};
+use crate::word::{CairoWord, FlagBits, FlagSets, Offsets};
 use ark_ff::Field;
 
+/// Number of memory accesses per instruction
+pub const ACC_PER_INS: usize = 4;
```
mimoo, line 10: where is this used?

kimchi/src/circuits/constraints.rs
```diff
@@ -146,6 +147,13 @@ pub struct ConstraintSystem<F: FftField> {
     #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub endomul_scalar8: E<F, D<F>>,
 
+    /// Cairo constraint selector polynomials
+    #[serde_as(as = "[o1_utils::serialization::SerdeAs; turshi::CIRCUIT_GATE_COUNT]")]
+    pub cairo: [DP<F>; turshi::CIRCUIT_GATE_COUNT],
+    /// Cairo selector evaluations over domain.d8
+    #[serde_as(as = "[o1_utils::serialization::SerdeAs; turshi::CIRCUIT_GATE_COUNT]")]
+    pub cairo8: [E<F, D<F>>; turshi::CIRCUIT_GATE_COUNT],
```
mimoo, line 155: since Cairo is optional, I think these should live under some `Option<Cairo>` field

kimchi/src/circuits/constraints.rs
```diff
@@ -649,6 +657,45 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
...
+            E::<F, D<F>>::from_vec_and_domain(
+                gates
+                    .iter()
+                    .map(|gate| if gate.typ == g { F::one() } else { F::zero() })
+                    .collect(),
+                domain.d1,
+            )
+            .interpolate()
+            .evaluate_over_domain_by_ref(domain.d8)
+        });
```
mimoo, line 697: move the cairo initialization to its own function which you can call here :o

kimchi/src/linearization.rs
```diff
@@ -28,9 +29,9 @@ pub fn constraints_expr<F: FftField + SquareRootField>(
     let mut powers_of_alpha = Alphas::<F>::default();
 
     // gates
-    let highest_constraints = VarbaseMul::<F>::CONSTRAINTS;
+    let highest_constraints = Instruction::<F>::CONSTRAINTS;
     powers_of_alpha.register(
-        ArgumentType::Gate(GateType::VarBaseMul),
+        ArgumentType::Gate(GateType::CairoInstruction),
         highest_constraints,
     );
```
mimoo, line 36: can we do it like the lookup stuff? This way if we don't use cairo the alpha ranges are not affected

kimchi/src/proof.rs
```diff
@@ -36,6 +36,8 @@ pub struct ProofEvaluations<Field> {
     pub generic_selector: Field,
     /// evaluation of the poseidon selector polynomial
     pub poseidon_selector: Field,
+    /// evaluation of the cairo selector polynomial
+    pub cairo_selector: [Field; crate::circuits::polynomials::turshi::CIRCUIT_GATE_COUNT],
```
mimoo, line 40: import turshi so you can write `turshi::CIRCUIT_GATE_COUNT`

mimoo: like lookup this should be an `Option`

mimoo (changes requested):
Could you separate the code that touches the proof system in a separate PR? I think they should be reviewed separately :o

kimchi/src/tests/turshi.rs
```diff
@@ -0,0 +1,178 @@
...
+        44,   // input
+        */
+    ];
+
+    let mut mem = CairoMemory::<F>::new(F::vec_to_field(&instrs));
+    // TODO(querolita): Need to know how to find out (hints)
+    mem.write(F::from(21u32), F::from(41u32)); // beginning of outputs
+    mem.write(F::from(22u32), F::from(44u32)); // end of outputs
+    mem.write(F::from(23u32), F::from(44u32)); //end of program
+    let prog = CairoProgram::new(&mut mem, 5, 24);
```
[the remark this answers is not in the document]
querolita: Every time you write to memory, the length gets automatically extended. Meaning, before line 72 the length was 21 (from 0 to 20) and then in line 75 the length of memory is already 24. the argument `24` in the creator of the program corresponds to the initial allocation pointer. This means, that I think you're right and this argument could be "omitted": the initial allocation pointer should be the first empty address of the memory, so it must be 24 in this case.

cairo/src/runner.rs
querolita: thanks!

cairo/src/helper.rs
querolita: You're right, good catch, I fixed this in the [i128 field helper](https://github.com/o1-labs/proof-systems/pull/493/commits/7dca96d7b3de965450947141c41ff95412def4e8) commit. Actually, I suggested that in that field_helper we could actually perform the casting using `u32` instead of `u64` to convert `i32` to fields.

cairo/src/runner.rs
querolita: Actually, this is being used only in the memory constraints, but I must have missed to remove this one from this PR.

kimchi/src/circuits/constraints.rs
querolita: amazing 😃

utils/src/field_helpers.rs
```diff
@@ -95,6 +95,14 @@ pub fn i32_to_field<F: From<u64> + Neg<Output = F>>(i: i32) -> F {
     }
 }
 
+pub fn i128_to_field<F: From<u128> + Neg<Output = F>>(i: i128) -> F {
+    if i >= 0 {
+        F::from(i as u128)
+    } else {
+        -F::from(-i as u128)
```
mimoo, line 102: this still looks dangerous to me :o casting `i128` to `u128`

jspada: The signed version should be about half the size of the unsigned version, but to be safe maybe use the `From` or `TryFrom` trait?

Example `-F::from(u128::from(i))`

Or guard with `std::u128::MAX`

cairo/src/word.rs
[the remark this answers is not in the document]
jspada: I just mean to make the code's form match the comment.

kimchi/src/tests/turshi.rs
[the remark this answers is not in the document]
jspada: Hmm... yeah, this is interesting.  It might be cool if the memory thing was "smart" and you can just fill it up with the initial memory and the `CairoProgram::new(&mut mem)` was smart enough to figure out what to tuse for `pc` and `ap`.  Maybe a thing for future though and not this PR :-). I was just concerned if there could be a bad offset there that could impact the test case

utils/src/field_helpers.rs
mimoo: there are no From implementations usually when the conversion can be dangerous

mimoo: you could use TryFrom and return an error (or panic if you know that this will never be given as a malicious input)

kimchi/src/linearization.rs
querolita: Good point, I will give it a thought

kimchi/src/proof.rs
querolita: I see, like above with the `cairo` and `cairo8`

utils/src/field_helpers.rs
querolita: Oh this chat is very enlightening... Nonetheless, this edge-case you're talking about is not the one we have here right? There you're saying:
`    let a = -i128::MAX;
    let b = a as u128;`
And then `a = -170141183460469231731687303715884105727`
and `b=170141183460469231731687303715884105729` 

But the conditional in the helper is instead doing this:
`    let a = -i128::MAX;
    let b = -a as u128;`
And then `a = -170141183460469231731687303715884105727`
and `b=170141183460469231731687303715884105727` 

Let's just use some smaller data so it is easier to imagine what's going on:
`let a = -i8::MAX;`
`let b = a as u8;`

In this case, `a = 0b1000_0001 as i8 => -127` and `b = 0b1000_0001 as u8 =>129`

But if we perform `b = -a as u8` , first, it negates obtaining +127 and then 127 is expressed as u8 with `0b0111_1111`.

Nonetheless, I think what you meant instead is the edge-case where `a=-i8::MIN`, meaning -128. In that case, performing `b=-a as u8` effectively tries to perform -(-128) which does not fit inside a `u8`, even if later on +128 fits inside `u8` just fine. 

So you're right! There's that issue when you try to convert the minimum value of the integer range. I am using this helper only to convert a bytecode program to field elements in the tests, without having the user write directly field elements, and I was not using `i128::MIN` for that. But if a helper is meant to work in any scenario, then I agree that this must be fixed or avoided somehow. Nonetheless, this should be upgraded so that indeed one could write field elements directly, as in the playground. I will fix this in the next commit of the pr.

kimchi/src/circuits/constraints.rs
querolita: I am not sure I get it. None of the other gates have an independent function for this, do they?

querolita:
Update: I have addressed @mimoo 's comments regarding the optional feature of the cairo gates, so I wrapped these fields in optional types. This way, powers of alpha are not altered if the circuit has no cairo gates. For this, I followed the same strategy as chacha, equipping the `constraints_expr` and `expr_linearization` with a boolean type cairo. OCaml types are updated to be optional as well, and the alphas have been updated on the specs.

This way, cairo gates have no effect on the rest of the proof system, but if you still think it is worth moving these things to another PR, that can be done.

utils/src/field_helpers.rs
mimoo: oh you're right, I actually didn't think of the edge case properly. So indeed this causes an integer overflow:

```rust
let a = i128::MIN;
let b = -a as u128;
```

 https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5770463ba987cd28eb7b0647fa397931

cairo/src/helper.rs
```diff
@@ -1,34 +1,31 @@
...
 /// Field element helpers for Cairo
 pub trait CairoFieldHelpers<F> {
     /// Return field element as byte, if it fits. Otherwise returns least significant byte
-    fn lsb(self) -> u8;
+    fn of_lsb(self) -> u8;
 
-    /// Return pos-th 16-bit chunk as another field element
-    fn chunk_u16(self, pos: usize) -> F;
+    /// Return `pos`-th 16-bit chunk as another field element
+    fn of_u16_chunk(self, pos: usize) -> F;
```
mimoo, line 14: please let's avoid `of_` which is an OCaml thing :D

book/src/specs/kimchi.md
```diff
@@ -1297,7 +1297,7 @@ pub struct ProofEvaluations<Field> {
     /// evaluation of the poseidon selector polynomial
     pub poseidon_selector: Field,
     /// evaluation of the cairo selector polynomial
-    pub cairo_selector: [Field; crate::circuits::polynomials::turshi::CIRCUIT_GATE_COUNT],
+    pub cairo_selector: Option<[Field; crate::circuits::polynomials::turshi::CIRCUIT_GATE_COUNT]>,
```
mimoo, line 1300: there's still this path that could be imported :o

mimoo:
> if you still think it is worth moving these things to another PR, that can be done.

yeah sorry can you do it ^^? I'd really like to be able to review the parts that affect kimchi in isolation so we make sure that it won't affect the hardfork

cairo/src/helper.rs
querolita: Ah ok! Both sounded good to me, I didn't know there was a standard

book/src/specs/kimchi.md
querolita: oh there! your eyes are well trained 👀

querolita:
> yeah sorry can you do it ^^? I'd really like to be able to review the parts that affect kimchi in isolation so we make sure that it won't affect the hardfork

sgtm, I'll do that then :)

utils/src/field_helpers.rs
mimoo: makes me think that the function above (`i32_to_field`) could also be simplified as a one-liner `F::from(i.abs() as u64)`

cairo/src/helper.rs
mimoo: yeah I think I replied on the comment there. But in rust we use `from` and `into` and `to` generally. (Usually `into` is used when `self` is consumed and does not implement `Copy`)

kimchi/src/circuits/constraints.rs
mimoo: if you look at lookup it does

mimoo: see https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/constraints.rs#L345

cairo/src/helper.rs
querolita: oh I was not aware of the `into` distinction for consuming the `self`! nice lesson

kimchi/src/circuits/constraints.rs
querolita: Oh I see, for lookup yeah... I was looking at chacha for instance. But fine, I can make it separately.

cairo/src/word.rs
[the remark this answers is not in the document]
querolita: oh well you mean to write the 4 directly and so... Yeah well this was to avoid some multiplications here and there

kimchi/src/verifier.rs
```diff
@@ -642,6 +642,9 @@ where
                             ChaCha1 => &index.chacha_comm.as_ref().unwrap()[1],
                             ChaCha2 => &index.chacha_comm.as_ref().unwrap()[2],
                             ChaChaFinal => &index.chacha_comm.as_ref().unwrap()[3],
+                            CairoClaim | CairoInstruction | CairoFlags | CairoTransition => {
+                                panic!("TODO")
```
mimoo, line 646: there's also `unimplemented!()` as well.

mimoo (approved):
Impressive work :)

kimchi/src/alphas.rs
```diff
@@ -332,7 +332,7 @@ mod tests {
                 .as_ref()
                 .map(|lcs| &lcs.configuration),
         );
-
+        println!("{:?}", powers_of_alpha);
```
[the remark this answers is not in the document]
querolita: absolutely right, that was a print I wrote for debugging, and seems like I forgot to remove it 😵‍💫

kimchi/src/circuits/constraints.rs
```diff
@@ -461,6 +477,46 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
+            matches!(
+                gate.typ,
+                GateType::CairoClaim
+                    | GateType::CairoInstruction
+                    | GateType::CairoFlags
+                    | GateType::CairoTransition
+            )
+        });
+        if !has_cairo_gate {
+            None
```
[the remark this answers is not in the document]
querolita: Oh you're right that's way cleaner! I will modify this in the `turshi_pfsys` branch 😃

kimchi/src/verifier.rs
querolita: indeed 😅

kimchi/src/circuits/polynomials/turshi.rs
```diff
@@ -0,0 +1,1025 @@
...
+        }
+    }
+
+    /// Gadget generator of the whole cairo circuits from an absolute row and number of instructions
+    pub fn create_cairo_gadget(
+        // the absolute row in the circuit
+        row: usize,
+        // number of instructions
+        num: usize,
+    ) -> Vec<Self> {
```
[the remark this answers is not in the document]
querolita: oh really good point! I was only foreseeing the use case where you only had a huge cairo gadget corresponding to a program, so it didn't have to interact with others. Nonetheless, adding next row is super simple to add and can be useful in the future! 👍🏻

kimchi/src/circuits/polynomials/turshi.rs
```diff
@@ -0,0 +1,1025 @@
...
+        // the final instruction
+        let ins_gate = Wire::new(row + 4 * last + 1);
+        let aux_gate = Wire::new(row + 4 * last + 2);
+        gates.push(CircuitGate::create_cairo_instruction(ins_gate));
+        gates.push(CircuitGate::zero(aux_gate));
+
+        gates
+    }
+
+    /// verifies that the Cairo gate constraints are solved by the witness depending on its type
```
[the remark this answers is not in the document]
querolita: Oh that's fine, but only for `ensure_`,  since `verify_` is the one called from `gates.rs` for all other gates. Nonetheless, I can also create a `mod witness` for all things related to creating the witness, again to put some order inside `turshi.rs`, which is too long otherwise.

kimchi/src/circuits/expr.rs
```diff
@@ -411,6 +411,10 @@ impl Variable {
             LookupTable => l.map(|l| l.table),
             Index(GateType::Poseidon) => Ok(evals.poseidon_selector),
             Index(GateType::Generic) => Ok(evals.generic_selector),
+            Index(GateType::CairoClaim)
+            | Index(GateType::CairoInstruction)
+            | Index(GateType::CairoFlags)
+            | Index(GateType::CairoTransition) => panic!("TODO"),
```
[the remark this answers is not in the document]
querolita: oh cute, didn't hear about this one

cairo/src/runner.rs
```diff
@@ -52,18 +79,178 @@ impl<F: Field> Default for CairoContext<F> {
...
+    word: CairoWord<F>,
+    /// pointers
+    ptrs: CairoState<F>,
+    /// auxiliary variables
+    vars: CairoContext<F>,
+}
+
+impl<F: Field> CairoInstruction<F> {
+    /// Creates a [CairoInstruction]
+    pub fn new(word: &CairoWord<F>, ptrs: &CairoState<F>, vars: &CairoContext<F>) -> Self {
```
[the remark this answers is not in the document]
querolita: Good point, yes, `CairoContext` is copiable as well, and `CairoWord`

cairo/src/runner.rs
```diff
@@ -427,42 +639,42 @@ mod tests {
...
+            F::from(0x48127ffb7fff8000u64),
+            F::from(0x1104800180018000u64),
+            F::from(-11),
+            F::from(0x48127ff87fff8000u64),
+            F::from(0x1104800180018000u64),
+            -F::from(14),
+            F::from(0x48127ff67fff8000u64),
+            F::from(0x1104800180018000u64),
+            -F::from(17),
+            F::from(0x208b7fff7fff7ffeu64),
```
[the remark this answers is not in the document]
querolita: Oh! I was afraid that it wouldn't be able to convert from `i64` directly, it seems like it is indeed implemented. Thanks for the tip!

querolita:
Merging Cairo instruction-related gates after addressing your comments (upcoming PRs about the proof system and the memory model)

poly-commitment/src/commitment.rs
```diff
@@ -613,7 +613,7 @@ impl<G: CommitmentCurve> SRS<G> {
     pub fn verify<EFqSponge, RNG>(
         &self,
         group_map: &G::Map,
-        batch: &mut Vec<BatchEvaluationProof<G, EFqSponge>>,
```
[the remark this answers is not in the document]
mimoo: yeah I started getting this error on my own PRs too, this is weird :D

kimchi/src/circuits/constraints.rs
```diff
@@ -461,6 +477,46 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
+        } else {
+            let cairo = array_init(|i| {
+                let g = match i {
+                    0 => GateType::CairoClaim,
+                    1 => GateType::CairoInstruction,
+                    2 => GateType::CairoFlags,
+                    3 => GateType::CairoTransition,
+                    _ => panic!("Invalid index"),
+                };
```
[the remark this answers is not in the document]
querolita: Yeah, combined with some tricks this helped reduce the function by 10 loc

PR #494, opened 2022-04-03
Compute linearization expression independently of domain size

mrmr1993:
This PR tweaks the representation of the normalized lagrange basis to be independent of the domain size.

This allows us to use the expression from OCaml without locking-in a particular choice of domain.

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -368,7 +368,7 @@ pub enum Expr<C> {
     VanishesOnLast4Rows,
     /// UnnormalizedLagrangeBasis(i) is
     /// (x^n - 1) / (x - omega^i)
-    UnnormalizedLagrangeBasis(usize),
+    UnnormalizedLagrangeBasis(i32),
     Pow(Box<Expr<C>>, u64),
     Cache(CacheId, Box<Expr<C>>),
 }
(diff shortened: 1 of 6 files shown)


kimchi/src/circuits/expr.rs
```diff
@@ -1396,10 +1397,19 @@ impl<F: FftField> Expr<F> {
...
-            Expr::UnnormalizedLagrangeBasis(i) => EvalResult::Evals {
-                domain: d,
-                evals: unnormalized_lagrange_evals(env.l0_1, *i, d, env),
-            },
+            Expr::UnnormalizedLagrangeBasis(i) => {
+                let i = {
+                    if *i >= 0 {
+                        *i as u32
+                    } else {
+                        (d1_size as u32) * (d as u32) - ((-*i) as u32)
```
mrmr1993, line 1405: TODO: the `d` factor is already accounted for in the function call below; it should be deleted here.

kimchi/src/circuits/expr.rs
```diff
@@ -368,7 +368,7 @@ pub enum Expr<C> {
     VanishesOnLast4Rows,
     /// UnnormalizedLagrangeBasis(i) is
     /// (x^n - 1) / (x - omega^i)
-    UnnormalizedLagrangeBasis(usize),
+    UnnormalizedLagrangeBasis(i32),
```
jspada, line 371: Not `u32`?

mrmr1993: That's what lets this PR work; otherwise, we have to fix the domain size.

mrmr1993:
Closing in favour of #579. I knew I'd already done this, but couldn't find the branch easily 🤦‍♂️

PR #498, opened 2022-04-04
[RFC] Integration of plookup in kimchi

mimoo:
looks like this: https://hackmd.io/LJQDwtJeSeKKg2zYBCw7-w?view

### DIFF
--- a/book/src/rfcs/3-lookup.md
+++ b/book/src/rfcs/3-lookup.md
@@ -1,0 +1,291 @@
+# RFC: Plookup in kimchi
+
+In 2020, [plookup](https://eprint.iacr.org/2020/315.pdf) showed how to create lookup proofs. Proofs that some witness values are part of a [lookup table](https://en.wikipedia.org/wiki/Lookup_table). Two years later, an independent team published [plonkup](https://eprint.iacr.org/2022/086) showing how to integrate Plookup into Plonk.
+
+This document specifies how we integrate plookup in kimchi. It assumes that the reader understands the basics behind plookup.
+
+## Overview
+
+We integrate plookup in kimchi with the following differences:
+
+* we snake-ify the sorted table instead of wrapping it around (see later)
+* we allow fixed-ahead-of-time linear combinations of columns of the queries we make
+* we only use a single table (XOR) at the moment of this writing
+* we allow several lookups (or queries) to be performed within the same row
+* zero-knowledgeness is added in a specific way (see later)
+
+The following document explains the protocol in more detail
+
+### Recap on the grand product argument of plookup
+
+As per the Plookup paper, the prover will have to compute three vectors:
+
+* $f$, the (secret) **query vector**, containing the witness values that the prover wants to prove are part of the lookup table.
+* $t$, the (public) **lookup table**.
+* $s$, the (secret) concatenation of $f$ and $t$, sorted by $t$ (where elements are listed in the order they are listed in $t$).
+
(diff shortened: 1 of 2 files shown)


book/src/rfcs/3-lookup.md
```diff
@@ -0,0 +1,291 @@
...
+
+Since we allow multiple queries per row, we define multiple **queries**, where each query is associated with a **lookup selector**. 
+
+At the moment of this writing, the `ChaCha` gates all perform $4$ queries in a row. Thus, $4$ is trivially the largest number of queries that happen in a row.
+
+**Important**: to make constraints work, this means that each row must make 4 queries. (Potentially some or all of them are dummy queries.)
+
+For example, the `ChaCha0`, `ChaCha1`, and `ChaCha2` gates will apply the following 4 XOR queries on the current and following rows:
+
+|   l   |   r   |   o    | -   |   l   |   r   |   o    | -   |   l   |   r   |   o    | -   |   l   |   r    |   o    |
```
querolita, line 134: This part with the table with 15 columns is a bit obscure to me in the sense that it is coincidentally the same number of kimchi columns so for me it was misleading. As I commented on the hackmd:

>  My understanding was that the XORs in the quarter rounds were checked against the lookup table with all of the XOR 4-bit entries but eventually they had to be linked to some witness cells in the rows somehow, but the way does not click in my head yet.

book/src/rfcs/3-lookup.md
```diff
@@ -0,0 +1,291 @@
...
+2. the prover can sort $s$ by $t$, so that the verifier doesn't have to do any sorting and can just rely on the commitment of the columns of $t$ (which the prover can evaluate in the protocol).
+
+We do the second one, but there is an edge-case: the combined $t$ entries can repeat.
+For some $i, l$ such that $i \neq l$, we might have
+
+$$
+t_0[i] + j t_1[i] + j^2 t_2[i] = t_0[l] + j t_1[l] + j^2 t_2[l]
+$$
+
+For example, if $f = \{1, 2, 2, 3\}$ and $t = \{2, 1, 2, 3\}$, then $\text{sorted}(f, t) = \{2, 2, 1, 1, 2, 3\}$ would be one way of sorting things out. But $\text{sorted}(f, t) = \{2, 2, 2, 1, 1, 3\}$ would be incorrect.
```
querolita, line 279: Here I really appreciate an example. Nonetheless, I got more questions from looking at it. I don't see the intuition of the "entries can repeat", and why the size of the sorted list has that size.

querolita (approved):
Beautiful explanation of plookups in kimchi. I still need to understand some parts better and connect to the full context, but the content of the writeup is great!

book/src/rfcs/3-lookup.md
mimoo: I answered there, but if you have an idea on how I can make this part clearer let me know :o I'll think about it

book/src/rfcs/3-lookup.md
mimoo: we can talk about it more during our 1:1, but quickly:

* entries can repeat because there is the "combination of the columns with `j` to make the lookup table a single column table" that might lead to this. If we sort `t` then this is not a problem, but we want the verifier to avoid doing this work (due to the verifier circuit) and so we want to keep `t` unsorted
* the sorted vector is the concatenation of `f` and `t`. Let's ignore `t`. `f` is the 4 query vectors, where each query vector is constructed from the witness columns (and thus spanning all the rows of the domain). So each row is a mix of dummy queries and real queries so that there are 4 queries per row. So you can see `f` as 4 columns, where each column behaves like a gate

PR #500, opened 2022-04-05
[easy][lookup] light cleanup

mimoo:
- added an `LookupCommitmentMissing` error instead of crashing if a commitment is missing in the proof
- added some spec comment


querolita (approved):
Great to address todos!

PR #505, opened 2022-04-05
[easy][specs][kimchi] update protocol diagram

mimoo:
- split the protocol diagram into section after discussion with @jspada 
- separate proof data structure into a separate section to easily refer to it

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -1213,7 +1213,7 @@ pub struct VerifierIndex<G: CommitmentCurve> {
 ```
 
 
-## Proof
+## Proof Construction & Verification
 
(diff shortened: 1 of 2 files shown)


book/src/specs/kimchi.md
```diff
@@ -1222,47 +1222,65 @@ For this reason, it can be useful to visualize the high-level interactive protoc
...
-A proof consists of:
-
-* a number of (hidden) polynomial commitments:
-  * the 15 registers/witness columns
-  * the permutation
-  * the quotient
-  * TODO: lookup
-  * TODO: public commitment is not here, but is in the sequence diagram
-* evaluations of these polynomials at two random points $\zeta$ and $\zeta \omega$
-* evaluations at the two random points of these additional polynomials:
```
querolita, line 1256: I liked the verbose explanation of the structures

mimoo:
> I am not a fan of removing the fine grained list of the proof components

The problem is that we need to keep this up to date with the code, and it's redundant information when we list the actual data structures after. I was thinking that ideally we should explain all of these things well in the comments of the structures we extract :o

PR #507, opened 2022-04-06
[lookup] refactor prover + spec

mimoo:
In this PR I do three things:

* I create a `LookupContext` struct in `prover.rs` to keep track of variables that are related to the lookup flow. This is so that we can track them across optional blocks (as lookup is an optional feature) instead of tracking a bunch of variables that pollute the scope
* I enclose each steps related to the lookup feature within a `if let Some(lcs) = &index.cs.lookup_constraint_system {`
* I add specification comments where relevant

(I tried to split the changes incrementally in commits)

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -3,12 +3,11 @@
 use crate::{
     circuits::{
         argument::{Argument, ArgumentType},
-        constraints::{LookupConstraintSystem, ZK_ROWS},
+        constraints::ZK_ROWS,
         expr::{l0_1, Constants, Environment, LookupEnvironment},
         gate::GateType,
         lookup::{
             self,
-            constraints::LookupConfiguration,
             lookups::LookupsUsed,
             tables::{combine_table_entry, CombinedEntry},
         },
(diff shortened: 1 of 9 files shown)


kimchi/src/prover.rs
```diff
@@ -45,7 +44,43 @@ use rayon::iter::{
...
+
+/// Contains variables needed for lookup in the prover algorithm.
+#[derive(Default)]
+struct LookupContext<G, F>
+where
+    G: CommitmentCurve,
+    F: FftField,
+{
+    /// The joint combiner used to join the columns of lookup tables
+    joint_combiner: Option<F>,
```
mrmr1993, line 57: It would be nice to use the type system to validate the transitions, rather than this runtime setup. I think we discussed using a trait like
```rust
trait LookupState {
  type JointCombiner;
   ...
}
struct LookupContext<State: LookupState> {
  joint_combiner: State::JointCombiner,
  ...
}
```
and then the helper functions in lookup can consume and construct the right contexts.

mimoo: yeah this is just the simplest way to move the lookup logic into optional blocks, we definitely should discuss more on how we can make the state transitions nicer :o

kimchi/src/verifier.rs
```diff
@@ -610,20 +613,19 @@ where
                             panic!("Attempted to use {:?}, but no lookup index was given", col)
                         }
                         Some(lindex) => {
+                            let joint_combiner =
+                                constants.joint_combiner.expect("missing joint combiner");
```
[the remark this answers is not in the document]
mimoo: It is only set to `None` if we didn't follow the flow correctly. It already defaults to `Some(zero)` in the case were we only do single column lookups.

Screen Shot 2022-05-03 at 8 35 51 AM
 
PS: I also believe we can't, today, do a single column lookup because `SingleLookup` doesn't have a `table_id` entry :D

PR #517, opened 2022-04-08
Refactor precomputation constraint system

chee-chyuan:
Another attempt at refactoring precomputation and closes https://github.com/o1-labs/proof-systems/pull/441

Issue
https://github.com/o1-labs/proof-systems/issues/394

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -1,5 +1,6 @@
 //! This module implements Plonk circuit constraint primitive.
 
+use crate::circuits::domain_constant_evaluation::DomainConstantEvaluations;
 use crate::circuits::{
     domains::EvaluationDomains,
     gate::{CircuitGate, GateType},
@@ -7,30 +8,27 @@ use crate::circuits::{
     wires::*,
 };
 use ark_ff::{FftField, SquareRootField, Zero};
-use ark_poly::UVPolynomial;
 use ark_poly::{
     univariate::DensePolynomial as DP, EvaluationDomain, Evaluations as E,
     Radix2EvaluationDomain as D,
 };
 use array_init::array_init;
 use blake2::{Blake2b512, Digest};
+use core::borrow::Borrow;
 use itertools::repeat_n;
 use o1_utils::{field_helpers::i32_to_field, ExtendedEvaluations};
 use oracle::poseidon::ArithmeticSpongeParams;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_with::serde_as;
 
-use super::lookup::{
-    constraints::LookupConfiguration,
-    lookups::{JointLookup, LookupInfo},
-    tables::LookupTable,
+use super::{
+    lookup::{
+        constraints::LookupConfiguration,
+        lookups::{JointLookup, LookupInfo},
+        tables::LookupTable,
+    },
+    polynomials::permutation::ZK_ROWS,
 };
-
-//
-// Constants
-//
-
-pub const ZK_ROWS: u64 = 3;
 
 //
 // ConstraintSystem
@@ -66,7 +64,7 @@ pub struct LookupConstraintSystem<F: FftField> {
 
 #[serde_as]
 #[derive(Clone, Serialize, Deserialize, Debug)]
-pub struct ConstraintSystem<F: FftField> {
+pub struct ConstraintSystem<F: FftField, B: Borrow<DomainConstantEvaluations<F>>> {
     // Basics
     // ------
     /// number of public inputs
@@ -83,9 +81,6 @@ pub struct ConstraintSystem<F: FftField> {
     /// permutation polynomial array
     #[serde_as(as = "[o1_utils::serialization::SerdeAs; PERMUTS]")]
     pub sigmam: [DP<F>; PERMUTS],
-    /// zero-knowledge polynomial
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub zkpm: DP<F>,
 
     // Coefficient polynomials. These define constant that gates can use as they like.
     // ---------------------------------------
@@ -146,31 +141,6 @@ pub struct ConstraintSystem<F: FftField> {
     #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub endomul_scalar8: E<F, D<F>>,
 
-    // Constant polynomials
-    // --------------------
-    /// 1-st Lagrange evaluated over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub l1: E<F, D<F>>,
-    /// 0-th Lagrange evaluated over domain.d4
-    // TODO(mimoo): be consistent with the paper/spec, call it L1 here or call it L0 there
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub l04: E<F, D<F>>,
-    /// 0-th Lagrange evaluated over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub l08: E<F, D<F>>,
-    /// zero evaluated over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub zero4: E<F, D<F>>,
-    /// zero evaluated over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub zero8: E<F, D<F>>,
-    /// zero-knowledge polynomial over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub zkpl: E<F, D<F>>,
-    /// the polynomial that vanishes on the last four rows
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub vanishes_on_last_4_rows: E<F, D<F>>,
-
     /// wire coordinate shifts
     #[serde_as(as = "[o1_utils::serialization::SerdeAs; PERMUTS]")]
     pub shift: [F; PERMUTS],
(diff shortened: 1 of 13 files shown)


kimchi/src/circuits/constraints.rs
```diff
@@ -146,31 +141,6 @@ pub struct ConstraintSystem<F: FftField> {
...
-    /// zero-knowledge polynomial over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub zkpl: E<F, D<F>>,
-    /// the polynomial that vanishes on the last four rows
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub vanishes_on_last_4_rows: E<F, D<F>>,
-
     /// wire coordinate shifts
     #[serde_as(as = "[o1_utils::serialization::SerdeAs; PERMUTS]")]
     pub shift: [F; PERMUTS],
```
mimoo, line 146: I think `shifts` should also be in precomputations btw, but this can be in another PR

kimchi/src/circuits/constraints.rs
```diff
@@ -185,6 +155,45 @@ pub struct ConstraintSystem<F: FftField> {
     /// lookup constraint system
     #[serde(bound = "LookupConstraintSystem<F>: Serialize + DeserializeOwned")]
     pub lookup_constraint_system: Option<LookupConstraintSystem<F>>,
+
+    /// precomputes
+    #[serde(skip)]
+    pub precomputations: Arc<DomainConstantEvaluations<F>>,
+}
+
+impl<F: FftField> Default for ConstraintSystem<F> {
```
mimoo, line 164: why implement `Default` for `ConstraintSystem`?

kimchi/src/circuits/constraints.rs
```diff
@@ -468,6 +420,7 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
         lookup_tables: Vec<LookupTable<F>>,
         fr_sponge_params: ArithmeticSpongeParams<F>,
         public: usize,
+        precompute: Option<Arc<DomainConstantEvaluations<F>>>,
```
mimoo, line 423: `precompute` -> `precomputations`?  Also worth adding in the documentation how this can be used by the caller (provide the precomputation or set it to None).

Now that I'm looking at this I find it a bit weird as well, the precomputation is based on the domain size and at this point in time we don't know the domain size... What if you pass a precomputation that's unrelated to the domain size we're using here?

At the very least it should perform a check that the domain are the same and return an error (or here `None` until https://github.com/o1-labs/proof-systems/pull/4907 gets in)

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,79 @@
...
+use ark_poly::{univariate::DensePolynomial as DP, Evaluations as E, Radix2EvaluationDomain as D};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+use super::polynomials::permutation::vanishes_on_last_4_rows;
+// use super::zk_polynomial::ZkPolynomial;
+
+#[serde_as]
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct DomainConstantEvaluations<F: FftField> {
```
mimoo, line 16: add documentation here (for example, "pre-computed polynomials that depend only on the chosen field and domain")

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,79 @@
+use crate::circuits::domains::EvaluationDomains;
```
mimoo, line 1: would be nice to add a module-level comment (with `//!`)

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,79 @@
...
+use crate::circuits::polynomials::permutation::ZK_ROWS;
+use ark_ff::FftField;
+use ark_poly::EvaluationDomain;
+use ark_poly::UVPolynomial;
+use ark_poly::{univariate::DensePolynomial as DP, Evaluations as E, Radix2EvaluationDomain as D};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+use super::polynomials::permutation::vanishes_on_last_4_rows;
+// use super::zk_polynomial::ZkPolynomial;
```
mimoo, line 12: delete

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,79 @@
...
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub vanishes_on_last_4_rows: E<F, D<F>>,
+    /// zero-knowledge polynomial over domain.d8
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub zkpl: E<F, D<F>>,
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub zkpm: DP<F>,
+}
+
+impl<F: FftField> Default for DomainConstantEvaluations<F> {
```
mimoo, line 37: why implement Default?

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,79 @@
...
+            .evaluate_over_domain_by_ref(domain.d8);
+        let constant_1_d4 =
+            E::<F, D<F>>::from_vec_and_domain(vec![F::one(); domain.d4.size as usize], domain.d4);
+        let constant_1_d8 =
+            E::<F, D<F>>::from_vec_and_domain(vec![F::one(); domain.d8.size as usize], domain.d8);
+
+        let vanishes_on_last_4_rows =
+            vanishes_on_last_4_rows(domain.d1).evaluate_over_domain(domain.d8);
+
+        assert!(domain.d1.size > ZK_ROWS);
```
mimoo, line 64: can you return an error here instead?

kimchi/src/circuits/constraints.rs
mimoo: oh and can you update the example in the `README.md`?

mimoo: From a user perspective it would be nice to have two functions:

* create
* create_with_shared_precomputations

where create calls create_with_shared_precomputations with precomputations set to `None`. Like that the example in the README doesn't have to change

kimchi/src/prover_index.rs
```diff
@@ -21,7 +21,7 @@ use std::sync::Arc;
 //~spec:startcode
 pub struct ProverIndex<G: CommitmentCurve> {
     /// constraints system polynomials
-    #[serde(bound = "ConstraintSystem<ScalarField<G>>: Serialize + DeserializeOwned")]
+    #[serde(skip)]
     pub cs: ConstraintSystem<ScalarField<G>>,
```
mimoo, line 25: oh, so that should definitely not be `skip` (I feel like I might be contradicting myself, did I say it should be `skip`?)

Instead the `precomputations` should be `skip`. But this introduces a new problem: now we need to know when something is set or not set after deserializing. I think there's a few options:

* either we set precomputations to be an `Option<Arc<...>>` and `unwrap()` it wherever we use it. So when you deserialize it is set to `None` automatically, and you have to manually call a function to set it (maybe two functions to offer the two different features: own it, or share it)
* or we set precomputations to be an `OnceCell<Arc<...>>` and if there's nothing in there when we access it we create it. But we also give the option to set the Arc. I think this is the cleaner path, but I'm not sure if I explained this very clearly :)

mimoo (commented):
This is great! I left you a number of comments, let me know what you think :o

kimchi/src/circuits/constraints.rs
mimoo: ok I see that you need it because of serde skip. So two things:

* in another comment I say that we shouldn't skip this, so we actually don't need Default here
* you could have done `#[derive(Default)]` and avoid that code :)

kimchi/src/prover_index.rs
mimoo: something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f4c82cfe98a03f59c5d7aa42039cf333

kimchi/src/circuits/constraints.rs
chee-chyuan: ok updated so the `create` calls `create_with_shared_precomputations` how shall i add this new fn to the Readme?

kimchi/src/circuits/domain_constant_evaluation.rs
chee-chyuan: added something but im not confident at my description. could you suggest something better?

kimchi/src/prover_index.rs
chee-chyuan: i did something slightly different. i created a method `get_or_set_domain_constant_evaluation`. what do you think?

kimchi/src/circuits/domain_constant_evaluation.rs
chee-chyuan: as in using `Result` ? or simply assert with a message?

kimchi/src/circuits/constraints.rs
```diff
@@ -637,7 +616,21 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
...
+        Some(constraints)
+    }
+
+    pub fn get_or_set_domain_constant_evaluation(
+        &self,
+        precomputations: Option<Arc<DomainConstantEvaluations<F>>>,
+    ) -> &Arc<DomainConstantEvaluations<F>> {
+        self.precomputations.get_or_init(|| match precomputations {
+            Some(t) => t,
+            None => Arc::new(DomainConstantEvaluations::create(self.domain).unwrap()),
```
mimoo, line 633: not really a fan of this approach for two reasons:

* the function name is long, and you're going to use it in multiple places
* it always takes a `None` everywhere you want to use it

I'd rather have two functions, once that I can call after deserialization, and one that I can call to get the value

kimchi/src/circuits/domain_constant_evaluation.rs
mimoo: maybe "This contains the [DomainConstantEvaluations] which is used to provide precomputations to a [ConstraintSystem]."

kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -207,7 +207,7 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
 
             // alpha
             let alpha_pow = {
-                let mut res = self.precomputations.constant_1_d4.clone();
+                let mut res = self.precomputations.get().unwrap().constant_1_d4.clone();
```
mimoo, line 210: I'm not sure how I feel about this `unwrap()`. The nice thing about OnceCell is that you can evaluate the value the first time you get it, and then cache it for subsequent calls.

I think it'd be nicer to have a `self.precomputations()` that never fails

chee-chyuan:
made precomputation into a non public property in ConstraintSystem so that user will need to call `precomputations()` method

kimchi/src/circuits/domain_constant_evaluation.rs
mimoo: since we're in a function that returns a Result it seems weird to panic here, I think returning an `Err("the domain is not large enough for zero-knowledge rows")`. Although this is outside the scope of this PR so don't worry about this :)

kimchi/src/circuits/constraints.rs
```diff
@@ -619,19 +619,27 @@ impl<F: FftField + SquareRootField> ConstraintSystem<F> {
...
-        &self,
-        precomputations: Option<Arc<DomainConstantEvaluations<F>>>,
-    ) -> &Arc<DomainConstantEvaluations<F>> {
-        self.precomputations.get_or_init(|| match precomputations {
-            Some(t) => t,
-            None => Arc::new(DomainConstantEvaluations::create(self.domain).unwrap()),
-        })
+    pub fn precomputations(&self) -> &Arc<DomainConstantEvaluations<F>> {
+        self.precomputations
+            .get_or_init(|| Arc::new(DomainConstantEvaluations::create(self.domain).unwrap()))
```
mimoo, line 636: can you use `expect` instead of `unwrap`?

kimchi/src/circuits/constraints.rs
mimoo: I think it's better to use the `create` function in the README :o which it already does

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// zero-knowledge polynomial over domain.d8
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub zkpl: E<F, D<F>>,
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub zkpm: DP<F>,
+}
+
+impl<F: FftField> DomainConstantEvaluations<F> {
+    pub fn create(domain: EvaluationDomains<F>) -> Option<Self> {
+        let poly_x_d1 = DP::from_coefficients_slice(&[F::zero(), F::one()])
```
mimoo, line 40: used to be l1

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -0,0 +1,65 @@
...
+            .evaluate_over_domain_by_ref(domain.d8);
+        let constant_1_d4 =
+            E::<F, D<F>>::from_vec_and_domain(vec![F::one(); domain.d4.size as usize], domain.d4);
```
mimoo, line 43: used to be l04

kimchi/src/circuits/domain_constant_evaluation.rs
mimoo: btw this should be called `x_poly` maybe, the `d1` here is useless as we're not in an evaluations type

mimoo (approved):
Made sure that all the code moves are legit :) this looks excellent to me so let's get that in!

thanks again for the many changes to our feedback!

PR #518, opened 2022-04-09
Feature - Range check gate

### LINKED ISSUE #497: Foreign field multiplication - proof generation and verification test
To be sure everything is integrated property we must go beyond the verification tests and create an end-to-end test for generating a proof for a circuit containing foreign field multiplication gates and verifying it.

### LINKED ISSUE #519: Foreign field multiplication - refactor into module

jspada:
The PR originally integrated the basic foreign field multiplication gate into kimchi.  The constraints are not yet complete, but this PR fully integrates it into the proof system.  Subsequently it was reduced to a range check gate because that is currently most of the functionality provided.

Closes #497
Closes #519


kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,107 @@
+// ForeignMul0 - Foreign field element constraints
```
querolita, line 1: todo: make these comments spec

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,107 @@
...
+        // 1) Apply range constraints on sublimbs
+
+        // Create 4 12-bit plookup range constraints
+        let mut constraints: Vec<E<F>> =
+            (1..5).map(|i| sublimb_plookup_constraint(&w(i))).collect();
+
+        // Create 8 2-bit chunk range constraints
+        constraints.append(
+            &mut (7..COLUMNS)
+                .map(|i| sublimb_plookup_constraint(&w(i)))
```
querolita, line 69: is this a typo? I was expecting this to be a call to `sublimb_crumb_constraint` instead...
I guess this verifies as Ok(()) because these are `zero()` in any case (the TODO inside plookup), so it is normal that it didn't complain before 😅

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_2.rs
```diff
@@ -0,0 +1,76 @@
...
+    //     these cells are equal to those
+    fn constraints() -> Vec<E<F>> {
+        let w = |i| E::cell(Column::Witness(i), Curr);
+
+        // Row structure
+        //  Column w(i) 0    1       ... 4       5     6     7     ... 14
+        //         Curr limb plookup ... plookup crumb crumb crumb ... crumb
+
+        // Apply range constraints on sublimbs (create 4 12-bit plookup range constraints)
+        // crumbs were constrained by ForeignMul1 circuit gate
```
querolita, line 63: ah right, in the "next" row, cool 👍🏻

querolita (approved):
Overall, it looks fantastic to me! Plenty of comments clarifying all of the steps and making review easier. Definitely should make these part of the specs when this is finished work. 

I think I spotted one typo. I am accepting the PR, but please try to address it so it will be fixed when merged 😃

book/specifications/kimchi/template.md
```diff
@@ -160,19 +160,19 @@ In this section, we describe all the constraints that make up the main polynomia
...
+It returns the sum of all the constraints, where each constraint has been multiplied by a power of alpha.
 In other words it returns:
 $$ \sum_i \alpha^i \cdot \text\{constraint}_i $$
 
 The different ranges of alpha are described as follows:
 
 <!-- generated using `cargo test -p kimchi --lib -- alphas::tests::get_alphas_for_spec --nocapture` -->
-* **gates**. Offset starts at 0 and 21 powers of $\alpha$ are used
-* **Permutation**. Offset starts at 21 and 3 powers of $\alpha$ are used
+* **gates**. Offset starts at 0 and 25 powers of $\alpha$ are used
```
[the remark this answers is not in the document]
jspada: Yes.  That's right.  I will update!

kimchi/src/circuits/constraints.rs
```diff
@@ -146,6 +148,10 @@ pub struct ConstraintSystem<F: FftField> {
     #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub endomul_scalar8: E<F, D<F>>,
 
+    /// Foreign field multiplication gate selector polynomials
+    #[serde(bound = "Vec<foreign_mul::SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
+    pub foreign_mul_selector_polys: Vec<foreign_mul::SelectorPolynomial<F>>,
```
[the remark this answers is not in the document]
jspada: Yes, exactly.   I decided to make it a vector of selector polynomials rather than an optional object containing arrays of polynomials.  This field is set by calling `foreign_mul::selector_polynomials(...)`.  When the vector is empty is means there are no foreign multiplication gates.

kimchi/src/circuits/constraints.rs
```diff
@@ -182,6 +188,10 @@ pub struct ConstraintSystem<F: FftField> {
     #[serde(skip)]
     pub fr_sponge_params: ArithmeticSpongeParams<F>,
 
+    /// Foreign field modulus parameter
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub foreign_modulus: Vec<F>,
```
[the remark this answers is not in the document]
jspada: Correct

kimchi/src/circuits/expr.rs
```diff
@@ -415,7 +417,14 @@ impl Variable {
             | Index(GateType::CairoInstruction)
             | Index(GateType::CairoFlags)
             | Index(GateType::CairoTransition) => todo!(),
+            Index(GateType::ForeignMul0) => Ok(evals.foreign_mul_selector[0]),
+            Index(GateType::ForeignMul1) => Ok(evals.foreign_mul_selector[1]),
+            Index(GateType::ForeignMul2) => Ok(evals.foreign_mul_selector[2]),
             Coefficient(_) | LookupKindIndex(_) | Index(_) => {
+                println!(
```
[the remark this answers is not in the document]
jspada: Good idea, will try

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
```diff
@@ -0,0 +1,107 @@
+// Foreign field multiplication
```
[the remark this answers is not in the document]
jspada: Yes, there will indeed be another PR for adding the spec comments.

kimchi/src/proof.rs
```diff
@@ -91,6 +96,7 @@ impl<F: Zero> ProofEvaluations<F> {
             lookup: None,
             generic_selector: F::zero(),
             poseidon_selector: F::zero(),
+            foreign_mul_selector: foreign_mul::off(),
```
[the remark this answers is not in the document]
jspada: I don't think so-- not for this.  This returns a vector of `F::zero()` (the foreign mul selector evaluations) one for each foreign mul circuit gate.  AFAIK, I only use `vec![]` for to specify no `foreign_modulus` and no `foreign_mul_selector_polys`

kimchi/src/proof.rs
```diff
@@ -204,6 +215,7 @@ pub mod caml {
         ),
         pub generic_selector: Vec<CamlF>,
         pub poseidon_selector: Vec<CamlF>,
+        pub foreign_mul_selector: Option<(Vec<CamlF>, Vec<CamlF>, Vec<CamlF>)>,
```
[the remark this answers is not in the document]
jspada: Because it's a tuple, rather than a vector and it cannot be empty, so needs to be an option.  I think it's expecting a tuple, so didn't want to change this.

kimchi/src/linearization.rs
```diff
@@ -22,13 +23,14 @@ use ark_poly::Radix2EvaluationDomain as D;
...
     chacha: bool,
+    foreign_mul: bool,
     lookup_constraint_system: Option<&LookupConfiguration<F>>,
 ) -> (Expr<ConstantExpr<F>>, Alphas<F>) {
     // register powers of alpha so that we don't reuse them across mutually inclusive constraints
     let mut powers_of_alpha = Alphas::<F>::default();
 
     // gates
-    let highest_constraints = VarbaseMul::<F>::CONSTRAINTS;
+    let highest_constraints = foreign_mul::ForeignMul1::<F>::CONSTRAINTS;
```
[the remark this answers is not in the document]
jspada: Yes.  I think this matches your first comment to only impact powers of alpha when foreign mul is enabled.  Good spot!

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
```diff
@@ -0,0 +1,207 @@
...
+use array_init::array_init;
+use num_bigint::BigUint;
+use o1_utils::FieldHelpers;
+
+use crate::circuits::polynomial::COLUMNS;
+
+const MAX_LIMBS: usize = 3;
+const LIMB_SIZE: usize = 88;
+
+enum WitnessCell {
```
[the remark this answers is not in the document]
jspada: I will change these significantly when I implement the witness computation, so would rather not document until then.

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
```diff
@@ -0,0 +1,207 @@
...
+
+struct ZeroWitnessCell;
+impl ZeroWitnessCell {
+    const fn create() -> WitnessCell {
+        WitnessCell::Zero
+    }
+}
+
+// Generate witness in shape that constraints expect (TODO: static for now, make dynamic)
+const WITNESS_SHAPE: [[WitnessCell; COLUMNS]; 4] = [
```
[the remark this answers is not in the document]
jspada: 👍🏼

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: will do in subsequent PR

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: Yes - good spot.  yeah, that's certainly why it verified fine.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_2.rs
jspada: 👍🏼

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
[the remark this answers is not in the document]
jspada: there's only a few so I'll comment

mimoo (commented):
this looks awesome o.o just giving you a light review before I forget posting this, I'll try to find the time to understand what's going on at a deeper level (would be nice if we spend some time together on that :o)

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
+//
+// Fundamental constraints and related helpers
+//
```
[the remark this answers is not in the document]
jspada: fixed

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use std::cell::RefCell;
+
+use ark_ff::{FftField, One, Zero};
+
+use crate::circuits::expr::{Cache, ConstantExpr, Expr, E};
+
+thread_local! {
+    static CACHE: std::cell::RefCell<Cache>  = RefCell::new(Cache::default());
+}
```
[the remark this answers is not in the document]
jspada: I want to avoid having a global shared cache with locks/mutexes and, instead, have one cache per thread.

kimchi/src/linearization.rs
```diff
@@ -30,11 +30,14 @@ pub fn constraints_expr<F: FftField + SquareRootField>(
...
-        highest_constraints,
-    );
+    let max_constraints = if foreign_mul {
+        foreign_mul::ForeignMul1::<F>::CONSTRAINTS
+    } else {
+        VarbaseMul::<F>::CONSTRAINTS
+    };
+    // Set up powers of alpha.  Only the max number of constraints matters.
+    // The gate type argument can just be the zero gate.
+    powers_of_alpha.register(ArgumentType::Gate(GateType::Zero), max_constraints);
```
querolita, line 40: Much better!

kimchi/src/circuits/polynomials/turshi.rs
```diff
@@ -222,16 +220,10 @@ impl<F: FftField> CircuitGate<F> {
...
-            z: F::rand(rng),
-            s: array_init(|_| F::rand(rng)),
-            generic_selector: F::zero(),
-            poseidon_selector: F::zero(),
-            lookup: None,
-            foreign_mul_selector: foreign_mul::off(),
-        };
-        let evals = vec![eval(curr), eval(next)];
+        let evals = vec![
+            ProofEvaluations::dummy_with_witness_evaluations(curr),
```
querolita, line 224: Ah, neat!

kimchi/src/circuits/expr.rs
```diff
@@ -421,25 +440,21 @@ impl Variable {
             Index(GateType::ForeignMul1) => Ok(evals.foreign_mul_selector[1]),
             Index(GateType::ForeignMul2) => Ok(evals.foreign_mul_selector[2]),
             Coefficient(_) | LookupKindIndex(_) | Index(_) => {
-                println!(
-                    "Cannot get index evaluation {:?} (should have been linearized away)",
-                    self.col
-                );
-                Err("Cannot get index evaluation (should have been linearized away)")
+                Err(ExprError::MissingIndexEvaluation(self.col))
```
querolita, line 443: Oh you found the way to customize the `Err` message, awesome!

querolita (approved):

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
```diff
@@ -91,7 +91,6 @@
 ///
 ///  Nb. each CircuitGate type corresponds to a unique polynomial and thus
 ///       is assigned its own unique powers of alpha
```
mimoo, line 93: move this to a module-level doc in constraints?

kimchi/src/verifier_index.rs
```diff
@@ -124,6 +128,10 @@ pub struct VerifierIndex<G: CommitmentCurve> {
     pub fr_sponge_params: ArithmeticSpongeParams<ScalarField<G>>,
     #[serde(skip)]
     pub fq_sponge_params: ArithmeticSpongeParams<BaseField<G>>,
+
+    // Foreign field modulus
+    #[serde(skip)]
+    pub foreign_modulus: Vec<ScalarField<G>>,
```
mimoo, line 134: why is it skipped here? but not in prover index?

kimchi/src/circuits/constraints.rs
```diff
@@ -153,6 +159,10 @@ pub struct ConstraintSystem<F: FftField> {
...
```
mimoo, line 164: same as in verifier_index, I think both of the foreign stuff should be behind a single `Option`

kimchi/src/circuits/constraints.rs
```diff
@@ -380,17 +390,23 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
+    /// Creates a constraint system from a vector of gates ([CircuitGate]),
+    ///   some sponge parameters ([ArithmeticSpongeParams]),
+    ///   an optional foreign modulus, and the number of public inputs.
+    ///   Note that passing an empty vector for the foreign modulus parameter
+    ///   disables the foreign field multiplication gate.
     pub fn create(
         gates: Vec<CircuitGate<F>>,
         lookup_tables: Vec<LookupTable<F>>,
         fr_sponge_params: ArithmeticSpongeParams<F>,
+        foreign_modulus: Vec<F>,
```
mimoo, line 402: I think this should be an `Option<Vec<F>>`. Also it sounded like your field needs to be split in 3 `F` no? Would it make sense to use `[F; 3]` for this (and what's in prover/verifier index?)

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+    x
+}
+
+pub(crate) fn two<F: FftField>() -> E<F> {
+    Expr::Constant(ConstantExpr::Literal(2u32.into()))
+}
+
+pub(crate) fn three<F: FftField>() -> E<F> {
+    Expr::Constant(ConstantExpr::Literal(3u32.into()))
+}
```
mimoo, line 26: I think most functions here should be moved to the expr module

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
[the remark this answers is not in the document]
mimoo: I guess there's `unsync` in once_cell that is not thread safe. I'm not sure what's the difference in practice. Also I'm still wondering why we're using a global here

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+    CACHE.with(|cache| x = cache.borrow_mut().cache(x.clone()));
+    x
+}
+
+pub(crate) fn two<F: FftField>() -> E<F> {
+    Expr::Constant(ConstantExpr::Literal(2u32.into()))
+}
+
+pub(crate) fn three<F: FftField>() -> E<F> {
+    Expr::Constant(ConstantExpr::Literal(3u32.into()))
```
mimoo, line 25: you have `constant` in `expr.rs` which you can use as `constant(3u32.into())`

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::circuits::expr::{Cache, ConstantExpr, Expr, E};
+
+thread_local! {
+    static CACHE: std::cell::RefCell<Cache>  = RefCell::new(Cache::default());
+}
+
+pub(crate) fn _cache<F: FftField>(mut x: E<F>) -> E<F> {
+    CACHE.with(|cache| x = cache.borrow_mut().cache(x.clone()));
+    x
+}
```
mimoo, line 18: I guess the cache and this function are not used anywhere?

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+    Expr::Constant(ConstantExpr::Literal(3u32.into()))
+}
+
+pub(crate) fn sublimb_plookup_constraint<F: FftField>(_sublimb: &E<F>) -> E<F> {
+    // TODO: implement plookup constraint for 12-bit sublimb
+    E::zero()
+}
+
+// Crumb constraint for 2-bit sublimb
+pub(crate) fn sublimb_crumb_constraint<F: FftField>(sublimb: &E<F>) -> E<F> {
```
mimoo, line 34: I don't think you need to have `sublimb` in the name here. Just `crumb` might be good enough no?

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,106 @@
...
+{
+    const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::ForeignMul0);
+    const CONSTRAINTS: u32 = 13;
+
+    // Constraints for ForeignMul0
+    //   * Operates on Curr row
+    //   * Range constrain all sublimbs except p4 and p5
+    //   * Constrain that combining all sublimbs equals the limb stored in column 0
+    fn constraints() -> Vec<E<F>> {
+        let w = |i| E::cell(Column::Witness(i), Curr);
```
mimoo, line 53: `expr.rs` has a function called `witness` that does exactly this

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        // Row structure
+        //  Column w(i) 0    1       ... 4       5    6    7     ... 14
+        //        Curr  limb plookup ... plookup copy copy crumb ... crumb
+
+        // 1) Apply range constraints on sublimbs
+
+        // Create 4 12-bit plookup range constraints
+        let mut constraints: Vec<E<F>> =
+            (1..5).map(|i| sublimb_plookup_constraint(&w(i))).collect();
```
mimoo, line 63: btw lookups won't be implemented in this file in this way, so I would just remove this line as it doesn't really serve as a placeholder

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,106 @@
...
+        // Check limb =  lp0*2^0 + lp1*2^{12}  + ... + p5*2^{60}   + lc0*2^{72}  + lc1*2^{74}  + ... + lc7*2^{86}
+        //       w(0) = w(1)*2^0 + w(2)*2^{12} + ... + w(6)*2^{60} + w(7)*2^{72} + w(8)*2^{74} + ... + w(14)*2^{86}
+        //            = \sum i \in [1,7] 2^{12*(i - 1)}*w(i) + \sum i \in [8,14] 2^{2*(i - 7) + 6*12}*w(i)
+        let combined_sublimbs = (1..COLUMNS).fold(E::zero(), |acc, i| {
+            match i {
+                1..=7 => {
+                    // 12-bit chunk offset
+                    acc + two().pow(12 * (i as u64 - 1)) * w(i)
+                }
+                8..=COLUMNS => {
```
mimoo, line 91: `8..COLUMS` no?

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,106 @@
...
+        //    s
+        //
+        // Check limb =  lp0*2^0 + lp1*2^{12}  + ... + p5*2^{60}   + lc0*2^{72}  + lc1*2^{74}  + ... + lc7*2^{86}
+        //       w(0) = w(1)*2^0 + w(2)*2^{12} + ... + w(6)*2^{60} + w(7)*2^{72} + w(8)*2^{74} + ... + w(14)*2^{86}
+        //            = \sum i \in [1,7] 2^{12*(i - 1)}*w(i) + \sum i \in [8,14] 2^{2*(i - 7) + 6*12}*w(i)
+        let combined_sublimbs = (1..COLUMNS).fold(E::zero(), |acc, i| {
+            match i {
+                1..=7 => {
+                    // 12-bit chunk offset
+                    acc + two().pow(12 * (i as u64 - 1)) * w(i)
```
mimoo, line 89: I guess you could speed that up with honer but I don't think it's an essential optimization here

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,106 @@
...
+                }
+                8..=COLUMNS => {
+                    // 2-bit chunk offset
+                    acc + two().pow(2 * (i as u64 - 7) + 6 * 12) * w(i)
+                }
+                _ => {
+                    panic!("Invalid column index {}", i)
+                }
```
mimoo, line 97: does it complain if you remove the `_` case?

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_1.rs
```diff
@@ -0,0 +1,125 @@
...
+    const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::ForeignMul1);
+    const CONSTRAINTS: u32 = 25;
+
+    // Constraints for ForeignMul1
+    //   * Operates on Curr and Next row
+    //   * Range constrain all sublimbs
+    //   * Constrain that combining all sublimbs equals the limb stored in row Curr, column 0
+    fn constraints() -> Vec<E<F>> {
+        let w_curr = |i| E::cell(Column::Witness(i), Curr);
+        let w_next = |i| E::cell(Column::Witness(i), Next);
```
mimoo, line 53: `witness_curr` and `witness_next` in `expr.rs`

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_1.rs
```diff
@@ -0,0 +1,125 @@
...
+        // (2nd part) + \sum i \in [5,14] 2^{2*(i - 5} + 68)*w_next(i)
+
+        // 1st part (Curr row): \sum i \in [1,5] 2^{12*(i - 1)}*w_curr(i) + \sum i \in [6,14] 2^{2*(i - 5) + 4*12}*w_curr(i)
+        let combined_sublimbs = (1..COLUMNS).fold(E::zero(), |acc, i| {
+            match i as usize {
+                1..=4 => {
+                    // 12-bit chunk
+                    acc + two().pow(12 * (i as u64 - 1)) * w_curr(i)
+                }
+                5..=COLUMNS => {
```
mimoo, line 104: same here `5..COLUMNS` I believe

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_2.rs
```diff
@@ -0,0 +1,75 @@
...
+        let mut constraints: Vec<E<F>> =
+            (1..5).map(|i| sublimb_plookup_constraint(&w(i))).collect();
+
+        // Temporary dummy constraint to avoid zero polynomial edge case
+        // and avoid verifier check that all commitments aren't identically zero
```
mimoo, line 67: not sure I understand this poseidon gate from the comment :D

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+        wires::GateWires,
+    },
+    proof::ProofEvaluations,
+};
+
+use super::{ForeignMul0, ForeignMul1, ForeignMul2};
+
+impl<F: FftField> CircuitGate<F> {
+    /// Create foreign multiplication gate
+    pub fn create_foreign_mul(wires: &[GateWires; 8]) -> Vec<Self> {
```
mimoo, line 28: two things:

* take the current `row: usize` as argument instead of wires, and return `(usize, Vec<Self>)` where `usize` is the next `row` that the caller can use to continue building their circuit (see poseidon)

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+                typ: GateType::ForeignMul1,
+                wires: wires[6],
+                coeffs: vec![],
+            },
+            CircuitGate {
+                typ: GateType::ForeignMul2,
+                wires: wires[7],
+                coeffs: vec![],
+            },
+        ]
```
mimoo, line 72: two things:

* use `Wire::new(row)` to generate each wire
* add the necessary wiring here as well

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+            poseidon_selector: F::zero(),
+            lookup: None,
+            foreign_mul_selector: match self.typ {
+                // Force this circuit gate on
+                GateType::ForeignMul0 => vec![F::one(), F::zero(), F::zero()],
+                GateType::ForeignMul1 => vec![F::zero(), F::one(), F::zero()],
+                GateType::ForeignMul2 => vec![F::zero(), F::zero(), F::one()],
+                _ => panic!("Invalid gate type {:?}", self.typ),
+            },
+        };
```
mimoo, line 131: we really need a `for_testing` for `ProofEvaluations` instead of rewriting this code all the time :)

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+        // Setup circuit constants
+        let constants = expr::Constants {
+            alpha: F::rand(rng),
+            beta: F::rand(rng),
+            gamma: F::rand(rng),
+            joint_combiner: F::rand(rng),
+            endo_coefficient: cs.endo,
+            mds: vec![],
+            foreign_modulus: cs.foreign_modulus.clone(),
+        };
```
mimoo, line 143: same here

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+pub fn combined_constraints<F: FftField>(alphas: &Alphas<F>) -> E<F> {
+    ForeignMul0::combined_constraints(alphas)
+        + ForeignMul1::combined_constraints(alphas)
+        + ForeignMul2::combined_constraints(alphas)
+}
+
+/// Foreign field multiplication CircuitGate selector polynomial
+#[serde_as]
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct SelectorPolynomial<F: FftField> {
```
mimoo, line 203: I think we should generalize this pattern to the other selector polynomials present in `ConstraintSystem`. Maybe create an issue for that if you don't want to do this here?

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+                .iter()
+                .map(|gate| {
+                    if gate.typ == *gate_type {
+                        F::one()
+                    } else {
+                        F::zero()
+                    }
+                })
+                .collect(),
+            domain.d1,
```
mimoo, line 230: we use this pattern so much that we might want to factor it out in a 

```rust
pub fn selector_from(gates: &[CircuitGate], typ: GateType) -> EV { ... }
```

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// Coefficient form
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub coeff: DP<F>,
+    /// Evaluation form (evaluated over domain d8)
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
+    pub eval8: EV<F, D<F>>,
+}
+
+/// Create foreign field multiplication circuit gates selector polynomials
+pub fn selector_polynomials<F: FftField>(
```
mimoo, line 213: implement as a method on `SelectorPolynomial`

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,528 @@
...
+        let eval8 = coeff.evaluate_over_domain_by_ref(domain.d8);
+
+        SelectorPolynomial { coeff, eval8 }
+    }))
+}
+
+/// Switch off foreign field multiplication, with zero circuit gate selector evaluations
+pub fn off<F: Zero>() -> Vec<F> {
+    circuit_gates().iter().map(|_| F::zero()).collect()
+}
```
mimoo, line 244: I find this function a bit odd, it's short and only used in one place I believe, not sure if really necessary

kimchi/src/linearization.rs
```diff
@@ -47,6 +52,10 @@ pub fn constraints_expr<F: FftField + SquareRootField>(
         expr += ChaChaFinal::combined_constraints(&powers_of_alpha);
     }
 
+    if foreign_mul {
+        expr += foreign_mul::combined_constraints(&powers_of_alpha);
```
mimoo, line 56: Chacha was actually a single call like this initially, but I felt like it was too much magic and not really showing us the cost of having the gate. Now we can clearly see that ChaCha has 4 gates/selector polynomials, whereas here it's not clear what the cost is. wdyt?

kimchi/src/proof.rs
```diff
@@ -112,6 +118,11 @@ impl<F: FftField> ProofEvaluations<Vec<F>> {
             }),
             generic_selector: DensePolynomial::eval_polynomial(&self.generic_selector, pt),
             poseidon_selector: DensePolynomial::eval_polynomial(&self.poseidon_selector, pt),
+            foreign_mul_selector: self
+                .foreign_mul_selector
+                .iter()
+                .map(|poly| DensePolynomial::eval_polynomial(poly, pt))
```
mimoo, line 124: I'm wondering why all these calls are not written as `poly.eval_polynomial(pt)`

kimchi/src/prover.rs
```diff
@@ -535,6 +542,40 @@ where
...
+                        panic!(
+                            "Bad evaluation domain size {} for {:?}",
+                            evals.domain().size,
+                            gate_type
+                        );
+                    }
+
+                    if cfg!(test) {
+                        let (_, res) = evals
+                            .clone()
```
mimoo, line 568: no need to clone?

kimchi/src/prover.rs
```diff
@@ -871,6 +935,11 @@ where
                     }),
                     generic_selector: DensePolynomial::eval_polynomial(&es.generic_selector, e1),
                     poseidon_selector: DensePolynomial::eval_polynomial(&es.poseidon_selector, e1),
+                    foreign_mul_selector: es
+                        .foreign_mul_selector
+                        .iter()
+                        .map(|poly| DensePolynomial::eval_polynomial(poly, e1))
+                        .collect(),
```
mimoo, line 942: I'm actually not sure I understand why you evaluate these selectors in the proof. Where does the linearization fail if you don't?

kimchi/src/verifier_index.rs
```diff
@@ -197,6 +205,16 @@ where
                 array_init(|i| self.srs.commit_evaluations_non_hiding(domain, &c[i], None))
             }),
 
+            foreign_mul_comm: self
+                .cs
+                .foreign_mul_selector_polys
+                .iter()
+                .map(|poly| {
+                    self.srs
+                        .commit_evaluations_non_hiding(domain, &poly.eval8, None)
```
mimoo, line 214: I believe this will crash because the domain of `domain` and `poly.eval8` won't match. Even if you use `d8` here it might not be a good idea as this means we'd have to precompute all the lagrange basis in d8 IIRC

ocaml/ocaml-gen/src/conv.rs
```diff
@@ -245,3 +245,23 @@ where
...
+            T2::ocaml_desc(env, generics),
+            T3::ocaml_desc(env, generics),
+        ];
+        format!("({})", v.join(" * "))
+    }
+
+    fn unique_id() -> u128 {
+        const_random!(u128)
+    }
+}
```
mimoo, line 267: unrelated to this PR: we really need a macro to generate all the tuples implementations

utils/src/lib.rs
```diff
@@ -4,9 +4,11 @@ pub mod evaluations;
 pub mod field_helpers;
 pub mod hasher;
 pub mod math;
+pub mod packed_modulus;
```
mimoo, line 7: this module name is not really descriptive imo

utils/src/packed_modulus.rs
```diff
@@ -0,0 +1,24 @@
...
+//!
+
+use ark_ff::{BigInteger, FftField, FpParameters, PrimeField};
+use num_bigint::BigUint;
+
+/// Get the modulus (as `BigUint`) for the field specified by type parameter `F`
+pub fn get_modulus<F: PrimeField>() -> BigUint {
+    let bytes = F::Params::MODULUS.to_bytes_le();
+    BigUint::from_bytes_le(&bytes)
+}
```
mimoo, line 12: I think this would be nicer implemented on `PrimeField` as an extension like the other utils. For example:

```
/// ...
trait ExtendedPrimeField : PrimeField {
  /// ...
  fn modulus_to_biguint() -> BigUint {
    BigUint::from_bytes_leSelf::Params::MODULUS.to_bytes_le())
  }
}

File: utils/src/packed_modulus.rs [hunk 51]
```diff
@@ -0,0 +1,24 @@
...
+pub fn packed_modulus<F: FftField>(modulus: BigUint) -> Vec<F> {
+    let bytes = modulus.to_bytes_le();
+    let chunks: Vec<&[u8]> = bytes
+        .chunks(<F::BasePrimeField as PrimeField>::size_in_bits() / 8)
+        .collect();
+    chunks
+        .iter()
+        .map(|chunk| F::from_random_bytes(chunk).expect("failed to deserialize"))
+        .collect()
+}
```
mimoo, line 24: this feels very specific to the foreign mul gate. Perhaps a good idea to move that there? It also feels like this should potentially pad the modulus and return a `[F; 3]` to make sure that you don't go above

mimoo (approved):
Looks really clean and organized and well-specified/documented! I left you a number of comments

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use ark_ff::{FftField, One, Zero};
+
+use crate::circuits::expr::{Cache, ConstantExpr, Expr, E};
+
+thread_local! {
+    static CACHE: std::cell::RefCell<Cache>  = RefCell::new(Cache::default());
+}
+
+pub(crate) fn _cache<F: FftField>(mut x: E<F>) -> E<F> {
```
mrmr1993, line 15: Can the cache be passed as an argument (to the `constraints` function of the `Argument` trait) instead of using this cache here?

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
```diff
@@ -0,0 +1,40 @@
...
+    CACHE.with(|cache| x = cache.borrow_mut().cache(x.clone()));
+    x
+}
+
+pub(crate) fn two<F: FftField>() -> E<F> {
+    Expr::Constant(ConstantExpr::Literal(2u32.into()))
```
mrmr1993, line 21: There's already a `From<u64> for Expr<ConstantExpr<F>>`, can you use 2u64.into() below instead?

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
mrmr1993: +1

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_2.rs
mrmr1993: It might make sense to add some more commentary (`we know the poseidon selector is evaluated and not identically zero, but will be zero whenever this gate is active, so use it as a placeholder`).

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: Cool! Noted for future enhancement, but yah, it doesn't seem like this will be the bottleneck.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: indubitably so
```
87 |             match i {
   |                   ^ pattern `_` not covered
```

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_2.rs
mrmr1993: If memory serves, this was the hack that I suggested. Basically, this constraint is comprised of only the Poseidon selector
```rust
        constraints.push(E::<F>::cell(
            Column::Index(GateType::Poseidon),
            CurrOrNext::Curr,
        ));
```
without further details.
I think @mimoo's comment was referring to the fact that it's confusing to use poseidon here, so I was suggesting that the comment above should be updated to shed some extra light on *why* the poseidon selector appears here as the body of the constraint. Concretely, I was proposing updating the comment to something like
```rust
        // Temporary placeholder constraint.
        // We cannot use any expression that evaluates directly to the zero polynomial
        // here, because the zero group element is not valid as a commitment point, and
        // thus will cause proof generation / verification to fail.
        // Instead, we abuse our knowledge that the poseidon gate is never active at the
        // same time as this gate, that its selector is not identically zero, and that
        // the selector's evaluation is provided in the proof. This allows us to set this
        // constraint as the poseidon selector, which satisfies the requirement of being
        // zero whenever this constraint is active, but importantly also which lets us
        // create a valid (ie. non-zero) commitment.
        // Naturally, this has nothing to do with the poseidon gate; the selector was
        // just an evaluated/committed-to column that matched the requirements to fill
        // this hole until this constraint is implemented.
```

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
jspada: Makes sense, but for next iteration's PR

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
jspada: Cool - nice.  Yes

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
jspada: Will include this on the selector polynomial issue and I could update other gates to use this pattern also.

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
jspada: Spoke with Matthew, this was a nit and will be done in next iteration's PR.

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
jspada: It's used in more than one place.. basically any place someone creates `ProofEvaluations` and wants foreign field multiplication switched off.  Normally you'd just have to pass a single `F::zero()` but with complex gates comprised of many `CircuitGates` like Chacha, FFMUL and Cairo you'd need to pass an array of `F::zero()` of some length, which is annoying and entails knowing internal stuff like how many `CircuitGate` there are... Thus, a helper like this makes it easy to do the right things without having to know the specific details.

kimchi/src/linearization.rs
jspada: I'm thinking of the gate (aka gadget) as a collection of `CircuitGate`.

I think a gate (i.e. foreign_mul, chacha, cairo) is atomic.  That is, the caller must not reach in and use only part of it, such as individual rows/CircuitGates.  They are designed to be used all together, not individually.

Therefore, I think it makes sense to encapsulate the internal details, such as how many CircuitGates is it comprised of, so that the caller doesn't need to concern itself with these concerns.

Following this line of reasoning, I don't think the above code should need to manually build the gate from it's individual CircuitGates each time it's called.  Instead, I feel like the caller should go "Oi -- give us the foreign field multiplication combined constraints"

I think counting or tracking stats how many constraints are added should be done elsewhere and just commented if it is relevant for performance reasons.

kimchi/src/prover.rs
jspada: Yes!  Dropped

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
jspada: I was trying to avoid passing it around everywhere, but sure can do.  However, it's not being used, so I will just remove it until I actually need it for something.

utils/src/packed_modulus.rs
jspada: Sure - good suggestion.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,106 @@
...
+        //     Columns
+        //    R        0      1    2    3    4    5    6    7    8    9    10   11   12   13   14
+        //    o  Curr  limb   lp0  lp1  lp2  lp3  lp4  lp5  lc0  lc1  lc2  lc3  lc4  lc5  lc6  lc7  <- LSB
+        //    w               76   64   52   40   28   16   14   12   10    8    6    4    2    0
+        //    s
+        //
+        // Check limb =  lp0*2^0 + lp1*2^{12}  + ... + p5*2^{60}   + lc0*2^{72}  + lc1*2^{74}  + ... + lc7*2^{86}
+        //       w(0) = w(1)*2^0 + w(2)*2^{12} + ... + w(6)*2^{60} + w(7)*2^{72} + w(8)*2^{74} + ... + w(14)*2^{86}
+        //            = \sum i \in [1,7] 2^{12*(i - 1)}*w(i) + \sum i \in [8,14] 2^{2*(i - 7) + 6*12}*w(i)
+        let combined_sublimbs = (1..COLUMNS).fold(E::zero(), |acc, i| {
```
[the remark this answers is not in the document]
jspada: Ah, yes. I couldn't remember exactly what your comment was, so I removed one of the cases... but yeah, I can split it.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_1.rs
```diff
@@ -0,0 +1,125 @@
...
+        //    s
+        //
+        // Check   l2 = lp0*2^0          + lp1*2^{12}       + ... + lp3*2^{36}       + lc0*2^{48}     + lc1*2^{50}     + ... + lc19*2^{66}
+        //       w(0) = w_curr(1)*2^0    + w_curr(2)*2^{12} + ... + w_curr(4)*2^{36} + w_curr(5)*2^48 + w_curr(6)*2^50 + ... + w_curr(14)*2^66
+        //            + w_next(5)*2^{68} + w_next(6)*2^{70} + ... + w_next(14)*2^{86}
+        // (1st part) = \sum i \in [1,5] 2^{12*(i - 1)}*w_curr(i) + \sum i \in [6,14] 2^{2*(i - 5) + 4*12}*w_curr(i)
+        // (2nd part) + \sum i \in [5,14] 2^{2*(i - 5} + 68)*w_next(i)
+
+        // 1st part (Curr row): \sum i \in [1,5] 2^{12*(i - 1)}*w_curr(i) + \sum i \in [6,14] 2^{2*(i - 5) + 4*12}*w_curr(i)
+        let combined_sublimbs = (1..COLUMNS).fold(E::zero(), |acc, i| {
```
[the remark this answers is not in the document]
jspada: Sure.

kimchi/src/prover.rs
jspada: The comment says it's evaluating the polynomials without chunking them.  They are then used to compute the `ft` polynomial for Maller's optimization.

kimchi/src/verifier_index.rs
jspada: Certainly doesn't crash in the unit tests.

jspada: It is the same approach used by other gates, such as ec variable base scalar multiplication, endoscalar mul and chacha.

I'm not really sure I can give the correct answer, but I think it is doing this.
* The selector polynomial is generated by interpolating a polynomial whose degree is related to the number of rows, i.e., domain.d1
* We then evaluate this polynomial again, but this time using the large d8 domain (I think this is for efficiency/performance reasons)

@mrmr1993 Can you please advise?

mrmr1993: I think @mimoo is right, but the actual bad behaviour is shielded by another error: the verifier never checks that the evaluations match with the commitment (which should happen [like this](https://github.com/o1-labs/proof-systems/blob/60617995a56533af71fb439341046579decec5b6/kimchi/src/verifier.rs#L753)). It would be better to remove the evaluations and just use the commitments, if you are able to.

kimchi/src/prover.rs
mimoo: not sure I understand, didn't you add the evaluations of your selectors in the proof?

kimchi/src/circuits/polynomials/foreign_mul/constraints.rs
mimoo: maybe a good idea to delete it here? So I can review its actual usage :D this looks weird to me

kimchi/src/prover.rs
jspada: I'm not sure I understand either.  I'm following the pattern that other gates are using, such as generic and poseidon on the proceeding two lines.

mrmr1993: The generic and poseidon gates are special cases, your gate should mirror the behaviour of the other gates (e.g. varbasemul)

jspada: OK, now it makes sense!  Thanks 👍🏼

kimchi/src/verifier_index.rs
jspada: I thought this is creating the commitments
```
            foreign_mul_comm: self
                .cs
                .foreign_mul_selector_polys
                .iter()
                .map(|poly| {
                    self.srs
                        .commit_evaluations_non_hiding(domain, &poly.eval8, None)
                })
                .collect(),
```

kimchi/src/prover.rs
mimoo: to add colors: there's no reason to evaluate something unless you have to (because it gets multiplied with another commitment)

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_2.rs
jspada: This is an interesting `CircuitGate` because it has no constraints other than lookups, which causes it to have a zero polynomial and require the Poseidon workaround.

It seems like we may want another way to handle situations like this without needing this workaround with Poseidon.  @mrmr1993 - are there any other solutions you can think of?

mrmr1993: You shouldn't need a selector or any constraints for this gate then, only the lookup pattern. I had basically the same scenario when implementing the `Lookup` gate (#399), so that PR might serve as a useful guide.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,91 @@
...
+        //     Columns
+        //    R        0      1    2    3    4    5    6    7    8    9    10   11   12   13   14
+        //    o  Curr  limb   lp0  lp1  lp2  lp3  lp4  lp5  lc0  lc1  lc2  lc3  lc4  lc5  lc6  lc7  <- LSB
+        //    w               76   64   52   40   28   16   14   12   10    8    6    4    2    0
+        //    s
+        //
+        // Check limb =  lp0*2^0 + lp1*2^{12}  + ... + p5*2^{60}   + lc0*2^{72}  + lc1*2^{74}  + ... + lc7*2^{86}
+        //       w(0) = w(1)*2^0 + w(2)*2^{12} + ... + w(6)*2^{60} + w(7)*2^{72} + w(8)*2^{74} + ... + w(14)*2^{86}
+        //            = \sum i \in [1,7] 2^{12*(i - 1)}*w(i) + \sum i \in [8,14] 2^{2*(i - 7) + 6*12}*w(i)
+        let combined_sublimbs = (1..8).fold(E::zero(), |acc, i| {
```
mrmr1993, line 78: It took me some thought to check that using `8` as the limit here was correct while `7` is the limit above. It would be good to add a comment explaining (or refactor as below).

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,91 @@
...
+    const CONSTRAINTS: u32 = 9;
+
+    // Constraints for ForeignMul0
+    //   * Operates on Curr row
+    //   * Range constrain all sublimbs except p4 and p5 (baring plookup constraints, which are done elsewhere)
+    //   * Constrain that combining all sublimbs equals the limb stored in column 0
+    fn constraints() -> Vec<E<F>> {
+        // Row structure
+        //  Column w(i) 0    1       ... 4       5    6    7     ... 14
+        //        Curr  limb plookup ... plookup copy copy crumb ... crumb
```
mrmr1993, line 54: What is `copy` here? It's not obvious where this is intended to copy to/from, which makes it tricky to understand or check that this gate is correct in isolation.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,91 @@
...
+    const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::ForeignMul0);
+    const CONSTRAINTS: u32 = 9;
+
+    // Constraints for ForeignMul0
+    //   * Operates on Curr row
+    //   * Range constrain all sublimbs except p4 and p5 (baring plookup constraints, which are done elsewhere)
+    //   * Constrain that combining all sublimbs equals the limb stored in column 0
+    fn constraints() -> Vec<E<F>> {
+        // Row structure
+        //  Column w(i) 0    1       ... 4       5    6    7     ... 14
```
mrmr1993, line 53: `Column w(i)` and `Curr` don't feel like they're adding much here apart from visual noise. Delete?

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,91 @@
+/// ForeignMul0 - Foreign field element constraints
+///
+///    Foreign field element F is comprised of three 88-bit limbs L0L1L2
+///
+///    * This circuit gate is used to partially constrain L0 or L1
+///    * The rest of L0 and L1 are constrained by a single ForeignMul2
+///    * It operates on the Curr row
+///
+/// Column | Curr
```
mrmr1993, line 9: Is it useful to have the layout here without talking about the constraints? I would prefer if you could move the description of the constraints here too.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_1.rs
```diff
@@ -0,0 +1,108 @@
+/// ForeignMul1 - Foreign field element constraints
+///
+///    Foreign field element F is comprised of three 88-bit limbs L0L1L2
+///
+///    * This circuit gate is used to fully constrain L2
```
mrmr1993, line 5: Much the same comments apply to this file as `foreign_mul_0.rs`

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
```diff
@@ -0,0 +1,505 @@
...
+                wires: wires[1],
+                coeffs: vec![],
+            },
+            CircuitGate {
+                typ: GateType::ForeignMul1,
+                wires: wires[2],
+                coeffs: vec![],
+            },
+            CircuitGate {
+                typ: GateType::ForeignMul2,
```
mrmr1993, line 49: Since `ForeignMul2` is only useful to apply the lookups in the row after `ForeignMul1`, it probably makes more sense to use `ForeignMul1, Next` as the trigger for the lookups vs `ForeignMul2, Curr`.

This can be done in the follow-up lookup PR, doesn't need to be here.

kimchi/src/circuits/expr.rs
```diff
@@ -37,6 +56,8 @@ pub struct Constants<F> {
     pub endo_coefficient: F,
     /// The MDS matrix
     pub mds: Vec<Vec<F>>,
+    /// Foreign field modulus
+    pub foreign_modulus: Vec<F>,
```
mrmr1993, line 60: This appears to be unused.. Delete it until it's needed?

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
```diff
@@ -0,0 +1,104 @@
+//! Foreign field multiplication gate
```
mrmr1993, line 1: As currently implemented, this collection of gates is more of an 88-bit range-check gate than anything to do with foreign field arithmetic. Worth renaming?

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// <https://hackmd.io/XZUHHGpDQsSOs0dUGugB5w>
+///
+/// Globals:
+///     * n: native field modulus
+///     * p: foreign field modulus
+///
+/// Inputs:
+///     * a: left foreign field element operand a \in Fp
```
mrmr1993, line 10: `a` isn't really 'in `Fp`', right? It's probably more correct to say 'as Fp limbs', this feels confusing as currently written.

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+///
+/// Witness:
+///     * q: \in Fp
+///     * r: such that a*b = q*p +r
+///
+/// Foreign field element structure:
+///
+///    Each foreign field element a is decomposed into three 88-bit limbs a0, a1, a2 s.t. a = a0a1a2 in
+///    little-endian byte order (i.e. a = a2*2^{2b} + a1*2^b + a0)
+///      a0 = a0p0a0p1a0p2a0p3a0p4a0p5a0c0a0c1a0c2a0c3a0c4a0c5a0c6a0c7
```
mrmr1993, line 21: I'm not sure that the numbering helps much here. For me, it would be clearer to have
```
L is a 12-bit lookup,
C is a 2-bit crumb.
     <----6----> <------8------>
a0 = L L L L L L C C C C C C C C
a1 = L L L L L L C C C C C C C C
     <--4--> <------------------20----------------->
a2 = L L L L C C C C C C C C C C C C C C C C C C C C
```

At a minimum, the repeated `a0`/`a1`/`a2` prefix adds a bunch of noise for not much benefit.

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+///     7 | crumb a0c0   | crumb a1c0   | crumb a2c2   | crumb a2c12
+///     8 | crumb a0c1   | crumb a1c1   | crumb a2c3   | crumb a2c13
+///     9 | crumb a0c2   | crumb a1c2   | crumb a2c4   | crumb a2c14
+///    10 | crumb a0c3   | crumb a1c3   | crumb a2c5   | crumb a2c15
+///    11 | crumb a0c4   | crumb a1c4   | crumb a2c6   | crumb a2c16
+///    12 | crumb a0c5   | crumb a1c5   | crumb a2c7   | crumb a2c17
+///    13 | crumb a0c6   | crumb a1c6   | crumb a2c8   | crumb a2c18
+///    14 | crumb a0c7   | crumb a1c7   | crumb a2c9   | crumb a2c19
+///
+/// Gate:   ForeignMul0    ForeignMul0    ForeignMul1    ForeignMul2
```
mrmr1993, line 73: Moving this above the table in place of the `0 1 2 3` header would make this a little cleaner.

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
```diff
@@ -0,0 +1,211 @@
...
+// An cell containing zero
+struct ZeroWitnessCell;
+impl ZeroWitnessCell {
+    const fn create() -> WitnessCell {
+        WitnessCell::Zero
+    }
+}
+
+// Generate witness in shape that constraints expect (TODO: static for now, make dynamic)
+const WITNESS_SHAPE: [[WitnessCell; COLUMNS]; 4] = [
```
mrmr1993, line 69: This spells out the layout of the gates much more clearly than the current code for the gates does. Is it possible to re-use / adapt this representation for that?

kimchi/src/circuits/constraints.rs
```diff
@@ -119,6 +125,10 @@ pub struct ConstraintSystem<F: FftField> {
     #[serde(skip)]
     pub fr_sponge_params: ArithmeticSpongeParams<F>,
 
+    /// Foreign field modulus parameter
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub foreign_modulus: Vec<F>,
```
mrmr1993, line 130: Currently unused; remove?

mrmr1993 (changes requested):

kimchi/src/circuits/constraints.rs
jspada: It's used.  I set it in the unit tests.

mrmr1993: Replace with a boolean flag? These gates aren't really doing any foreign field arithmetic, and they don't use this modulus..

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
jspada: Agree - thanks!

kimchi/src/circuits/constraints.rs
jspada: Which gates do you mean?

I could remove it, but then I will need to put it right back in the next PR because the next constraints will be using it.  My goal for this PR was that it hooks up everything I need for the next set of constraints using the foreign modulus.

The circuit developer needs a way to specify the foreign modulus.  How do you propose this is specified and passed to the gate?  Izaak indicated he didn't want it passed as a gate input nor in the coefficients.  If there's a better or more desirable way, I'll do it.  What's your idea?

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
jspada: I'm not sure.  I guess this would be an alternative format to using the expr framework or some kind of wrapper around it.

It's an interesting idea.  Perhaps I could traverse this structure in the `constraints()` function and auto-generate the constraints from it.  It's a cool idea, but would be a big change.  I propose to do it in the subsequent PR.

Also, it feels like it would be good to re-use the same code for both witness generation and generating the constraints.  I think this is very similar to the code you and Izzy have work on for creating circuits, which I would rather use than re-invent.  So maybe it makes more sense I convert things to the new circuit interface once it's ready?

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
```diff
@@ -0,0 +1,211 @@
...
+    /* row 1, ForeignMul0 row */
+    [
+        LimbWitnessCell::create(),
+        /* 12-bit plookups */
+        SublimbWitnessCell::create(0, 0, 0, 12),
+        SublimbWitnessCell::create(0, 0, 12, 24),
+        SublimbWitnessCell::create(0, 0, 24, 36),
+        SublimbWitnessCell::create(0, 0, 36, 48),
+        /* 12-bit copies */
```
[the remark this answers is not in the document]
jspada: Yes!

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
jspada: 👍🏼

kimchi/src/circuits/polynomials/foreign_mul/common.rs
```diff
@@ -0,0 +1,24 @@
...
+pub(crate) fn two<F: FftField>() -> E<F> {
+    2u64.into()
+}
+
+pub(crate) fn three<F: FftField>() -> E<F> {
+    3u64.into()
+}
+
+// Crumb constraint for 2-bit sublimb
+pub(crate) fn sublimb_crumb_constraint<F: FftField>(sublimb: &E<F>) -> E<F> {
```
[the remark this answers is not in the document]
jspada: I was thinking of moving this as an Expr helper in a subsequent PR.  AFAIK, it only makes sense to use degree-n constraints up to n=4, so we could have a helper for these to avoid re-implementing all the time.  wdyt?

kimchi/src/circuits/gate.rs
```diff
@@ -149,13 +153,13 @@ impl<F: FftField> CircuitGate<F> {
             VarBaseMul => self.verify_vbmul(row, witness),
             EndoMul => self.verify_endomul(row, witness, cs),
             EndoMulScalar => self.verify_endomul_scalar(row, witness, cs),
-            // TODO: implement the verification for chacha
```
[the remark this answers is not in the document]
jspada: Ah, I screwed up a merge.  Fixed!

kimchi/src/circuits/polynomials/foreign_mul/common.rs
```diff
@@ -0,0 +1,24 @@
+//!
+//! Common constraints and related helpers
+//!
+
+use ark_ff::{FftField, One};
+
+use crate::circuits::expr::E;
+
+pub(crate) fn two<F: FftField>() -> E<F> {
```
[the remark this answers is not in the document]
jspada: Let me try.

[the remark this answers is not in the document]
jspada: No, but `E::<F>::from(2u64)` works and seems nicer.

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
jspada: Well Fp is the foreign field and Fn is native (see above).

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: It's described in the comments for the whole layout, but I can also comment it here.

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
mrmr1993: Note that Izzy's circuit DSL is for creating circuits, not laying out individual gates. I don't think you'll be able to use it for this.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: I think so.  So you can read the comment and understand each individually.  Added description of constraints.

kimchi/src/circuits/polynomials/foreign_mul/gate.rs
jspada: Thanks - noted!

kimchi/src/circuits/expr.rs
jspada: Will convert this PR & branch into `RangeCheck` and move the foreign mul code into another branch.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
jspada: Fixed by adding comments above.

kimchi/src/circuits/polynomials/foreign_mul/foreign_mul_0.rs
```diff
@@ -0,0 +1,91 @@
+/// ForeignMul0 - Foreign field element constraints
+///
+///    Foreign field element F is comprised of three 88-bit limbs L0L1L2
+///
+///    * This circuit gate is used to partially constrain L0 or L1
```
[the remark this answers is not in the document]
jspada: Even if I put them in the same file they'd still be separate `CircuitGates` and these comments would still be needed describe how they are related.  The developer would have to jump around inside the same file to find them.  I really feel breaking them down and fully documenting them in separate files is is cleaner, so I'd much prefer to to keep it that way.

kimchi/src/circuits/constraints.rs
jspada: I see you mean to extract everything that's range check related into a `RangeCheck` gate.  For the range check gate this indeed can be removed.  I'll do this.

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
jspada: Interesting.  So something similar to that for specifying the gate layout and generating constraints perhaps.

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
jspada: Fixed

kimchi/src/circuits/polynomials/foreign_mul/witness.rs
jspada: This will be done in a subsequent PR.

kimchi/src/circuits/polynomials/foreign_mul/mod.rs
jspada: Done!

kimchi/src/circuits/polynomials/range_check/mod.rs
```diff
@@ -0,0 +1,97 @@
+//! Range check gate
+
+///```text
```
mrmr1993, line 3: Can you move this comment into the `circuitgates.rs` file?

kimchi/src/circuits/polynomials/range_check/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+/// Range check field element structure:
+///
+///    Each field element a is decomposed into three 88-bit limbs a0, a1, a2 s.t. a = a0a1a2 in
```
mrmr1993, line 6: This is no longer true of this gate; all it's doing is 3 88-bit range checks. Can you rewrite this as `Every batch of 4 range-check gates performs 3 88-bit range checks` or equivalent?

kimchi/src/circuits/polynomials/range_check/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+///    a2 = L L L L C C C C C C C C C C C C C C C C C C C C
+///
+/// Input structure:
+///
+///   Row*  Contents**
+///     0   a0
+///     1   a1
+///     2   a2
+///     3   a0,a1,a2
+///     4   b0
```
mrmr1993, line 25: There's no need for rows 4-7 here

kimchi/src/circuits/polynomials/range_check/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+///     5   b1
+///     6   b2
+///     7   b0,b1,b2
+///
+///    (*)  Row offsets
+///    (**) Some part of the limb is contained in this row
```
mrmr1993, line 31: This footnote is important to understanding the table above. I would be tempted to rewrite the whole thing as brief prose, e.g.
```
Each of the first 3 gates checks most of a different range-check input. The final gate performs the remaining checks for all 3 inputs.
```

kimchi/src/circuits/polynomials/range_check/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+///    10 | crumb a0c3   | crumb a1c3   | crumb a2c5   | crumb a2c15
+///    11 | crumb a0c4   | crumb a1c4   | crumb a2c6   | crumb a2c16
+///    12 | crumb a0c5   | crumb a1c5   | crumb a2c7   | crumb a2c17
+///    13 | crumb a0c6   | crumb a1c6   | crumb a2c8   | crumb a2c18
+///    14 | crumb a0c7   | crumb a1c7   | crumb a2c9   | crumb a2c19
+///
+///   The 12-bit chunks are constrained with plookups and the 2-bit crumbs constrained with
+///   degree-4 constraints of the form x*(x - 1)*(x - 2)*(x - 3).
+///
+///   Note that copy denotes a plookup that is deferred to the RangeCheck2 gate.
```
mrmr1993, line 70: Nit: mention `via the permutation argument`?

kimchi/src/circuits/polynomials/range_check/mod.rs
jspada: Fixed!

kimchi/src/circuits/polynomials/range_check/mod.rs
jspada: 👍🏼

kimchi/src/circuits/polynomials/range_check/mod.rs
jspada: 👍🏼

kimchi/src/circuits/polynomials/range_check/mod.rs
jspada: 👍🏼  true story. I could not find a way to constraint the whole 264-bit element since it doesn't fit anywhere.  I will comment this.

kimchi/src/circuits/polynomials/range_check/mod.rs
jspada: Done!

mrmr1993 (approved):

PR #523, opened 2022-04-29
fixing mdbook admonish error

querolita:
There seems to be a problem regarding the version of `mdbook-admonish`, this PR tries to fix it. 

The details of the error are here: 
https://github.com/o1-labs/proof-systems/runs/6231484369?check_suite_focus=true


jspada (approved):
lgtm!

PR #524, opened 2022-04-29
renaming of ZK_ROWS

bufferhe4d:
Splitted ZK_ROWS into ZK_ROWS and ACC_ROWS with EVAL_ROWS = ZK_ROWS + ACC_ROWS.

### DIFF
--- a/kimchi/src/circuits/polynomials/permutation.rs
+++ b/kimchi/src/circuits/polynomials/permutation.rs
@@ -56,11 +56,16 @@ use rand::{CryptoRng, RngCore};
 
 /// Number of constraints produced by the argument.
 pub const CONSTRAINTS: u32 = 3;
-pub const ZK_ROWS: u64 = 3;
+/// Number of randomized rows (1 per evaluation).
+pub const ZK_ROWS: u64 = 2;
+/// Number of rows for the final accumulator value.
+pub const ACC_ROWS: u64 = 1;
+/// Total number of evaluation rows.
+pub const EVAL_ROWS: u64 = ZK_ROWS + ACC_ROWS;
 /// Evaluates the polynomial
 /// (x - w^{n - 4}) (x - w^{n - 3}) * (x - w^{n - 2}) * (x - w^{n - 1})
 pub fn eval_vanishes_on_last_4_rows<F: FftField>(domain: D<F>, x: F) -> F {
-    let w4 = domain.group_gen.pow(&[domain.size - (ZK_ROWS + 1)]);
+    let w4 = domain.group_gen.pow(&[domain.size - (EVAL_ROWS + 1)]);
     let w3 = domain.group_gen * w4;
     let w2 = domain.group_gen * w3;
     let w1 = domain.group_gen * w2;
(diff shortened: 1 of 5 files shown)


kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -56,11 +56,16 @@ use rand::{CryptoRng, RngCore};
 
 /// Number of constraints produced by the argument.
 pub const CONSTRAINTS: u32 = 3;
-pub const ZK_ROWS: u64 = 3;
+/// Number of randomized rows (1 per evaluation).
+pub const ZK_ROWS: u64 = 2;
+/// Number of rows for the final accumulator value.
+pub const ACC_ROWS: u64 = 1;
+/// Total number of evaluation rows.
+pub const EVAL_ROWS: u64 = ZK_ROWS + ACC_ROWS;
```
mimoo, line 64: a few things here:

* `EVAL_ROWS` is not a good name, as it is the number of evaluation points, not evaluation rows. `EVALS` could be better. This is supposed to be set to 2 because we evaluate things at zeta, and zeta * omega (see the PLONK video series for an explanation on that)
* `EVALS` is not related to the permutation, it's also not related to the `circuits/` as it is a protocol value. Maybe a good idea to move it to `kimchi/src/lib.rs`

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -56,11 +56,16 @@ use rand::{CryptoRng, RngCore};
...
 /// Evaluates the polynomial
 /// (x - w^{n - 4}) (x - w^{n - 3}) * (x - w^{n - 2}) * (x - w^{n - 1})
 pub fn eval_vanishes_on_last_4_rows<F: FftField>(domain: D<F>, x: F) -> F {
-    let w4 = domain.group_gen.pow(&[domain.size - (ZK_ROWS + 1)]);
+    let w4 = domain.group_gen.pow(&[domain.size - (EVAL_ROWS + 1)]);
```
mimoo, line 68: maybe this value here should be called `PERM_FINAL_CONSTRAINTS`, as it represents the final constraints as explained here https://o1-labs.github.io/proof-systems/plonk/zkpm.html

so I think it'd be make sense to define something like


`const PERM_FINAL_CONSTRAINTS: usize = EVALS /* zk rows */ + 1 /* end of accumulator */;`

kimchi/src/prover.rs
```diff
@@ -110,8 +110,8 @@ where
             // padding
             w.extend(std::iter::repeat(ScalarField::<G>::zero()).take(length_padding));
 
-            // zk-rows
-            for row in w.iter_mut().rev().take(ZK_ROWS as usize) {
+            // padding for eval-rows = zk-rows + acc-rows
```
mimoo, line 113: this is not padding, this is the zero-knowledge rows

kimchi/src/circuits/polynomials/permutation.rs
mimoo: (also, might be a good idea to just have these values be u32 instead of u64)

mimoo (commented):
Left you a few comments, thanks for the PR :)

kimchi/src/prover.rs
bufferhe4d: Can't we use `PERM_FINAL_CONSTRAINTS` here since it is also equal to `EVALS + 1` with the new approach?

mimoo: I guess there's a number of different polynomials/arguments that we zk-ify:

- witness (here)
- permutation
- lookup

and they all rely on the global number of evaluations (which is 2, but we can set it to 3 just to make sure we have enough random values)

mimoo: (btw don't hesitate to hit me up on twitter if you want to talk more about this)

PR #525, opened 2022-05-02
[lookup] handle different table id 0

mimoo:
This PR moves the lookup constraint system to its own module so that we can specify it. It also refactors it as explained below.

We currently use an XOR table with id = 0 and the last dummy entry `(0, 0, 0)`. We then generalize based on these assumptions:

* we will always use the XOR table when doing lookups
* the XOR table will always be the only table with id = 0

This is in part so that we can use the dummy entry. 

This PR tries to improve the situation to avoid users shooting themselves in the foot:

* it enforces that we cannot use different lookup tables with the same ID
* it enforces that whatever is the table with ID = 0 then it has a dummy entry that is all zeros

This should allow us to do lookups without having to use the XOR table, as long as the first table has an entry with all-zeroes (BTW do we need this entry to be the last one? This is what we check in the XOR table but I don't see how it's important anymore)
If the tables you want to use don't have a single entry of all zeros, then create a DummyTable with ID=0 that has a single entry of all zeros.

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -1,10 +1,12 @@
 //! This module implements Plonk circuit constraint primitive.
 
-use crate::circuits::domain_constant_evaluation::DomainConstantEvaluations;
 use crate::circuits::{
+    domain_constant_evaluation::DomainConstantEvaluations,
     domains::EvaluationDomains,
     gate::{CircuitGate, GateType},
+    lookup::{index::LookupConstraintSystem, tables::LookupTable},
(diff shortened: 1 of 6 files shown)


mrmr1993 (changes requested):
Blocking this, as discussed out-of-band. In particular, I am opposed to this:
> it enforces that we cannot use different lookup tables with the same ID

PRs also shouldn't move code around and add specs while also making logic changes.

Happy to approve a PR that does just this
> it enforces that whatever is the table with ID = 0 then it has a dummy entry that is all zeros

mimoo:
cc @mrmr1993  I created this PR: https://github.com/o1-labs/proof-systems/pull/536 (in order to preserve this one)

PR #532, opened 2022-05-03
[book] add more to domain section

mimoo:


mimoo:
cc @querolita @jspada

book/src/plonk/domain.md
```diff
@@ -1,14 +1,62 @@
...
+```
+
+The 2-adicity of 32 means that there's a multiplicative subgroup of size $2^{32}$ that exists in the field.
+The code above also defines a generator $g$ for it, such that $g^{2^{32}} = 1$ and $g^i \neq 1$ for $i \in [[1, 2^{32}]]$ (so it's a **primitive** $2^{32}$-th root of unity).
+
+[Lagrange's theorem](https://en.wikipedia.org/wiki/Lagrange%27s_theorem_(group_theory\)) tells us that if we have a group of order $n$, then we'll have subgroups with orders dividing $n$. So in our case, we have subgroups with all the powers of 2, up to the 32-th power of 2.
+
+To find any of these groups, it is pretty straight forward as well. Notice that:
+
+* let $h = g^2$, then $h^{2^{31}} = g^{2^{32}} = 1$ and so $h$ generates a subgroup of order 31
```
querolita, line 36: perhaps I would use different letters for these two examples, because in the first line h = g^2 and in the next one h=g^4, perhaps just h' could be more clear

book/src/plonk/domain.md
```diff
@@ -1,14 +1,62 @@
...
+for _ in log_size_of_group..Self::TWO_ADICITY {
+    omega.square_in_place();
+}
+```
+
+this allows you to easily find subgroups of different sizes of powers of 2, which is useful in zero-knowledge proof systems as FFT optimizations apply well on domains that are powers of 2. You can read more about that in the [mina book](https://o1-labs.github.io/proof-systems/fundamentals/zkbook_fft.html).
+
+## Efficient computation of the vanishing polynomial
+
+For each curve, we generate a domain $H$ as the set $H = {1, \omega, \omega^2, \cdots}$.
```
querolita, line 55: add the final \omega^{n-1} perhaps? otherwise it may look infinite

querolita (approved):
Very nice addition, I really like having this as part of the book 😄

PR #536, opened 2022-05-04
[lookup] move and specify index

mimoo:


kimchi/src/circuits/lookup/index.rs
```diff
@@ -0,0 +1,197 @@
...
+
+                //~ 4. Get the highest number of columns `max_table_width`
+                //~    that a lookup table can have.
+                let max_table_width = lookup_tables
+                    .iter()
+                    .map(|table| table.data.len())
+                    .max()
+                    .unwrap_or(0);
+
+                //~ 5. Add the table ID stuff
```
mrmr1993, line 100: I'm not sure that this comment is adding value (to the code or the spec). Delete?

kimchi/src/circuits/lookup/index.rs
```diff
@@ -0,0 +1,197 @@
...
+                let mut lookup_table = vec![Vec::with_capacity(d1_size); max_table_width];
+                let mut table_ids: Vec<F> = Vec::with_capacity(d1_size);
+                let mut non_zero_table_id = false;
+                //~ 6. For each table:
+                for table in lookup_tables.iter() {
+                    let table_len = table.data[0].len();
+
+                    //~ b. Make sure that if table with id 0 is used, then it's the XOR table.
```
mrmr1993, line 108: Nit: `a.`?

kimchi/src/circuits/lookup/index.rs
```diff
@@ -0,0 +1,197 @@
...
+                    //~    We do this because we use a table with id 0 and
+                    //~
+                    if table.id == 0 {
+                        if !table.has_zero_entry() {
+                            return Err(LookupError::TableIDZeroMustHaveZeroEntry);
```
mrmr1993, line 113: Don't do this here, this will mess with any non-fixed or partial table with ID 0 even if it has a counterpart that is fixed and contains 0.

If you really want to add this check, accumulate outside of the loop with
```rust
let mut has_table_id_0 = false;
let mut has_table_id_0_with_zero_entry = false;
...
  if table.id == 0 {
    has_table_id_0 = true;
    if table.has_zero_entry() {
      has_table_id_0_with_zero_entry = true;
    }
  }
...
if (has_table_id_0 && !has_table_id_0_with_zero_entry) {
  return Err(LookupError::TableIDZeroMustHaveZeroEntry);
}
```

kimchi/src/circuits/lookup/index.rs
```diff
@@ -0,0 +1,197 @@
...
+                        }
+                    }
+
+                    // Update table IDs
+                    if table.id != 0 {
+                        non_zero_table_id = true;
+                    }
+                    //~ c. Update table IDs
```
mrmr1993, line 121: Nit: `b`

mrmr1993: This seems under-detailed for a spec. Shouldn't this be something like 'extend the table IDs column with the table ID for this table, repeated for each entry in the table' or some such?

kimchi/src/circuits/lookup/index.rs
```diff
@@ -0,0 +1,197 @@
...
+                    let table_id: F = i32_to_field(table.id);
+                    table_ids.extend(repeat_n(table_id, table_len));
+
+                    //~ d. Update lookup_table values
```
mrmr1993, line 125: Nit: `c`

kimchi/src/circuits/lookup/index.rs
```diff
@@ -0,0 +1,197 @@
...
+                    for (i, col) in table.data.iter().enumerate() {
+                        if col.len() != table_len {
+                            return Err(LookupError::InconsistentTableLength);
+                        }
+                        lookup_table[i].extend(col);
+                    }
+
+                    //~ e. Fill in any unused columns with 0 to match the dummy value
```
mrmr1993, line 133: Nit: `d`

mrmr1993 (changes requested):

kimchi/src/circuits/lookup/index.rs
mimoo: I thought we said that non-fixed tables should have negative IDs?

mrmr1993: That's a detail for the library user (ie. snarky / SnarkyJS) to decide.

kimchi/src/circuits/lookup/index.rs
mimoo: what do you mean by "lifting to the lifted check"?

kimchi/src/circuits/lookup/index.rs
mimoo: I think we're creating a lot of footguns for users here. Maybe let's chat about that tomorrow as well.

mrmr1993 (approved):

PR #539, opened 2022-05-05
create srs in verifier_index if it is not passed in

chee-chyuan:
Verifier_index has `serde_skip` on srs and it requires user to pass in srs to make it usable. 
This pr allows user to use verifier_index without having user to pass it into the `from_file` function


chee-chyuan:
a doubt i have in mind is that if user pass in srs without lagrange_basis, the verification progress will fail. im wondering how can i make the arc mutable so i can also call `add_lagrange_basis`

mimoo:
> a doubt i have in mind is that if user pass in srs without lagrange_basis, the verification progress will fail. im wondering how can i make the arc mutable so i can also call add_lagrange_basis

I never understood why we don't generate these lagrange basis lazily on-demand. Maybe that's something we should do...

mrmr1993:
> I never understood why we don't generate these lagrange basis lazily on-demand. Maybe that's something we should do...

I'd rather we didn't. This is essentially building in a memory leak. Generally the library users should make decisions about memory management, rather than the library itself.

mimoo:
Yeah that makes sense

kimchi/src/tests/framework.rs
```diff
@@ -134,4 +139,51 @@ impl TestFramework {
...
+        verifier_index_deserialize.powers_of_alpha = index.powers_of_alpha;
+        verifier_index_deserialize.linearization = index.linearization;
+        verifier_index_deserialize.w = zk_w3(verifier_index_deserialize.domain);
+
+        // verify the proof
+        let start = Instant::now();
+        verify::<Affine, BaseSponge, ScalarSponge>(&group_map, &verifier_index_deserialize, &proof)
+            .unwrap();
+        println!("- time to verify: {}ms", start.elapsed().as_millis());
+    }
```
mimoo, line 188: don't think this should be in `framework.rs` btw, as this looks more like a test than boilerplate code that is useful for creating other tests

kimchi/src/tests/framework.rs
```diff
@@ -134,4 +139,51 @@ impl TestFramework {
...
+        public: &[Fp],
+    ) {
+        // create the index
+        let start = Instant::now();
+        let index = new_index_for_test(gates, public.len());
+        let verifier_index = index.verifier_index();
+        println!("- time to create index: {:?}s", start.elapsed().as_secs());
+
+        let verifier_index_serialize =
+            serde_json::to_string(&verifier_index).expect("couldn't serialize index");
```
mimoo, line 156: thinking out loud: wondering how big the `.json` will be :o will it even be readable? (I'm assuming a lot of the binary stuff doesn't encode nicely to json, and will take a huge amount of space)

kimchi/src/tests/framework.rs
```diff
@@ -134,4 +139,51 @@ impl TestFramework {
...
+        // add srs with lagrange bases
+        let mut srs = SRS::<GroupAffine<VestaParameters>>::create(verifier_index.max_poly_size);
+        srs.add_lagrange_basis(verifier_index.domain);
+        verifier_index_deserialize.srs = Arc::new(srs);
+        verifier_index_deserialize.fq_sponge_params = oracle::pasta::fq_kimchi::params();
+        verifier_index_deserialize.fr_sponge_params = oracle::pasta::fp_kimchi::params();
+        verifier_index_deserialize.zkpm = zk_polynomial(verifier_index_deserialize.domain);
+        verifier_index_deserialize.powers_of_alpha = index.powers_of_alpha;
+        verifier_index_deserialize.linearization = index.linearization;
+        verifier_index_deserialize.w = zk_w3(verifier_index_deserialize.domain);
```
mimoo, line 181: Nice! I think we can have a PR that adds this test. This is super useful in itself if the test runs. 

Then we should try to remove most of these steps to make the deserialization easy to the user

kimchi/src/verifier_index.rs
```diff
@@ -237,7 +237,15 @@ where
...
         // fill in the rest
-        verifier_index.srs = srs;
+        verifier_index.srs = match srs {
+            Some(t) => t,
+            None => {
+                let mut srs = SRS::<G>::create(verifier_index.max_poly_size);
+                srs.add_lagrange_basis(verifier_index.domain);
+                Arc::new(srs)
+            }
+        };
```
mimoo, line 247: just FYI this works too and is a bit shorter:

```rust
srs.unwrap_or_else(|| {
  let mut srs = SRS::<G>::create(verifier_index.max_poly_size);
  srs.add_lagrange_basis(verifier_index.domain);
  Arc::new(srs)
});
```

mimoo (commented):
Nice :) I think this serves as a nice start for PRs that can improve deserialization (and for people who want to see how to deserialize an index atm)

kimchi/src/tests/framework.rs
chee-chyuan: shall i extract this part out and place it in the newly created `serialization.rs` ?

mimoo: yeah I think that's a good idea

kimchi/src/verifier_index.rs
```diff
@@ -232,13 +232,13 @@ where
     }
 }
 
-impl<G> VerifierIndex<G>
+impl<G: CommitmentCurve> VerifierIndex<G>
 where
-    G: CommitmentCurve,
+    G::BaseField: PrimeField,
```
mimoo, line 237: I'm not sure I understand why this change ^

mimoo (approved):
LGTM! Thanks :)

PR #547, opened 2022-05-09
[lookup] simplify code

mimoo:
check the commits, they:

1. remove the type `UncombinedEntry` as it was deadcode
2. remove the type `CombinedEntry` (which allows us to remove the trait `Entry`) as it doesn't seem to serve much purpose
3. simplify `sorted` to take an iterator instead of a closure, this simplifies the signature and the code


mrmr1993 (approved):

mimoo:
thanks! I'll get this in to avoid a rebase on the other PR, but will work on a PR to address comments right away

PR #548, opened 2022-05-09
[lookup] compute the joint lookup table ONCE and only once

mimoo:
The prover computes the lookup argument by computing:

1. the joint table
2. the numerator part of the argument (using an `f_part` and `t_part`)
3. the denominator part of the argument (called `sorted`)

Both the numerator and the denominator make use of the joint table. Looking at the spec, we compute them in this order: 3 - 1 - 2. 

But why not do 1 - 3 - 2 and use the computed joint table in the computation of the denominator as well?

This PR fixes this.

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -34,10 +34,7 @@ use commitment_dlog::commitment::{b_poly_coefficients, CommitmentCurve, PolyComm
 use itertools::Itertools;
 use o1_utils::{types::fields::*, ExtendedDensePolynomial as _};
 use oracle::{sponge::ScalarChallenge, FqSponge};
-use rayon::iter::{
-    IndexedParallelIterator, IntoParallelRefIterator, IntoParallelRefMutIterator, ParallelIterator,
-};
-use std::{collections::HashMap, iter};
+use std::collections::HashMap;
 
 /// The result of a proof creation or verification.
 type Result<T> = std::result::Result<T, ProverError>;


kimchi/src/prover.rs
```diff
@@ -234,27 +234,41 @@ where
...
-                    // If there is no `table_ids8` in the constraint system,
-                    // every table ID is identically 0.
-                    {
-                        ScalarField::<G>::zero()
-                    }
-                };
-                combine_table_entry(&joint_combiner, &table_id_combiner, row, &table_id)
-            });
+            //~     - Compute the lookup table values as the combination of the lookup table entries.
+            let joint_lookup_table_d8 = {
```
mimoo, line 238: instead of computing a iterator of the joint lookup table, we compute the evaluation on d8

kimchi/src/prover.rs
```diff
@@ -296,6 +310,7 @@ where
             lookup_context.sorted_coeffs = Some(sorted_coeffs);
             lookup_context.sorted_comm = Some(sorted_comm);
             lookup_context.sorted8 = Some(sorted8);
+            lookup_context.joint_lookup_table_d8 = Some(joint_lookup_table_d8);
```
mimoo, line 313: and then we save this in our context, for use later

kimchi/src/prover.rs
```diff
@@ -305,32 +320,22 @@ where
...
-                    combine_table_entry(
-                        &lookup_context.joint_combiner.unwrap(),
-                        &lookup_context.table_id_combiner.unwrap(),
-                        row,
-                        &table_id,
-                    )
-                })
+            let joint_lookup_table_d8 = lookup_context.joint_lookup_table_d8.as_ref().unwrap();
+
+            let joint_lookup_table = {
```
mimoo, line 327: ignore this block, as I only did this to see if the tests would still pass. If I get a 👍 for the approach taken in this PR I'll implement the changes to `aggregation()` in order to leverage this `joint_lookup_table_d8` (the same way I modified `sorted()`)

kimchi/src/prover.rs
```diff
@@ -382,44 +387,13 @@ where
 
         //~ 20. If using lookup:
         let lookup_env = if let Some(lcs) = &index.cs.lookup_constraint_system {
-            let lookup_table_combined = {
```
mimoo, line 385: this is the second computation of the lookup table, we can get rid of it now :)

mimoo:
moved to https://github.com/o1-labs/proof-systems/pull/549

PR #549, opened 2022-05-09
[lookup] compute the joint lookup table ONCE and only once

mimoo:
The prover computes the lookup argument by computing:

1. the joint table
2. the numerator part of the argument (using an `f_part` and `t_part`)
3. the denominator part of the argument (called `sorted`)

Both the numerator and the denominator make use of the joint table. Looking at the spec, we compute them in this order: 3 - 1 - 2. 

But why not do 1 - 3 - 2 and use the computed joint table in the computation of the denominator as well?

This PR fixes this.

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -5,11 +5,7 @@ use crate::{
         argument::{Argument, ArgumentType},
         expr::{l0_1, Constants, Environment, LookupEnvironment},
         gate::GateType,
-        lookup::{
-            self,
-            lookups::LookupsUsed,
-            tables::{combine_table_entry, CombinedEntry},
-        },
+        lookup::{self, lookups::LookupsUsed, tables::combine_table_entry},
         polynomials::{
             chacha::{ChaCha0, ChaCha1, ChaCha2, ChaChaFinal},
             complete_add::CompleteAdd,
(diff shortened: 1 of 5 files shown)


kimchi/src/prover.rs
```diff
@@ -382,44 +375,13 @@ where
 
         //~ 20. If using lookup:
         let lookup_env = if let Some(lcs) = &index.cs.lookup_constraint_system {
-            let lookup_table_combined = {
```
mimoo, line 385: that's the second computation of the combined table that we can skip now :)

kimchi/src/prover.rs
```diff
@@ -305,32 +317,13 @@ where
...
-                    };
-                    combine_table_entry(
-                        &lookup_context.joint_combiner.unwrap(),
-                        &lookup_context.table_id_combiner.unwrap(),
-                        row,
-                        &table_id,
-                    )
-                })
-            };
+            let joint_lookup_table_d8 = lookup_context.joint_lookup_table_d8.as_ref().unwrap();
```
mimoo, line 322: that's the second computation of the table that we can skip now :)

kimchi/src/prover.rs
```diff
@@ -296,6 +307,7 @@ where
             lookup_context.sorted_coeffs = Some(sorted_coeffs);
             lookup_context.sorted_comm = Some(sorted_comm);
             lookup_context.sorted8 = Some(sorted8);
+            lookup_context.joint_lookup_table_d8 = Some(joint_lookup_table_d8);
```
mimoo, line 310: this time we store the computation of the combined table the first time we compute it

mrmr1993 (approved):
There's a trade-off here with memory usage. I think this should be fine, but we may need to back this out if SnarkyJS's memory usage ends up too high as a result.

PR #554, opened 2022-05-11
[prover] move tests to debug_assertions

mimoo:
We previously tested each gate/argument constraints via something like this:

```rust
        if cfg!(test) {
            let (_, res) = evaluation
                .interpolate_by_ref()
                .divide_by_vanishing_poly(index.cs.domain.d1)
                .unwrap();
            assert!(res.is_zero());
        }
```

I moved this to a macro to simplify the code, and used `cfg!(debug_assertions)` so that we get this behavior whenever we are not running in release.

Also, I removed a bunch of `drop` to replace them with scoped `{ ... }` which is exactly the same but cleaner

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -39,6 +39,24 @@ use std::collections::HashMap;
 /// The result of a proof creation or verification.
 type Result<T> = std::result::Result<T, ProverError>;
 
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):
Love it!

Issue #527, opened 2022-05-02
Make sure one can easily use the library from the README


mimoo:
There's a few things that we should do:

* make sure the code example in README is run as test in cargo (so it doesn't break)
* give some guidelines on how to use this library

right now you have to import things like this:

```toml
[dependencies]
kimchi = { git = "https://github.com/o1-labs/proof-systems" }
mina-curves = { git = "https://github.com/o1-labs/proof-systems" }
oracle = { git = "https://github.com/o1-labs/proof-systems" }
commitment_dlog = { git = "https://github.com/o1-labs/proof-systems" }
```

which imo is not clean. Kimchi should re-export these crates so you don't have to import them yourself

mimoo:
I think the README should include a simple circuit example (like that we can compile it). For example something that takes a hash output and makes sure that you know the hash input.

Something like this:

```rust
use itertools::iterate;

/// create a simple circuit to hash an input
fn create_circuit() -> Vec<CircuitGate<Fp>> {
    let gates = vec![];
    let mut gates_row = iterate(0, |&i| i + 1);
    let row = || gates_row.next().unwrap();

    // the output to the hash function is a public input
    gates.push(CircuitGate::create_generic_gadget(
        Wire::new(row()),
        GenericGateSpec::Pub,
        None,
    ));
    gates.push(CircuitGate::create_generic_gadget(
        Wire::new(row()),
        GenericGateSpec::Pub,
        None,
    ));
    gates.push(CircuitGate::create_generic_gadget(
        Wire::new(row()),
        GenericGateSpec::Pub,
        None,
    ));

    // hash a private input
    let poseidon_params = oracle::pasta::fp_kimchi::params();
    let round_constants = &poseidon_params.round_constants;
    let row = row();
    let (g, final_row) = CircuitGate::<Fp>::create_poseidon_gadget(
        row,
        [Wire::new(row), Wire::new(row + 11)], // TODO: this argument should be deleted from the fn
        round_constants,
    );
    gates.extend(g);

    // wire the output to the public input
    // TODO: it'd be nice if we had functions that would do this for us, and panic if a permutation is already in place
    let last_row = gates.iter_mut().last().unwrap();
    last_row.wires[0] = Wire { row: 0, col: 0 };
    last_row.wires[1] = Wire { row: 0, col: 1 };
    last_row.wires[2] = Wire { row: 0, col: 2 };

    gates[0].wires[0] = Wire {
        row: final_row,
        col: 0,
    };
[...]

PR #555, opened 2022-05-12
[kimchi][doc] update readme example with circuit-construction API

katat:
Refer to https://github.com/o1-labs/proof-systems/issues/527

This PR aims to fix the compiling of the example in the Kimchi readme in the `cargo test` command while demonstrating how to use the library from a high-level perspective.

At the moment, the updates were inspired by the tests in the kimchi package, mainly to get the compiling passes. So probably it has not yet reflected the original goal.

If I may, here are a couple of questions about the goals of the example:
1. **public inputs and witness setups** - Can I assume the code for these setups would need to be included in the readme as well so that the `gates` variable can be valid in the scope? The problem is setting up these variables would make the script a bit long, but it would make it easier to follow where these variables come from. 

2. **b_poly_coefficients, ceil_log2** - These modules are in the example, but they are not being used. I guess they were intended to demonstrate some math with the library?

3. **URS** - The comment says to create URS, but the actual code is to create SRS. I may need to clarify if the code was diverging from the original purposes.

4. As you see from the updates in the current PR, the public and witness inputs are empty. I would assume it'd better come up with an example with non-empty inputs so as to demonstrate the math behind how the prover and verifier interact.


README.md
```diff
@@ -19,32 +19,50 @@ We assume that you already have:
...
 type ScalarSponge = DefaultFrSponge<Fp, SpongeParams>;
 
 // compile the circuit
+let gates = create_circuit(0, 0);
+
 let fp_sponge_params = oracle::pasta::fp_kimchi::params();
-let cs = ConstraintSystem::<Fp>::create(gates, vec![], fp_sponge_params, public_size).unwrap();
+let public = vec![Fp::from(0); 0];
+let mut witness: [Vec<Fp>; COLUMNS] = array_init(|_| vec![Fp::zero(); gates.len()]);
+fill_in_witness(0, &mut witness, &public);
```
mimoo, line 47: idea we wouldn't use testing functions like `create_circuit` and `fill_in_witness` but I guess without the circuit writer PR in we don't have much choices. cc @imeckler

README.md
```diff
@@ -19,32 +19,50 @@ We assume that you already have:
...
+let cs = ConstraintSystem::<Fp>::create(
+    gates, 
+    vec![], 
+    fp_sponge_params, 
+    public.len()
+).unwrap();
 
 // create an URS
-let mut urs = SRS::<Affine>::create(cs.domain.d1.size as usize);
+let mut srs = SRS::<Affine>::create(cs.domain.d1.size as usize);
```
mimoo, line 57: if you rename this you can rename the comment just above as well :o

mimoo:
Thanks a lot for the PR! This is great :) here are some answers:

> public inputs and witness setups - Can I assume the code for these setups would need to be included in the readme as well so that the gates variable can be valid in the scope? The problem is setting up these variables would make the script a bit long, but it would make it easier to follow where these variables come from.

the problem is that we don't have the circuit writer yet (I believe it's a matter of days though). For now, it's probably fine to use the testing functions to create the circuit and generate the witness, but once we have the circuit writer we'll need to revisit this README :)

> b_poly_coefficients, ceil_log2 - These modules are in the example, but they are not being used. I guess they were intended to demonstrate some math with the library?

ah, we should remove them from the example. They're here for the recursion but we removed it from the example as it's not really exposed in kimchi yet.

> URS - The comment says to create URS, but the actual code is to create SRS. I may need to clarify if the code was diverging from the original purposes.

This is actually an URS (as there's no trusted setup) but we call it SRS in the code for legacy reason. Might be a good idea to rename SRS here and not care too much about that distinction. I think it should be called something user friendly anyway, like CommonParameters or something.

> As you see from the updates in the current PR, the public and witness inputs are empty. I would assume it'd better come up with an example with non-empty inputs so as to demonstrate the math behind how the prover and verifier interact.

that would mean writing an example of writing a circuit without the circuit writer, so same answer, maybe better if we wait for the PR to get in

katat:
Thanks @mimoo for the feedback :)

While waiting for the circuit writer APIs to be ready, I have been trying to understand the maths behind kimchi, concepts such as polynomial commitment schemes, and how plonk works thanks to your videos and the book "the real world cryptography". I have to say these learning materials are really helpful, especially since there is a codebase from kimchi to confirm the understanding of the math concepts. Now at least I think I got some basic understanding of the pieces of kimchi that works together and mapping in my mind between the code and the "mina book".

Based on what I can see relevant to kimchi so far, here are a couple of suggestions relevant to the referenced issue #527, so that the readme can remain simple and serve as an introductory overview for the references to the examples and relevant documentation.

**examples in readme**
Alternatively, I think these examples could be provided under a specific folder (maybe a folder called `examples`?) or a separate Github repo. The reasoning is that even with the circuit construction API, based on what I see in the code in the PR, I guess it might be still too verbose to put all the example code into the readme since there are both high and low levels of concepts that need to explain to the starters to fully understand the how to use the kimchi API at both levels.

For the high level, examples with the circuit construction API for some common programs, such as hashing or sudoku, would be helpful for the starters to get an overview and dive into the implementation code or automated tests themselves.

For the low level, examples of how to create gates, piecing them up in a circuit for a simple program, and finally open a proof for verification, would be helpful. Although there are some existing automated tests against different gate types, adding more smaller granular tests would be also helpful as examples for the developers to understand the concepts along with the mina book.

In comparison with the snarkyjs, I think kimchi offers the flexibility to integrate with the other systems instead of mina. Some examples of these would be really cool, such as using kimchi to generate proofs for an ethereum contract to verify (not sure if this is possible yet, a research topic I have in mind at the moment. I know there is a [bba](https://github.com/imeckler/bba/) proposal, and it seems something to relevant to this topic).

**unit tests**
Currently, from my understanding, most automated tests are at the integration level, which tests for workflows in which a number of different functions are put together. Integration tests are indeed cost-effective to create and serve demonstrative purposes at a high level. On the other hand, unit tests in a smaller granularity may provide additional benefits.

Facilitate understanding of how a small function works by decoupling it from the other functions during tests and making a variety of assertions for a single function. This makes it convenient for the developers to review the implementation expectations and their relationship with the code quality itself. 

It may make the code easier to refactor in the long run since unit tests help decouple the code right the way at the beginning and could provide more "signals" in refactoring processes. 

It may help understand the math concepts if math equations can be decoupled into smaller functions so that it would be easier for anyone, the ones without a math degree, to easily play around with the inputs and outputs to confirm the understanding of the maths from the [[mina book]] or somewhere else.

**example ideas**
As the circuit construction API is getting ready, I am currently thinking of starter use cases that I can experiment with in addition to the hash example in the circuit API PR. One of them is sudoku, which once implemented can be used as a reference to compare with the snarkyjs version while demonstrating the full life cycle of using circuit API.

mimoo:
Sorry for the late answers! And thanks for the kind words :)

> examples in readme

I'm on the side of "every README should show some examples first". They don't have to be exhaustive examples, but they should give an idea of how the library is to be used. We can add more verbose and useful examples in an example folder.

> unit tests

agree with everything you said :o 

> example ideas

I was actually writing one for fun, I can push my code somewhere if you want to work on that and use some inspiration?

katat:
> I was actually writing one for fun, I can push my code somewhere if you want to work on that and use some inspiration?
> 
> EDIT: pushed it [here](https://github.com/o1-labs/proof-systems/commit/82332e5b2ddb3ba64ed58fa5557ad5ab62de2c12). It was a heavy WIP that wasn't working yet :D

After reading the code in your branch, I think it would be nice for the circuit API to add gates such as `and` `or`, in addition to your implemented `equals`. Then the `equals` might be better to be decoupled into `sub`, `invert` `is_zero` etc.

Please let me know if this is the right path, then I will try to get them implemented in the next few days :)

mimoo:
Yeah these sounds like excellent additions!

katat:
I am trying to make a simple example so as to keep the readme example minimum, as there will be other examples, such as the existing hash example and the upcoming sudoku in your branch, located in a separated model.

There are a couple of questions commented as "TODO" in the example of the PR, wondering if it makes sense further simplify the setups and the arguments.

atm, the example in the readme fails to compile due to the cyclic dependency issue. That is because the readme is located in the kimchi package, to get the example compiled, kimchi needs to use the circuit-construction package, while the circuit-construction package already uses kimchi package the way around. I am figuring out if there is a better way to resolve this.

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
 
 Then, you can create an URS for your circuit in the following way:
 
-```rust,ignore
```
mimoo, line 22: I guess the description above the example could be updated

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
+    });
+    sys.assert_eq(result, three);
+}
+
+// create SRS
+let srs = {
+    //TODO how to determine depth value? it throws error when the depth is too large
+    let mut srs = SRS::<Affine>::create(1 << 3); // 2^3 = 8
+    srs.add_lagrange_basis(D::new(srs.g.len()).unwrap());
+    Arc::new(srs)
```
mimoo, line 65: In the future, it’d be nice if we could omit the SRS and let the proving function below create it for us (if we don’t care about sharing the SRS)

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
+// generate circuit and index
+let prover_index = generate_prover_index::<FpInner, _>(
+    //TODO do these arguments have to be provided?
+    srs,
+    &fp_constants(),
+    &oracle::pasta::fq_kimchi::params(),
+    //TODO should this be encapsulated?
+    public_inputs.len(),
+    |sys, p| circuit::<_, _>(None, sys, p),
+);
```
mimoo, line 79: In the future it’ll be nice if we have a builder pattern for that as well

kimchi/Cargo.toml
```diff
@@ -37,6 +37,7 @@ groupmap = { path = "../groupmap" }
 mina-curves = { path = "../curves" }
 o1-utils = { path = "../utils" }
 oracle = { path = "../oracle" }
+circuit-construction = { path = "../circuit-construction" }
```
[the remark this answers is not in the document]
mimoo: See https://github.com/o1-labs/proof-systems/blob/14af2ed3037b3e14fe6b05f28701d60b9d21e737/kimchi/src/lib.rs#L1

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
+    //TODO replace the following free variable once the "add" gate constraint is ready to use.
+    let result = sys.var(|| {
+        first_public_input.val() + witness.val()
+    });
+    sys.assert_eq(result, three);
+}
+
+// create SRS
+let srs = {
+    //TODO how to determine depth value? it throws error when the depth is too large
```
mimoo, line 62: Yeah basically lol, the usecase we’d want to support is either 1) the user doesn’t care and we can generate the correct srs for them or 2) the user creates a large srs and this will be the upperbound for all circuits that use it (that’s what mina does)

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
+>(
+    witness: Option<F>,
+    sys: &mut Sys,
+    public_input: Vec<Var<F>>,
+) {
+    let three = sys.constant(3u32.into());
+    let first_public_input = public_input[0];
+    let witness = sys.var(|| witness.unwrap());
+
+    //TODO replace the following free variable once the "add" gate constraint is ready to use.
```
mimoo, line 53: Remove todo in example

README.md
mimoo: And to not have the user have to figure out all of these params

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
 
-// verify a proof
-verify::<Affine, BaseSponge, ScalarSponge>(&group_map, verifier_index, proof).unwrap();
+// generate circuit and index
+let prover_index = generate_prover_index::<FpInner, _>(
+    //TODO do these arguments have to be provided?
+    srs,
+    &fp_constants(),
+    &oracle::pasta::fq_kimchi::params(),
+    //TODO should this be encapsulated?
```
mimoo, line 76: Lgtm

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
+let public_inputs = vec![1i32.into()];
 let group_map = <Affine as CommitmentCurve>::Map::setup();
-let proof =  ProverProof::create::<BaseSponge, ScalarSponge>(
-    &group_map, witness, &prover_index);
 
-// verify a proof
-verify::<Affine, BaseSponge, ScalarSponge>(&group_map, verifier_index, proof).unwrap();
+// generate circuit and index
+let prover_index = generate_prover_index::<FpInner, _>(
+    //TODO do these arguments have to be provided?
```
mimoo, line 72: Yeah unfortunately for now

kimchi/Cargo.toml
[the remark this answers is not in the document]
katat: > Perhaps that README should be included in the lib.rs of circuit-writer not kimchi

Thanks for pointing this out. I think if the `circuit-writer` is going to be the main library, then the main README can be located in the `circuit-construction` folder instead and get compiled without concern about this cycle dependency issue.

README.md
katat: Updated. Please have a check. 
Note I still use the name `circuit-construction` and the name `kimchi` still remains. I guess they can remain until a new name for `circuit-construction` is determined.

README.md
katat: Yes, I think a builder pattern makes sense.

kimchi/Cargo.toml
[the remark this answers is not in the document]
katat: Updated to have `circuit-construction` include the main readme, while `kimchi` refers to the readme under its own folder.

README.md
```diff
@@ -19,44 +19,80 @@ We assume that you already have:
...
+use commitment_dlog::{commitment::CommitmentCurve, srs::SRS};
+use groupmap::GroupMap;
+use kimchi::verifier::verify;
+use mina_curves::pasta::{
+    fp::Fp,
+    vesta::{Affine, VestaParameters},
+};
 use oracle::{
-    constants::PlonkSpongeConstantsKimchi,
+    constants::*,
```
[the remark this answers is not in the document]
katat: Updated the `circuit-construction` to do the re-export. Please have a check if that makes sense, otherwise I can revert this commit and leave it to a dedicated future refactor task. 

btw, it seems to me it would look cleaner to have `circuit-construction/src/lib.rs` file separated into several different files and leave the `lib.rs` only holds the definitions of `mods` and `re-exports`.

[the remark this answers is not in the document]
mimoo: > btw, it seems to me it would look cleaner to have circuit-construction/src/lib.rs file separated into several different files and leave the lib.rs only holds the definitions of mods and re-exports.

agree, if you want to take a stab at it be my guest, otherwise I might get to it in a week or so

README.md
```diff
@@ -11,54 +11,104 @@ You can read more about this project on the [Kimchi book](https://o1-labs.github
...
+3. Generate prover index
+4. Generate a proof
+5. Verify the proof
+
+```rust
+use std::sync::Arc;
+use ark_ff::{FftField, PrimeField};
+use ark_poly::{EvaluationDomain, Radix2EvaluationDomain};
+use circuit_construction::{
+    *,
```
mimoo, line 27: I think we should avoid importing `*`, unless it's a well-defined module (like `prologue`) that is meant to be opened in the root scope

circuit-construction/tests/example_proof.rs
```diff
@@ -1,21 +1,23 @@
...
-    fp::Fp,
-    pallas::Affine as Other,
-    vesta::{Affine, VestaParameters},
-};
-use oracle::{
-    constants::*,
-    poseidon::{ArithmeticSponge, Sponge},
-    sponge::{DefaultFqSponge, DefaultFrSponge},
+use circuit_construction::{
+    *,
```
mimoo, line 6: same here

mimoo (approved):
Amazing :) thanks for working on making this library more usable! Let's see if CI passes and I'll merge. Thanks again!

README.md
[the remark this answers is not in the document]
katat: I am happy to :D
I will create a PR for it.

README.md
katat: Here is an attempt to use a prologue. It does simplify the import a lot, just use one line. But I am not sure if there are any potential issues when it comes to practice :D

https://github.com/katat/kimchi/pull/1

README.md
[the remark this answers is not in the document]
katat: Here is an attempt to refactor the file structure. Let me know any suggestions :D https://github.com/katat/kimchi/pull/1

PR #563, opened 2022-05-20
Implemented range-check gate wiring for permutation argument

### LINKED ISSUE #558: Range check gate - add copy constraints
We need to update the create function to

* be usable from a given row
* create the wiring for the copy constraints which we will constrained by the permutation argument
* return the ending row

jspada:
* Add connect_cell_pair() helper
    * Wire up copy constraints
    * Change create_range_check interface to take start row and return next row
    * Improve verify_range_check() to handle perm_aggreg exceptions
    * Fix witness creation (size from 8 to 4 rows) to work on single set of three 88-bit limbs

Closes #558

### DIFF
--- a/kimchi/src/circuits/polynomials/range_check/gate.rs
+++ b/kimchi/src/circuits/polynomials/range_check/gate.rs
@@ -26,7 +26,7 @@ use super::{RangeCheck0, RangeCheck1};
 
 // Connect the pair of cells of the witness specified by the cell1 and cell2 parameters
 // cell1 --> cell2 && cell2 --> cell1
-fn connect_cell_pair(wires: &mut Vec<GateWires>, cell1: (usize, usize), cell2: (usize, usize)) {
+fn connect_cell_pair(wires: &mut [GateWires], cell1: (usize, usize), cell2: (usize, usize)) {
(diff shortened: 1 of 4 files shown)


kimchi/src/circuits/polynomials/range_check/gate.rs
```diff
@@ -18,17 +18,42 @@ use crate::{
...
         polynomial::COLUMNS,
-        wires::GateWires,
+        wires::{GateWires, Wire},
     },
 };
 
 use super::{RangeCheck0, RangeCheck1};
 
+// Connect the pair of cells of the witness specified by the cell1 and cell2 parameters
+// cell1 --> cell2 && cell2 --> cell1
```
mrmr1993, line 28: Can you add a warning to this comment, saying that this will split the permutation if the two elements are transitively already part of the same permutation

mrmr1993 (approved):
One nit, but looks great!

kimchi/src/circuits/polynomials/range_check/gate.rs
jspada: 👍🏼

PR #564, opened 2022-05-20
Use dune lock to avoid racing access to cargo / rustup

mrmr1993:
`rustup` (and, by association `cargo` when using a `rust-toolchain.toml`) is [not safe to run concurrently](https://github.com/rust-lang/rustup/issues/988). In particular, this means that runs triggered from `dune build` or `dune runtest` can put the rust toolchain in an unusable state for one of the instances. See for example [this Mina CI failure](https://buildkite.com/o-1-labs-2/mina/builds/22966#12db5d05-1db0-479e-9b07-0d14969cd552).

By using a global `/cargo-lock` lock, we can also attach the same lock on the Mina side and prevent this issue.


jspada (approved):
lgtm!

PR #565, opened 2022-05-20
Put CI on same version of rust

jspada:


mrmr1993 (approved):

jspada:
> why is this needed since we have `rust-toolchain.toml`?

It was overriding the that version with the latest.  By simply removing the `toolchain` parameter `actions-rs` was supposed to default to the `rust-toolchain.toml`, but that would not work (I think it's a bug), so I had to specify the version.

PR #567, opened 2022-05-23
enforce length of SRS to be at least the domain size

mimoo:
in practice it is possible that using SRS that have a smaller size than the domain could work, as we support splitting every polynomials we use, but I'm not sure we did think of every corner cases. It's probably safer to just not allow this edge-case.

### DIFF
--- a/kimchi/src/error.rs
+++ b/kimchi/src/error.rs
@@ -21,7 +21,7 @@ pub enum ProverError {
     #[error("the lookup failed to find a match in the table")]
     ValueNotInTable,
 
-    #[error("kimchi does not support SRS of size smaller than the domain")]
+    #[error("SRS size is smaller than the domain size required by the circuit")]
     SRSTooSmall,
 }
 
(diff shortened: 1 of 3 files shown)


kimchi/src/error.rs
```diff
@@ -20,6 +20,9 @@ pub enum ProverError {
 
     #[error("the lookup failed to find a match in the table")]
     ValueNotInTable,
+
+    #[error("kimchi does not support SRS of size smaller than the domain")]
```
mrmr1993, line 24: Nit: remove `kimchi` from the message? Something like `SRS size is smaller than the domain size required by the circuit`?

kimchi/src/error.rs
```diff
@@ -39,6 +42,12 @@ pub enum VerifyError {
 
     #[error("lookup used in circuit, but proof has inconsistent number of lookup evaluations and commitments")]
     ProofInconsistentLookup,
+
+    #[error("kimchi does not support batching verification of proofs of different SRS")]
```
mrmr1993, line 46: Nit: `Cannot batch proofs using different SRSes`?

kimchi/src/error.rs
```diff
@@ -39,6 +42,12 @@ pub enum VerifyError {
...
+    DifferentSRS,
+
+    #[error("kimchi does not support SRS of size smaller than the domain")]
```
mrmr1993, line 49: `SRS size is smaller than the domain size`?

mrmr1993 (approved):
A few nits, but lgtm

Issue #434, opened 2022-03-18
Make the proof serializable


mimoo:
Right now there's no way to serialize the `ProverProof` struct. We should implement `serde::{Serialize, Deserialize}` on it.

I've tried doing this here: https://github.com/o1-labs/proof-systems/commit/7e8ae435738b88eda926392dae07b8588ba450fd

The blocker right now is this one: https://github.com/o1-labs/proof-systems/issues/231 which I'm not sure how to solve. I think having abstractions for chunked polynomials/commitments/evaluations, and functions to linearize them, would have to be solved first. (which is in https://github.com/o1-labs/proof-systems/issues/432)

Issue #231, opened 2021-11-23
ProofEvaluations is too generic


mimoo:
```rust
pub struct ProofEvaluations<Field> {
    /// witness polynomials
    pub w: [Field; COLUMNS],
    /// permutation polynomial
    pub z: Field,
    /// permutation polynomials
    /// (PERMUTS-1 evaluations because the last permutation is only used in commitment form)
    pub s: [Field; PERMUTS - 1],
    /// lookup-related evaluations
    pub lookup: Option<LookupEvaluations<Field>>,
    /// evaluation of the generic selector polynomial
    pub generic_selector: Field,
    /// evaluation of the poseidon selector polynomial
    pub poseidon_selector: Field,
}
```

should be 

```rust
pub struct ProofEvaluations<Field> {
    /// witness polynomials
    pub w: [Vec<Field>; COLUMNS],
    /// permutation polynomial
    pub z: Vec<Field>,
    /// permutation polynomials
    /// (PERMUTS-1 evaluations because the last permutation is only used in commitment form)
    pub s: [Vec<Field>; PERMUTS - 1],
    /// lookup-related evaluations
    pub lookup: Option<LookupEvaluations<Vec<Field>>>,
    /// evaluation of the generic selector polynomial
    pub generic_selector: Vec<Field>,
    /// evaluation of the poseidon selector polynomial
    pub poseidon_selector: Vec<Field>,
}
```

it'd be nice if we actually abstracted chunked evaluations, chunked polynomials, chunked commitments, etc.

mimoo:
we might not be able to tackle this one without https://github.com/o1-labs/proof-systems/issues/432 first

mimoo:
I think we might be able to just write this struct as:

```rust
pub struct ProofEvaluations<F> where F: FftField {
    /// witness polynomials
    pub w: [F; COLUMNS],
    /// permutation polynomial
    pub z: F,
    /// permutation polynomials
    /// (PERMUTS-1 evaluations because the last permutation is only used in commitment form)
    pub s: [F; PERMUTS - 1],
    /// lookup-related evaluations
    pub lookup: Option<LookupEvaluations<F>>,
    /// evaluation of the generic selector polynomial
    pub generic_selector: F,
    /// evaluation of the poseidon selector polynomial
    pub poseidon_selector: F,
}
```

as none of these evaluations are chunked (unless I'm misunderstanding something)

querolita:
Let's try something like: 

```rust
[...]

PR #570, opened 2022-05-24
Making the proof serializable

querolita:
This PR makes the kimchi proof serializable. It addresses [this issue](https://github.com/o1-labs/proof-systems/issues/434) and [this one](https://github.com/o1-labs/proof-systems/issues/231). It treats `ProofEvaluations` with vector fields, both for chunked and non-chunked (single-sized vectors). I started a different approach with specialized types for chunked vs non-chunked variants in a different branch, but it was far more verbose. 

Perhaps it is not needed to have all of the fields in the struct being vectors ([here](https://github.com/o1-labs/proof-systems/pull/567/)). Plus, I am not a huge fan of accessing single-sized fields with `[0]`, not very friendly either. I can improve those type of things in a different PR.

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -1394,7 +1394,9 @@ pub struct LookupEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
     pub runtime: Option<Vec<F>>,
 }
 
-// TODO: this should really be vectors here, perhaps create another type for chunked evaluations?
+/// Polynomial evaluations contained in a `ProverProof`.
+/// - **Chunked evaluations** use vectors with a length that equals the length of the chunk
+/// - **Non chunked evaluations** use single-sized vectors, so the single evaluation appears in the first position of each field of the struct
 #[serde_as]
 #[derive(Clone, Deserialize, Serialize)]
 pub struct ProofEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
@@ -1423,11 +1425,12 @@ pub struct ProofEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
 #[serde_as]
 #[derive(Clone, Deserialize, Serialize)]
 pub struct LookupCommitments<G: AffineCurve> {
+    /// Commitments to the sorted lookup table polynomial (may have chunks)
     #[serde(bound = "PolyComm<G>: Serialize + DeserializeOwned")]
     pub sorted: Vec<PolyComm<G>>,
+    /// Commitment to the lookup aggregation polynomial
     #[serde(bound = "PolyComm<G>: Serialize + DeserializeOwned")]
     pub aggreg: PolyComm<G>,
-
     /// Optional commitment to concatenated runtime tables
     #[serde(bound = "Option<PolyComm<G>>: Serialize + DeserializeOwned")]
     pub runtime: Option<PolyComm<G>>,
@@ -1480,13 +1483,14 @@ where
     pub public: Vec<ScalarField<G>>,
 
     /// The challenges underlying the optional polynomials folded into the proof
-    #[serde(bound = "Challenge<G>: Serialize + DeserializeOwned")]
-    pub prev_challenges: Vec<Challenge<G>>,
+    #[serde(bound = "RecursionChallenge<G>: Serialize + DeserializeOwned")]
+    pub prev_challenges: Vec<RecursionChallenge<G>>,
 }
 
+/// A struct to store the challenges inside a `ProverProof`
 #[serde_as]
 #[derive(Clone, Deserialize, Serialize)]
-pub struct Challenge<G>
+pub struct RecursionChallenge<G>
 where
     G: AffineCurve,
 {
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -1376,86 +1376,128 @@ You can find these operations under the [proof creation](#proof-creation) and [p
 A proof consists of the following data structures:
 
 ```rs
-#[derive(Clone)]
-pub struct LookupEvaluations<F> {
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
+pub struct LookupEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
     /// sorted lookup table polynomial
+    #[serde_as(as = "Vec<Vec<o1_utils::serialization::SerdeAs>>")]
     pub sorted: Vec<Vec<F>>,
     /// lookup aggregation polynomial
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
     pub aggreg: Vec<F>,
     // TODO: May be possible to optimize this away?
     /// lookup table polynomial
-    pub table: Field,
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub table: Vec<F>,
     /// Optionally, a runtime table polynomial.
-    pub runtime: Option<Field>,
+    #[serde_as(as = "Option<Vec<o1_utils::serialization::SerdeAs>>")]
+    pub runtime: Option<Vec<F>>,
 }
 
 // TODO: this should really be vectors here, perhaps create another type for chunked evaluations?
-//#[serde_as]
-//#[derive(Clone, Deserialize, Serialize)]
-#[derive(Clone)]
-pub struct ProofEvaluations<F> {
-    //: CanonicalDeserialize + CanonicalSerialize> {
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
+pub struct ProofEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
     /// witness polynomials
+    #[serde_as(as = "[Vec<o1_utils::serialization::SerdeAs>; COLUMNS]")]
     pub w: [Vec<F>; COLUMNS],
     /// permutation polynomial
-    //#[serde(bound = "EvalEnum<Field>: Serialize")]
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
     pub z: Vec<F>,
     /// permutation polynomials
     /// (PERMUTS-1 evaluations because the last permutation is only used in commitment form)
+    #[serde_as(as = "[Vec<o1_utils::serialization::SerdeAs>; PERMUTS - 1]")]
     pub s: [Vec<F>; PERMUTS - 1],
     /// lookup-related evaluations
+    #[serde(bound = "LookupEvaluations<F>: Serialize")]
     pub lookup: Option<LookupEvaluations<F>>,
     /// evaluation of the generic selector polynomial
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
     pub generic_selector: Vec<F>,
     /// evaluation of the poseidon selector polynomial
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
     pub poseidon_selector: Vec<F>,
 }
 
 /// Commitments linked to the lookup feature
-#[derive(Clone)]
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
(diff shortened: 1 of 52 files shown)


book/src/specs/kimchi.md
```diff
@@ -1376,83 +1376,128 @@ You can find these operations under the [proof creation](#proof-creation) and [p
 A proof consists of the following data structures:
 
 ```rs
-#[derive(Clone)]
-pub struct LookupEvaluations<Field> {
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
+pub struct LookupEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
```
mimoo, line 1381: might be nicer to just have `F: Field` in the trait bound (this is what we do everywhere else)

book/src/specs/kimchi.md
```diff
@@ -1376,83 +1376,128 @@ You can find these operations under the [proof creation](#proof-creation) and [p
...
 
     /// The challenges underlying the optional polynomials folded into the proof
-    pub prev_challenges: Vec<(Vec<ScalarField<G>>, PolyComm<G>)>,
+    #[serde(bound = "Challenge<G>: Serialize + DeserializeOwned")]
+    pub prev_challenges: Vec<Challenge<G>>,
 }
+
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
+pub struct Challenge<G>
```
mimoo, line 1489: missing doc

kimchi/src/bench.rs
```diff
@@ -138,6 +138,10 @@ mod tests {
         let proof = ctx.create_proof();
         println!("proof created in {}", start.elapsed().as_millis());
 
+        // small check of proof being serializable
+        let ser_pf = rmp_serde::to_vec(&proof).unwrap();
+        println!("proof size: {} bytes", ser_pf.len());
```
mimoo, line 143: probably not the right place to test this :)

kimchi/src/circuits/expr.rs
```diff
@@ -436,14 +436,14 @@ impl Variable {
...
-            Index(GateType::Poseidon) => Ok(evals.poseidon_selector),
-            Index(GateType::Generic) => Ok(evals.generic_selector),
+            Witness(i) => Ok(evals.w[i][0]),
+            Z => Ok(evals.z[0]),
+            LookupSorted(i) => l.map(|l| l.sorted[i][0]),
+            LookupAggreg => l.map(|l| l.aggreg[0]),
+            LookupTable => l.map(|l| l.table[0]),
+            LookupRuntimeTable => l.and_then(|l| l.get_runtime().ok_or(ExprError::MissingRuntime)),
+            Index(GateType::Poseidon) => Ok(evals.poseidon_selector[0]),
+            Index(GateType::Generic) => Ok(evals.generic_selector[0]),
```
mimoo, line 446: this feels very unsafe to me. If we're going to this route I think it'd be better to have two distinct type, one with a vec and one without

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -539,21 +539,24 @@ mod tests {
...
-            lookup: Some(LookupEvaluations {
-                sorted: (0..(lookup_info.max_per_row + 1))
-                    .map(|_| F::rand(rng))
-                    .collect(),
-                aggreg: F::rand(rng),
-                table: F::rand(rng),
-                runtime: None,
-            }),
+        let mut eval = || {
+            ProofEvaluations::new(
```
mimoo, line 543: there really ought to be a `ProofEvaluations::rand_for_test()` or smthg

kimchi/src/proof.rs
```diff
@@ -2,126 +2,225 @@
...
+    #[serde(bound = "LookupEvaluations<F>: Serialize")]
+    pub lookup: Option<LookupEvaluations<F>>,
     /// evaluation of the generic selector polynomial
-    pub generic_selector: Field,
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub generic_selector: Vec<F>,
     /// evaluation of the poseidon selector polynomial
-    pub poseidon_selector: Field,
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub poseidon_selector: Vec<F>,
```
mimoo, line 55: I think we should try to remove the `Vec` here and see if it works

mimoo:
Nice! I think this is going in the right direction :) 

As the `ProofEvaluations` struct is exposed to Mina, we need to update the bindings and see if it can work on the Mina side. Because of that, I'd suggest creating a PR that singles out only the changes on `ProofEvaluations` so we can review & test that in isolation in Mina before tackling the serialization part. Also, since there's binding work to be done, I'd suggest trying to remove the vectors now so that we don't have to play with the OCaml bindings twice :o

book/src/specs/kimchi.md
jspada: I was thinking the same, but maybe it's also good the way Anais has it (to keep the struct as general as possible and then put the `Field` bound on the functions in the impl)?

kimchi/src/bench.rs
```diff
@@ -96,7 +96,7 @@ impl BenchmarkCtx {
                 let b = DensePolynomial::from_coefficients_vec(coeffs);
                 self.index.srs.commit_non_hiding(&b, None)
             };
-            (chals, comm)
+            Challenge { chals, comm }
```
jspada, line 99: Maybe create a `::new` or `::from_` function

kimchi/src/circuits/expr.rs
jspada: Perhaps use a `is_chunked()` function to check the situation and handle any errors?  Maybe I missed somewhere were you already do it.

kimchi/src/circuits/polynomials/chacha.rs
jspada: Perhaps it should allow you to optionally set the witness, selector and lookups for the test also

kimchi/src/proof.rs
```diff
@@ -2,126 +2,225 @@
...
     // TODO: May be possible to optimize this away?
     /// lookup table polynomial
-    pub table: Field,
-
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub table: Vec<F>,
     /// Optionally, a runtime table polynomial.
-    pub runtime: Option<Field>,
+    #[serde_as(as = "Option<Vec<o1_utils::serialization::SerdeAs>>")]
+    pub runtime: Option<Vec<F>>,
```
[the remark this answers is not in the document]
jspada: Lookups are never chunked?

kimchi/src/proof.rs
jspada: Just on the poseidon selector or the whole thing?

kimchi/src/proof.rs
```diff
@@ -2,126 +2,225 @@
...
 }
 
 /// Commitments linked to the lookup feature
-#[derive(Clone)]
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
 pub struct LookupCommitments<G: AffineCurve> {
+    #[serde(bound = "PolyComm<G>: Serialize + DeserializeOwned")]
     pub sorted: Vec<PolyComm<G>>,
+    #[serde(bound = "PolyComm<G>: Serialize + DeserializeOwned")]
```
jspada, line 64: Add doc comments for these.

kimchi/src/prover.rs
```diff
@@ -1176,6 +1186,47 @@ pub mod caml {
...
+        fn from(caml_ch: CamlChallenge<CamlG, CamlF>) -> Challenge<G> {
+            Challenge {
+                chals: caml_ch.chals.into_iter().map(Into::into).collect(),
+                comm: caml_ch.comm.into(),
+            }
+        }
+    }
+
+    //
+    // CamlProverProof<CamlG, CamlF>
```
jspada, line 1227: Just to be clear will these changes cause any breakage on mina side?

kimchi/src/prover.rs
```diff
@@ -1184,9 +1235,13 @@ pub mod caml {
         pub evals: (CamlProofEvaluations<CamlF>, CamlProofEvaluations<CamlF>),
         pub ft_eval1: CamlF,
         pub public: Vec<CamlF>,
-        pub prev_challenges: Vec<(Vec<CamlF>, CamlPolyComm<CamlG>)>,
+        pub prev_challenges: Vec<CamlChallenge<CamlG, CamlF>>, //Vec<(Vec<CamlF>, CamlPolyComm<CamlG>)>,
     }
 
+    //
+    // CamlProverCommitments<CamlG>
```
jspada, line 1242: I think this comment should be more descriptive, giving new info.

kimchi/src/verifier.rs
```diff
@@ -307,7 +308,10 @@ where
             .prev_challenges
             .iter()
             .zip(self.prev_chal_evals(index, &evaluation_points, &powers_of_eval_points_for_chunks))
-            .map(|(c, e)| (c.1.clone(), e))
+            .map(|(c, e)| {
```
jspada, line 311: Suggest replace one letter variable names with something more readable.

poly-commitment/src/evaluation_proof.rs
```diff
@@ -310,13 +312,19 @@ impl<G: CommitmentCurve> SRS<G> {
...
-#[derive(Clone, Debug)]
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct OpeningProof<G: AffineCurve> {
     /// vector of rounds of L & R commitments
+    #[serde_as(as = "Vec<(o1_utils::serialization::SerdeAs, o1_utils::serialization::SerdeAs)>")]
     pub lr: Vec<(G, G)>,
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub delta: G,
+    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
```
jspada, line 323: Please doc comment all of these public members

utils/src/serialization.rs
```diff
@@ -79,3 +79,38 @@ where
         T::deserialize(&mut &bytes[..]).map_err(serde::de::Error::custom)
     }
 }
+
+/*
```
jspada, line 83: Remove this

book/src/specs/kimchi.md
```diff
@@ -1376,83 +1376,128 @@ You can find these operations under the [proof creation](#proof-creation) and [p
 A proof consists of the following data structures:
 
 ```rs
-#[derive(Clone)]
-pub struct LookupEvaluations<Field> {
+#[serde_as]
```
jspada, line 1379: Really great work!  This code looks so much simpler than the earlier versions where you split the enums and structs into chunked and non-chunked types.

kimchi/src/proof.rs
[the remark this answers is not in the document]
querolita: I set this one as `Vec<F>` because in the previous version, `Field` could potentially be a vector. Nonetheless, I will check now which of them could be just single elements, both for this one and for other fields of the structs.

kimchi/src/prover.rs
querolita: Mmm yeah they will, definitely will have to modify these on mina side

kimchi/src/prover.rs
querolita: I think these are just "separators" whenever we define a new Caml struct

kimchi/src/proof.rs
querolita: Actually, when defining `chunked_evals` in the proof, it assumes selectors can as well be chunked.

mrmr1993:
I've opened #599 as an alternative. It doesn't make any changes to types; it only adds annotations and the test from this PR.

querolita:
This PR has been replaced for [this one](https://github.com/o1-labs/proof-systems/pull/599), thus I am closing the current one. May be reopened in the future to address the `RecursionChallenge` type.

PR #574, opened 2022-05-25
[easy][test-framework] remove redundant code

mimoo:


kimchi/src/tests/lookup.rs
```diff
@@ -157,5 +173,16 @@ fn test_runtime_table() {
     print_witness(&witness, 0, 20);
 
     // run test
-    TestFramework::run_test_lookups(gates, witness, &[], vec![], Some(runtime_tables));
+    TestFramework::default()
+        .gates(gates)
+        .witness(witness)
+        .runtime_tables_cfg(runtime_tables_cfg)
+        .runtime_tables(runtime_tables)
```
mrmr1993, line 180: Nit: `runtime_tables` should probably come after `setup`, since it's not required for the setup.

kimchi/src/tests/framework.rs
```diff
@@ -31,128 +31,122 @@ type BaseSponge = DefaultFqSponge<VestaParameters, SpongeParams>;
...
+    #[must_use]
+    pub(crate) fn recursion(mut self, recursion: Vec<(Vec<Fp>, PolyComm<Affine>)>) -> Self {
+        self.recursion = recursion;
+        self
+    }
 
-        // verify the proof
+    /// creates the indexes
+    #[must_use]
+    pub(crate) fn setup(mut self) -> Self {
```
mrmr1993, line 96: It feels like this should return a different type, since you shouldn't continue 'setting it up' after calling this.

kimchi/src/tests/framework.rs
```diff
@@ -31,128 +31,122 @@ type BaseSponge = DefaultFqSponge<VestaParameters, SpongeParams>;
...
-                    len: table.data.len(),
-                })
-                .collect()
-        });
-        let index =
-            new_index_for_test_with_lookups(gates, public.len(), lookup_tables, runtime_tables_cfg);
-        let verifier_index = index.verifier_index();
-        println!("- time to create index: {:?}s", start.elapsed().as_secs());
+    /// Create and verify a proof
+    pub(crate) fn prove_and_verify(self) {
```
mrmr1993, line 120: Then this function should hang off that new type.

kimchi/src/tests/framework.rs
```diff
@@ -31,128 +31,122 @@ type BaseSponge = DefaultFqSponge<VestaParameters, SpongeParams>;
...
-            let chals: Vec<_> = (0..k).map(|_| Fp::rand(rng)).collect();
-            let comm = {
-                let coeffs = b_poly_coefficients(&chals);
-                let b = DensePolynomial::from_coefficients_vec(coeffs);
-                index.srs.commit_non_hiding(&b, None)
-            };
-            (chals, comm)
-        };
+    #[must_use]
+    pub(crate) fn runtime_tables(mut self, runtime_tables: Vec<RuntimeTable<Fp>>) -> Self {
```
mrmr1993, line 83: As should this function.

mrmr1993 (approved):
This is already a nice improvement, but left a few comments about splitting the framework into 2 types (1 for setup, the other for proving).

mimoo:
good points! Let me start that in another PR :o

PR #576, opened 2022-05-25
[test-framework] introduce new TestRunner type

mimoo:


kimchi/src/tests/lookup.rs
```diff
@@ -181,12 +181,14 @@ fn runtime_table(num: usize, indexed: bool) {
     print_witness(&witness, 0, 20);
 
     // run test
-    TestFramework::default()
+    let test_runner = TestFramework::default()
```
mrmr1993, line 184: I'm not sure this `let` binding is adding much, it's probably fine to keep it as a sequence of function calls and just change the order

mrmr1993 (approved):

mrmr1993 (approved):

PR #579, opened 2022-05-28
Make lookup constraints independent of choice of domain

mrmr1993:
This PR tweaks the behaviour of `UnnormalizedLagrangeBasis` so that it can take positive or negative values (relative to the domain size). This allows us to make the lookup argument constraints independent of the domain size.

As a side-effect, this also gives us 'negative' powers that we can apply to the domain's generator. In the verifier circuit, this lets us perform a much smaller number of multiplications by calculating `1/g^(-1)`, rather that performing a large number of squarings.


jspada (approved):
Looks good to me.

kimchi/src/circuits/expr.rs
```diff
@@ -475,9 +475,15 @@ impl<F: FftField> PolishToken<F> {
                 EndoCoefficient => stack.push(c.endo_coefficient),
                 Mds { row, col } => stack.push(c.mds[*row][*col]),
                 VanishesOnLast4Rows => stack.push(eval_vanishes_on_last_4_rows(d, pt)),
-                UnnormalizedLagrangeBasis(i) => stack.push(
-                    d.evaluate_vanishing_polynomial(pt) / (pt - d.group_gen.pow(&[*i as u64])),
-                ),
+                UnnormalizedLagrangeBasis(i) => {
+                    // Renormalize negative values to wrap around at domain size
+                    let i = if *i < 0 {
+                        ((*i as i64) + (d.size as i64)) as u64
```
mimoo, line 481: this is great!

mimoo: would still be nice to add tests related to this, off by ones scare me :)

jspada: good idea

PR #581, opened 2022-05-30
Factor out lookup layouts by LookupPattern

mrmr1993:
This PR builds upon #579, separating out the different built-in lookup patterns as an explicit enum that maps into the patterns.

### DIFF
--- a/kimchi/src/circuits/lookup/lookups.rs
+++ b/kimchi/src/circuits/lookup/lookups.rs
@@ -263,6 +263,85 @@ impl<F: Copy> JointLookup<SingleLookup<F>, LookupTableID> {
     }
 }
 
+pub enum LookupPattern {
+    ChaCha,
+    ChaChaFinal,
+    LookupGate,
+}
+
+impl LookupPattern {
+    pub fn lookups<F: Field>(&self) -> Vec<JointLookupSpec<F>> {
+        let curr_row = |column| LocalPosition {
+            row: CurrOrNext::Curr,
+            column,
+        };
+        match self {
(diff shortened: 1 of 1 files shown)


kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -263,6 +263,85 @@ impl<F: Copy> JointLookup<SingleLookup<F>, LookupTableID> {
     }
 }
 
+pub enum LookupPattern {
```
jspada, line 266: This is a really nice improvement!

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -263,6 +263,85 @@ impl<F: Copy> JointLookup<SingleLookup<F>, LookupTableID> {
...
+    ChaCha,
+    ChaChaFinal,
+    LookupGate,
+}
+
+impl LookupPattern {
+    pub fn lookups<F: Field>(&self) -> Vec<JointLookupSpec<F>> {
```
jspada, line 273: It might be nicer to add the lookup pattern to the `Argument` interface, so that the lookup pattern definition can be specified on the definition of the respective circuitgates.  For example, `Argument::lookup_pattern()` could either return a pattern or None if the circuit gate doesn't use any plookups.

jspada (approved):

kimchi/src/circuits/lookup/lookups.rs
mimoo: agree :o but this is a step in the right direction!

PR #582, opened 2022-05-30
Refactors using explicit lookup patterns

mrmr1993:
This PR builds upon #581 to simplify the logic around the 'lookup constraint system' using the explicit lookup patterns.

PR #583, opened 2022-05-30
Move `runtime_tables`, `runtime_table_offset` into `LookupConstraintSystem`

### LINKED ISSUE #580: Move `runtime_tables`, `runtime_table_offset` out of `LookupConfiguration`
`runtime_tables` and `runtime_table_offset` do not belong in [`LookupConfiguration`](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/constraints.rs#L340-L345), which is used to set up the general structure of the constraints without specifying the layout of the actual circuit. It would be best to move them to [`LookupConstraintSystem`](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/index.rs#L41), and store only a `max_runtime_table_width` in `LookupConfiguration`.

mrmr1993:
This PR builds upon #582. This fixes #580.

### DIFF
--- a/kimchi/src/circuits/lookup/index.rs
+++ b/kimchi/src/circuits/lookup/index.rs
@@ -76,7 +76,7 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
         runtime_tables: Option<Vec<RuntimeTableCfg<F>>>,
         domain: &EvaluationDomains<F>,
     ) -> Result<Option<Self>, LookupError> {
-        let lookup_info = LookupInfo::create();
+        let lookup_info = LookupInfo::create(runtime_tables.is_some());
 
         //~ 1. If no lookup is used in the circuit, do not create a lookup index
         match lookup_info.lookup_used(gates) {
--- a/kimchi/src/circuits/lookup/index.rs
+++ b/kimchi/src/circuits/lookup/index.rs
@@ -64,6 +64,12 @@ pub struct LookupConstraintSystem<F: FftField> {
     #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
     pub runtime_selector: Option<E<F, D<F>>>,
 
+    /// Optional runtime tables, listed as tuples `(length, id)`.
+    pub runtime_tables: Option<Vec<RuntimeTableSpec>>,
+
+    /// The offset of the runtime table within the concatenated table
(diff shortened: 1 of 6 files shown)


kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -37,11 +37,13 @@ pub struct LookupInfo {
     pub max_per_row: usize,
     /// The maximum joint size of any joint lookup in a constraint in `kinds`. This can be computed from `kinds`.
     pub max_joint_size: u32,
+    /// True if runtime lookup tables are used.
+    pub uses_runtime_tables: bool,
```
mimoo, line 41: this looks unnecessary, but I like it. But am I wrong in saying that this is unnecessary?

kimchi/src/circuits/lookup/index.rs
```diff
@@ -64,6 +64,12 @@ pub struct LookupConstraintSystem<F: FftField> {
     #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
     pub runtime_selector: Option<E<F, D<F>>>,
 
+    /// Optional runtime tables, listed as tuples `(length, id)`.
+    pub runtime_tables: Option<Vec<RuntimeTableSpec>>,
+
+    /// The offset of the runtime table within the concatenated table
+    pub runtime_table_offset: Option<usize>,
```
mimoo, line 71: I think at this point it'd be interesting to create an `runtime_index: Option<RuntimeIndex>` to avoid all the options there

mimoo (approved):
LGTM!

kimchi/src/circuits/lookup/lookups.rs
mrmr1993: We need this on the proof-systems side, to be able to turn on `uses_runtime_tables` in the generated plonk scalar calculations.

mimoo: I meant that this information is also present in the lookup index (if the Option is set for runtime table)

kimchi/src/circuits/lookup/index.rs
jspada: good idea

jspada (commented):
looks good to me

PR #584, opened 2022-05-31
Use explicitly-labelled lookup selectors

mrmr1993:
This PR builds on top of #583.

This refactors the `lookup_selectors` store to use a structured `LookupSelectors` record. This allows us finer-grained control over which selectors are used or unused, which we couldn't previously have with the vector-position system.

### DIFF
--- a/kimchi/src/circuits/lookup/index.rs
+++ b/kimchi/src/circuits/lookup/index.rs
@@ -119,6 +119,9 @@ impl<T> LookupSelectors<T> {
             chacha_final,
             lookup_gate,
         } = self;
+        // This closure isn't really redundant -- it shields the parameter from a copy -- but
+        // clippy isn't smart enough to figure that out..
+        #[allow(clippy::redundant_closure)]
         let f = |x| f(x);
         LookupSelectors {
             chacha: chacha.map(f),
--- a/kimchi/src/circuits/lookup/index.rs
+++ b/kimchi/src/circuits/lookup/index.rs
@@ -91,13 +91,13 @@ impl<'de, F: FftField> serde_with::DeserializeAs<'de, LookupSelectors<E<F, D<F>>
 }
 
 impl<T> std::ops::Index<LookupPattern> for LookupSelectors<T> {
-    type Output = T;
+    type Output = Option<T>;
 
     fn index(&self, index: LookupPattern) -> &Self::Output {
         match index {
-            LookupPattern::ChaCha => self.chacha.as_ref().expect("has chacha"),
-            LookupPattern::ChaChaFinal => self.chacha_final.as_ref().expect("has chacha_final"),
-            LookupPattern::LookupGate => self.lookup_gate.as_ref().expect("has lookup_gate"),
+            LookupPattern::ChaCha => &self.chacha,
+            LookupPattern::ChaChaFinal => &self.chacha_final,
+            LookupPattern::LookupGate => &self.lookup_gate,
         }
     }
 }
@@ -105,9 +105,9 @@ impl<T> std::ops::Index<LookupPattern> for LookupSelectors<T> {
 impl<T> std::ops::IndexMut<LookupPattern> for LookupSelectors<T> {
     fn index_mut(&mut self, index: LookupPattern) -> &mut Self::Output {
         match index {
-            LookupPattern::ChaCha => self.chacha.as_mut().expect("has chacha"),
-            LookupPattern::ChaChaFinal => self.chacha_final.as_mut().expect("has chacha_final"),
-            LookupPattern::LookupGate => self.lookup_gate.as_mut().expect("has lookup_gate"),
+            LookupPattern::ChaCha => &mut self.chacha,
+            LookupPattern::ChaChaFinal => &mut self.chacha_final,
+            LookupPattern::LookupGate => &mut self.lookup_gate,
         }
     }
 }
--- a/kimchi/src/circuits/lookup/index.rs
+++ b/kimchi/src/circuits/lookup/index.rs
@@ -36,6 +36,108 @@ pub enum LookupError {
     TableIDZeroMustHaveZeroEntry,
 }
 
+#[derive(Clone, Serialize, Deserialize, Debug, Default)]
+pub struct LookupSelectors<T> {
+    pub chacha: Option<T>,
+    pub chacha_final: Option<T>,
+    pub lookup_gate: Option<T>,
+}
+
+#[serde_as]
+#[derive(Clone, Serialize, Deserialize, Debug, Default)]
+struct LookupSelectorsSerdeAs<F: FftField> {
+    #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
+    pub chacha: Option<E<F, D<F>>>,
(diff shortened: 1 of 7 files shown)


kimchi/src/circuits/lookup/index.rs
```diff
@@ -36,6 +36,108 @@ pub enum LookupError {
     TableIDZeroMustHaveZeroEntry,
 }
 
+#[derive(Clone, Serialize, Deserialize, Debug, Default)]
+pub struct LookupSelectors<T> {
+    pub chacha: Option<T>,
+    pub chacha_final: Option<T>,
+    pub lookup_gate: Option<T>,
```
mimoo, line 43: not sure I understand why this struct is so generic. Why not make `T` -> `Evaluations<F, D<F>>` ?

mimoo: also all of these fields are `pub` and missing doc

kimchi/src/circuits/lookup/index.rs
```diff
@@ -36,6 +36,108 @@ pub enum LookupError {
...
+            lookup_gate: lookup_gate.map(f),
+        }
+    }
+
+    pub fn as_ref(&self) -> LookupSelectors<&T> {
+        LookupSelectors {
+            chacha: self.chacha.as_ref(),
+            chacha_final: self.chacha_final.as_ref(),
+            lookup_gate: self.lookup_gate.as_ref(),
+        }
```
mimoo, line 138: implement this trait instead no? https://doc.rust-lang.org/std/convert/trait.AsRef.html

kimchi/src/circuits/lookup/index.rs
```diff
@@ -36,6 +36,108 @@ pub enum LookupError {
...
+}
+
+impl<T> LookupSelectors<T> {
+    pub fn map<U, F: Fn(T) -> U>(self, f: F) -> LookupSelectors<U> {
+        let LookupSelectors {
+            chacha,
+            chacha_final,
+            lookup_gate,
+        } = self;
+        let f = |x| f(x);
```
mimoo, line 125: why this line?

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -83,9 +84,14 @@ impl LookupInfo {
         &self,
         domain: &EvaluationDomains<F>,
         gates: &[CircuitGate<F>],
-    ) -> (Vec<Evaluations<F>>, Vec<LookupTable<F>>) {
+    ) -> (LookupSelectors<Evaluations<F>>, Vec<LookupTable<F>>) {
         let n = domain.d1.size();
-        let mut selector_values: Vec<_> = self.kinds.iter().map(|_| vec![F::zero(); n]).collect();
+
+        let mut selector_values: LookupSelectors<_> = Default::default();
+        for kind in self.kinds.iter() {
```
mimoo, line 92: `for kind in &self.kinds {`

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -108,14 +114,11 @@ impl LookupInfo {
...
-                E::<F, D<F>>::from_vec_and_domain(v, domain.d1)
-                    .interpolate()
-                    .evaluate_over_domain(domain.d8)
-            })
-            .collect();
+        let selector_values8: LookupSelectors<_> = selector_values.map(|v| {
+            E::<F, D<F>>::from_vec_and_domain(v, domain.d1)
+                .interpolate()
+                .evaluate_over_domain(domain.d8)
+        });
```
mimoo, line 129: I gotta say, I don't like the overly generic `LookupSelectors` and the `map` function. I'd rather we have more verbosity here than more complexity.

mimoo:
```rust

let mut selector_values = HashMap::new();
for kind in &self.kinds {
    selector_values.insert(kind, vec![F::zero(); n]);
}

// ...

let to_e8 = |pat| {
    selector_values.get(pat).map(|values| 
        E::<F, D<F>>::from_vec_and_domain(values, domain.d1)
            .interpolate()
            .evaluate_over_domain(domain.d8))
};

use LookupPattern::*;
let lookup_selectors = LookupSelectors {
    chacha: to_e8(ChaCha),
    chacha_final: to_e8(ChaChaFinal),
    lookup_gate: to_e8(LookupGate),
};
```

kimchi/src/circuits/lookup/index.rs
```diff
@@ -119,6 +119,9 @@ impl<T> LookupSelectors<T> {
             chacha_final,
             lookup_gate,
         } = self;
+        // This closure isn't really redundant -- it shields the parameter from a copy -- but
```
mimoo, line 122: ah I see, but I'm not sure I'm more clever than clippy here because I don't get how this is true

mimoo (approved):
approving, but I think it is weird that expr still uses indexing. I think ideally expr should not use indexing

kimchi/src/circuits/lookup/index.rs
mrmr1993: No. It has the wrong type.

kimchi/src/circuits/lookup/index.rs
mrmr1993: I wouldn't have added the generic parameter if it was always `Evaluations<F, D<F>>`. It's used with `PolyComm<G>` in `VerifierIndex`.

kimchi/src/circuits/lookup/index.rs
mrmr1993: Try deleting it :man_shrugging:

kimchi/src/circuits/lookup/lookups.rs
mrmr1993: > I gotta say, I'm not fond of the overly generic `LookupSelectors` and the `map` function. They seem like complex hack (that forces us to have SerdeAs code) to make this function (and this function only) more clever.

Nope, they're there for the the `VerifierIndex`.

> I'd rather we have more verbosity here than more complexity, and build the LookupSelectors struct explicitly in this function. It'll be a bit more verbose but much clearer and easier to maintain imo. wdyt?

Not a fan; leaving as is.

kimchi/src/circuits/lookup/index.rs
```diff
@@ -36,6 +36,125 @@ pub enum LookupError {
     TableIDZeroMustHaveZeroEntry,
 }
 
+/// Lookup selectors
+#[derive(Clone, Serialize, Deserialize, Debug, Default)]
+pub struct LookupSelectors<T> {
+    /// Chacha pattern lookup selector
+    pub chacha: Option<T>,
+    /// ChachaFinal pattern lookup selector
+    pub chacha_final: Option<T>,
```
jspada, line 45: Maybe,  as part of the (future) gate interface, have each gate define a vector of LookupSelectors (Some for each of the gate's CircuiteGates that has a lookup, otherwise None).  Thus, all the ChaCha lookups could be defined under the "gate".

jspada (commented):
Looks good - left a future idea comment.

PR #585, opened 2022-05-31
Only add lookup selectors for used gates

mrmr1993:
This PR builds upon #584.

This tweaks the logic in `ConstraintSystem::create` so that it only uses the lookups for gates that are present in the circuit. In particular, this allows us to migrate the pickles recursion layer to using these lookups without needing to support the `LookupPattern`s for all possible gates.

### DIFF
--- a/kimchi/src/circuits/lookup/index.rs
+++ b/kimchi/src/circuits/lookup/index.rs
@@ -182,12 +182,14 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
         runtime_tables: Option<Vec<RuntimeTableCfg<F>>>,
         domain: &EvaluationDomains<F>,
     ) -> Result<Option<Self>, LookupError> {
-        let lookup_info = LookupInfo::create(runtime_tables.is_some());
-
         //~ 1. If no lookup is used in the circuit, do not create a lookup index
-        match lookup_info.lookup_used(gates) {
+        match LookupInfo::create_from_gates(gates, runtime_tables.is_some()) {
             None => Ok(None),
-            Some(lookup_used) => {
+            Some(lookup_info) => {
+                let lookup_used = match lookup_info.lookup_used() {
+                    Some(lookup_used) => lookup_used,
+                    None => return Ok(None),
+                };
                 let d1_size = domain.d1.size();
 
                 // The maximum number of entries that can be provided across all tables.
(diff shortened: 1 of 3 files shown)


kimchi/src/circuits/lookup/index.rs
```diff
@@ -185,12 +185,14 @@ impl<F: FftField + SquareRootField> LookupConstraintSystem<F> {
...
         domain: &EvaluationDomains<F>,
     ) -> Result<Option<Self>, LookupError> {
-        let lookup_info = LookupInfo::create(runtime_tables.is_some());
-
         //~ 1. If no lookup is used in the circuit, do not create a lookup index
-        match lookup_info.lookup_used(gates) {
+        match LookupInfo::create_from_gates(gates, runtime_tables.is_some()) {
             None => Ok(None),
-            Some(lookup_used) => {
+            Some(lookup_info) => {
```
mimoo, line 191: to avoid all the indented code, this should be replaced with

```rust
let runtime_tables_used = runtime_tables.is_some();
let lookup_info = LookupInfo::create_from_gates(gates, runtime_tables_used).ok_or(...)?;
// ...
```

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -60,22 +61,35 @@ impl LookupInfo {
...
+            None
+        } else {
+            Some(Self::create(kinds, uses_runtime_tables))
+        }
+    }
+
+    /// Check what kind of lookups, if any, are used by this circuit.
+    pub fn lookup_used(&self) -> Option<LookupsUsed> {
+        let mut lookups_used = None;
+        for lookup_pattern in self.kinds.iter() {
```
mimoo, line 86: `for lookup_pattern in &self.kinds`

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -60,22 +61,35 @@ impl LookupInfo {
...
         }
+        if kinds.is_empty() {
+            None
+        } else {
+            Some(Self::create(kinds, uses_runtime_tables))
+        }
+    }
+
+    /// Check what kind of lookups, if any, are used by this circuit.
+    pub fn lookup_used(&self) -> Option<LookupsUsed> {
```
mimoo, line 84: why not store that information in `LookupInfo`?

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -60,22 +61,35 @@ impl LookupInfo {
...
+        let mut lookups_used = None;
```
mimoo, line 85: why not make the default `LookupsUsed::Single` instead of an `Option`?

mimoo (approved):

PR #586, opened 2022-05-31
use oncecell on w, zkpm and srs

chee-chyuan:


mimoo (dismissed):
Beautiful! There's a number of manual deserialization of `zk_w3` on the mina side. I can take a look shortly.

mimoo:
looks like this is still a draft, so won't merge now

chee-chyuan:
> looks like this is still a draft, so won't merge now

yea. i also plan to do the same to other parameters which inly depends on verifier_index. let me do that before i mark it as ready

chee-chyuan:
i also changed abit on how srs is being loaded `from_file` and added an error for that. lmk if you need i need to amend it

kimchi/src/verifier_index.rs
```diff
@@ -236,6 +238,24 @@ impl<G: CommitmentCurve> VerifierIndex<G>
 where
     G::BaseField: PrimeField,
 {
+    pub fn srs(&self) -> &Arc<SRS<G>> {
+        self.srs.get_or_init(|| {
+            let mut srs = SRS::<G>::create(self.max_poly_size);
```
mrmr1993, line 243: I don't think we want to create a new SRS here, for memory usage / performance reasons. Can we continue to use the one from the prover index?

kimchi/src/verifier_index.rs
```diff
@@ -64,7 +66,7 @@ pub struct VerifierIndex<G: CommitmentCurve> {
     pub max_quot_size: usize,
     /// polynomial commitment keys
     #[serde(skip)]
-    pub srs: Arc<SRS<G>>,
+    srs: OnceCell<Arc<SRS<G>>>,
```
mrmr1993, line 69: This field (and the others below) need to be public, otherwise the mina side won't be able to create a verifier index from one it receives over the network.

mrmr1993 (changes requested):

kimchi/src/verifier_index.rs
chee-chyuan: my bad on this. not aware how mina does things

kimchi/src/verifier_index.rs
```diff
@@ -177,7 +179,7 @@ where
             max_poly_size: self.max_poly_size,
             max_quot_size: self.max_quot_size,
             powers_of_alpha: self.powers_of_alpha.clone(),
-            srs: Arc::clone(&self.srs),
+            srs: OnceCell::new(),
```
mrmr1993, line 182: If you change this to
```rust
srs: {
    let cell = OnceCell::new();
    cell.set(Arc::clone(&self.srs));
    cell
},
```
then we get back the old behaviour while still allowing for deserialisation.

chee-chyuan: arr ok i gotcha!

kimchi/src/verifier_index.rs
mimoo: I think you should change the field to be `Option<Arc<SRS>>`, like that it will deserialize to `None`. You can then implement `srs()` as:

```rust
fn srs(&self) -> Arc<SRS> {
  self.srs.expect("You must use `set_srs` to initialize the SRS")
}

  fn set_srs(&mut self, srs: Arc<SRS>) {
    self.srs = Some(srs)
  }
```

chee-chyuan: but that loses the ‘auto’ deserialization property though. or we dont really need that for srs?

mimoo: Oh no you’re right. I was thinking that we’d never want to do this, but actually we’d probably want to do it. We can probably use max_poly_size in the verifier index to init it if it’s not initialized

kimchi/src/error.rs
```diff
@@ -65,3 +65,10 @@ pub enum SetupError {
     #[error("the domain could not be constructed: {0}")]
     DomainCreation(&'static str),
 }
+
+/// Errors that can arise when creating a verifier index
+#[derive(Error, Debug, Clone)]
+pub enum VerifierIndexError {
+    #[error("unable to set srs")]
```
mimoo, line 72: maybe say something like "SRS already set"

mimoo:
this looks good to me now! But @mrmr1993 I'll let you approve

mrmr1993 (approved):
Looks great, thank you!

Issue #541, opened 2022-05-05
Builder pattern for setup


mimoo:
The setup is starting to get convoluted. It might be nice to explore being able to create it like this:

```rust
let cs = ConstraintSystem::create(gates, fr_sponge_params).public(5).lookup_tables(t).runtime_tables(rt).shared_precomputations(sp).build();
```

PR #588, opened 2022-06-01
[easy] Builder for setup

querolita:
This PR addresses [Issue 541](https://github.com/o1-labs/proof-systems/issues/541). 

Before, creating a constraint system was done like this:

```rust
let cs = ConstraintSystem::create(gates, fq_sponge_params, public, lookup_tables, runtime_tables);
```
which used yet another method called `create_with_shared_precomputations` that introduced a `None` additional parameter. 

We are increasingly adding parameters that are optional, or default to empty vectors, but the static-sized `create()` method cannot take less arguments.

After these modifications, we can create our constraint systems step by step, choosing between all of the things that we want to include. The necessary fields are the gates and the parameters for the sponge. The others can have values by default, as follows:

```rust
let cs = ConstraintSystem::create(gates, fq_sponge_params).public(p).lookup(lt).runtime(rt).precomputations(pc).build();
```

Even if it may look just as long, when any of the above are not used, the builder looks shorter and more readable. 

The way I modelled it is by introducing a `Builder` type that is the output type of the `create()` method, then `public(), lookup(), runtime(), precomputations(), and build()` are implemented for the builder, and the `build()` finally returns a `Result<ConstraintSystem<F>, SetupError>`, just like the original `create()` method used to do.

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -1148,10 +1148,10 @@ To create the index, follow these steps:
 
    To do this, for each table:
 
-~ - Update the corresponding entries in a table id vector (of size the domain as well)
+	- Update the corresponding entries in a table id vector (of size the domain as well)
    with the table ID of the table.
-~ - Copy the entries from the table to new rows in the corresponding columns of the concatenated table.
-~ - Fill in any unused columns with 0 (to match the dummy value)
+	- Copy the entries from the table to new rows in the corresponding columns of the concatenated table.
+	- Fill in any unused columns with 0 (to match the dummy value)
 6. Pad the end of the concatened table with the dummy value.
 7. Pad the end of the table id vector with 0s.
 8. pre-compute polynomial and evaluation form for the look up tables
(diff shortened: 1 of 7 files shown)


kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -112,6 +114,78 @@ pub fn zk_polynomial<F: FftField>(domain: D<F>) -> DensePolynomial<F> {
     ])
 }
 
+/// Shifts represent the shifts required in the permutation argument of PLONK.
```
querolita, line 117: I moved this to the permutation module as suggested in a TODO comment

kimchi/src/circuits/constraints.rs
```diff
@@ -136,129 +135,205 @@ pub struct ConstraintSystem<F: FftField> {
...
+            public: 0,
+            lookup_tables: vec![],
+            runtime_tables: None,
+            precomputations: None,
+        }
+    }
 
-        //
-        Self { shifts, map }
+    pub fn precomputations(&self) -> &Arc<DomainConstantEvaluations<F>> {
```
querolita, line 185: these changes are just moving the code in the file, not real modifications

kimchi/src/circuits/constraints.rs
```diff
@@ -136,129 +135,205 @@ pub struct ConstraintSystem<F: FftField> {
...
-#[derive(Debug)]
-pub enum GateError {
-    /// Some connected wires have different values
-    DisconnectedWires(Wire, Wire),
-    /// A public gate was incorrectly connected
-    IncorrectPublic(usize),
-    /// A specific gate did not verify correctly
-    Custom { row: usize, err: String },
-}
+impl<F: FftField + SquareRootField> Builder<F> {
```
querolita, line 290: the actual changes start here

book/src/specs/kimchi.md
```diff
@@ -1148,10 +1148,10 @@ To create the index, follow these steps:
 
    To do this, for each table:
 
-	- Update the corresponding entries in a table id vector (of size the domain as well)
+~ - Update the corresponding entries in a table id vector (of size the domain as well)
```
mrmr1993, line 1151: You need to update your `cargo spec`

kimchi/src/circuits/polynomials/range_check/gate.rs
```diff
@@ -285,8 +285,12 @@ mod tests {
     fn create_test_constraint_system() -> ConstraintSystem<PallasField> {
         let (_, gates) = CircuitGate::<PallasField>::create_range_check(0);
 
-        ConstraintSystem::create(gates, vec![], None, oracle::pasta::fp_kimchi::params(), 0)
+        ConstraintSystem::create(gates, oracle::pasta::fp_kimchi::params())
+            .build()
             .unwrap()
+
+        //ConstraintSystem::create(gates, vec![], None, oracle::pasta::fp_kimchi::params(), 0)
```
mrmr1993, line 292: Delete?

mrmr1993 (changes requested):
Looks good, but left a few comments

book/src/specs/kimchi.md
querolita: Oh thanks!! So that was it, CI was failing and I was not sure what was happening with the specs 👍🏻

kimchi/src/circuits/constraints.rs
```diff
@@ -136,129 +135,204 @@ pub struct ConstraintSystem<F: FftField> {
...
+    /// Set up the number of public inputs.
+    /// If not invoked, it equals `0` by default.
+    pub fn public(mut self, public: usize) -> Self {
+        self.public = public;
+        self
+    }
```
mimoo, line 296: unrelated to the PR, but a thought: it'd be interesting if we had public input gates (or even public input and public output gates) so that we wouldn't have to pass that argument. The gates wouldn't really exist (they would still use the double generic gate underneath) but could allow nicer abstractions

querolita: oh that is very cute, definitely a good idea

mimoo, line 296: unrelated to the PR, but a thought: it'd be interesting if we had public input gates (or even public input and public output gates) so that we wouldn't have to pass that argument. The gates wouldn't really exist (they would still use the double generic gate underneath) but could allow nicer abstractions

mrmr1993: I'm not sure that adding this at the `GateType` level is a good idea; then, we have to remember to handle this `PublicInput` kind everywhere, including where it appears part-way through the circuit. It might still make sense to add this to snarky / Izzy's circuit builder API, but I'm not sure whether it actually makes things cleaner than e.g. snarky's input `Typ.t` interface.

mrmr1993 (approved):

PR #590, opened 2022-06-02
Compute unnormalized lagrange basis in a faster way

### LINKED ISSUE #589: bugfix - add missing places where faster lagrange basis computation should be used
* Add missing places where faster lagrange basis computation should be used
* Also switch to faster way to compute it using `inverse`, i.e., `g^{n-i} = 1/g^i`

jspada:
* Switch to faster way of computing lagrange basis
      g^{n-i} = 1/g^i
    * Fix missing spots where we should use this
    
   closes #589

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -145,6 +145,16 @@ pub fn l0_1<F: FftField>(d: D<F>) -> F {
         .fold(F::one(), |acc, omega_j| acc * (F::one() - omega_j))
 }
 
+// Compute the ith unnormalized lagrange basis
+fn unnormalized_lagrange_basis<F: FftField>(domain: &D<F>, i: i32, pt: &F) -> F {
+    let omega_i = if i < 0 {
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):

kimchi/src/circuits/expr.rs
```diff
@@ -145,6 +145,16 @@ pub fn l0_1<F: FftField>(d: D<F>) -> F {
...
 
+// Compute the ith unnormalized lagrange basis
+fn unnormalized_lagrange_basis<F: FftField>(domain: &D<F>, i: i32, pt: &F) -> F {
+    let omega_i = if i < 0 {
+        domain.group_gen.pow(&[-i as u64]).inverse().unwrap()
+    } else {
+        domain.group_gen.pow(&[i as u64])
+    };
+    domain.evaluate_vanishing_polynomial(*pt) / (*pt - omega_i)
+}
```
mimoo, line 156: <3

PR #595, opened 2022-06-03
Add unnormalized lagrange basis unit test

### LINKED ISSUE #594: Unit test for check both ways of computing unnormalized lagrange basis give same result
David has suggested that we should create a unit test for verifying that both ways of computing unnormalized lagrange basis give same result

jspada:
Closes #594


mimoo (approved):
nice!

PR #596, opened 2022-06-03
[snarky-rs] add Sys to Var

mimoo:
This is a proposal to add `Sys: Cs` to `Var` in the circuit writer.

Essentially, we need to pass a main `sys` around all the time to do interesting things with our variables (like add two variables together). It's a bit annoying, so this PR adds a pointer to `sys` in vars so you can just do things like:

```rust
    let x1 = sys.var(|| F::one());
    let x2 = sys.var(|| F::one());
    let x3 = x1 + x2;
    let two = sys.constant(F::from(2u32));
    sys.assert_eq(x3, two);
```

instead of

```rust
    let x1 = sys.var(|| F::one());
    let x2 = sys.var(|| F::one());
    let x3 = sys.add(x1, x2);
    let two = sys.constant(F::from(2u32));
    sys.assert_eq(x3, two);
```

I'm actually not convinced that it's the best idea, as it makes the code a bit more complicated/verbose for a marginal gain. But perhaps it is still a good idea if we push this further. What do people think? Maybe I took a more complicated than necessary approach also.

The approach can be summarized as:

```diff
// a var now holds a Sys

- pub struct Var<F>
+ pub struct Var<Sys, F>
where
    Sys: Clone,
{
    /// A unique identifier for this variable.
    pub index: usize,
    /// ends up holding a value when generating the witness.
    pub value: Option<F>,
+    /// A pointer to the system.
+    pub sys: Sys,
}

// now Cs is implemented on a Rc<RefCell<Sys>>

- impl<F: PrimeField> Cs<F> forWitnessGenerator<F> {
+ impl<F: PrimeField> Cs<F> for Rc<RefCell<WitnessGenerator<F>>> {

// So these changes need to happen on Cs itself:

- pub trait Cs<F: PrimeField> {
+ pub trait Cs<F: PrimeField>: Sized + Clone {
-    fn var<G>(&mut self, g: G) -> Var<F>
+    fn var<G>(&self, g: G) -> Var<Self, F>

```

where

### DIFF
--- a/circuit-construction/src/lib.rs
+++ b/circuit-construction/src/lib.rs
@@ -15,6 +15,9 @@ use mina_curves::pasta::{fp::Fp, fq::Fq, pallas::Affine as Other, vesta::Affine}
 use oracle::{constants::*, permutation::full_round, poseidon::ArithmeticSpongeParams, FqSponge};
 use std::{cell::RefCell, collections::HashMap, ops::Add, rc::Rc};
 
+#[cfg(test)]
+mod tests;
+
 pub const GENERICS: usize = 3;
 pub const ZK_ROWS: usize = kimchi::circuits::polynomials::permutation::ZK_ROWS as usize;
 
--- a/circuit-construction/src/lib.rs
+++ b/circuit-construction/src/lib.rs
@@ -13,7 +13,7 @@ use kimchi::circuits::{
 use kimchi::{plonk_sponge::FrSponge, proof::ProverProof, prover_index::ProverIndex};
 use mina_curves::pasta::{fp::Fp, fq::Fq, pallas::Affine as Other, vesta::Affine};
 use oracle::{constants::*, permutation::full_round, poseidon::ArithmeticSpongeParams, FqSponge};
-use std::{cell::RefCell, collections::HashMap, rc::Rc};
+use std::{cell::RefCell, collections::HashMap, ops::Add, rc::Rc};
 
 pub const GENERICS: usize = 3;
 pub const ZK_ROWS: usize = kimchi::circuits::polynomials::permutation::ZK_ROWS as usize;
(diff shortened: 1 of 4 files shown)


mimoo:
cc @katat

circuit-construction/src/lib.rs
```diff
@@ -101,24 +105,47 @@ impl Cycle for FqInner {
...
+pub struct Var<Sys, F>
+where
+    Sys: Clone,
+{
+    /// A unique identifier for this variable.
     pub index: usize,
+    /// ends up holding a value when generating the witness.
     pub value: Option<F>,
+    /// A pointer to the system.
+    pub sys: Sys,
```
mrmr1993, line 121: If this is supposed to be a pointer, make it one. The `Arc<Rc<Sys>>` should be an implementation detail.

circuit-construction/src/lib.rs
```diff
@@ -101,24 +105,47 @@ impl Cycle for FqInner {
...
+    F: Copy,
+    Sys: Clone,
+{
     pub fn val(&self) -> F {
         self.value.unwrap()
     }
 }
 
-pub struct ShiftedScalar<F>(Var<F>);
+/// ?
```
mrmr1993, line 134: This is not a useful change; delete

circuit-construction/src/lib.rs
```diff
@@ -101,24 +105,47 @@ impl Cycle for FqInner {
...
+pub struct ShiftedScalar<Sys: Clone, F>(Var<Sys, F>);
 
-pub struct GateSpec<F: FftField> {
+/// Represents a gate before compilation.
```
mrmr1993, line 137: Don't add doc-comments as part of other unrelated changes.

circuit-construction/src/lib.rs
```diff
@@ -101,24 +105,47 @@ impl Cycle for FqInner {
     type OuterProj = <Other as AffineCurve>::Projective;
 }
 
-#[derive(Hash, Eq, PartialEq, Debug, Clone, Copy)]
-pub struct Var<F> {
+type Ref<Sys> = Rc<RefCell<Sys>>;
```
mrmr1993, line 108: Shouldn't we try to be thread-safe?

mrmr1993 (changes requested):
If we're doing this, we should make sure that the `sys` field is the same for both variables when we use it.

mimoo:
waiting on feedback for https://github.com/o1-labs/proof-systems/pull/605 which should simplify this approach

PR #601, opened 2022-06-07
[easy][commitment] introduce BlindedCommitment struct

mimoo:
introduces this struct:

```rust
pub struct BlindedCommitment<G>
where
    G: CommitmentCurve,
{
    pub commitment: PolyComm<G>,
    pub blinders: PolyComm<G::ScalarField>,
}
```


mrmr1993 (approved):
Love it!

PR #602, opened 2022-06-07
[easy] remove ScalarField and BaseField helpers

mimoo:
Do we really need to write `ScalarField<G>` and `BaseField<G>` when we can just write `G::ScalarField` and `G::BaseField`? Nope!


mrmr1993 (approved):
Personally I thought it looked a bit nicer, but I'm not that attached. You'll need to update the spec though.

mimoo:
my thinking: it removes code, reduces the amount of `<<>>>`, and even reduces the amount of code when you write things like 

```diff
- ScalarField::<G>::zero()
+ G::ScalarField::zero()
```

PR #603, opened 2022-06-07
[easy] Add `RecursionChallenge` type for `ProverProof`

querolita:
This PR adds a type called `RecursionChallenge` to substitute the tuple of `prev_evals` inside `ProverProof`. This was proposed in the older [PR for serialization](https://github.com/o1-labs/proof-systems/pull/570) that was substituted by [this one](https://github.com/o1-labs/proof-systems/pull/599/files), already merged. If there's interest for this change, I already had [this PR](https://github.com/MinaProtocol/mina/pull/11130) addressing the mina side, that I would reopen.

It also compresses the two test files about serialization into one file only, as they were redundant.

### DIFF
--- a/kimchi/src/proof.rs
+++ b/kimchi/src/proof.rs
@@ -25,6 +25,9 @@ pub struct LookupEvaluations<F: CanonicalSerialize + CanonicalDeserialize> {
     /// lookup table polynomial
     #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
     pub table: Vec<F>,
+    /// Optionally, a runtime table polynomial.
+    #[serde_as(as = "Option<Vec<o1_utils::serialization::SerdeAs>>")]
+    pub runtime: Option<Vec<F>>,
 }
 
 // TODO: this should really be vectors here, perhaps create another type for chunked evaluations?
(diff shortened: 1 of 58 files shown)


kimchi/src/prover.rs
```diff
@@ -1050,7 +1051,8 @@ where
 
         let polys = prev_challenges
             .iter()
-            .map(|(chals, comm)| {
+            .map(|c| {
+                let RecursionChallenge { chals, comm } = c;
```
mrmr1993, line 1055: Nit: inline this pattern in the argument?

kimchi/src/proof.rs
```diff
@@ -35,6 +36,9 @@ pub struct LookupEvaluations<Field> {
 }
 
 // TODO: this should really be vectors here, perhaps create another type for chunked evaluations?
+/// Polynomial evaluations contained in a `ProverProof`.
```
mrmr1993, line 39: Please add unrelated doc-comments in their own PR!

kimchi/src/verifier.rs
```diff
@@ -72,9 +72,10 @@ where
     ) -> Vec<Vec<Vec<ScalarField<G>>>> {
         self.prev_challenges
             .iter()
-            .map(|(chals, _poly)| {
+            .map(|c| {
                 // No need to check the correctness of poly explicitly. Its correctness is assured by the
                 // checking of the inner product argument.
+                let RecursionChallenge { chals, comm: _ } = c;
```
mrmr1993, line 78: Nit: inline in function argument?

querolita: Thanks!

kimchi/src/prover.rs
querolita: Thanks!

kimchi/src/proof.rs
querolita: 🤦🏻 oh, I usually forget

PR #605, opened 2022-06-08
[circuit-writer] remove Cs trait

mimoo:
turns out we don't need to be clever with the `Cs` trait implemented on two different structs, we can just have an enum and a single implementation. Removes a bunch of code and reduces complexity.

### DIFF
--- a/circuit-construction/src/lib.rs
+++ b/circuit-construction/src/lib.rs
@@ -114,7 +114,7 @@ impl<F: Copy> Var<F> {
 
 pub struct ShiftedScalar<F>(Var<F>);
 
-pub struct GateSpec<F: FftField> {
+pub struct GateSpec<F> {
     pub typ: GateType,
     pub row: [Var<F>; COLUMNS],
     pub c: Vec<F>,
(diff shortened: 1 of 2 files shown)


mimoo:
What's the rationale?

mimoo:
I was trying to think of why the generics could be nice in the long run. I couldn't think of much, but I thought of another argument against: extending on this PR, we could enforce that the CircuitGenerator is first ran (via a `compile()` function) before we generate a witness.

```rust
impl CircuitGenerator {
  fn compile(self) -> WitnessGenerator {
    // logic
  }
}
```

this way we cannot, by mistake, do changes to a circuit after generating a witness for it.

Another thing that is nice, is that we can also avoid duplication of some logic that needs to be performed in both the witness and the circuit generator. For example, for the double generic gate we queue generic gates until we accumulate two before actually adding a gate to the circuit. We must not forget to flush any queued generic gate when we compile the circuit AND produce the witness. This is duplicated logic. Another example is cached constants, I believe we would also have to also track these on the witness side.

If we could compile the circuit down to just a number of vars, it'd be much easier to generate the witness: we'd just read the values of each var (or unwrap to 0). There would be no logic there.

PR #607, opened 2022-06-08
[easy] Changing array of 2 (zeta and omegaz) for `ConsecutiveEvals` inside `ProverProof`

querolita:
This PR addresses a `TODO` comment that was suggesting to create a type for the `evals` fiend inside `ProverProof`. 

~~Because we were using the tuple structure to iterate over the zeta and omegazeta evaluations, I created a method that returns the array version of the struct. I don't love it, but I thought rephrasing the prover would be less concise.~~

I defined the type generically enough to be instantiated with single fields or vectors of fields. This is useful to reuse the `combine()` method of `ProofEvaluations` to replace 25 loc in `prover.rs`. `ConsecutiveEvals` is now used more in the code, substituting tuples of `ProofEvaluations`. This includes the methods inside `expr.rs`. I created three constants representing indices inside the struct (0 for zeta, 1 for zetaomega). Implemented `Index` for this type, because I didn't want to change the logics inside `evaluate()` for `Variable`. 

Any suggestions are welcome.


book/src/specs/kimchi.md
```diff
@@ -1474,8 +1474,7 @@ pub struct ProverProof<G: AffineCurve> {
     pub proof: OpeningProof<G>,
 
     /// Two evaluations over a number of committed polynomials
-    // TODO(mimoo): that really should be a type Evals { z: PE, zw: PE }
-    pub evals: [ProofEvaluations<Vec<ScalarField<G>>>; 2],
+    pub evals: ConsecutiveEvals<G>,
```
mimoo, line 1477: `ConsecutiveEvals<G::ScalarField>` would be more correct

kimchi/src/proof.rs
```diff
@@ -144,6 +157,12 @@ impl<G: AffineCurve> RecursionChallenge<G> {
     }
 }
 
+impl<G: AffineCurve> ConsecutiveEvals<G> {
+    pub fn array(&self) -> [&ProofEvaluations<Vec<ScalarField<G>>>; 2] {
```
mimoo, line 161: I don't necessarily like these functions because they are error prone (what if we add an evaluation at some point and forget to update this function?)

ideas: deriving this with a macro, or using something [like this](https://stackoverflow.com/questions/54177438/how-to-programmatically-get-the-number-of-fields-of-a-struct) as an easier solution to make sure we created an array of the right size, or just accepting this tiny risk

kimchi/src/prover.rs
```diff
@@ -919,8 +922,9 @@ where
         let evals = {
             let power_of_eval_points_for_chunks = [zeta_to_srs_len, zeta_omega_to_srs_len];
             &chunked_evals
+                .array()
```
mimoo, line 925: I suggest re-implementing this without the `array()`, which would make this part of the code clearer :o

kimchi/src/verifier.rs
```diff
@@ -278,7 +278,7 @@ where
         //~~ - poseidon selector
         //~~ - the 15 register/witness
         //~~ - 6 sigmas evaluations (the last one is not evaluated)
-        for (p, e) in p_eval.iter().zip(&self.evals) {
+        for (p, e) in p_eval.iter().zip(&self.evals.array()) {
```
mimoo, line 281: same here, we don't need to call `array`

kimchi/src/verifier.rs
```diff
@@ -396,18 +396,24 @@ where
             es.push((p_eval.clone(), None));
             es.push((vec![ft_eval0, ft_eval1], None));
             es.push((
-                self.evals.iter().map(|e| e.z.clone()).collect::<Vec<_>>(),
+                self.evals
+                    .array()
```
mimoo, line 400: create two push here :)

mimoo:
I think it's a good idea to use a strong type instead of a vec/array, but if we do use a strong type I think we should go all the way and avoid just iterating on it like it's an array :o 

Also, I think there should be an associated PR on the mina side to make sure this works

kimchi/src/proof.rs
querolita: I see the point, if we do it, it should better be a constant. The reason that I convert to an array is because in the prover we tend to iterate over the two elements of `ConsecutiveEvals` (meaning the zeta and the zeta_omega), together with other tuples in the code (for example, using them in `zip` calls). Other solutions that come to mind seem more verbose 🤔

kimchi/src/prover.rs
querolita: Yeah, or that is another option. I wouldn't like it to look "repetitive", since the code is the same both for zeta and zetaomega. I see the point of using the iterators there, I will think of more elegant ways for this part then.

kimchi/src/verifier.rs
querolita: Yeah... This solution with the array is the one with fewer modifications in the original code. But if we are fine with being more explicit about zeta and zeta_omega, then I can change it.

querolita:
Answering @mimoo , I will write the mina side as soon as the mina side for `RecursionChallenge` is merged. I am having issues building the repo, so it is taking me a while to configure everything. In the meantime I can think of refactors for the prover code without relying so much on arrays.

kimchi/src/prover.rs
mimoo: I think there's different ways to handle that, either create a closure and call it on both, or do something like

```rust
let CombinedEvals { z: chunked_z, zw: chunked_zw } = chunked_evals;
for chunked_evals in [z, zw] { ... }
```

I think even better is to implement most of the logic on the `CombinedEvals` type as methods.

kimchi/src/verifier.rs
```diff
@@ -313,10 +312,16 @@ where
...
 
-        let evals = vec![
-            self.evals.z.combine(powers_of_eval_points_for_chunks[0]),
-            self.evals.zw.combine(powers_of_eval_points_for_chunks[1]),
-        ];
+        let evals = ConsecutiveEvals {
+            zeta: self
+                .evals
+                .zeta
+                .combine(powers_of_eval_points_for_chunks[Z_IDX]),
```
mimoo, line 319: might want to fix powers_of_eval_points_for_chunks to be a type as well, instead of an array :p

querolita: Good point

Issue #466, opened 2022-03-26
Enforce the public input length


mimoo:
I don't think the public input length is enforced in any way on the verifier side. It'd be good to make sure that this is the case. (the length of the public input should probably be in the verifier index)

cc @querolita if we want to make this library usable for people with different usecases

mrmr1993:
I don't think this is necessary. The verifier absorbs the public input into the sponge, so we'll notice pretty quickly if the prover sends a string of the wrong length.

mimoo:
BTW @mrmr1993 this is the code on the verifier side:

```rust
    //~ 1. Commit to the negated public input polynomial.
    let lgr_comm = index
        .srs
        .lagrange_bases
        .get(&index.domain.size())
        .expect("pre-computed committed lagrange bases not found");
    let com: Vec<_> = lgr_comm
        .iter()
        .map(|c| PolyComm {
            unshifted: vec![*c],
            shifted: None,
        })
        .take(proof.public.len())
        .collect();
    let com_ref: Vec<_> = com.iter().collect();
    let elm: Vec<_> = proof.public.iter().map(|s| -*s).collect();
    let p_comm = PolyComm::<G>::multi_scalar_mul(&com_ref, &elm);
```

So I'm making a few proposals here:

* specify that better, it wasn't clear what happened on the spec side and had to go back to this
* add public_input_size to the verifier index, and check that the public input is of the correct size in verify
* remove the public input from the proof itself, make that an argument of verify

mimoo:
reopening as we still haven't moved the public input to an argument of verify(). I'll change the title to reflect that.

PR #609, opened 2022-06-08
[kimchi] absorb verifier index + recursion data

mimoo:
- [x] adds number of proofs verifier in a recursion circuit to `ConstraintSystem`
- [x] adds public input size to verifier index
- [x] absorbs verifier index in proof
- [x] absorbs prev challenges in proof
- [x] specify

closes https://github.com/o1-labs/proof-systems/issues/466 https://github.com/o1-labs/proof-systems/issues/552

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -191,6 +191,15 @@ where
         //~ 1. Setup the Fq-Sponge.
         let mut fq_sponge = EFqSponge::new(index.fq_sponge_params.clone());
 
+        //~ 1. Absorb the verifier index
+        let verifier_index = index.verifier_index();
+        verifier_index.absorb(&mut fq_sponge);
+
+        //~ 1. If recursion commitments are passed, absorb them
+        for RecursionChallenge { comm, .. } in &prev_challenges {
+            fq_sponge.absorb_g(&comm.unshifted);
+        }
+
         //~ 1. Compute the negated public input polynomial as
         //~    the polynomial that evaluates to $-p_i$ for the first `public_input_size` values of the domain,
         //~    and $0$ for the rest.
@@ -1033,6 +1042,13 @@ where
 
         //~ 1. Absorb the unique evaluation of ft: $ft(\zeta\omega)$.
         fr_sponge.absorb(&ft_eval1);
+
+        //~ 1. If recursion challenges are passed, absorb them
+        for RecursionChallenge { chals, .. } in &prev_challenges {
+            for chal in chals {
+                fr_sponge.absorb(chal);
+            }
+        }
 
         //~ 1. Sample $v'$ with the Fr-Sponge
         let v_chal = fr_sponge.challenge();
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -196,9 +196,12 @@ where
         verifier_index.absorb(&mut fq_sponge);
 
         //~ 1. If recursion commitments are passed, absorb them
+        let mut temp_fq_sponge = EFqSponge::new(index.fq_sponge_params.clone()); // TODO: no domain separation from the main sponge?
         for RecursionChallenge { comm, .. } in &prev_challenges {
-            fq_sponge.absorb_g(&comm.unshifted);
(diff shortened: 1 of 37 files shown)


kimchi/src/prover.rs
```diff
@@ -191,6 +191,15 @@ where
         //~ 1. Setup the Fq-Sponge.
         let mut fq_sponge = EFqSponge::new(index.fq_sponge_params.clone());
 
+        //~ 1. Absorb the verifier index
+        let verifier_index = index.verifier_index();
```
mrmr1993, line 195: Doesn't need to be done now, but we should compute this once while we're creating the index and cache the initialised sponge in the index. We don't really want to recommit to everything in the verifier index in every proof.

kimchi/src/verifier.rs
```diff
@@ -63,6 +63,7 @@ impl<G: CommitmentCurve> ProverProof<G>
 where
     G::BaseField: PrimeField,
 {
+    /// ?
```
mrmr1993, line 66: Stop doing this.

kimchi/src/verifier_index.rs
```diff
@@ -248,6 +258,58 @@ impl<G: CommitmentCurve> VerifierIndex<G>
...
+            for commitment in commitments {
+                sponge.absorb_g(&commitment.unshifted);
+            }
+
+            // TODO: should we absorb:
+            // - lookup_used
+            // - max_joint_size
+        }
+
+        // TODO: should we absorb:
```
mrmr1993, line 306: No

kimchi/src/verifier_index.rs
```diff
@@ -248,6 +258,58 @@ impl<G: CommitmentCurve> VerifierIndex<G>
...
+                .chain(lcs.lookup_selectors.lookup_gate.iter())
+                .chain(lcs.lookup_selectors.range_check_gate.iter())
+                .chain(lcs.table_ids.iter())
+                .chain(lcs.runtime_tables_selector.iter());
+
+            for commitment in commitments {
+                sponge.absorb_g(&commitment.unshifted);
+            }
+
+            // TODO: should we absorb:
```
mrmr1993, line 301: No

kimchi/src/verifier_index.rs
```diff
@@ -248,6 +258,58 @@ impl<G: CommitmentCurve> VerifierIndex<G>
...
+            .chain([&self.psm_comm])
+            .chain([&self.complete_add_comm])
+            .chain([&self.mul_comm])
+            .chain([&self.emul_comm])
+            .chain([&self.endomul_scalar_comm])
+            .chain(self.chacha_comm.iter().flatten())
+            .chain(&self.range_check_comm);
+
+        for commitment in commitments {
+            sponge.absorb_g(&commitment.unshifted);
```
mrmr1993, line 282: I think (for pickles reasons) we want to hash the verifier index and then absorb that hash. Can you change this to do that?

kimchi/src/verifier_index.rs
```diff
@@ -183,6 +190,9 @@ where
                 cell
             },
 
+            public_input_size: self.cs.public,
+            recursion_challenges: todo!(),
```
mrmr1993, line 194: Take it as an input to the constraint system builder, pass it through from there.

mrmr1993 (changes requested):

kimchi/src/verifier.rs
mimoo: it's a WIP PR, I can do whatever I want :P

mrmr1993: You did ask me to look at it!

mimoo:
Interestingly, I expected being able to create a proof that has prev_challenges of random length (at least as long as it's smaller than our SRS size, why would there be an issue?) 
But tests don't pass when I try that. From a quick glance I think the verifier's code can handle prev_challenges from a higher domain, but not from a smaller domain:  https://github.com/o1-labs/proof-systems/blob/e123cf8362c57191c4ab96501eb80a069e3a8aaa/kimchi/src/verifier.rs#L66

investigating...

Issue #591, opened 2022-06-02
Necessity of sponge parameters within the `ConstraintSystem` setup `create()` function


querolita:
(Continuation of [this PR](https://github.com/o1-labs/proof-systems/pull/588))

As suggested by @mimoo in [Issue 541](https://github.com/o1-labs/proof-systems/issues/541), creating a constraint system used to be done like this:

```rust
let cs = ConstraintSystem::create(gates, fq_sponge_params, public, lookup_tables, runtime_tables);
```
Right now, this is simplified by having the ability to configure **most** of these parameters individually and incrementally like this:

```rust
let cs = ConstraintSystem::create(gates, fq_sponge_params).public(p).lookup(lt).runtime(rt).precomputations(pc).build();
```
The issue is:
When are we really using the sponge parameters? Are they always necessary to be included in the very first `create()` method together with the gates? Should they have an additional setter, which is optional as the rest like `public()` or `runtime()`? Are there any parts of the code where the sponge parameters should be an `Option` type?

fabrizio-m:
I'll take it.

fabrizio-m:
Those parameters are necessary for: 
- The Poseidon gate.
- To generate the `u` and `v` challenges.

For the first item the optionality of the parameters depends on the optionality of the Poseidon gate.
More importantly, for the second item the challenges are used to create the opening proofs, I will look at SRS::open() a bit more but it seems that those challenges can't be optional.

mimoo:
Actually IIUC the poseidon gate doesn't use this param, it uses the `fr_sponge_params` in the `ConstraintSystem` instead. But yeah the sponge we use for Fiat-Shamir does use `fq_sponge_params`. I'm wondering if it would make sense to move this to the SRS?

fabrizio-m:
There seems to be some inconsistencies with the names, but this `fq_sponge_params` seems to be the `fr_sponge_params` that is used in poseidon.
the builder created for the benchmark takes `fp_sponge_params` from `oracle::pasta::fp_kimchi::params()`, then Builder::build() stores it in `ConstraintSystem`, field `fr_sponge_params`.
By the way, is there any meaning to fp, fr and fp beyond differentiating fields?

About moving the params to the SRS, I think it depends on how normal it will to change the params.
Are there any reason to use more than one sponge for the same curve?

mimoo:
> By the way, is there any meaning to fp, fr and fp beyond differentiating fields?

[...]

PR #610, opened 2022-06-09
move sponge from constraint system to srs

fabrizio-m:
Should solve #591


circuit-construction/tests/example_proof.rs
```diff
@@ -64,23 +64,20 @@ const PUBLIC_INPUT_LENGTH: usize = 3;
 #[test]
 fn test_example_circuit() {
     // create SRS
+    let proof_system_constants = fp_constants();
     let srs = {
-        let mut srs = SRS::<Affine>::create(1 << 7); // 2^7 = 128
+        let mut srs = SRS::<Affine>::create(1 << 7, proof_system_constants.poseidon.clone()); // 2^7 = 128
```
mimoo, line 69: I'm wondering if we really have to pass that as argument. In the same vein we generate `endo_q` and `endo_r`, couldn't we generate this automatically in `SRS::create`?

kimchi/src/prover_index.rs
```diff
@@ -93,6 +94,19 @@ where
             fq_sponge_params,
         }
     }
+
+    /// This function verifies the consistency of the wire
+    /// assignements (witness) against the constraints
+    ///     witness: wire assignement witness
+    ///     RETURN: verification status
+    pub fn verify(
```
mimoo, line 102: maybe `verify_witness`?

kimchi/src/verifier_index.rs
```diff
@@ -238,7 +238,7 @@ where
             endo: self.cs.endo,
             lookup_index,
             linearization: self.linearization.clone(),
-            fr_sponge_params: self.cs.fr_sponge_params.clone(),
+            fr_sponge_params: self.srs.scalar_sponge_params.clone(),
             fq_sponge_params: self.fq_sponge_params.clone(),
```
mimoo, line 242: I guess `fq_sponge_params` could follow the same treatment right? The SRS could contain both parameters

kimchi/src/verifier_index.rs
```diff
@@ -238,7 +238,7 @@ where
             endo: self.cs.endo,
             lookup_index,
             linearization: self.linearization.clone(),
-            fr_sponge_params: self.cs.fr_sponge_params.clone(),
+            fr_sponge_params: self.srs.scalar_sponge_params.clone(),
```
mimoo, line 241: if this is present in the SRS, then we can remove it from the verifier index unless I'm missing something

kimchi/src/prover_index.rs
fabrizio-m: Sounds good.

kimchi/src/verifier_index.rs
fabrizio-m: The problem with that is VerifierIndex::srs() that returns the srs or tries to create it, in that case VerifierIndex won't be able to provide the params to SRS::create(). VerifierIndex::srs() could take the params but I think that will just be pushing the problem.
But my proposal about using a trait should solve this too.

fabrizio-m:
Closing in favor of the more general #639 that will address what this tried to solve in a future PR.

PR #612, opened 2022-06-10
Use kimchi/README.md in the kimchi crate rustdoc

yorickvP:
As part of MinaProtocol/mina#11212, I'm moving some rust deps from a git submodule to a cargo git dep. However, the `kimchi` crate fails to build, since it can't find ../../README.md in this case.

I've moved the example into the kimchi crate README.md and used that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,58 +9,6 @@ You can read more about this project on the [Kimchi book](https://o1-labs.github
 
 [See here for the rust documentation](https://o1-labs.github.io/proof-systems/rustdoc).
 
-## Example
-
-We assume that you already have:
-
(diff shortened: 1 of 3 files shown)


mimoo:
https://github.com/o1-labs/proof-systems/pull/555 should fix your issue I believe. Also, we should really create releases and publish them on crates.io. I think we'll investigate that next.

yorickvP:
#555 will fix the kimchi crate, but break the circuit-construction crate in exactly the same way. Publishing it on crates.io would be ideal, but it would also be nice to have a fix for the interim.

yorickvP:
This works, thanks!

PR #620, opened 2022-06-13
64-bit range check support

### LINKED ISSUE #592: range check gate - relocate lookups to MSB

jspada:
This adds 64-bit range check support.  This can be done with a single `RangeCheck0` circuit gate / row by copying the cells in columns 1 and 2 to zero, i.e. by adding two copy constraints.

Closes #592


mrmr1993 (approved):
Nice!

PR #622, opened 2022-06-13
refactor the circuit construction module into smaller modules

katat:
Here is an attempt to refactor the circuit construction module into smaller modules, and expose a prologue so as to simplify importing modules downstream.

- refactor the circuit-construction/src/lib.rs into constants.rs, prover.rs and writer.rs
- now the lib.rs only defines the modules and a prologue


mimoo:
This is awesome! But one problem is that others (like @rot256 ) are currently modifying this file and so they might have some trouble with merges later. Would it be possible to make the first commit actually just be a `git mv lib.rs writer.rs`? This way merge/rebase should understand that more easily (and this will make this PR easier to review as well, if each commits building on top of that is moving only a single part of the file)

mimoo (approved):
ok actually, let's bite the bullet, this is to useful to wait on :) thanks again!

(reviewed the change, no alteration of text detected!)

katat:
> Would it be possible to make the first commit actually just be a git mv lib.rs writer.rs? This way merge/rebase should understand that more easily (and this will make this PR easier to review as well, if each commits building on top of that is moving only a single part of the file)

Ah, I will try this in the next chance. Thanks for the tip!

PR #625, opened 2022-06-14
[easy] fix admonish error

querolita:
Fix admonish error after merging


mimoo (approved):
thanks for the fix :)

PR #627, opened 2022-06-15
installing mdbook-admonish again

querolita:
I have updated mdbook-admonish to version 1.7.0 which changes book.toml to 2.0.0. I think this should solve the error.


jspada (approved):
lgtm!

PR #629, opened 2022-06-16
[easy] clean up kombucha

mimoo:


circuit-construction/src/writer.rs
```diff
@@ -689,14 +691,14 @@ impl<F: FftField + PrimeField> Cs<F> for WitnessGenerator<F> {
     }
 }
 
-impl<F: FftField> WitnessGenerator<F> {
+impl<F: PrimeField> WitnessGenerator<F> {
```
querolita, line 694: Why this change?

querolita: mmm ok!

PR #630, opened 2022-06-16
[kombucha] allows None as well as shorter rows / wires in a circuit

mimoo:


circuit-construction/src/writer.rs
```diff
@@ -26,10 +26,26 @@ pub struct ShiftedScalar<F>(Var<F>);
 
 pub struct GateSpec<F> {
     pub typ: GateType,
-    pub row: [Var<F>; COLUMNS],
+    pub row: Vec<Option<Var<F>>>,
     pub coeffs: Vec<F>,
 }
 
+impl<F: FftField> GateSpec<F> {
+    pub fn get_var_val_or(&self, col: usize, default: F) -> F {
```
querolita, line 34: Just curious. In what cases you would like to set a default value that is different from zero?

querolita (approved):
Well done!

circuit-construction/src/writer.rs
mimoo: not sure :D maybe this could be changed to only add zero...

PR #631, opened 2022-06-16
[kombucha] add support for double generic gate

mimoo:


circuit-construction/src/writer.rs
```diff
@@ -122,50 +126,62 @@ pub trait Cs<F: PrimeField> {
...
 
+    /// Stores a generic gate until it can combine two of them
+    /// into a double generic gate.
+    fn generic_queue(&mut self, gate: GateSpec<F>) -> Option<GateSpec<F>>;
+
+    /// Adds a generic gate.
+    ///
+    /// Warning: this assumes that some finalization occurs to flush
+    /// any queued generic gate.
+    fn generic(&mut self, coeffs: [F; GENERIC_COEFFS], vars: [Option<Var<F>>; GENERIC_REGISTERS]) {
```
querolita, line 161: Why do you want to be explicit about the size of the array here instead of having vectors directly? When you create the GateSpec you convert to vector in any case.

circuit-construction/src/writer.rs
```diff
@@ -122,50 +126,62 @@ pub trait Cs<F: PrimeField> {
...
+        }
+    }
+
     // TODO
     fn scale(&mut self, x: F, v: Var<F>) -> Var<F> {
         let xv = self.var(|| v.val() * x);
-        let row = vec![Some(v), Some(xv)];
 
-        let mut coeffs = vec![F::zero(); DOUBLE_GENERIC_COEFFS];
+        let vars = [Some(v), Some(xv), None];
```
querolita, line 177: If it was a vector already you could leave the `vec![Some(v), Some(xv)]` as before

querolita (approved):
Very cool functionality

circuit-construction/src/writer.rs
mimoo: because they need to be of the right size! If I would pass a vector at this point, I'd need to assert that they're the right size

querolita: Well yes, I was assuming you would include an assert for the size in that case. Anyway, just a comment.

PR #633, opened 2022-06-16
[kombucha] implement cached constants

mimoo:
Cache constants the same way the OCaml code does.

When writing this, and the double generic gate, it feels like I'm writing redundant or unnecessary code sometimes. It makes me think that if `Cs` was implemented on a single struct, combining both `System` and `WitnessGenerator`, it would be leaner. IIUC this is what the OCaml code is doing, but this is also what arkworks r1cs std is doing

### DIFF
--- a/circuit-construction/src/writer.rs
+++ b/circuit-construction/src/writer.rs
@@ -1,4 +1,4 @@
-use ark_ff::{BigInteger, FftField, PrimeField, Zero};
+use ark_ff::{BigInteger, FftField, PrimeField};
 use array_init::array_init;
 use kimchi::circuits::{
     gate::{CircuitGate, GateType},
--- a/circuit-construction/src/writer.rs
+++ b/circuit-construction/src/writer.rs
@@ -1,4 +1,4 @@
-use ark_ff::{BigInteger, FftField, PrimeField};
+use ark_ff::{BigInteger, FftField, PrimeField, Zero};
 use array_init::array_init;
 use kimchi::circuits::{
     gate::{CircuitGate, GateType},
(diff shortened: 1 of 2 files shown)


circuit-construction/src/writer.rs
```diff
@@ -789,6 +826,17 @@ impl<F: PrimeField> Cs<F> for System<F> {
             None
         }
     }
+
+    fn cached_constants(&mut self, x: F) -> Var<F> {
+        match self.cached_constants.get(&x) {
```
querolita, line 831: What is the desired behaviour? Could you have different variables with coincidentally the same value?

circuit-construction/src/writer.rs
```diff
@@ -744,6 +770,17 @@ impl<F: PrimeField> Cs<F> for WitnessGenerator<F> {
             None
         }
     }
+
+    fn cached_constants(&mut self, x: F) -> Var<F> {
```
querolita, line 774: This method seems to behave exactly the same both for WitnessGenerator and for System. Perhaps you can move it to the trait itself to avoid duplication.

querolita (approved):
I am approving these changes, but please take a look at the comments.

circuit-construction/src/writer.rs
mimoo: The problem is that I can't access `self.cached_constants` because the trait doesn't know any of the fields of `Self`.

That's why I think we should move to a single struct instead of two, this is really redundant code

circuit-construction/src/writer.rs
mimoo: so when I want to create a variable that will hold a constant (like 3), I first check if I created such a variable in the past, and if I did I just return that instead. This way, the wiring will take care of ensuring that all of them are the same value.

PR #651, opened 2022-06-21
WIP: Snarky intf

mrmr1993:


snarky/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    where
+        CS: SnarkyConstraintSystem<F, Var = Self::Var, Constraint = Self::Constraint>;
+}
+
+pub enum HandleResponse<A, PrevRequest> {
+    Respond(A),
+    Via(PrevRequest),
+}
+
+pub trait SnarkyConstraintSystem<F> {
```
mimoo, line 34: that looks like the one we have in circuit-writer. Except I'm not sure what Request/Response is, or why it's generic over constraint/var

PR #652, opened 2022-06-21
[easy][CI] cache bench

mimoo:


mimoo:
the bench in CI is really not that useful considering the lack of accuracy... I might just kill it for now

PR #653, opened 2022-06-22
[kombucha] docs

querolita:
I have written a preliminary version of the docs of kombucha. In a future iteration, we may want to move some of these comments to the specs, if we wanted to have specs for it as well.


circuit-construction/src/writer.rs
```diff
@@ -531,12 +617,16 @@ pub trait Cs<F: FftField + PrimeField> {
...
     fn assert_pack(&mut self, zero: Var<F>, x: Var<F>, bits_lsb: &[Var<F>]) {
         let crumbs_per_row = 8;
         let bits_per_row = 2 * crumbs_per_row;
         assert_eq!(bits_lsb.len() % bits_per_row, 0);
         let num_rows = bits_lsb.len() / bits_per_row;
 
+        // Reverse string of bits to have MSB first in the vector
         let bits_msb: Vec<_> = bits_lsb.iter().rev().collect();
 
         let mut a = self.var(|| F::from(2u64));
```
querolita, line 632: Here, I don't see what `a` and `b` are being used for. They seem to be accumulating partially the crumb value, but they are never checked.

circuit-construction/src/writer.rs
```diff
@@ -577,9 +672,11 @@ pub trait Cs<F: FftField + PrimeField> {
                     }
                 });
 
+                // Accumulated chunk value
                 n = self.var(|| n.val().double().double() + crumb.val());
             }
 
+            // In final row, this is the input value, otherwise the accumulated value
             row[1] = if i == num_rows - 1 { x } else { n };
```
querolita, line 680: Specially here, I think we should have an additional assert_eq between the intermediate values, and eventually, obtain an equality between the input value `x` and the final computed `n`.

circuit-construction/src/writer.rs
```diff
@@ -546,17 +636,22 @@ pub trait Cs<F: FftField + PrimeField> {
...
+                // Remember the MSB of each crumb is in the 0 index
+                let b0 = crumb_bits[1]; // less valued
+                let b1 = crumb_bits[0]; // more valued
 
+                // Value of the 2-bit crumb in MSB
                 let crumb = self.var(|| b0.val() + b1.val().double());
+                // Stores the 8 of them in positions [6..13] of the row
                 row[6 + j] = crumb;
 
                 a = self.var(|| {
```
querolita, line 657: Here, I don't see why this is useful. There are no 4 different branches for the 4 different values of the crumb, but instead just 3. And again, we don't use `a` nor `b` later, so this code seems to have no effects later.

mimoo (approved):

PR #655, opened 2022-06-22
[circuit-construction] deprecating the implementation

mimoo:
We're going to sunset this implementation and go with another approach. Sorry for anyone who has contributed to this. 

cc @katat


mimoo:
postponing this so @rot256 is not slowed down by the change of API

PR #663, opened 2022-06-27
[lookup][runtime] specify

mimoo:
specify some parts of the runtime tables

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -282,10 +282,10 @@ where
 
         //~ 1. If using lookup:
         if let Some(lcs) = &index.cs.lookup_constraint_system {
-            // if using runtime table
+            //~~ - if using runtime table:
             if let Some(cfg_runtime_tables) = &lcs.runtime_tables {
-                // check that all the provided runtime tables have length and IDs that match the runtime table configuration of the index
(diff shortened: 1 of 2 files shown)


kimchi/src/prover.rs
```diff
@@ -1135,22 +1135,22 @@ where
...
+            //~~ - add the lookup sorted polynomials
             let sorted_poly = lookup_context.sorted_coeffs.as_ref().unwrap();
             let sorted_comms = lookup_context.sorted_comms.as_ref().unwrap();
 
             for (poly, comm) in sorted_poly.iter().zip(sorted_comms) {
                 polynomials.push((poly, None, comm.blinders.clone()));
             }
 
-            // add the aggreg polynomial
+            //~~ - add the lookup aggreg polynomial
```
fabrizio-m, line 1148: Maybe it would be better as "aggregation" now that it will be in the spec, if it isn't mention as "aggreg" in other part of the spec it feels as an implementation detail.

fabrizio-m (approved):
I left a comment, but shouldn't be a problem to merge.

kimchi/src/prover.rs
mimoo: yeah I think there's a lot more we need to do to make the lookup spec look better :o just wanted to get that in before I forget, thanks!

PR #666, opened 2022-06-30
Minor API changes to support mina-signer-wasm

hanabi1224:
This PR is for supporting [mina-signer-wasm](https://github.com/ChainSafe/mina-signer-wasm)

Changes:
- Add APIs to `Keypair` to support `derivePublicKey`, `verifyKeypair` and `publicKeyToRaw` APIs in `mina-signer`
- Update `Schnorr` API in `signer` crate to support pooling, we see ~10x performance boost by reusing its inner hasher in our [benchmark](https://github.com/ChainSafe/mina-signer-wasm#benchmark-result-from-ci-job)


signer/src/seckey.rs
```diff
@@ -25,4 +25,9 @@ impl SecKey {
     pub fn into_scalar(self) -> ScalarField {
         self.0
     }
+
+    /// Borrows the scalar field element of the secret key
+    pub fn scalar(&self) -> &ScalarField {
+        &self.0
```
mimoo, line 31: why are both into_scalar and scalar a thing btw? Usually you want to prevent people from accessing the secret key directly

signer/src/keypair.rs
```diff
@@ -73,10 +69,38 @@ impl Keypair {
...
+        &public == self.public.point()
+    }
+
+    /// Borrows the secret
+    pub fn secret(&self) -> &SecKey {
+        &self.secret
+    }
+
+    /// Derive public key curve point from the secret
+    fn derive_pubkey_curve_point(secret: ScalarField) -> CurvePoint {
```
mimoo, line 99: we already have `self.public` no?

signer/src/keypair.rs
```diff
@@ -73,10 +69,38 @@ impl Keypair {
...
+        Ok(Keypair::from_parts_unsafe(secret, public))
+    }
+
     /// Obtain the Mina address corresponding to the keypair's public key
     pub fn get_address(self) -> String {
         self.public.into_address()
     }
+
+    /// Validates the keypair
+    pub fn validate(&self) -> bool {
```
mimoo, line 88: it sounds like this check should happen automatically, not with a function that can be forgotten

mimoo (commented):
I find a lot of these functions a bit too low-level for a safe API, but I guess this might be necessary to expose things to the wasm side? Otherwise it LGTM, but I'll let @jspada give the final stamp since he knows this place better o.o

signer/src/seckey.rs
hanabi1224: I think a function that returns a borrow is sufficient enough, caller can do clone if it needs a owned value.  into_scalar requires an owned self which the caller may or may not have, but it forces the caller to clone when the caller does not own self and only needs a borrowed scalar value, which is not ideal performance wise

hanabi1224: In mina-signer, genkeys gives string representation of the secret key, so it requires reading the secret from keypair struct

signer/src/keypair.rs
hanabi1224: in mina-signer, one can construct an arbitrary keypair and validate that later, so it does require a separate API

e.g.

```javascript
const client = new Client({network: 'mainnet'})
const keypair = {privateKey: 'bla', publicKey: 'bla'}
expect(client.verifyKeypair(keypair)).toBe(true)
```

signer/src/seckey.rs
hanabi1224: That's the javascript SDK for mina  https://www.npmjs.com/package/mina-signer

signer/src/keypair.rs
hanabi1224: It doable, but the goal is to create a wasm version of [mina-signer](https://www.npmjs.com/package/mina-signer) as a drop-in replacement, all APIs have to be compatible

signer/src/schnorr.rs
```diff
@@ -83,23 +84,41 @@ impl<H: 'static + Hashable> Signer<H> for Schnorr<H> {
...
+    /// Creates a schnorr instance
+    pub fn new(hasher: T, domain_param: H::D) -> Self {
+        Self {
+            hasher,
+            domain_param,
+        }
+    }
+
+    /// Initiates inner hasher with the given domain param
+    pub fn init_domain_param(&mut self, domain_param: H::D) {
```
hanabi1224, line 117: This API is for supporting pooling, I believe it will be no longer relevant if #675  and #677  are accepted

jspada (changes requested):
I agree with David that adding low-level unsafe APIs like this is not desirable.  Indeed, it's against one of the main design goals for this module, which is safety.

On the other hand, it is understandable that the SDK may require some low-level unsafe functionality.  For this, we already have the `from_parts_unsafe` interface. Rather than add interfaces to adjust this API around unsafe usage, we prefer you to perform verification externally and then call `from_parts_unsafe`, which is fine in this limited case.

PR #668, opened 2022-07-01
Update FieldHelper API to not claim ownership

hanabi1224:
This PR updates the `FieldHelper` API to not claim ownership of `self` when not necessary so that the caller won't have to either give up the ownership or make unnecessary `clone`(s) 

It also tries to remove `Copy` derives in `signer` and `hasher` crates. I think in general it's not a good idea to derive `Copy` trait because it simply disables the powerful borrow checker and lifetime checker in the Rust compiler, and the combination of deriving `Copy` trait and overclaim of the ownership leads to unnecessary implicit copies which the rust compiler should have been capable of detecting.  I will tag some examples inline.

### DIFF
--- a/kimchi/src/verifier.rs
+++ b/kimchi/src/verifier.rs
@@ -166,7 +166,8 @@ where
             //~~ - Derive the scalar joint combiner challenge $j$ from $j'$ using the endomorphism.
             //~~   (TODO: specify endomorphism)
             let joint_combiner = ScalarChallenge(joint_combiner);
-            let joint_combiner = (joint_combiner, joint_combiner.to_field(&index.srs().endo_r));
+            let joint_combiner_field = joint_combiner.to_field(&index.srs().endo_r);
+            let joint_combiner = (joint_combiner, joint_combiner_field);
 
             //~~ - absorb the commitments to the sorted polynomials.
             for com in &lookup_commits.sorted {
@@ -367,7 +368,7 @@ where
                 alpha,
                 beta,
                 gamma,
-                joint_combiner: joint_combiner.map(|j| j.1),
+                joint_combiner: joint_combiner.clone().map(|j| j.1),
                 endo_coefficient: index.endo,
                 mds: index.fr_sponge_params.mds.clone(),
             };
@@ -575,7 +576,7 @@ where
                 alpha: oracles.alpha,
                 beta: oracles.beta,
                 gamma: oracles.gamma,
-                joint_combiner: oracles.joint_combiner.map(|j| j.1),
+                joint_combiner: oracles.joint_combiner.as_ref().map(|j| j.1),
                 endo_coefficient: index.endo,
                 mds: index.fr_sponge_params.mds.clone(),
             };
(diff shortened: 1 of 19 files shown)


kimchi/src/verifier.rs
```diff
@@ -166,7 +166,8 @@ where
             //~~ - Derive the scalar joint combiner challenge $j$ from $j'$ using the endomorphism.
             //~~   (TODO: specify endomorphism)
             let joint_combiner = ScalarChallenge(joint_combiner);
-            let joint_combiner = (joint_combiner, joint_combiner.to_field(&index.srs().endo_r));
```
hanabi1224, line 169: joint_combiner is implicitly copied at pos 0 before when `ScalarChallenge` derives `Copy`

hanabi1224: And now it transfers the ownership instead, no copies occur

signer/src/schnorr.rs
```diff
@@ -107,10 +111,10 @@ impl<H: 'static + Hashable> Schnorr<H> {
         let mut blake_hasher = Blake2bVar::new(32).unwrap();
 
         let mut roi: ROInput = input.to_roinput();
-        roi.append_field(kp.public.into_point().x);
```
hanabi1224, line 110: `kp.public` is implicitly copied twice before, now only `x` and `y` are implicitly once respectively

signer/src/schnorr.rs
```diff
@@ -135,8 +139,8 @@ impl<H: 'static + Hashable> Schnorr<H> {
     fn message_hash(&mut self, pub_key: &PubKey, rx: BaseField, input: &H) -> ScalarField {
         let schnorr_input = Message::<H> {
             input: input.clone(),
-            pub_key_x: pub_key.into_point().x,
```
hanabi1224, line 138: kp.public is implicitly copied twice before, now only x and y are implicitly copied once respectively

utils/src/field_helpers.rs
```diff
@@ -152,7 +152,7 @@ mod tests {
         let field_hex = "f2eee8d8f6e5fb182c610cae6c5393fce69dc4d900e7b4923b074e54ad00fb36";
         assert_eq!(
             BaseField::to_hex(
-                BaseField::from_hex(field_hex).expect("Failed to deserialize field hex")
+                &BaseField::from_hex(field_hex).expect("Failed to deserialize field hex")
```
hanabi1224, line 155: Now users don't need to give up the ownership of BaseField so that it can be reused for other purposes without cloning which are common in business logic

signer/src/keypair.rs
```diff
@@ -47,12 +47,13 @@ impl Keypair {
     /// Generate a random keypair
     pub fn rand(rng: &mut (impl RngCore + CryptoRng)) -> Self {
         let sec_key: SecKey = SecKey::rand(rng);
+        let scalar = sec_key.into_scalar();
```
mimoo, line 50: this could be a one-liner if merged with the line above

mimoo (approved):
In general I still think it's fine to derive `Copy` as it makes the code more readable and brings negligible overhead

hanabi1224:
Thanks, that makes sense!  My only concern is that, if the public types in a library derives `Copy`, it forces disabling compiler borrow checker for its users unless they make a newish type

PR #669, opened 2022-07-01
Fix cargo clippy warnings in rust 1.62.0

hanabi1224:
This PR fixes the Cargo Clippy warnings detected by rust 1.62.0 and it's compatible with current fixed rust 1.58.0.


mimoo (approved):
nice!

PR #675, opened 2022-07-06
Improve performance of creating hasher and signer

hanabi1224:
Currently, it's quite expensive to create a hasher or a signer so it's recommended to use pooling in user's code.  After some benchmarking, I figured the time cost is mostly from `fp_{legacy/kimchi}::params`, this PR proposes a way of speeding it up to make it fairly cheap to create a hasher or a signer.

>/// Interface for hashing [`Hashable`] inputs
///
/// Mina uses a unique hasher configured with domain separation for each type of object hashed.
/// The underlying hash parameters are large and costly to initialize, so the [`Hasher`] interface
/// provides a reusable context for efficient hashing with domain separation.

> /// Poseidon hasher context
//
//  The arithmetic sponge parameters are large and costly to initialize,
//  so we only want to do this once and then re-use the Poseidon context
//  for many hashes. Also, following approach of the mina code we store
//  a backup of the initialized sponge state for efficient reuse.

From below benchmark result, we can see time cost of `create_legacy` is almost equivalent to that of `fp_legacy::params()`, and with this change, the time cost is reduced by ~99.64% which means pooling is no longer needed and we can remove the `sponge_state` field, `state` field and `reset` method from `Poseidon` and `Hasher` trait
```
legacy/fp_legacy::params()                                                                            
                        time:   [815.50 us 840.61 us 870.33 us]
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) high mild
  8 (8.00%) high severe
legacy/fp_legacy_params()                                                                             
                        time:   [2.9636 us 2.9934 us 3.0315 us]
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
legacy/create_legacy()  time:   [3.1849 us 3.1949 us 3.2063 us]                                    

kimchi/fp_kimchi::params()                                                                            
                        time:   [445.45 us 448.63 us 452.90 us]
Found 10 outliers among 100 measurements (10.00%)
  10 (10.00%) high severe
kimchi/fp_kimchi_params()                                                                             
                        time:   [1.6615 us 1.7135 us 1.7805 us]
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) high mild
  6 (6.00%) high severe
kimchi/create_kimchi()  time:   [1.9265 us 1.9483 us 1.9739 us]                                    
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high severe
```


oracle/src/pasta/mod.rs
```diff
@@ -2,3 +2,27 @@ pub mod fp_kimchi;
 pub mod fp_legacy;
 pub mod fq_kimchi;
 pub mod fq_legacy;
+
+use crate::poseidon::ArithmeticSpongeParams;
+use mina_curves::pasta::{Fp, Fq};
+use once_cell::sync::OnceCell;
+
+pub fn fp_kimchi_params() -> ArithmeticSpongeParams<Fp> {
```
mrmr1993, line 10: In general, I'm not a fan of bloating the library with memory leaks like these. IMO, it would be better to separate out the parameters from the constructor, and have a cachable `PoseidonConfig` that can be passed by reference to `Poseidon::new`, instead of the `new_legacy` and `new_kimchi` functions. Then, you can choose to cache the parameters -- or not -- in your use of the library.

It looks like you've done the reference side of that in #677 already, so it may make sense to start with that as the base for these changes.

mrmr1993 (changes requested):
Nice!

oracle/src/pasta/mod.rs
hanabi1224: I think the same improvement is done separately by [another commit](https://github.com/o1-labs/proof-systems/commit/cb9a8877d4def4d8f22630a06d217c02a5c6c0fd#diff-857e792c5b38117d759c3d896c29ab7f3c2ffe06fd388b9e08f4cdb12ac46ae2R50) so I will close my 2 PRs.  

btw, the benchmark in the other PR https://github.com/o1-labs/proof-systems/pull/677 should still be valid and I think the `reset` functionality now creates non-trivial overhead, it's probably better to just remove the `reset`

PR #672, opened 2022-07-05
Absorb the previous challenge evaluations before computing the IPA

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/671, modifying the prover and verifier to absorb the previous challenge evaluations before sampling the parameters used in the inner-product argument.

This is part of the work to ensure that Kimchi is sound when not used with the pickles library.

PR #676, opened 2022-07-06
Absorb commitments from the verifier index

mrmr1993:
This PR builds upon #672 to absorb the verifier index commitments at the start of proving and verification. This also adds an API to cache the verifier index hash inside the prover index, so that we can avoid re-computing it each time.

This PR currently only introduces the commitments that are exposed to pickles, in an order that matches the order there. I intend to add the remaining commitments in a follow-up PR.


oracle/src/sponge.rs
```diff
@@ -156,6 +156,10 @@ where
         }
     }
 
+    fn absorb_fq(&mut self, x: &[P::BaseField]) {
+        self.sponge.absorb(x)
```
mimoo, line 160: I don't remember too much since I wrote this a while ago, but in my version (simply called `absorb`) since it's implemented on the FqSponge, it resets `last_squeezed` first: https://github.com/o1-labs/proof-systems/pull/609/files#diff-aa0e65214a7b00c016dc4f8f67fdeae2dbfd7b77e2c18d01e74994f92fbe3efdR145

oracle/src/sponge.rs
```diff
@@ -199,6 +203,10 @@ where
         P::ScalarField::from_repr(x.into()).unwrap_or_else(P::ScalarField::zero)
     }
 
+    fn digest_fq(mut self) -> P::BaseField {
```
mimoo, line 206: just `digest` since it's implemented for SpongeFq

oracle/src/lib.rs
```diff
@@ -12,9 +12,11 @@ use ark_ff::Field;
 pub trait FqSponge<Fq: Field, G, Fr> {
     fn new(p: poseidon::ArithmeticSpongeParams<Fq>) -> Self;
     fn absorb_g(&mut self, g: &[G]);
+    fn absorb_fq(&mut self, x: &[Fq]);
     fn absorb_fr(&mut self, x: &[Fr]);
     fn challenge(&mut self) -> Fr;
     fn challenge_fq(&mut self) -> Fq;
 
     fn digest(self) -> Fr;
+    fn digest_fq(self) -> Fq;
```
mimoo, line 21: we should document all these functions btw

kimchi/src/verifier_index.rs
```diff
@@ -325,4 +329,25 @@ where
...
+        for comm in self.coefficients_comm.iter() {
+            fq_sponge.absorb_g(&comm.unshifted)
+        }
+        fq_sponge.absorb_g(&self.generic_comm.unshifted);
+        fq_sponge.absorb_g(&self.psm_comm.unshifted);
+        fq_sponge.absorb_g(&self.complete_add_comm.unshifted);
+        fq_sponge.absorb_g(&self.mul_comm.unshifted);
+        fq_sponge.absorb_g(&self.emul_comm.unshifted);
+        fq_sponge.absorb_g(&self.endomul_scalar_comm.unshifted);
+        fq_sponge.digest_fq()
```
mimoo, line 351: I think this function should be called absorb, and you're missing range_check, lookup, chacha, etc. see https://github.com/o1-labs/proof-systems/pull/609/files#diff-c89c45dc84ecd1f409b94d69ca374ed88678982f04f419b79435fd5ebff85b4fR264

mimoo: and you don't need to return the digest IIUC

mimoo: I think I understand what you're doing, you're adding more rows in the verifier circuit but optimizing the absorption of the verifier index by caching its digest. IMO we can have both world by absorbing the verifier index first, and caching the state of the sponge at this point.

kimchi/src/prover_index.rs
```diff
@@ -91,6 +100,45 @@ where
...
+    >(
+        &mut self,
+    ) -> G::BaseField {
+        if let Some(verifier_index_digest) = self.verifier_index_digest {
+            return verifier_index_digest;
+        }
+
+        if let None = &self.verifier_index {
+            self.verifier_index = Some(self.verifier_index());
+        }
```
mimoo, line 121: why not just do `let verifier_index = self.verifier_index()`?

kimchi/src/prover_index.rs
```diff
@@ -91,6 +100,45 @@ where
...
+        if let Some(verifier_index_digest) = self.verifier_index_digest {
+            return verifier_index_digest;
+        }
+
+        match &self.verifier_index {
+            None => {
+                let verifier_index = self.verifier_index();
+                verifier_index.digest::<EFqSponge>()
+            }
+            Some(verifier_index) => verifier_index.digest::<EFqSponge>(),
```
mimoo, line 141: same here

```rust
let verifier_index = self.verifier_index();
verifier_index.digest()
```

kimchi/src/verifier_index.rs
```diff
@@ -146,6 +146,10 @@ where
 {
     /// Produces the [VerifierIndex] from the prover's [ProverIndex].
     pub fn verifier_index(&self) -> VerifierIndex<G> {
+        if let Some(verifier_index) = &self.verifier_index {
+            return verifier_index.clone();
+        }
```
mimoo, line 151: at the end of this function, cache the verifier_index into self.verifier_index before returning it

kimchi/src/prover_index.rs
```diff
@@ -91,6 +100,45 @@ where
...
             max_quot_size,
             fq_sponge_params,
+            verifier_index: None,
+            verifier_index_digest: None,
+        }
+    }
+
+    /// Retrieve or compute the digest for the corresponding verifier index.
+    /// If the digest is not already cached inside the index, store it.
+    pub fn compute_verifier_index_digest<
```
mimoo, line 110: this function is unused

mimoo: looks like it's used on the ocaml side

kimchi/src/prover_index.rs
```diff
@@ -91,6 +100,45 @@ where
...
+            self.verifier_index = Some(self.verifier_index());
+        }
+
+        let verifier_index_digest = self.verifier_index_digest::<EFqSponge>();
+        self.verifier_index_digest = Some(verifier_index_digest);
+        verifier_index_digest
+    }
+
+    /// Retrieve or compute the digest for the corresponding verifier index.
+    pub fn verifier_index_digest<EFqSponge: Clone + FqSponge<G::BaseField, G, G::ScalarField>>(
```
mimoo, line 129: this function should fill the cache if it's not already filled. This way it can be merged with `compute_verifier_index_digest`

mimoo (changes requested):
a number of comments, but the most important is that it's missing some commitments

kimchi/src/prover_index.rs
mrmr1993: Because then we don't get caching. This code is designed to only compute the `verifier_index` term when it's not already present.

kimchi/src/prover_index.rs
mrmr1993: As above, so below.

kimchi/src/verifier_index.rs
mrmr1993: We don't have a `&mut` reference.

kimchi/src/prover_index.rs
mrmr1993: We don't always have a mutable reference -- in particular, in the kimchi prover, the reference is immutable, and I don't think we want to change it to be mutable. This extra complexity is to deal with that.

kimchi/src/prover_index.rs
mrmr1993: Same as all of the above: we need a version that doesn't mutate so that we don't require a mutable pointer to the index every time we want to create a proof.

oracle/src/sponge.rs
mrmr1993: There's already a `digest`, called it this to avoid clobbering it. See the next review comment.

oracle/src/lib.rs
mrmr1993: Agreed, but not now and not here :)

kimchi/src/verifier_index.rs
mrmr1993: > and you don't need to return the digest if you pass the sponge and absorb it from here, currently this code recreates a sponge on top of the already existing fq sponge :o

As I mentioned (I think on a call), we want to limit the scope in which we have to use the 'optional sponge' in the verifier circuit. We can do this by using a fresh optional sponge here, squeezing it, and then absorbing the commitment into the regular, non-optional sponge.

> I think I understand what you're doing, you're adding more rows in the verifier circuit but optimizing the absorption of the verifier index by caching its digest. IMO we can have both world by absorbing the verifier index first, and caching the state of the sponge at this point.

We could also cache the state of the main sponge, yes. Looks like it's marginally more efficient to do it this way (we only need to store a single field element rather than the sponge state), but I'm fine with doing either.

mrmr1993: > you're missing range_check, lookup, chacha, etc

Added these (and expanded the verifier index explicitly to make sure we don't miss any).

oracle/src/lib.rs
mimoo: In general I'm on the board that if you're adding a new public function, you need to document it

mimoo (approved):
approving, but:

* I insist that we document new functions in the PR that introduces them :)
* I don't agree with the cache handling, I think it leaks an internal detail to the user, I think it'd be much simpler if we just cache it automatically when we create the prover index

kimchi/src/verifier_index.rs
mimoo: > the 'optional sponge'

I guess I don't understand what the optional sponge is to understand why this makes sense. Would you mind writing an explanation for it somewhere?

kimchi/src/verifier_index.rs
```diff
@@ -321,18 +321,111 @@ impl<G: KimchiCurve> VerifierIndex<G> {
         &self,
     ) -> G::BaseField {
         let mut fq_sponge = EFqSponge::new(G::OtherCurve::sponge_params());
-        for comm in self.sigma_comm.iter() {
+        // We fully expand this to make the compiler check that we aren't missing any commitments
```
mimoo, line 324: clever!

kimchi/src/verifier_index.rs
```diff
@@ -321,18 +321,111 @@ impl<G: KimchiCurve> VerifierIndex<G> {
...
+            shift: _,
+            zkpm: _,
+            w: _,
+            endo: _,
+
+            linearization: _,
+            powers_of_alpha: _,
+        } = &self;
+
+        // Always present
```
mimoo, line 359: what does this comment mean?

Issue #221, opened 2021-11-09
[kimchi] absorbing all the values!


mimoo:
in [plonk_sponge.rs](https://github.com/o1-labs/proof-systems/blob/master/dlog/kimchi/src/plonk_sponge.rs#L38), it's too easy to forget to absorb some of the evaluation values. So here we should discuss:

- is this really needed? It's not clear to me, as we're already absorbing the digest of the other sponge and it authenticates these values already IIUC.
- if it is needed, can be make it less error-prone? Perhaps use some generated-code (via macros).

mimoo:
I think there's actually an quick 'n easy way to address this issue atm without resorting to macros! @mrmr1993  used a technique earlier which really inspired me and I implemented it here as part of a larger PR that might never see the light of day: https://github.com/o1-labs/proof-systems/pull/680/files#diff-7d26d412f0cf053bc5fbbc590ab9c3760a3f2bfd0ef7ff640781d7ba127bc905R201

the technique is to deconstruct whatever it is we really want to use all of, and if we don't use one of the things the linter will yell at us (and CI won't let us merge the PR):

```rust
impl<F: Field> ProofEvaluations<F> {
    /// Returns an iterator over all evaluations contained in [Self].
    pub fn iter(&self) -> impl Iterator<Item = F> {
        // important: we use a deconstruction pattern so that the compiler can tell us if we forget one of the evaluations
        let ProofEvaluations {
            w,
            z,
            s,
            lookup,
            generic_selector,
            poseidon_selector,
        } = self.clone();

        let mut values: Vec<F> = chain![
            std::iter::once(z),
            std::iter::once(generic_selector),
            std::iter::once(poseidon_selector),
            w,
            s,
        ]
        .collect();

        // important: at this point we assume that Self is well-formed and contains a lookup
        if let Some(lookup_evals) = lookup {
            // important: we use a deconstruction pattern so that the compiler can tell us if we forget one of the evaluations
            let LookupEvaluations {
                sorted,
                aggreg,
                table,
                runtime,
            } = lookup_evals;

            values.extend(chain![
                std::iter::once(aggreg),
                std::iter::once(table),
                sorted,
                runtime,
            ]);
        }

        values.into_iter()
    }
}

impl<F> AbsorbableScalarField<F> for [&ProofEvaluations<F>; 2]
where
    F: Field,
{
    fn absorb(&self, sponge: &mut impl FrSponge<F>) {
[...]

PR #680, opened 2022-07-07
[proof] converting vectors of F to single F element

mimoo:
This PR removes vectors from several proof structs (e.g. `ProofEvaluations`). The vectors were initially used because some of the evaluations were "chunked", but we don't do this anymore. Proof:

```
[kimchi/src/verifier.rs:483] proof.evals[0].w[0].len() = 1
[kimchi/src/verifier.rs:484] proof.evals[0].z.len() = 1
[kimchi/src/verifier.rs:485] proof.evals[0].s[0].len() = 1
[kimchi/src/verifier.rs:486] proof.evals[0].generic_selector.len() = 0
[kimchi/src/verifier.rs:487] proof.evals[0].poseidon_selector.len() = 0
[kimchi/src/verifier.rs:488] proof.evals[0].[lookup.as](http://lookup.as/)_ref().unwrap().sorted[0].len() = 1
[kimchi/src/verifier.rs:489] proof.evals[0].[lookup.as](http://lookup.as/)_ref().unwrap().aggreg.len() = 1
[kimchi/src/verifier.rs:490] proof.evals[0].[lookup.as](http://lookup.as/)_ref().unwrap().table.len() = 1
[kimchi/src/verifier.rs:491] proof.evals[0].[lookup.as](http://lookup.as/)_ref().unwrap().[runtime.as](http://runtime.as/)_ref().unwrap().len() = 1
```

The changes are pretty straight forward: I simply change things like `Vec<F>` to `F` and remove code that's not relevant anymore. 

There's one thing that seems like a logic change (but is not): I do not include a polynomial in the evaluation proof if its associated commitment in the verifier index is empty. This is because of the following:

> so, ran into an interesting edge case in our codebase, for committing to the zero polynomial we just have an empty commit (literally an empty vec), and when we do the evaluation proof we just omit such zero polynomials from the combined inner product. So for example, if you don't use poseidon in your circuit, the poseidon selector will be all zero and we will not add it to the aggregated evaluation proof. It also means we can't hide such a polynomial commitment.

> It sounds sensible since pedersen commitments work by doing an inner product of a polynomial coeff with the URS' generators, and so we would get the point at infinity (and we can't encode that since we're in affine coordinates). But that also means we have some issue if a private polynomial evaluates to all zeros (I think the codebase doesn't address that edgecase). Would a private polynomial ever evaluate to all zeros? I'm guessing not at least due to the random rows (the probability would be very low).

Note that it would be nice to have two follow up to this PR:

* refactor of the combine inner product function: https://github.com/o1-labs/proof-systems/pull/448
* make `absorb_evaluations` less error-prone by using a macro https://github.com/o1-labs/proof-systems/issues/221


mimoo:
I think I narrowed the issue: 

```
other:
[Fp256(BigInteger256([0, 0, 0, 0]))]
[Fp256(BigInteger256([0, 0, 0, 0]))]
orign:
[]
[]
```

the PCS doesn't like it when I pass [0] instead of []

kimchi/src/proof.rs
```diff
@@ -138,52 +144,16 @@ where
 
 //~ spec:endcode
 
-impl<F> ProofEvaluations<F> {
-    pub fn transpose<const N: usize>(
```
mimoo, line 142: decided to get rid of this as it is highly error prone and doesn't add much. We need to implement a macro to do the job.

kimchi/src/proof.rs
```diff
@@ -1,69 +1,76 @@
...
-    pub runtime: Option<Field>,
+    #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
+    pub runtime: Option<F>,
 }
 
 // TODO: this should really be vectors here, perhaps create another type for chunked evaluations?
 /// Polynomial evaluations contained in a `ProverProof`.
 /// - **Chunked evaluations** `Field` is instantiated with vectors with a length that equals the length of the chunk
-/// - **Non chunked evaluations** `Field` is instantiated with a field, so they are single-sized#[serde_as]
+/// - **Non chunked evaluations** `Field` is instantiated with a field, so they are single-sized
```
querolita, line 43: Shouldn't this be updated to remove chunked evaluations?

kimchi/src/proof.rs
```diff
@@ -107,7 +116,7 @@ pub struct ProverProof<G: AffineCurve> {
 
     /// Two evaluations over a number of committed polynomials
     // TODO(mimoo): that really should be a type Evals { z: PE, zw: PE }
-    pub evals: [ProofEvaluations<Vec<G::ScalarField>>; 2],
+    pub evals: [ProofEvaluations<G::ScalarField>; 2],
```
querolita, line 119: Will it still be relevant to have the `ConsecutiveEvals` type? [PR 607](https://github.com/o1-labs/proof-systems/pull/607), I would have to adapt it to the current master, but it should be easy

kimchi/src/prover.rs
```diff
@@ -855,124 +842,49 @@ where
...
-                w: array_init(|i| {
-                    witness_poly[i]
-                        .to_chunked_polynomial(index.max_poly_size)
-                        .evaluate_chunks(zeta_omega)
-                }),
-
-                z: z_poly
-                    .to_chunked_polynomial(index.max_poly_size)
-                    .evaluate_chunks(zeta_omega),
+            let evals_zeta_omega = ProofEvaluations::<G::ScalarField> {
```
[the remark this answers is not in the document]
querolita: Makes sense, otherwise it is duplicate code, right?

kimchi/src/prover.rs
```diff
@@ -821,39 +821,26 @@ where
         //~ 1. If lookup is used, evaluate the following polynomials at $\zeta$ and $\zeta \omega$:
         if index.cs.lookup_constraint_system.is_some() {
             //~~ - the aggregation polynomial
-            let aggreg = lookup_context
-                .aggreg_coeffs
-                .as_ref()
-                .unwrap()
-                .to_chunked_polynomial(index.max_poly_size);
+            let aggreg = lookup_context.aggreg_coeffs.as_ref().unwrap();
```
querolita, line 824: What is now the flow? Don't we check anymore that the polynomials fit in the max_poly_size?

kimchi/src/verifier.rs
```diff
@@ -734,53 +719,43 @@ where
...
-    evaluations.push(Evaluation {
-        commitment: index.psm_comm.clone(),
-        evaluations: proof
-            .evals
-            .iter()
-            .map(|e| e.poseidon_selector.clone())
-            .collect(),
-        degree_bound: None,
-    });
+    if !index.generic_comm.is_empty() {
```
querolita, line 727: Are you making this check because what you found out a few weeks ago that the PCS was complaining when you were passing [0]?

querolita (approved):
Excellent work! So much cleaner now 😄

kimchi/src/prover.rs
mimoo: I don't think we ever performed that check, but one thing we still need to do is enforce the length of the commitments we receive

kimchi/src/verifier.rs
mimoo: yes! (see what I wrote on slack or in the description of this PR as well, it's a problem with zero commitments)

kimchi/src/prover.rs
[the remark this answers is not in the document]
mimoo: yeah, but also moving some logic out of prover.rs would make the code easier to follow :D

mimoo:
closes https://github.com/o1-labs/proof-systems/issues/231

mrmr1993 (approved):

mimoo:
added an `AbsorbableScalarField` trait in the last commit (for lack of a better name)

mimoo:
sounds like I can't get consensus on this PR. I still believe it's the best route, but at the very least I'll leave this open for when I have time to submit smaller PRs that implement some of the ideas here.

PR #705, opened 2022-07-26
[utils] enforce docs in utils

mimoo:
Since everything is well commented in the `utils` crate, let's enforce docs there while we still can


querolita (approved):
Good!

PR #706, opened 2022-07-26
Range check changes

querolita:
This PR includes the changes to the range check made for foreign field arithmetics.


kimchi/src/circuits/polynomials/range_check/gate.rs
```diff
@@ -1,1400 +0,0 @@
-//! Range check gate
```
mimoo, line 1: it stil shows this file as deleted, and the gadget file as created from scratch. I think you need to do a `git mv`

mimoo: I think it's because you do the `git mv` too late in your commits. You need to have the `git mv` in the first commit

querolita: Never mind, it is named now again as gate.rs, so it will show the diff

mimoo: btw if I were you, to fix the issue I would:

1. create a clean branch where I `git mv gate.rs gadget.rs`
1. then merge this branch into the clean branch (or better, squash all these commits, and then merge into the clean branch)

kimchi/src/circuits/polynomials/range_check/gate.rs
```diff
@@ -113,6 +63,18 @@ impl<F: PrimeField> CircuitGate<F> {
             },
         ];
 
+        // copy v0p0
+        circuit_gates.connect_cell_pair((0, 1), (3, 3));
```
mimoo, line 67: I much prefer this! I think ideally we should really remove the wiring from `CircuitGate` and keep track of it separately, to make it safer to connect cells (by just keeping track of cycles)

kimchi/src/circuits/polynomials/range_check/gate.rs
```diff
@@ -76,21 +38,9 @@ impl<F: PrimeField> CircuitGate<F> {
     ///       next_row      - next row after this gate
     ///       circuit_gates - vector of circuit gates comprising this gate
     pub fn create_multi_range_check(start_row: usize) -> (usize, Vec<Self>) {
-        let mut wires: Vec<GateWires> = (0..4).map(|i| Wire::new(start_row + i)).collect();
+        let wires: Vec<GateWires> = (0..4).map(|i| Wire::new(start_row + i)).collect();
```
mimoo, line 41: can you inline this within each CircuitGate? Instead of creating this here in advance

kimchi/src/circuits/polynomials/range_check/mod.rs
```diff
@@ -4,7 +4,3 @@ mod circuitgates;
 
 pub mod gate;
 pub mod witness;
-
-pub use circuitgates::{RangeCheck0, RangeCheck1};
-pub use gate::*;
-pub use witness::*;
```
mimoo, line 10: nice

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -2,39 +2,40 @@
 
 use ark_ff::PrimeField;
 use array_init::array_init;
-use o1_utils::FieldHelpers;
+//use num_bigint::BigUint;
```
mimoo, line 5: delete

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -2,39 +2,40 @@
...
         WitnessCell::Copy(CopyWitnessCell { row, col })
     }
 }
 
 // Witness cell for a range check field element limb
-struct ValueWitnessCell;
+pub struct ValueWitnessCell;
 impl ValueWitnessCell {
-    const fn create() -> WitnessCell {
+    pub const fn create() -> WitnessCell {
```
mimoo, line 32: doc

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -2,39 +2,40 @@
...
         WitnessCell::Value
     }
 }
 
 // Witness cell for a range check field element sub-limb
```
mimoo, line 37: doc

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -2,39 +2,40 @@
...
-struct LimbWitnessCell {
+pub struct LimbWitnessCell {
```
mimoo, line 38: document the fields

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -43,7 +44,7 @@ struct LimbWitnessCell {
 
 impl LimbWitnessCell {
     // Params: source (row, col), starting bit offset and ending bit offset (exclusive)
```
mimoo, line 46: doc

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -53,10 +54,10 @@ impl LimbWitnessCell {
     }
 }
 
-// An cell containing zero
-struct ZeroWitnessCell;
+// A cell containing zero
```
mimoo, line 57: doc

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -53,10 +54,10 @@ impl LimbWitnessCell {
...
+pub struct ZeroWitnessCell;
 impl ZeroWitnessCell {
-    const fn create() -> WitnessCell {
+    pub const fn create() -> WitnessCell {
```
mimoo, line 60: doc

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -72,7 +73,7 @@ impl ZeroWitnessCell {
 //     For example, we can convert the RangeCheck0 circuit gate into
 //     a 64-bit lookup by adding two copy constraints to constrain
 //     columns 1 and 2 to zero.
```
mimoo, line 75: doc

utils/src/field_helpers.rs
```diff
@@ -95,6 +101,20 @@ impl<F: Field> FieldHelpers<F> for F {
...
+        // Pad with zeros if necessary because the BigUint to_bytes_le function gives the smallest possible vector of bytes
+        bytes.resize(32, 0);
+
+        F::deserialize(&mut &bytes[..]).map_err(|_| FieldHelpersError::DeserializeBytes)
+    }
+
+    fn to_big(self) -> BigUint {
+        let bytes = self.to_bytes();
+        BigUint::from_bytes_le(&bytes)
+    }
```
mimoo, line 117: as I said in the other review, these functions already exists in `PrimeField`, and the `from_big` has a bug (`bytes.resize` will truncate the length if it's > 32)

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -151,35 +152,45 @@ const fn range_check_row(row: usize) -> [WitnessCell; COLUMNS] {
...
-                    witness[limb_cell.col][limb_cell.row], // limb cell (row, col)
-                    limb_cell.start,                       // starting bit
-                    limb_cell.end,                         // ending bit (exclusive)
-                );
-            }
-            WitnessCell::Zero => {
-                witness[col][row] = F::zero();
-            }
-        }
+        handle_standard_witness_cell(witness, &WITNESS_SHAPE[row][col], row, col, value);
```
mimoo, line 188: why this change?

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -192,11 +203,20 @@ pub fn create_multi_witness<F: PrimeField>(v0: F, v1: F, v2: F) -> [Vec<F>; COLU
...
 }
+
+/// Extend an existing witness with a multi-range-check gate for foreign field
+/// elements fe
+pub fn extend_witness<F: PrimeField>(witness: &mut [Vec<F>; COLUMNS], fe: ForeignElement<F, 3>) {
+    let limbs_witness = create_multi_witness(*fe.lo(), *fe.mi(), *fe.hi());
+    for col in 0..COLUMNS {
+        witness[col].extend(limbs_witness[col].iter())
+    }
+}
```
mimoo, line 222: not used anywhere no?

utils/src/foreign_field.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mut bytes = vec![];
+        for limb in self.limbs {
+            bytes.extend_from_slice(&limb.to_bytes());
+        }
+        BigUint::from_bytes_le(&bytes)
+    }
+
+    /// Split a foreign field element into a vector of `LIMB_BITS` bits field elements of type `F` in little-endian.
+    /// Right now it is written so that it gives `LIMB_COUNT` limbs, even if it fits in less bits.
+    fn big_to_vec(fe: BigUint) -> Vec<F> {
```
mimoo, line 73: if we only support elements of size 3, imo we should return `[F; 3]` here

utils/src/foreign_field.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub fn new_from_big(big: BigUint) -> Self {
+        let vec = ForeignElement::<F, N>::big_to_vec(big);
+
+        // create an array of N native elements containing the limbs
+        // until the array is full in big endian, so most significant
+        // limbs may be zero if the big number is smaller
+        let limbs = {
+            let mut ini = [F::zero(); N];
+            for i in 0..N {
+                if i < vec.len() {
```
mimoo, line 40: I pointed that in the other PR, but if vec.len() > N?

kimchi/src/circuits/constraints.rs
```diff
@@ -112,8 +112,12 @@ pub struct ConstraintSystem<F: PrimeField> {
...
 
     /// Range check gate selector polynomials
-    #[serde(bound = "Vec<range_check::SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
-    pub range_check_selector_polys: Vec<range_check::SelectorPolynomial<F>>,
+    #[serde(bound = "Vec<SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
+    pub range_check_selector_polys: Vec<SelectorPolynomial<F>>,
+
+    /// Foreign field modulus
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub foreign_field_modulus: Vec<F>,
```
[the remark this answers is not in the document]
mimoo: (btw I'm fine if this gets in, in this PR, if it makes the rest of the PRs easier! I was wondering when I started reviewing it tho)

kimchi/src/circuits/constraints.rs
```diff
@@ -147,6 +151,48 @@ pub struct Builder<F: PrimeField> {
...
+    .interpolate();
+
+    // Evaluation form (evaluated over d8)
+    let eval8 = coeff.evaluate_over_domain_by_ref(domain.d8);
+
+    SelectorPolynomial { eval8 }
+}
+
+/// Create selector polynomials for a gate (i.e. a collection of circuit gates)
+pub fn selector_polynomials<F: PrimeField>(
```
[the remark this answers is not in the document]
mimoo: and to provide some rationale: the more we factor out logic into other functions, the more the reader has to jump functions to follow some logic. So doing it too much impacts readability, not doing it enough also impacts readability, there's a trade off but I'm thinking that this is probably already too much :D not a big deal tho

kimchi/src/circuits/constraints.rs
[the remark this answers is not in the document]
querolita: Or perhaps store it as a ForeignElement<F, 3> ?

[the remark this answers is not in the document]
querolita: Well perhaps I can add this later in either of FFAdd or FFMul (or both).

kimchi/src/circuits/constraints.rs
```diff
@@ -112,8 +112,12 @@ pub struct ConstraintSystem<F: PrimeField> {
     pub endomul_scalar8: E<F, D<F>>,
 
     /// Range check gate selector polynomials
-    #[serde(bound = "Vec<range_check::SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
-    pub range_check_selector_polys: Vec<range_check::SelectorPolynomial<F>>,
+    #[serde(bound = "Vec<SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
+    pub range_check_selector_polys: Vec<SelectorPolynomial<F>>,
```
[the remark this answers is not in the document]
querolita: Ok I went too fast here. It is a Vec because it can sometimes be empty. Maybe I can switch it to an optional array...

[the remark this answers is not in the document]
querolita: I followed the same approach the chacha params both for the selector polynomials and the commitments

kimchi/src/circuits/constraints.rs
```diff
@@ -112,8 +112,12 @@ pub struct ConstraintSystem<F: PrimeField> {
     pub endomul_scalar8: E<F, D<F>>,
 
     /// Range check gate selector polynomials
-    #[serde(bound = "Vec<range_check::SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
-    pub range_check_selector_polys: Vec<range_check::SelectorPolynomial<F>>,
+    #[serde(bound = "Vec<SelectorPolynomial<F>>: Serialize + DeserializeOwned")]
```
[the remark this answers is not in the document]
querolita: Done

kimchi/src/circuits/constraints.rs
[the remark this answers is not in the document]
querolita: I will keep this in mind, as Matthew was not very keen on creating that type either. Nonetheless, I wonder if I should postpone this change to a later PR, modifying it for both ffmul and ffadd gates.

kimchi/src/circuits/constraints.rs
```diff
@@ -330,6 +377,13 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
         self
     }
 
+    /// Set up the foreign field modulus.
+    /// If not invoked, it is `vec![]` by default.
+    pub fn foreign_field_modulus(mut self, foreign_field_modulus: Vec<F>) -> Self {
+        self.foreign_field_modulus = foreign_field_modulus;
```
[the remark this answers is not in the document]
querolita: In general I have some doubts regarding BigUint (versus BigInt), as I am not sure how to handle negative values (subtractions) in the FFAdd PR. Basically, using BigUint and/or field elements immediately gives my negative values as "positive", and I never get to obtain a negative carry term in the FFAdd PR.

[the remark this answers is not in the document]
querolita: In that sense, I agree with having the modulus of type `ForeignElement<F, 3>` for the moment until we decide on big arithmetics.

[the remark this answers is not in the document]
querolita: And in any case, this should be moved to FFAdd as it is not directly related to Range Check

kimchi/src/circuits/constraints.rs
```diff
@@ -503,9 +557,10 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
 
         // Range check constraint selector polynomials
         let range_check_selector_polys = {
-            if !circuit_gates_used.is_disjoint(&range_check::circuit_gates().into_iter().collect())
+            if !circuit_gates_used
+                .is_disjoint(&range_check::gate::circuit_gates().into_iter().collect())
             {
-                range_check::selector_polynomials(&gates, &domain)
+                range_check::gate::selector_polynomials(&gates, &domain)
```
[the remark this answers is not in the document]
querolita: It is an array now 😃

kimchi/src/circuits/expr.rs
```diff
@@ -2037,6 +2045,7 @@ impl<F: PrimeField> ConstantExpr<F> {
             JointCombiner => "joint\\_combiner".to_string(),
             EndoCoefficient => "endo\\_coefficient".to_string(),
             Mds { row, col } => format!("mds({row}, {col})"),
+            ForeignFieldModulus(i) => format!("foreign_field_modulus({i})"),
```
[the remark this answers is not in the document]
querolita: oh thank you!

kimchi/src/circuits/gate.rs
```diff
@@ -91,11 +94,48 @@ pub enum GateType {
     CairoInstruction = 13,
     CairoFlags = 14,
     CairoTransition = 15,
-    // Range check (16-24)
+    /// Range check (16-24)
```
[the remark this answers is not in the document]
querolita: For some reason we were leaving the numbers between 16 and 24 for range check and foreign arithmetics. Taking into account the latest version of these gates, we are taking from 16 to 19. Matthew wanted to leave that comment as is, so I haven't changed the interval just in case.

kimchi/src/circuits/gate.rs
```diff
@@ -168,6 +208,24 @@ impl<F: PrimeField> CircuitGate<F> {
...
+    fn connect_cell_pair(&mut self, cell1: (usize, usize), cell2: (usize, usize));
+}
+
+impl<F: PrimeField> Connect for Vec<CircuitGate<F>> {
+    /// Connect the pair of cells specified by the cell1 and cell2 parameters
+    /// cell_pre --> cell_new && cell_new --> wire_tmp
+    ///
+    /// Note: This function assumes that the targeted cells are freshly instantiated
+    ///       with self-connections.  If the two cells are transitively already part
+    ///       of the same permutation then this would split it.
```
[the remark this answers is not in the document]
querolita: Oh there it was, thanks for pointing out!

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
```diff
@@ -1,95 +1,100 @@
...
-///     2   RangeCheck1   Fully constrain v2 (and trigger plookups constraints on row 3)
-///     3   Zero          Complete the constraining of v0 and v1 using lookups
-///
-///  Nb. each CircuitGate type corresponds to a unique polynomial and thus
-///       is assigned its own unique powers of alpha
-///```
+//~ Range check circuit gates:
+//~
+//~    The range check gate is comprised of three circuit gates (RangeCheck0, RangeCheck1
+//~    and Zero) and can perform range checks on up to three 88-bit values: v0, v1 and v2.
```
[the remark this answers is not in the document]
querolita: Oh but in theory one could use it to range check on smaller limbs. Perhaps it is better to say on three values of up to 88-bit?

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
```diff
@@ -1,95 +1,100 @@
...
+//~
+//~    Byte-order:
+//~      * Each value is in little-endian byte order
```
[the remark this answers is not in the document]
querolita: Because we want most significant sublimbs on the first few columns of the rows to allow for copies of zeroes to convert the 88-bit checks to 64-bit easily.

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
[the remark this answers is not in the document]
querolita: I think for readability of the specs when reading it from the code. Like: section, and then inside the text.

kimchi/src/circuits/constraints.rs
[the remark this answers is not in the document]
mimoo: > In general I have some doubts regarding BigUint (versus BigInt), as I am not sure how to handle negative values (subtractions) in the FFAdd PR. Basically, using BigUint and/or field elements immediately gives my negative values as "positive", and I never get to obtain a negative carry term in the FFAdd PR.

what I mean here is that the user should pass the value as a `BigUint` and then somewhere you do the conversion to `[F; 3]` or `ForeignElement<F, 3>`. But it feels weird for the user to have to do that conversion themselves no?

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
```diff
@@ -1,95 +1,100 @@
...
+//~    7 | crumb   v0c0 | crumb   v1c0 | crumb   v2c2 | crumb   v2c12
+//~    8 | crumb   v0c1 | crumb   v1c1 | crumb   v2c3 | crumb   v2c13
+//~    9 | crumb   v0c2 | crumb   v1c2 | crumb   v2c4 | crumb   v2c14
+//~   10 | crumb   v0c3 | crumb   v1c3 | crumb   v2c5 | crumb   v2c15
+//~   11 | crumb   v0c4 | crumb   v1c4 | crumb   v2c6 | crumb   v2c16
+//~   12 | crumb   v0c5 | crumb   v1c5 | crumb   v2c7 | crumb   v2c17
+//~   13 | crumb   v0p6 | copy    v1c6 | crumb   v2c8 | crumb   v2c18
+//~   14 | crumb   v0p7 | copy    v1c7 | crumb   v2c9 | crumb   v2c19 <- LSB
+//~
+//~ ```
```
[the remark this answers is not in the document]
querolita: since those were not actual tables, if we parse them as single-spaced text-format they will look "kind of" like a table in the specs

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
```diff
@@ -1,95 +1,100 @@
...
+//~    5 | plookup v0p4 | plookup v1p4 | plookup v2p2 | plookup  v1p0
+//~    6 | plookup v0p5 | plookup v1p5 | plookup v2p3 | plookup  v1p1
+//~    7 | crumb   v0c0 | crumb   v1c0 | crumb   v2c2 | crumb   v2c12
+//~    8 | crumb   v0c1 | crumb   v1c1 | crumb   v2c3 | crumb   v2c13
+//~    9 | crumb   v0c2 | crumb   v1c2 | crumb   v2c4 | crumb   v2c14
+//~   10 | crumb   v0c3 | crumb   v1c3 | crumb   v2c5 | crumb   v2c15
+//~   11 | crumb   v0c4 | crumb   v1c4 | crumb   v2c6 | crumb   v2c16
+//~   12 | crumb   v0c5 | crumb   v1c5 | crumb   v2c7 | crumb   v2c17
+//~   13 | crumb   v0p6 | copy    v1c6 | crumb   v2c8 | crumb   v2c18
+//~   14 | crumb   v0p7 | copy    v1c7 | crumb   v2c9 | crumb   v2c19 <- LSB
```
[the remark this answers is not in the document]
querolita: Yeah perhaps 😅

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
```diff
@@ -1,95 +1,100 @@
...
+//~
+//~ ```
+//~   The 12-bit chunks are constrained with plookups and the 2-bit crumbs
+//~   constrained with degree-4 constraints of the form $x · (x - 1) · (x - 2) · (x - 3)$.
```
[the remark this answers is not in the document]
querolita: Sad... using \cdot was ugly when reading the code, I will write * then.

kimchi/src/circuits/polynomials/range_check/gate.rs
querolita: Thanks! In any case, I agree that there should be something separate for it. But at least this way is simpler, and it reuses the same function for both range check and ffadd (which was not the case initially)

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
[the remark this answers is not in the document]
mimoo: ah yeah true, I guess it's more granular that that: you can enforce values of the limbs of your value. You could say "the last 2 bits must be equal to 3" as well "the higher 24 bits must be equal to 0" (64-bit check).

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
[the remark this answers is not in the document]
mimoo: I can understand for the fake tables, but for the text? This will also prevent the latex stuff to render

kimchi/src/circuits/constraints.rs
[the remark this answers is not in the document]
mimoo: optional array is better yeah

kimchi/src/circuits/gate.rs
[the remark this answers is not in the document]
mimoo: I think it's fine to reserve these ranges, but if we do that then ffmul and ffadd should start at 25 no?

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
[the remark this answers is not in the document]
mimoo: I'd rather we use markdown everywhere :o this is what markdown is for

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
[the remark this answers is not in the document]
mimoo: `*` is still not valid latex :D why not `$x(x-1)(x-2)(x-3)$`

kimchi/src/circuits/gate.rs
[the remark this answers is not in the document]
querolita: Yeah that's what I thought... I was not sure what was going to be the standard here. Perhaps Matthew had some ideas that did not include the FFMul and FFAdd gates. If this is the case, then I will start ffmul and ffadd in 25 just in case. Good catch.

kimchi/src/circuits/constraints.rs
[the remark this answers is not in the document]
querolita: Addressed in the FFAdd PR (foreign_field_modulus was removed from the range check PR)

kimchi/src/circuits/expr.rs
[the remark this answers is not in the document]
querolita: addressed in FFAdd

utils/src/foreign_field.rs
querolita: The current version panics, but we can be more elegant here.

kimchi/src/circuits/polynomials/range_check/witness.rs
querolita: Oh well, you're right it is not used in this PR, but it will be used for FFMul and FFAdd, I can remove it from this PR.

kimchi/src/circuits/polynomials/range_check/witness.rs
querolita: Because in FFAdd and FFMul we will have both FF-related cells and range-check witness cells. Then for the latter type, we wanted to call that piece of code that handles the range-check cells independently. But I agree that for the scope of this PR, this could be postponed to FFAdd. I will revert it.

utils/src/field_helpers.rs
querolita: Fair enough... Yeah, in fact the only thing I needed was a pad to 32 when the length was smaller.

utils/src/foreign_field.rs
mimoo: I'm not sure I understand where it would panic.

(Also in general it's not good practice to do `for i in 0..N`, it's better to iterate through an iterator)

kimchi/src/circuits/polynomials/range_check/circuitgates.rs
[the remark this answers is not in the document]
querolita: Yeah not sure, Joseph likes some ASCII-style here and there, I guess he likes the mono-spaced style of it.

utils/src/foreign_field.rs
querolita: Mmm right now it panics in the FFAdd branch if we were to instantiate a foreign_field_modulus parameter that does not fit in 3 limbs -> [here](https://github.com/o1-labs/proof-systems/blob/21feaf439fc0c7b353fd35cae93d5c3288e3f9f9/kimchi/src/circuits/constraints.rs#L394)

book/src/specs/kimchi.md
```diff
@@ -1029,167 +1029,166 @@ This equation is translated as the constraint:
 
 #### Range Check
 
-Range check gadget:
+#### Range check gadget:
```
mimoo, line 1032: I think there's a header issue here :D (should be `#####`)

mimoo: (maybe at this point we can use `**Range check gadget**` instead of headers)

kimchi/src/prover.rs
```diff
@@ -588,9 +588,14 @@ where
...
                 });
-            if !index.cs.range_check_selector_polys.is_empty() {
-                index_evals.extend(range_check::circuit_gates().iter().enumerate().map(
-                    |(i, gate_type)| (*gate_type, &index.cs.range_check_selector_polys[i].eval8),
+            if index.cs.range_check_selector_polys.is_some() {
+                index_evals.extend(range_check::gadget::circuit_gates().iter().enumerate().map(
+                    |(i, gate_type)| {
+                        (
+                            *gate_type,
+                            &index.cs.range_check_selector_polys.as_ref().unwrap()[i].eval8,
```
mimoo, line 596: best practice is to avoid unwrap if you can, for example here you can rewrite this as:

```rust
if let Some(polys) = &index.cs.range_check_selector_polys {
 
 index_evals.extend(range_check::gadget::circuit_gates().iter().enumerate().map(
                    |(i, gate_type)| {
                        (
                            *gate_type,
                            &polys[i].eval8,
```

utils/src/field_helpers.rs
```diff
@@ -100,6 +106,22 @@ impl<F: Field> FieldHelpers<F> for F {
             bits
         })
     }
+
+    fn from_big(big: BigUint) -> Result<F> {
+        let mut bytes = big.to_bytes_le();
+
+        // Pad with zeros until multiple of 32 if necessary because the BigUint to_bytes_le function gives the smallest possible vector of bytes
+        if bytes.len() % 32 != 0 {
+            bytes.resize(32 + bytes.len() / 32, 0);
```
mimoo, line 115: this will also not work if bytes is like >= 2 * 32 of length ?

mimoo: if we really want to write our own (and not use the ones exposed by `PrimeField`) then I think we should just pad:

```rust
if bytes.len() < 32 {
   bytes.extend(std::iter::repeat(0).take(32 - bytes.len());
}
```

mimoo (approved):
Nice! Let's get this in : )

utils/src/field_helpers.rs
querolita: the new version pads to the smallest larger multiple of 32 bytes

querolita:
TODO: 
 * [ ] check what's the smallest domain to commit to range-check polys
 * [ ] decide whether to remove `SelectorPolynomial` struct or adapt functions
 * [ ] move wiring out of `CircuitGate` 
 * [ ] from big pads to zeros, do we want to adapt our functions to use the one already in `PrimeField`?
 * [ ] what amount of generic constants do we want in ForeignElement? Should we just stick to arrays of length 3?

utils/src/field_helpers.rs
```diff
@@ -112,7 +112,9 @@ impl<F: Field> FieldHelpers<F> for F {
 
         // Pad with zeros until multiple of 32 if necessary because the BigUint to_bytes_le function gives the smallest possible vector of bytes
         if bytes.len() % 32 != 0 {
-            bytes.resize(32 + bytes.len() / 32, 0);
+            // smallest larger multiple of 32 bytes
+            let goal = 32 + bytes.len() / 32;
```
mimoo, line 116: this is not going to be a multiple of 32 :D

mimoo: btw why does it have to be a multiple of 32?

querolita: Oh you're absolutely right! I forgot to multiply by 32 later 😧 I meant:

```
let goal = 32 * (1 + bytes.len() / 32 as usize );
````

The reason for the 32 bytes is to have a uniform format for foreign field elements of the 256-bit field when transforming from and to BigUint. Basically when the elements need less than those 256 bits because the serialization gives the smallest length in bytes possible. If they didn't all have the same format at some point, then the conversions do not lead to the same values.

I will make a quick PR just to fix the 32-multiple issue, and think about other alternative which do not have to pad.

querolita: Solved in the new branch -> [PR](https://github.com/o1-labs/proof-systems/pull/711)

PR #708, opened 2022-07-27
[ocaml] add missing ocaml derivations

mimoo:
https://github.com/MinaProtocol/mina/pull/11576

### DIFF
--- a/kimchi/src/oracles.rs
+++ b/kimchi/src/oracles.rs
@@ -47,6 +47,7 @@ pub mod caml {
 
     use super::*;
 
+    #[derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::Struct)]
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):

kimchi/src/oracles.rs
```diff
@@ -47,6 +47,7 @@ pub mod caml {
 
     use super::*;
 
+    #[derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::Struct)]
```
mrmr1993, line 50: This may need to be feature flagged, but I'll assume that you `cargo build`-ed locally to confirm.

mimoo: we're inside `#[cfg(feature = "ocaml_types")]pub mod caml { ... }`

PR #709, opened 2022-08-02
fn CompressedPubKey::empty()

hanabi1224:
The default value aligns with `empty` in [`non_zero_curve_point.ml`](https://github.com/MinaProtocol/mina/blob/develop/src/lib/non_zero_curve_point/non_zero_curve_point.ml#L90), it's useful for implementing default account etc.

```OCaml
let empty = Poly.{ x = Field.zero; is_odd = false }
```

### DIFF
--- a/signer/src/pubkey.rs
+++ b/signer/src/pubkey.rs
@@ -2,7 +2,7 @@
 //!
 //! Definition of public key structure and helpers
 
-use ark_ff::{BigInteger, PrimeField};
+use ark_ff::{BigInteger, PrimeField, Zero};
 use bs58;
 use core::fmt;
 use sha2::{Digest, Sha256};
@@ -148,6 +148,15 @@ pub struct CompressedPubKey {
 
     /// Parity of y-coordinate
     pub is_odd: bool,
+}
+
+impl Default for CompressedPubKey {
+    fn default() -> Self {
+        Self {
+            x: BaseField::zero(),
+            is_odd: false,
+        }
+    }
 }
 
 fn into_address(x: &BaseField, is_odd: bool) -> String {


mimoo:
Hey! Thanks for the PR :) I'm curious, can you explain what is the usecase for this? Is this mostly for testing or is there a prod usecase? Also, this point is not valid right?

hanabi1224:
Like when we merge hashes of empty nodes of the merkle ledger, it uses hash value of a default empty account, which uses this pk point.

This is the `Public_key.Compressed.empty` in empty account in [`account.ml`](https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_base/account.ml#L685)

```OCaml
let empty =
  { Poly.public_key = Public_key.Compressed.empty
  ; token_id = Token_id.default
  ; token_permissions = Token_permissions.default
  ; token_symbol = Token_symbol.default
  ; balance = Balance.zero
  ; nonce = Nonce.zero
  ; receipt_chain_hash = Receipt.Chain_hash.empty
  ; delegate = None
  ; voting_for = State_hash.dummy
  ; timing = Timing.Untimed
  ; permissions =
      Permissions.user_default
      (* TODO: This should maybe be Permissions.empty *)
  ; zkapp = None
  ; zkapp_uri = ""
  }

let empty_digest = digest empty
```

hanabi1224:
That makes sense. `Default` is not best trait in stdlib I can find to serve this purpose but I agree it's not the best fit for this case.  My goal is to avoid manually defining this empty pk everywhere, besides empty account, it's also used for calculating the hash of `Public_key.Compressed.t option` or `Option<CompressedPubkey>`, maybe I should just make it a non-trait function

mimoo (approved):
LGTM

PR #710, opened 2022-08-02
Update ROInput API to use builder pattern

hanabi1224:
This PR proposes API change in `ROInput` to use the [builder(-like) pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) to remove boilerplate code of defining mutable local variable in most cases.

before:
```rust
fn to_roinput(&self) -> ROInput {
    let mut roi = self.input.to_roinput();
    roi.append_field(self.pub_key_x);
    roi.append_field(self.pub_key_y);
    roi.append_field(self.rx);

    roi
}
```

after:
```rust
fn to_roinput(&self) -> ROInput {
    self.input
        .to_roinput()
        .append_field(self.pub_key_x)
        .append_field(self.pub_key_y)
        .append_field(self.rx)
}
```

### DIFF
--- a/hasher/README.md
+++ b/hasher/README.md
@@ -24,9 +24,9 @@ impl Hashable for Example {
     type D = u32;
 
     fn to_roinput(&self) -> ROInput {
-        let mut roi = ROInput::new();
-        roi.append_u32(self.x).append_u64(self.y);
-        roi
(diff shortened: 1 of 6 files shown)


mimoo (approved):
nice!

PR #711, opened 2022-08-02
[easy] Fix BigUint converters

querolita:
This PR fixes the to/from_big functions for field elements that used our own functions inside and now just uses Into/From implementations in PrimeField as suggested by David.

### DIFF
--- a/utils/src/field_helpers.rs
+++ b/utils/src/field_helpers.rs
@@ -1,7 +1,7 @@
 //! Useful helper methods to extend [ark_ff::Field].
 
-use ark_ff::{BigInteger, BigInteger256, Field, FpParameters, PrimeField};
-use num_bigint::{BigInt, BigUint, Sign, ToBigUint};
+use ark_ff::{BigInteger, Field, FpParameters, PrimeField};
+use num_bigint::BigUint;
 use std::ops::Neg;
 use thiserror::Error;
 
@@ -30,9 +30,6 @@ pub trait FieldHelpers<F> {
     /// Deserialize from bits
     fn from_bits(bits: &[bool]) -> Result<F>;
 
-    /// Deserialize from big unsigned integer
-    fn from_big_f(big: BigUint) -> Result<F>;
-
     /// Serialize to bytes
     fn to_bytes(&self) -> Vec<u8>;
 
@@ -41,9 +38,6 @@ pub trait FieldHelpers<F> {
 
     /// Serialize to bits
     fn to_bits(&self) -> Vec<bool>;
-
-    /// Serialize field element as a BigUint
-    fn to_big_f(self) -> BigUint;
 
     /// Field size in bytes
     fn size_in_bytes() -> usize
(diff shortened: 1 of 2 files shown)


utils/src/field_helpers.rs
```diff
@@ -106,23 +100,25 @@ impl<F: Field> FieldHelpers<F> for F {
...
-            // smallest larger multiple of 32 bytes
-            let goal = 32 + bytes.len() / 32;
-            bytes.extend(std::iter::repeat(0).take(goal - bytes.len()));
-        }
+    /// Serialize field element as a big unsigned integer
+    fn to_big(self) -> BigUint;
+}
 
-        F::deserialize(&mut &bytes[..]).map_err(|_| FieldHelpersError::DeserializeBytes)
+impl<F: PrimeField> BigFieldHelpers<F> for F {
```
querolita, line 114: I had to move them to a new trait because the other one is just for Field. But I would prefer it to be in the same trait as before if I could cast F as PrimeField just for these two.

utils/src/foreign_field.rs
```diff
@@ -67,7 +70,10 @@ impl<F: FftField, const N: usize> ForeignElement<F, N> {
     }
 
     /// Initializes a new foreign element from an element in the native field
-    pub fn new_from_field(field: F) -> Self {
+    pub fn new_from_field(field: F) -> Self
+    where
+        F: PrimeField,
```
mimoo, line 75: maybe we can require this for the entire impl?

utils/src/field_helpers.rs
```diff
@@ -106,23 +100,25 @@ impl<F: Field> FieldHelpers<F> for F {
...
+    fn from_big(big: BigUint) -> Result<F> {
+        big.try_into()
+            .map_err(|_| FieldHelpersError::DeserializeBytes)
     }
 
     fn to_big(self) -> BigUint {
-        let bytes = self.to_bytes();
-        BigUint::from_bytes_le(&bytes)
+        self.into_repr().into()
```
mimoo, line 121: I'm wondering why we need to wrap them in our own trait? If they're already in the PrimeField trait?

mimoo (approved):

utils/src/field_helpers.rs
mimoo: (by that I mean, why not just use the PrimeField functions directly where needed?)

querolita: I thought to_big() was easier to remember/more readable than the `into_repr().into()`, and similar for the from_big(). At least for now where I am using them in some functions of FFMul, it feels more immediate for me to use our wrapper. When the gates are finished though, we may realise that they are not being used that much and we may be able to get rid of them.

mimoo: you can just do `into()` wihtout the `into_repr()` though, similar with the other one

mimoo: for example:

```rust
let b: BigUint = Fp::from(3u32).into();
```

querolita: yeah it worked! I removed the to_big but left from_big for now

PR #712, opened 2022-08-02
[book] added more to pasta spec

mimoo:


querolita (approved):
Nice addition of the sage commands :)

PR #714, opened 2022-08-08
Keccak XOR gate

querolita:
This is a first version of the XOR gate for the Keccak gadget. There are still a bunch of TO-DO's in the code as they depend on the rest of gates needed for Keccak and how they will interact. 

In particular for a future PR:
 * [x] wiring
 * [x] rotation gates
 * [x] keccak selector
 * [x] tests 

For this PR:
 * [x] xor circuitgate
 * [x] xor gadget
 * [x] xor witness
 * [x] xor pattern
 * [x] xor tests

Addresses [issue 702](https://github.com/o1-labs/proof-systems/issues/702)


querolita:
I have a suggestion to modify the witness layout in a way that the whole 64-bit values for inputs and output are kept as well within the Xor/Zero gates. Meaning, right now the 32-bit halves are obtained with the `LimbWitnessCell` creator by selecting the bit intervals in the range check gates. Nonetheless, since these are not "full" copies, I think these do not produce copy constraints (wires) between the range check rows and the xor rows. 

Thus, shouldn't we store the "full" 64-bit values in columns <7 in the witness, apart from the 32-bit halves and the 4-bit crumbs? One possibility could be to store the current in1, in2, out in perhaps the curr rows (the xor gates), and then in the next rows (the zero gates) the full 64bits.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
+//~ The keccak gadget is comprised of _ circuit gates (KeccakXOR, .., and Zero)
```
jspada, line 1: Please link this up in `book/specifications/kimchi/Specification.toml` and `book/specifications/kimchi/template.md`

jspada, line 1: Pls define $A[x, i]$ notation above.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~
+//~ Keccak works with 64-bit words. The state is represented using $5\times 5$ matrix
+//~ of 64 bit words. Each compression step of Keccak consists of 24 rounds. Let us
+//~ denote the state matrix with A. Each round then consists of the following 5 steps:
+//~
+//~ \begin{align}
+//~ C[x] &= A[x,0] \oplus A[x,1] \oplus A[x,2] \oplus A[x,3] \oplus A[x,4] \\
```
jspada, line 8: Maybe define $A[x, 0]$ notation

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~ Kimchi framework that dictated those approaches.
+//~ * only 4 lookups per row
+//~ * only first 7 columns are available to the permutation polynomial
+//~
+//~ ##### 32-bit decomposition gate
+//~
+//~ This is a basic operation that is typically done for 64-bit initial state and
+//~ intermediate values.
+//~
+//~ Let `inp` be a 64-bit word. The constraint is: `in` $= 2^32 \cdot$ `in_hi` $+$ `in_lo`.
```
jspada, line 46: typo: 2^{32}

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
+//~ The keccak gadget is comprised of _ circuit gates (KeccakXOR, .., and Zero)
+//~
+//~ Keccak works with 64-bit words. The state is represented using $5\times 5$ matrix
+//~ of 64 bit words. Each compression step of Keccak consists of 24 rounds. Let us
+//~ denote the state matrix with A. Each round then consists of the following 5 steps:
+//~
+//~ \begin{align}
```
jspada, line 7: Need wrapping
$$
$$
here for github to render.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~ * only 4 lookups per row
+//~ * only first 7 columns are available to the permutation polynomial
+//~
+//~ ##### 32-bit decomposition gate
+//~
+//~ This is a basic operation that is typically done for 64-bit initial state and
+//~ intermediate values.
+//~
+//~ Let `inp` be a 64-bit word. The constraint is: `in` $= 2^32 \cdot$ `in_hi` $+$ `in_lo`.
+//~ It takes 3 cells for values `in`, `in_hi`, `in_lo`. We have not yet placed them w.r.t
```
jspada, line 47: Typo: w.r.t.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~ other rows of the Keccak computation; the only requirement is that all these cells be
+//~ within the first 7 columns for permutation equation accessibility.
```
jspada, line 49: typo: permutation **argument**?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~ |     7 | plookup0 `in1_0` | plookup4 `in1_4` |
+//~ |     8 | plookup1 `in1_1` | plookup5 `in1_5` |
+//~ |     9 | plookup2 `in1_2` | plookup6 `in1_6` |
+//~ |    10 | plookup3 `in1_3` | plookup7 `in1_7` |
+//~ |    11 | plookup0 `out_0` | plookup4 `out_4` |
+//~ |    12 | plookup1 `out_1` | plookup5 `out_5` |
+//~ |    13 | plookup2 `out_2` | plookup6 `out_6` |
+//~ |    14 | plookup3 `out_3` | plookup7 `out_7` |
+//~
+//~ Now we apply this gate twice to obtain a XOR gadget for 64-bit words by halving:
```
jspada, line 76: Why are we leaving those empty spaces?  Might be good to say now.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~  for 8 bits at a time, but for now we will use the 4-bit XOR table that we have.
+//~ ```
+//~
+//~ ##### Gate types:
+//~
+//~ Different rows are constrained using different CircuitGate types
+//~
+//~  | Row | `CircuitGate` | Purpose                        |
+//~  | --- | ------------- | ------------------------------ |
+//~  |   0 | `Xor`         | Xor first 2 bytes of low  half |
```
jspada, line 114: nit: KeccakXor

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~ ##### `KeccakXor` - XOR constraints for 32-bit words
+//~
+//~ * This circuit gate is used to constrain that `in1` xored with `in2` equals `out`.
+//~ * This gate operates on the `Curr` row and the `Next` row.
+//~
+//~ It uses three different types of constraints
+//~ * copy    - copy to another cell (32-bits)
+//~ * plookup - xor-table plookup (4-bits)
+//~
+//~ The 4-bit crumbs are assumed to be laid out with `0` being the least significant crumb.
```
jspada, line 138: column or cell `0`

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct KeccakXor<F>(PhantomData<F>);
+
+impl<F> Argument<F> for KeccakXor<F>
+where
+    F: PrimeField,
+{
+    const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::KeccakXor);
+    const CONSTRAINTS: u32 = 3;
+
+    // Constraints for Xor
```
jspada, line 170: KeccakXor

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Computes the decomposition of a 32-bit word whose most significant 4-bit crumb
+/// is located in the `max` column of `witness_next`. The layout is the following:
+///
+/// |        | max - 3 | max - 2 | max - 1 |     max |
+/// | ------ | ------- | ------- | ------- | ------- |
+/// | `Curr` |  crumb0 |  crumb1 |  crumb2 |  crumb3 |
+/// | `Next` |  crumb4 |  crumb5 |  crumb6 |  crumb7 |
+///
+fn four_bit<F: PrimeField, T: ExprOps<F>>(env: &ArgumentEnv<F, T>, max: usize) -> T {
```
jspada, line 200: This name doesn't convey a lot of meaning to the reader of the code.  Maybe `four_bit_decomoposition` or `four_bit_sum` would be better?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+    let mut sum = T::zero();
+    let two = T::one() + T::one();
```
jspada, line 202: `T::from(2u32)`

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+    let four_bit = two.pow(4);
```
jspada, line 203: Or just do `let four_bit = T::from(16u32)`

kimchi/src/circuits/polynomials/keccak/gadget.rs
```diff
@@ -0,0 +1,105 @@
...
+                coeffs: vec![],
+            },
+            CircuitGate {
+                typ: GateType::Zero,
+                wires: Wire::new(new_row + 4),
+                coeffs: vec![],
+            },
+        ]);
+
+        // TODO: copies when other gates are added using connect_cell_pair
```
jspada, line 70: Put `todo!()` here?

kimchi/src/circuits/polynomials/keccak/witness.rs
```diff
@@ -0,0 +1,120 @@
+//! Foreign field multiplication witness computation
```
jspada, line 1: Update to new framework will be easy once it's merged.

jspada (commented):
Looks good so far.  Requested some changes.

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -390,13 +394,39 @@ impl LookupPattern {
...
                     .collect()
             }
+            LookupPattern::Xor => {
+                (0..4)
+                    .map(|i| {
+                        // each row represents for XOR operations
+                        // where l XOR r = o
+                        //
+                        // 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
+                        // - - - l l l l r r r r  o  o  o  o
```
mrmr1993, line 404: This pattern doesn't look quite right; shouldn't it be
```
// - - - l - - - r - - - o - - -
// - - - - l - - - r - - - o - -
// - - - - - l - - - r - - - o -
// - - - - - - l - - - r - - - o
```

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
mrmr1993, line 46: You're using `inp` and `in` here. Switch both to `input` perhaps?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~
+//~ First we consider a XOR gate that checks that a 32-bit word `out` is the XOR of `in1` and `in2`.
+//~ This gate will use 2 rows, with a `Xor` row followed by a `Zero` row.
+//~
+//~ | Gates |          `Xor`   |          `Zero`  |
+//~ | ----- | ---------------- | ---------------- |
+//~ | Rows  |           0      |           1      |
+//~ | Cols  |                  |                  |
+//~ |     0 | copy     `in1`   | copy     `out`   |
+//~ |     1 |                  | copy     `in2`   |
```
mrmr1993, line 61: Comment about why `in2` is in the next column?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~
+//~ To fit within 15 wires, we first decompose each word into its lower and upper 32-bit
+//~ components. A gate for an atomic operation works with those 32-bit components at a time.
+//~
+//~ Before we describe the specific gate design approaches, below are some constraints in the
+//~ Kimchi framework that dictated those approaches.
+//~ * only 4 lookups per row
+//~ * only first 7 columns are available to the permutation polynomial
+//~
+//~ ##### 32-bit decomposition gate
```
mrmr1993, line 41: Do we need a separate gate for this? It would be nice to use the space in `XOR` below to (at least partially) extract the low 32-bytes.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
mrmr1993, line 1: Since this is a general-purpose XOR gate -- nothing about it seems keccak-specific -- doesn't it make sense to separate it out? I know that e.g. it would be handy to use this for SHA256.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,211 @@
...
+//~ |    14 | plookup3 `out_3` | plookup7 `out_7` |
+//~
+//~ Now we apply this gate twice to obtain a XOR gadget for 64-bit words by halving:
+//~
+//~ Consider the following operations:
+//~ * `out_lo` $=$ `in1_lo` $\oplus$ `in2_lo` and
+//~ * `out_hi` $=$ `in1_hi` $\oplus$ `in2_hi`,
+//~ where each element is 32 bits long.
+//~
+//~ | Gates |    `Xor` |   `Zero` |    `Xor` |   `Zero` |
```
mrmr1993, line 83: It feels rather tempting to have a 'chainable' gate here, along the lines of

```
        0   1   2   3   4   5   6   7   8   9   10  11  12  13  14
XOR     x   y   z   x_0 y_0 z_0 x_1 y_1 z_1 x_2 y_2 z_2 x_3 y_3 z_3
        x'  y'  z'
```

where you have x, y, z arbitrary-size values, and constrain
```
x_0 XOR y_0 = z_0
x_1 XOR y_1 = z_1
x_2 XOR y_2 = z_2
x_3 XOR y_3 = z_3
x = x_0 + 2^4 * x_1 + 2^(2*4) * x_2 + 2^(3*4) * x_3 + 2^(8*4) * x'
y = y_0 + 2^4 * y_1 + 2^(2*4) * y_2 + 2^(3*4) * y_3 + 2^(8*4) * y'
z = z_0 + 2^4 * z_1 + 2^(2*4) * z_2 + 2^(3*4) * z_3 + 2^(8*4) * z'
```

(that is, x', y', and z' are the remaining high bits of x, y, and z respectively)

With this, you would witness x, y, z, claiming that `x XOR y = z` in 64-bits, then run the gate twice and constrain the final outputs to be 0. (Likewise, you would run it 4 times for an 128-bit XOR.)

This also appears to use 3 rows (1 partial, could be reused for a generic) rather than 4.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
jspada: agree

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
jspada: nice idea

kimchi/src/circuits/lookup/lookups.rs
querolita: Oh! I didn't know how to represent the fact that at each row there was 4 lookups: one for 3 xor 7 = 11, other for 4 xor 8 = 12, other for 5 xor 9 = 13, a final one for 6 xor 10 = 14. But if this is represented as you suggest, then it is exactly the same pattern as ChaCha and then we can remove this one completely 😄 Actually the code in this pattern and that in ChaCha is identical... Yeah, both are doing the same thing!

kimchi/src/circuits/lookup/index.rs
```diff
@@ -109,6 +116,7 @@ impl<T> std::ops::Index<LookupPattern> for LookupSelectors<T> {
             LookupPattern::ChaChaFinal => &self.chacha_final,
             LookupPattern::LookupGate => &self.lookup_gate,
             LookupPattern::RangeCheckGate => &self.range_check_gate,
+            LookupPattern::Xor => &self.keccak_xor,
```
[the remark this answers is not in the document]
querolita: Actually after talking to Matthew it feels like we can reuse the chacha pattern so that's great

kimchi/src/circuits/gate.rs
```diff
@@ -230,6 +232,7 @@ impl<F: PrimeField> CircuitGate<F> {
             RangeCheck0 | RangeCheck1 => self
                 .verify_range_check::<G>(row, witness, cs)
                 .map_err(|e| e.to_string()),
+            KeccakXor => Ok(()), // TODO
```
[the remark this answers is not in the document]
querolita: You mean in KeccakXor or all of them?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: Oh well yeah. I have barely modified the explanation wrt Vanishree's design document. It surely can be extended and explained differently but for now until all the gates are settled I preferred to keep it succinct.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: iirc it is pretty arbitrary

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: Mmm I like the power, gives the reason why it is called four bit

kimchi/src/circuits/polynomials/keccak/gadget.rs
querolita: well not really, i don't want it to halt there yet because i want to still be able to use it

kimchi/src/circuits/polynomials/keccak/witness.rs
querolita: done!

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: I think it is pretty arbitrary and we could move `out` and `in2` to the current row. At first I thought it was for chaining (specially if only `out` was in the next row), but it is never the case...

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: absolutely agree, each of these pairs of gates have a total of 3 empty cells in permutable rows.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: Oh well, indeed it has nothing to do with Keccak at this level. It's simply not just any generic XOR, but only 32-bit XOR. So If you're fine with renaming it like Xor32 or similar, that would be nice as well.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: I really really like this approach! Specially because it does save in number of rows. Meaning, the original proposal needs 4 rows and 1,5 decomposition gates, for a total of 5.5 rows. Whereas here as you say, it is 4,5 or 5, depending on whether we reuse the final row of three leading zeros for something else (or we create a XORfinal gate that does not need to access Next, but costs one gate type more). 

Nonetheless, I think the offset for the implicit decomposition constrain would not be a fixed value but a power of 2^16 each time. Meaning, for each x y and z:

``` 
XOR1: x = x_0 + 2^4 * x_1 + 2^8 * x_2 + 2^12 * x_3 + 2^16 * x' (where x' has 48 bits)
XOR2: x' = x'_0 + 2^4 * x'_1 + 2^8 * x'_2 + 2^12 * x'_3 + 2^32 * x'' (where x'' has 32 bits) 
XOR3: x'' = x''_0 + 2^4 * x''_1 + 2^8 * x''_2 + 2^12 * x''_3 + 2^32 * x''' (where x''' has 16 bits) 
XOR4: x''' = x'''_0 + 2^4 * x'''_1 + 2^8 * x'''_2 + 2^12 * x'''_3 + 2^32 * x'''' (where x'''' is zero) 
```
Thus, I guess the offset could be a coefficient of the xor gate. What do you think?

querolita: I am pasting a chat with Matthew so everyone can read:

This is a different question, but if in the future we had an 8-bit XOR table instead, we would just need 2 rows per xor, period. No need for the 0,5 final one.

Just to have some numbers, each keccak gadget will need about 3 thousand 64-bit XORs. And going from the current 4-bit XOR table (256 rows) to an 8-bit one would mean 65k more rows. The new design will then need about 15k xor rows for 1 keccak, and the other one 6k. So as soon as we needed 8 keccaks in one single circuit, then it would make sense to have an 8-bit xor table. Right?

Oh, this reminds me of a question about plookup tables. Since XOR is commutative, do we have any means to check whether (a,b) belongs to the table, or (b,a)? I am asking because if we had, then we could halve the rows of the 8-bit xor table, and thus it could make sense to have it as soon as we needed as little as 4 keccaks in one circuit.

mrmr1993: > Since XOR is commutative, do we have any means to check whether (a,b) belongs to the table, or (b,a)?

Not that I'm aware of -- the entries in the table are encoded as a single value that represents the *ordered* tuple of values, so changing the order gives an entirely different value. I'm interested to hear any suggestions if you think we can tweak the plookup argument to allow for this though!

kimchi/src/circuits/lookup/index.rs
```diff
@@ -39,7 +39,7 @@ pub enum LookupError {
 #[derive(Clone, Serialize, Deserialize, Debug, Default)]
 pub struct LookupSelectors<T> {
     /// Chacha pattern lookup selector
-    pub chacha: Option<T>,
+    pub chacha_xor: Option<T>,
```
mrmr1993, line 42: Nit: rename to XOR?

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -279,7 +279,7 @@ impl<F: Copy> JointLookup<SingleLookup<F>, LookupTableID> {
     Copy, Clone, Serialize, Deserialize, Debug, EnumIter, PartialEq, Eq, PartialOrd, Ord, Hash,
 )]
 pub enum LookupPattern {
-    ChaCha,
+    ChaChaXor,
```
mrmr1993, line 282: Nit: rename to XOR?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -0,0 +1,148 @@
...
+use std::marker::PhantomData;
+
+use crate::circuits::{
+    argument::{Argument, ArgumentEnv, ArgumentType},
+    expr::constraints::ExprOps,
+    gate::GateType,
+};
+use ark_ff::PrimeField;
+
+//~ ##### `Xor16` - Chainable XOR constraints for words of multiples of 16 bits.
```
mrmr1993, line 54: Nit: move to `xor` module?

kimchi/src/circuits/polynomials/keccak/gadget.rs
```diff
@@ -0,0 +1,118 @@
...
+                coeffs: vec![],
+            },
+            CircuitGate {
+                typ: GateType::Xor16,
+                wires: Wire::new(new_row + 3),
+                coeffs: vec![],
+            },
+            CircuitGate {
+                typ: GateType::Zero,
+                wires: Wire::new(new_row + 4),
```
mrmr1993, line 56: Note that here we could use a generic gate to finish the range check on the output (`o = 0`) and then permute the first and second entries with it to check that they are also 0, entirely within the row.

kimchi/src/tests/keccak.rs
```diff
@@ -0,0 +1,119 @@
...
+#[test]
+// Test a XOR of 64bit whose output is all ones with alternating inputs
+fn test_xor64_alternating() {
+    let cs = create_test_constraint_system();
+
+    let zero_ones: u64 = 6510615555426900570;
+    let one_zeros: u64 = 11936128518282651045;
+    let witness = create(zero_ones, one_zeros);
+
+    for row in 0..=5 {
```
mrmr1993, line 33: Nit: this seems common to all the tests, worth factoring it into a helper?

kimchi/src/verifier.rs
```diff
@@ -656,6 +656,7 @@ where
                             }
                             RangeCheck0 => &index.range_check_comm.as_ref().unwrap()[0],
                             RangeCheck1 => &index.range_check_comm.as_ref().unwrap()[1],
+                            Xor16 => unimplemented!(),
```
mrmr1993, line 659: We need to add this, otherwise the verifier will crash!

book/src/specs/kimchi.md
```diff
@@ -1299,6 +1299,103 @@ for just this check requiring less constrains, but the cost of adding one more s
...
+
+To fit within 15 wires, we first decompose each word into its lower and upper 32-bit
+components. A gate for an atomic operation works with those 32-bit components at a time.
+
+Before we describe the specific gate design approaches, below are some constraints in the
+Kimchi framework that dictated those approaches.
+* only 4 lookups per row
+* only first 7 columns are available to the permutation polynomial
+
+##### `Xor16` - Chainable XOR constraints for words of multiples of 16 bits.
```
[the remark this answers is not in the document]
querolita: Sure, it is no longer a Keccak-only gate 👍🏻

kimchi/src/tests/keccak.rs
querolita: Done!

kimchi/src/circuits/polynomials/xor.rs
```diff
@@ -0,0 +1,288 @@
...
+                typ: GateType::Xor16,
+                wires: Wire::new(new_row + i),
+                coeffs: vec![],
+            })
+            .collect::<Vec<_>>();
+        let zero_row = new_row + num_xors;
+        gates.push(CircuitGate {
+            typ: GateType::Generic,
+            wires: Wire::new(zero_row),
+            coeffs: vec![],
```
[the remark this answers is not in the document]
querolita: Oh, because otherwise it will not do the operation, right!

querolita:
Small update: I have to reorder the layout of the witness to fully align with ChaCha Xor plookup pattern. The pattern assumes all left input first, then all right input, then all output. Here I am laying them out alternating. Will do this tomorrow morning and merge!

querolita:
UPDATE: All suggestions addressed. In particular, the end-to-end prove is working, and all the proof-systems related parameters have been included (verifier, prover).

kimchi/src/circuits/polynomials/xor.rs
```diff
@@ -0,0 +1,546 @@
...
+                typ: GateType::Xor16,
+                wires: Wire::new(new_row + i),
+                coeffs: vec![],
+            })
+            .collect::<Vec<_>>();
+        let zero_row = new_row + num_xors;
+        gates.push(CircuitGate {
+            typ: GateType::Generic,
+            wires: Wire::new(zero_row),
+            coeffs: vec![F::one(), F::one(), -F::one()],
```
mrmr1993, line 53: I don't think we want to check that `fin_in1 + fin_in2 - fin_out = 0`. Just constrain `fin_in = 0` with `vec[F::one()]`

kimchi/src/circuits/polynomials/xor.rs
```diff
@@ -0,0 +1,546 @@
...
+    fn constraint_checks<T: ExprOps<F>>(env: &ArgumentEnv<F, T>) -> Vec<T> {
+        // Returns the constraints:
+        // in1 = in1_0 + in1_1 * 2^4 + in1_2 * 2^8 + in1_3 * 2^12 + next_in1 * 2^16
+        // in2 = in2_0 + in2_1 * 2^4 + in2_2 * 2^8 + in2_3 * 2^12 + next_in2 * 2^16
+        // out = out_0 + out_1 * 2^4 + out_2 * 2^8 + out_3 * 2^12 + next_out * 2^16
+        (0..3)
+            .map(|i| {
+                env.witness_curr(i)
+                    - chunk_sum(env, 3 + 4 * i)
+                    - T::from(2u64).pow(16) * env.witness_next(i)
```
mrmr1993, line 418: I'm not sure that it helps to have `chunk_sum` here, it might be clearer to write out the constraints explicitly like they are above.

mrmr1993 (approved):

kimchi/src/circuits/polynomials/xor.rs
querolita: Oh well, I was trying to avoid code duplication, but i guess it is not that bad since it's just 3 times. I will change it prior merge.

jspada (commented):
approved. will leave some comments in next pr

Issue #701, opened 2022-07-25
Support empty commitments


mimoo:
Currently, commiting to zero polynomials will return an empty commitment: 

https://github.com/o1-labs/proof-systems/blob/master/poly-commitment/src/commitment.rs#L594

from @imeckler 

> The way to really fix it is to add the hiding generator to all the index commitments (like 1*h so the blinding factor is known when we open)

mimoo:
^ when this is done, we should remove the specific handling of https://github.com/o1-labs/proof-systems/pull/717

mimoo:
I think https://github.com/o1-labs/proof-systems/pull/795 was the last PR that implemented this manually, it would be great to have an API that does this and remove the "manual-ness"

PR #717, opened 2022-08-11
fix all-zeros public input for zkapps

mimoo:
This is a fix for https://github.com/o1-labs/snarkyjs/issues/344 that implements https://github.com/o1-labs/proof-systems/issues/701 for the public commitment/polynomial only

The mina PR associated with this is here https://github.com/MinaProtocol/mina/pull/11652

---

I think we should implement https://github.com/o1-labs/proof-systems/issues/701 next to generalize this pattern

### DIFF
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -500,6 +500,34 @@ where
 }
 
 impl<G: CommitmentCurve> SRS<G> {
+    /// Commits a polynomial, and adds the blinding factor to it.
+    /// It is not a blinding commitment, as the blinding factor is scaled by 1 instead of a random value.
+    /// This is useful for when the polynomial can be zero,
+    /// in which case the commitment would be the point at infinity if no blinding factor was added.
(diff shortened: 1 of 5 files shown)


poly-commitment/src/commitment.rs
```diff
@@ -500,6 +500,34 @@ where
 }
 
 impl<G: CommitmentCurve> SRS<G> {
+    /// Commits a polynomial, and adds the blinding factor to it.
+    /// It is not a blinding commitment, as the blinding factor is scaled by 1 instead of a random value.
+    /// This is useful for when the polynomial can be zero,
+    /// in which case the commitment would be the point at infinity if no blinding factor was added.
+    /// (And points at infinity are annoying to handle in circuits.)
+    pub fn commit_public(&self, poly: &DensePolynomial<G::ScalarField>) -> BlindedCommitment<G> {
```
mrmr1993, line 508: It seems like you've reinvented a apecialised `mask_custom`. Can we use that instead?

mrmr1993 (commented):
A comment and some questions

PR #722, opened 2022-08-23
Replace body of `mask` with call to `mask_custom`

mrmr1993:
This PR replaces the two distinct implementations of masking with a single implementation.


jspada (approved):
Looks good to me.  We should fix the potential, albeit unlikely, leak mentioned in the comment.

PR #724, opened 2022-08-26
Allow an all-zero public input

mrmr1993:
This PR tweaks the handling of public inputs, so that the public input commitment can be allowed to be all-zero.

The handling of zero polynomials and commitments is special-cased in lots of places in the code, so this was a rather tricky change to make. In particular, the 'commitments' test is still broken (via a probabilistic flake), since it depends more closely on some of those special cases.

This is ready for review, but should not be merged until the flaking test is fixed.

### DIFF
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -232,7 +232,7 @@ impl<C: AffineCurve> PolyComm<C> {
             },
             unshifted: {
                 if com.is_empty() || elm.is_empty() {
-                    Vec::new()
+                    vec![C::zero()]
                 } else {
                     let n = Iterator::max(com.iter().map(|c| c.unshifted.len())).unwrap();
                     (0..n)
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -563,7 +563,7 @@ impl<G: CommitmentCurve> SRS<G> {
 
         // committing all the segments without shifting
         let unshifted = if is_zero {
-            Vec::new()
+            vec![G::zero()]
         } else {
             (0..p / n + if p % n != 0 { 1 } else { 0 })
                 .map(|i| {
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -435,7 +435,7 @@ pub fn combined_inner_product<F: PrimeField>(
 
         if let Some(m) = shifted {
             // polyscale^i sum_j evalscale^j elm_j^{N - m} f(elm_j)
-            let last_evals = if *m > evals.len() * srs_length {
+            let last_evals = if *m >= evals.len() * srs_length {
                 vec![F::zero(); evaluation_points.len()]
             } else {
                 evals[evals.len() - 1].clone()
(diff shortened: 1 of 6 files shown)


poly-commitment/src/evaluation_proof.rs
```diff
@@ -112,7 +112,7 @@ impl<G: CommitmentCurve> SRS<G> {
             let mut scale = G::ScalarField::one();
 
             // iterating over polynomials in the batch
-            for (p_i, degree_bound, omegas) in plnms.iter().filter(|p| !p.0.is_zero()) {
+            for (p_i, degree_bound, omegas) in plnms.iter() {
```
mimoo, line 115: in `&plnms`

poly-commitment/src/commitment.rs
```diff
@@ -435,7 +435,7 @@ pub fn combined_inner_product<F: PrimeField>(
 
         if let Some(m) = shifted {
             // polyscale^i sum_j evalscale^j elm_j^{N - m} f(elm_j)
-            let last_evals = if *m > evals.len() * srs_length {
+            let last_evals = if *m >= evals.len() * srs_length {
```
mimoo, line 438: I really think we ought to move to this simplified function: https://github.com/o1-labs/proof-systems/pull/448/files and handle chunked polynomials outside of the API

mimoo: btw the reason I couldn't move to that code is the empty polynomials (like selectors when you don't use them)

poly-commitment/src/commitment.rs
```diff
@@ -518,15 +518,9 @@ impl<G: CommitmentCurve> SRS<G> {
...
-                    // We should change this so that we still mask in this case
-                    g
-                } else {
-                    let mut g_masked = self.h.mul(b);
-                    g_masked.add_assign_mixed(&g);
-                    g_masked.into_affine()
-                }
+                let mut g_masked = self.h.mul(b);
+                g_masked.add_assign_mixed(&g);
+                g_masked.into_affine()
```
mimoo, line 523: I'm still not sure how this fixes the issue, as your public polynomial or the selector polynomials are not masked

jspada: good question

poly-commitment/src/commitment.rs
mrmr1993: Perhaps, but I don't think we should block this PR on it.

poly-commitment/src/commitment.rs
mrmr1993: The tests pass :man_shrugging: This PR makes the handling of zero commitments correct; I'm going to add blinding and propagate it to the pickles side separately.

mimoo (approved):

poly-commitment/src/evaluation_proof.rs
```diff
@@ -130,9 +130,11 @@ impl<G: CommitmentCurve> SRS<G> {
                     scale *= &polyscale;
                     offset += self.g.len();
                     if let Some(m) = degree_bound {
-                        if offset > *m {
-                            // mixing in the shifted segment since degree is bounded
-                            plnm.add_shifted(scale, self.g.len() - m % self.g.len(), segment);
+                        if offset >= *m {
```
mimoo, line 133: perhaps it'd be a good idea to start avoiding "degree" and use `size` or `length`

cf https://twitter.com/cronokirby/status/1568268593946677248

PR #725, opened 2022-08-26
Add a deterministic version of the commitment unit test

mrmr1993:
This PR adds a deterministic version of the commitment test, to ensure that we are reliably testing commitments that feature a 'zero' commitment.

This is part of the work on #724, but separated since it is orthogonal and good-to-have anyway.


jspada (approved):
Looks great!

fabrizio-m (approved):

PR #727, opened 2022-08-26
Add interface for reusing constraint code for witness computation

### LINKED ISSUE #726: Framework for reusing constraint code for witness computation
We often find ourselves rewriting code that we already implemented for the constraints again when implementing witness computation.  This is error prone, time consuming and creates more code to maintain.

Since the constraints are written in `Expr` and both `Expr` and `ark_ff::fields::Field` implement the standard arithmetic traits we should be able to produce generic code that can be reused with either `Expr` (when generating constraints) or `ark_ff::fields::Field` (when computing the witness).

This issue makes this possible by introducing an `ArithmeticOps` trait.

jspada:
* Create ArithmeticOps trait
    * Fixed some unrelated cargo doc warnings
    
    Closes #726

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -2146,15 +2146,54 @@ where
 pub mod constraints {
     use super::*;
 
+    /// This trait defines a common arithmetic operations interface
+    /// that can be used by constraints.  It allows us to reuse
+    /// constraint code for witness computation.
+    pub trait ArithmeticOps:
+        std::ops::Add<Output = Self>
+        + std::ops::Sub<Output = Self>
+        + std::ops::Neg<Output = Self>
+        + std::ops::Mul<Output = Self>
+        + Clone
+        + Zero
+        + One
+        + From<u64>
+    // Add more as necessary
+    where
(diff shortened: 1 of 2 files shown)


mrmr1993 (approved):

kimchi/src/circuits/expr.rs
```diff
@@ -2146,15 +2146,54 @@ where
...
+    pub trait ArithmeticOps:
+        std::ops::Add<Output = Self>
+        + std::ops::Sub<Output = Self>
+        + std::ops::Neg<Output = Self>
+        + std::ops::Mul<Output = Self>
+        + Clone
+        + Zero
+        + One
+        + From<u64>
+    // Add more as necessary
```
fabrizio-m, line 2161: just to know, is there an specific reason to no have Copy too? like a limitation in some implementor.

fabrizio-m (approved):

kimchi/src/circuits/expr.rs
jspada: No reason that I'm aware of.  I just added the minimal set that the current codebase requires 👍🏼

PR #729, opened 2022-09-01
[Trivial] Rename function parameters to be consistent between different functions

mrmr1993:
This PR renames `xi` and `r` to `polyscale` and `evalscale` respectively, so that the same descriptive names are used for `open` in `evaluation_proof.rs` and `verify` in `commitment.rs`.


jspada (approved):
Love this kind of improvement!

PR #733, opened 2022-09-02
Remove lookup context creation function

jspada:


mimoo:
@jspada what was the conclusion of the meeting on wed? Are we getting this in or are we removing LookupContext through a refactor?

jspada:
> @jspada what was the conclusion of the meeting on wed? Are we getting this in or are we removing LookupContext through a refactor?

I think the conclusion was to keep it in for now (due to time constraints & priorities), but the conclusion was also that I'm not going to use it anyway (also due to time constraints & priorities).  Furthermore, we plan to eventually do the refactor of unit tests to the TestFramework and witness verification the better way.  Therefore, as fas as I'm concerned, since I'm not going to use it, then it's completely up to you and Matthew if you want to remove it-- pick your PR! :-)

mimoo:
ah, well in this case let's get this in and fix the spec :) can you fix the merge conflicts?

mimoo (approved):

querolita (approved):
I checked the diff was correct.

PR #734, opened 2022-09-06
Add `double` and `pow` to `ArithmeticOps`

mrmr1993:


jspada (approved):
Awesome - just what we need!

PR #741, opened 2022-09-08
Allow constraints to be reused for witness verification

### LINKED ISSUE #745: Allow constraints to be reused for witness verification
Currently, for witness verification we typically reimplement the constraints rather than reusing them, so we end up with two implementations.  This is because the constraints are written in the expression framework as an AST, but not as equations over field elements that can be evaluated directly.  They can only be evaluated by the proof system by the time consuming process of generating a proof.

When we design gates and need to debug the constraints we either have to run the prover (which is a slow debug process) or create another implementation of the constraints that can verify the witness elements directly (which is duplication of work and can be a source of bugs, in which case, we must revert to running the prover to track down).

What would be better is to be able to implement the constraints once and then have the ability to automatically instantiate them as either
* An expression framework AST (e.g. for the proof system)
* Equations over field elements (e.g. for directly verifying the witness without using the proof system)

This will save implementation and debug time when developing custom gates.

To this end, we extend the Expression Framework with the `ExprOps` trait (which is implemented by both `Field` and `Expr<ConstantExpr<Field>>`) .  We also extend the `Argument` interface so that the constraints definition is parameterised by an environment that allows the developer to specify at runtime whether to render the constraints as equations of  `Expr<ConstantExpr<Field>>` or `Field` elements.

jspada:
Closes #745


kimchi/src/circuits/argument.rs
```diff
@@ -33,13 +39,24 @@ pub trait Argument<F: FftField> {
     /// The number of constraints created by the argument.
     const CONSTRAINTS: u32;
 
+    /// Constraints for this argument
+    fn constraints<T: ArithmeticOps<F>>(witness: &GateWitness<T>, constants: Vec<T>) -> Vec<T>;
```
mimoo, line 43: btw have we considered implementing `Field` on the types in `expr.rs` instead of creating an `ArithmeticOps`? Could be cleaner

mimoo: also semantically I don't understand this change, why would a gate take a witness?

kimchi/src/circuits/expr.rs
```diff
@@ -2170,39 +2175,64 @@ pub mod constraints {
 
         /// Raise the value to the given power
         fn pow(&self, p: u64) -> Self;
+
+        fn literal(x: F) -> Self;
+
+        /// Cache item
+        fn cache(&self, cache: &mut Cache) -> Self;
```
mimoo, line 2182: these don't seem to be arithmetic ops

kimchi/src/circuits/argument.rs
jspada: It's an idea I have considered, but Field seems to have a lot of other things we would not want to implement.  I will consider again and post back here.

kimchi/src/circuits/expr.rs
jspada: Indeed - I think the name should change now that we get a better idea of what it is.  Any suggestions?

kimchi/src/circuits/argument.rs
jspada: > why would a gate take a witness?
The gate operates on the witness to see whether it's relation is satisfied by the witness assignment.  This allows the definition of the gate, i.e., its constraints to be reused for witness verification.  That is we can have a generic `verify_witness()` that can not only reuse the constraints (no second implementation required), but also even pinpoint which constraints failed, thus, aiding the debug process for gate developers.

jspada: > also semantically I don't understand this change, why would a gate take a witness?


The gate operates on the witness to see whether its relation is satisfied by the witness assignment.  This allows the definition of the gate (i.e., its constraints) to be reused for witness verification (not only expr generation).  That is, we can have a generic `verify_witness()` that can not only reuse the constraints (no second implementation required), but also even pinpoint which constraints failed, thus, aiding the debug process for gate developers.

mimoo: > It's an idea I have considered, but Field seems to have a lot of other things we would not want to implement. I will consider again and post back here.

yeah but we can easily just add `unimplemented!()` for all the functions we don't care about in the implementation. It could make a lot of this code cleaner potentially

jspada: no - it's a work in progress, but I put it here to get early feedback.

Hmm interesting idea to use `unimplemented!()`, but not only would we need to implement `Field` on `expr.rs`, but we need some expression functions' behaviour to change when used on real field elements instead of expressions.

kimchi/src/circuits/expr.rs
mrmr1993: `ExprOperations`?

kimchi/src/circuits/polynomials/varbasemul.rs
```diff
@@ -242,24 +256,78 @@ fn single_bit<F: FftField>(
...
+
+trait FromWitness<F, T> {
+    fn new_from_env(&self, env: &ArgumentEnv<F, T>) -> T;
+}
+
+impl<F: Field, T: ArithmeticOps<F>> FromWitness<F, T> for Variable {
+    fn new_from_env(&self, env: &ArgumentEnv<F, T>) -> T {
+        let column_to_index = |_| match self.col {
+            Column::Witness(i) => i,
+            _ => panic!("Can get index from witness columns"),
```
mrmr1993, line 288: Nit: `Can't`?

mrmr1993 (approved):
Really nicely done!

kimchi/src/circuits/polynomials/varbasemul.rs
jspada: Ty!

mrmr1993 (approved):

kimchi/src/circuits/argument.rs
jspada: n.b. we are not mixing test code with production.  We are making it so that we can directly test the some of the production code without running the prover or needing to create an alternative implementation to check the witness.

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -477,7 +477,7 @@ mod tests {
         curve::KimchiCurve,
         proof::{LookupEvaluations, ProofEvaluations},
     };
-    use ark_ff::UniformRand;
+    use ark_ff::{UniformRand, Zero};
```
querolita, line 480: why was it not needed before?

jspada: I was needed, but it was located near the top of the file.  However, the new version of clippy says it's only needed for the tests module and so it's moved there.

querolita (approved):
This will be a game changer 🥳

kimchi/src/circuits/polynomials/varbasemul.rs
```diff
@@ -171,22 +187,22 @@ fn single_bit_witness<F: FftField>(
...
     b: Variable,
-    base: CurveVar,
+    base: &Point<Variable>,
     s1: Variable,
-    input: CurveVar,
-    output: CurveVar,
+    input: &Point<Variable>,
+    output: &Point<Variable>,
     b_value: F,
     base_value: (F, F),
```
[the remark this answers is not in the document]
jspada: No, because those are `expr::Variable`, which constains only `col,row`.  Think of them as placeholders (expr stuff) and this function translates them into values using environment which will either be `F`s or `Expr<ConstantExpr<F>>`s depending on the context.

jspada:
> This will be a game changer 🥳

kimchi/src/circuits/expr.rs
jspada: sgtm!

kimchi/src/circuits/polynomials/endomul_scalar.rs
```diff
@@ -158,29 +158,38 @@ where
...
+        let a8 = env.witness_curr(4);
+        let b8 = env.witness_curr(5);
 
         // x0..x7
-        let xs: [_; 8] = array::from_fn(|i| witness_curr(6 + i));
+        let xs: [_; 8] = array::from_fn(|i| env.witness_curr(6 + i));
 
         let mut cache = Cache::default();
 
         let c_coeffs = [
```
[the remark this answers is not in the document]
jspada: I like this idea.  I've updated it.

kimchi/src/circuits/polynomials/endomul_scalar.rs
```diff
@@ -45,11 +45,11 @@ impl<F: PrimeField> CircuitGate<F> {
     }
 }
 
-fn polynomial<F: Field>(coeffs: &[F], x: &E<F>) -> E<F> {
+fn polynomial<F: Field, T: ArithmeticOps<F>>(coeffs: &[T], x: &T) -> T {
     coeffs
         .iter()
         .rev()
-        .fold(E::zero(), |acc, c| acc * x.clone() + E::literal(*c))
+        .fold(T::zero(), |acc, c| acc * x.clone() + c.clone())
```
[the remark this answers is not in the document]
jspada: Good idea, I'll do it.

kimchi/src/circuits/polynomials/varbasemul.rs
```diff
@@ -157,7 +157,23 @@ impl<F: PrimeField> CircuitGate<F> {
...
+    }
+}
+
+impl Point<Variable> {
+    pub fn new_from_env<F: Field, T: ArithmeticOps<F>>(&self, env: &ArgumentEnv<F, T>) -> Point<T> {
+        Point::create(self.x.new_from_env(env), self.y.new_from_env(env))
+    }
+}
 
 fn set<F>(w: &mut [Vec<F>; COLUMNS], row0: usize, var: Variable, x: F) {
```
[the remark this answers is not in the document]
jspada: No, `x` an `expr::Variable` and gets converted to either `F` or `Expr<ConstantExpr<F>>` depending on the environment.

mrmr1993 (approved):

kimchi/src/circuits/gate.rs
```diff
@@ -206,6 +228,126 @@ impl<F: PrimeField> CircuitGate<F> {
...
+        &self,
+        row: usize,
+        witness: &[Vec<F>; COLUMNS],
+        cs: &ConstraintSystem<F>,
+        _public: &[F],
+    ) -> CircuitGateResult<()> {
+        // Grab the relevant part of the witness
+        let argument_witness = self.argument_witness(row, witness)?;
+        // Set up the constants.  Note that alpha, beta, gamma and joint_combiner
+        // are zero because this function is not running the prover.
```
mrmr1993, line 243: `s/zero/one`

mrmr1993 (approved):

PR #742, opened 2022-09-08
Tidy up `open`

mrmr1993:
This PR tidies up the code in `open`, unifying the two implementations and using more idiomatic rust.


jspada (approved):
good refactor; looks correct to me

PR #743, opened 2022-09-08
applied cargo clippy -- -W clippy::explicit_iter_loop

mimoo:


mrmr1993 (approved):
LGTM.

Do we need to add this flag on CI too? I know I keep getting this wrong :)

mimoo:
not a bad idea, I remember making a list of pedantic flags (that are not necessarily listed in pedantic) that are nice to have, but lost that list : ( (also the list keeps on growing with each releases!)

PR #744, opened 2022-09-12
Replace array_init with array::from_fn in zero publc input test

mrmr1993:


jspada (approved):
perfect!

mimoo:
oh I guess that means array_init is still present in `Cargo.toml` files?

Issue #716, opened 2022-08-11
remove array_init dependency


mimoo:
Looks like [array::from_fn](https://doc.rust-lang.org/stable/std/array/fn.from_fn.html) got stabilized in the latest Rust, so we don't need array_init anymore

duguorong009:
Can I take this one? @mimoo 
In fact, I already done replacement in my local machine.
But, I am not sure if it is okay to create PR without any inform.
(maybe, just following contribute guideline is enough?)

duguorong009:
@mimoo 
Created a first contribution PR. 
Please check & give me the review.

mimoo:
hey! Sorry for the late response, but I see that you already did it so thanks a lot : )

mimoo:
reopening as it looks like array-init is still part of the `Cargo.toml`

duguorong009:
@mimoo 
I also noticed that `array-init` still exists in some `Cargo.toml` files.
Will re-create the PR to fix it.

duguorong009:
Created a new PR #748

PR #748, opened 2022-09-14
Remove the `array-init` crates from toml files

duguorong009:
**Reference**: #716


mimoo (approved):
Awesome thanks a lot again :)

Issue #513, opened 2022-04-07
Fix pedantic lints


mimoo:
run:

```
$ cargo clippy -- -W clippy::pedantic
```

and fix some of the warnings

duguorong009:
@mimoo 
I will also take care of this issue.

mimoo:
cool thanks : ) don't hesitate to ping me here or on twitter if you have questions.

duguorong009:
@mimoo 
I think the following line is for creating the `cell(s)` for setting up the `prover_index(prover_key)`.
https://github.com/o1-labs/proof-systems/blob/master/circuit-construction/src/prover.rs#L175
Atm, it does have one parameter(`FnOnce`) for creating the default empty value.(like ` let v = system.var(|| panic!("fail"));`)
Can I update the `Writer::var()` trait to have no parameter?
The reason is that I think it is unnecessary.
Also, the parameter value is, in most cases, `|| panic("fail!")`.

duguorong009:
@mimoo 
Please go for reviewing the PR #752 .

PR #752, opened 2022-09-16
Fix some warnings from clippy pedantic lints

duguorong009:
**What's been done:**
Fix some warnings from rust clippy pedantic lints(`cargo clippy -- -W clippy::pedantic`).

**Reference:** #513

### DIFF
--- a/circuit-construction/src/writer.rs
+++ b/circuit-construction/src/writer.rs
@@ -6,7 +6,10 @@ use kimchi::circuits::{
     },
     wires::{Wire, COLUMNS},
 };
-use oracle::{constants::*, permutation::full_round};
+use oracle::{
+    constants::{PlonkSpongeConstantsKimchi, SpongeConstants},
+    permutation::full_round,
+};
 use std::array;
 use std::collections::HashMap;
 
@@ -23,7 +26,10 @@ pub struct Var<F> {
 
 impl<F: Copy> Var<F> {
     /// Returns the value inside a variable [Var].
-    /// It panics if it is `None`.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if it is `None`.
     pub fn val(&self) -> F {
         self.value.unwrap()
     }
@@ -50,6 +56,7 @@ impl<F: FftField> GateSpec<F> {
         }
     }
 
+    #[must_use]
     pub fn get_var_idx(&self, col: usize) -> Option<usize> {
         match self.row.get(col) {
             Some(Some(var)) => Some(var.index),
@@ -60,7 +67,7 @@ impl<F: FftField> GateSpec<F> {
 
 /// A set of gates within the circuit.
 /// It carries the index for the next available variable,
-/// and the vector of [GateSpec] created so far.
+/// and the vector of [`GateSpec`] created so far.
 /// It also keeps track of the queue of generic gates and cached constants.
 #[derive(Default)]
 pub struct System<F> {
@@ -145,7 +152,7 @@ pub trait Cs<F: PrimeField> {
         })
     }
 
-    /// This function creates a [ShiftedScalar] variable from a field element that is
+    /// This function creates a [`ShiftedScalar`] variable from a field element that is
     /// returned by function `g()`, and a length that should be a multiple of 5.
     fn scalar<G, Fr: PrimeField>(&mut self, length: usize, g: G) -> ShiftedScalar<F>
     where
@@ -264,7 +271,7 @@ pub trait Cs<F: PrimeField> {
         let same_x = self.var(|| {
             let same_x = x1.val() == x2.val();
             same_x_bool = same_x;
-            F::from(same_x as u64)
+            F::from(u64::from(same_x))
         });
 
         let inf = zero;
@@ -341,7 +348,7 @@ pub trait Cs<F: PrimeField> {
         let same_x = self.var(|| {
             let same_x = x1.val() == x2.val();
             same_x_bool = same_x;
-            F::from(same_x as u64)
+            F::from(u64::from(same_x))
         });
 
         let inf = zero;
@@ -407,7 +414,7 @@ pub trait Cs<F: PrimeField> {
     /// For (1):
     /// - Creates a row with left wire `t`, right wire `f`, and output wire `delta`
     /// - Assigns `1` to the left coefficient, `-1` to the right coefficient, and `-1` to the output coefficient.
-    /// - That way, it creates a first gate constraining: `1 * t - 1 * f - delta = 0``
+    /// - That way, it creates a first gate constraining: `1 * t - 1 * f - delta = 0`
     /// For (2):
     /// - Creates a row with left wire `b`, right wire `delta`, and output wire `res1`.
     /// - Assigns `-1` to the output coefficient, and `1` to the multiplication coefficient.
@@ -455,7 +462,7 @@ pub trait Cs<F: PrimeField> {
         res
     }
 
-    /// Performs a scalar multiplication between a [ShiftedScalar] and a point `(xt, yt)`.
+    /// Performs a scalar multiplication between a [`ShiftedScalar`] and a point `(xt, yt)`.
     /// This function creates 51 rows pairs of rows.
     fn scalar_mul(
         &mut self,
@@ -521,7 +528,7 @@ pub trait Cs<F: PrimeField> {
                 row: row2.into_iter().map(Some).collect(),
                 typ: GateType::Zero,
                 coeffs: vec![],
-            })
+            });
         }
 
         res.unwrap()
@@ -538,6 +545,7 @@ pub trait Cs<F: PrimeField> {
     /// |  r  | `EndoMul`  |
     /// | r+1 | `Zero`     |
     ///
+    #[allow(clippy::too_many_lines)]
     fn endo(
         &mut self,
         zero: Var<F>,
@@ -563,9 +571,9 @@ pub trait Cs<F: PrimeField> {
                             .take(length_in_bits)
                             .copied()
                             .rev()
-                            .collect()
+                            .collect();
                     }
-                    F::from(bits_[i] as u64)
+                    F::from(u64::from(bits_[i]))
                 })
             })
             .collect();
@@ -759,7 +767,7 @@ pub trait Cs<F: PrimeField> {
     /// Creates a Poseidon gadget for given constants and a given input.
     /// It generates a number of `Poseidon` gates followed by a final `Zero` gate.
     fn poseidon(&mut self, constants: &Constants<F>, input: Vec<Var<F>>) -> Vec<Var<F>> {
-        use kimchi::circuits::polynomials::poseidon::*;
+        use kimchi::circuits::polynomials::poseidon::{POS_ROWS_PER_HASH, ROUNDS_PER_ROW};
 
         let params = constants.poseidon;
         let rc = &params.round_constants;
@@ -782,7 +790,7 @@ pub trait Cs<F: PrimeField> {
                                         // Do one full round on the previous value
                                         let mut acc = states[states.len() - 1]
                                             .iter()
-                                            .map(|x| x.val())
+                                            .map(crate::writer::Var::val)
                                             .collect();
                                         full_round::<F, PlonkSpongeConstantsKimchi>(
                                             params,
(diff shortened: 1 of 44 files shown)


circuit-construction/src/writer.rs
```diff
@@ -782,7 +790,7 @@ pub trait Cs<F: PrimeField> {
                                         // Do one full round on the previous value
                                         let mut acc = states[states.len() - 1]
                                             .iter()
-                                            .map(|x| x.val())
+                                            .map(crate::writer::Var::val)
```
mimoo, line 793: don't think this is a good change :o

duguorong009: Do you think that way?
So, you mean better to keep original style? like `(|x| x.val())`

kimchi/src/verifier.rs
```diff
@@ -616,7 +627,11 @@ where
                         panic!("runtime lookup table is unused in the linearization")
                     }
                     Index(t) => {
-                        use GateType::*;
+                        use GateType::{
+                            CairoClaim, CairoFlags, CairoInstruction, CairoTransition, ChaCha0,
+                            ChaCha1, ChaCha2, ChaChaFinal, CompleteAdd, EndoMul, EndoMulScalar,
+                            Generic, Lookup, Poseidon, RangeCheck0, RangeCheck1, VarBaseMul, Zero,
+                        };
```
mimoo, line 634: not a fan of that change also, I find that in general it is fine to open with a wildcard if it's an enum and right before using the enum

mimoo:
This is great!

So, going through this list I have several reactions:

1. a lot of these fixes are useful (e.g. documenting panics, removing `_` from match statements, removing wildcard imports, etc.)
2. some of these seem less useful and I'm wondering if we should keep them

I guess 2. prompts me to ask: should we enforce pedantic clippy lints in CI? I'm thinking that it's probably too much, and a better approach would be to selectively enable _some_ lints manually in CI. (Back in my previous job, we would review new clippy lints from times to times and figure out which one would be useful to run, it was neat!)

If we don't enable pedantic lints in CI, then a number of lints here are debatable. Here's some thoughts:

* are the `#[must_use]` annotations really useful? CI already enforce a clean compiler output, and the compiler will warn you if you don't use a return value already
* the new `#[allow(clippy::_)]` would not be necessary
* I find importing enums with a wildcard OK if it's right before using it personally, I find it pretty verbose to exhaustively import enum variants

duguorong009:
First, thanks for the reviews.
It helps me to understand the code style you(or team) follow.
Also, personally, I learnt much of the codebase while checking & fixing the lint warnings, especially documentation. 😄 

1. Yes, I also think we don't need pedantic lints in CI. That's too much, and let developers use up much more time.
2. About your thoughts(in case of don't enable pedantic lints in CI):
   - I also think `#[must_use]` and `#[allow(clippy::_)] should be used in limit way.
     Atm, I just tried to fix the lint warnings as much as possible.
   - I also think wildcard is okay for importing several enum or struct from other packages.
     Maybe, it is better to use the wildcard if the number of imported enum/struct is bigger than 3, I think.

duguorong009:
So, @mimoo , let me know your idea on my suggestions.
Simply, I accept your ideas, and added some more detail.
About the current pedantic lints fixture commits, do you think I should correct in the way we are talking, before merge?

mimoo:
Ah yeah, I guess if you agree with me can you remove all the ones we said were not necessary and then I can rereview it to make sure it all makes sense?

duguorong009:
okay, got it.
Will update the PR as discussed.

duguorong009:
@mimoo 
Please go for re-review the PR.

kimchi/src/circuits/expr.rs
```diff
@@ -2168,12 +2174,15 @@ pub mod constraints {
         Self: std::marker::Sized,
     {
         /// Double the value
+
```
mimoo, line 2177: why the spaces there :o ?

kimchi/src/circuits/expr.rs
```diff
@@ -2333,7 +2346,7 @@ pub fn coeff<F>(i: usize) -> E<F> {
 
 /// You can import this module like `use kimchi::circuits::expr::prologue::*` to obtain a number of handy aliases and helpers
 pub mod prologue {
-    pub use super::{coeff, constant, index, witness, witness_curr, witness_next, E};
+    pub use super::*;
```
mimoo, line 2349: we actually want to be explicit here

kimchi/src/circuits/gate.rs
```diff
@@ -43,6 +43,7 @@ impl CurrOrNext {
     /// Compute the offset corresponding to the `CurrOrNext` value.
     /// - `Curr.shift() == 0`
     /// - `Next.shift() == 1`
+
```
mimoo, line 46: I don't think it's idiomatic to put a space between a doc comment and a function :o

kimchi/src/circuits/gate.rs
```diff
@@ -190,6 +191,7 @@ impl<F: PrimeField> ToBytes for CircuitGate<F> {
 
 impl<F: PrimeField> CircuitGate<F> {
     /// this function creates "empty" circuit gate
+
```
mimoo, line 194: here as well

kimchi/src/circuits/gate.rs
```diff
@@ -200,14 +202,22 @@ impl<F: PrimeField> CircuitGate<F> {
...
         witness: &[Vec<F>; COLUMNS],
         cs: &ConstraintSystem<F>,
         public: &[F],
     ) -> Result<(), String> {
-        use GateType::*;
+        use GateType::{
+            CairoClaim, CairoFlags, CairoInstruction, CairoTransition, ChaCha0, ChaCha1, ChaCha2,
+            ChaChaFinal, CompleteAdd, EndoMul, EndoMulScalar, Generic, Lookup, Poseidon,
+            RangeCheck0, RangeCheck1, VarBaseMul, Zero,
+        };
```
mimoo, line 220: better with a wildcard

kimchi/src/circuits/gate.rs
```diff
@@ -477,6 +487,8 @@ pub mod caml {
 
 #[cfg(test)]
 mod tests {
+    use crate::circuits::wires::PERMUTS;
```
mimoo, line 490: ?

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -1,6 +1,9 @@
 use crate::{
     circuits::{
-        expr::{prologue::*, Column, ConstantExpr},
+        expr::{
+            prologue::{witness, E},
```
mimoo, line 4: btw prologue is supposed to be prelude here (I think I was mixing up the two words), and as such is supposed to be imported with a wildcard (that's what `prelude`s are for)

kimchi/src/circuits/lookup/runtime_tables.rs
```diff
@@ -1,9 +1,9 @@
 //! Runtime tables are tables (or arrays) that can be produced during proof creation.
-//! The setup has to prepare for their presence using [RuntimeTableCfg].
-//! At proving time, the prover can use [RuntimeTable] to specify the actual tables.
+//! The setup has to prepare for their presence using [`RuntimeTableCfg`].
+//! At proving time, the prover can use [`RuntimeTable`] to specify the actual tables.
 
 use crate::circuits::{
-    expr::{prologue::*, Column},
+    expr::{prologue::E, Column},
```
mimoo, line 6: same here

kimchi/src/prover.rs
```diff
@@ -135,7 +139,20 @@ where
...
+    /// Will give error if inputs(like `lookup_context.joint_lookup_table_d8`) are None.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if `lookup_context.joint_lookup_table_d8` is None.
+    #[allow(
+        clippy::needless_pass_by_value,
+        clippy::too_many_lines,
+        clippy::cast_possible_truncation
+    )]
```
mimoo, line 155: there's still some clippy stuff :o

kimchi/src/tests/framework.rs
```diff
@@ -52,37 +52,31 @@ pub(crate) struct TestFramework {
...
     }
 
-    #[must_use]
     pub(crate) fn witness(mut self, witness: [Vec<Fp>; COLUMNS]) -> Self {
         self.witness = Some(witness);
         self
     }
 
-    #[must_use]
     pub(crate) fn public_inputs(mut self, public_inputs: Vec<Fp>) -> Self {
```
mimoo, line 65: actually, now I realize that I don't think the compiler will yell at you if there isn't a `#[must_use]`, I think the compiler yells at you when you don't use a type `Result`. In this API I think it makes sense to leave the #[must_use]

tools/kimchi-visu/src/lib.rs
```diff
@@ -44,7 +44,10 @@ where
     F: PrimeField,
 {
     fn latex() -> Vec<Vec<String>> {
-        Self::constraints().iter().map(|c| c.latex_str()).collect()
+        Self::constraints()
+            .iter()
+            .map(kimchi::circuits::expr::Expr::latex_str)
```
mimoo, line 49: ^

duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/tests/framework.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/prover.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/lookup/runtime_tables.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/lookup/constraints.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/gate.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/gate.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/gate.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/expr.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/expr.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/e74530b7cf67a9cf66787be09ff530dff090630e

kimchi/src/circuits/gate.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/752/commits/a3f5b22770f98cd7bb2c6a188814eab4371d723e

duguorong009:
Thanks for the comments! @mimoo 
In fact, I pushed the quick PR with what I could do instantly.

Note: I have not removed all the clippy allow lints. I just left the original ones.
Simply, I just tried to remove the allow Iints I pushed.

duguorong009:
Finally, updated the PR as best as possible. :sweat_smile: 
Plz go for re-review. @mimoo

mimoo (approved):
woop, that was a big PR, thanks a lot for fixing all of these things :)

PR #753, opened 2022-09-19
first commit for snarky-rs

mimoo:
This introduce some amount of code almost directly translated from the OCaml counterpart. Some decisions were made to replace GADTs in some places, or the translation might not be exact in other places, and we're still missing things like a good API and handling of public output.


kimchi/src/snarky/cvar.rs
```diff
@@ -0,0 +1,516 @@
...
+            (F::one(), F::zero())
+        }
+    }
+
+    pub fn equal(&self, state: &mut RunState<F>, other: &CVar<F>) -> Boolean<F> {
+        match (self, other) {
+            (CVar::Constant(x), CVar::Constant(y)) => {
+                let res = if x == y { F::one() } else { F::zero() };
+                let cvars = vec![CVar::Constant(res)];
+                Boolean::from_cvars_unsafe(cvars, ())
```
mrmr1993, line 212: Nit: Can you add an explicit helper function `of_cvar_unsafe`, instead of using the `Typ.t` internal one?

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn from_cvars_unsafe(cvars: Vec<CVar<F>>, aux: Self::Auxiliary) -> Self;
+
+    /// Checks that the circuit variables behind this type are valid.
+    /// For some definition of valid.
+    /// For example, a Boolean snarky type would check that the field element representing it is either 0 or 1.
+    /// The function does this by adding constraints to your constraint system.
+    fn check(&self, cs: &mut RunState<F>);
+
+    /// Deserializes this type into the out-of-circuit type (along with some auxiliary data).
+    fn deserialize(&self) -> (Self::OutOfCircuit, Self::Auxiliary);
```
mrmr1993, line 41: Delete?

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Serializes the out-of-circuit type (along with some auxiliary data) into this type.
+    fn serialize(out_of_circuit: Self::OutOfCircuit, aux: Self::Auxiliary) -> Self;
```
mrmr1993, line 44: Delete?

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+};
+
+/// A snarky type is a type that can be used in a circuit.
+/// It references an equivalent "out-of-circuit" type that one can use outside of the circuit.
+/// (For example, to construct private or public inputs, or a public output, to the circuit.)
+pub trait SnarkyType<F>: Sized
+where
+    F: PrimeField,
+{
+    /// Some additional information. (TODO: add more documentation.)
```
mrmr1993, line 18: > Some 'out-of-circuit' data, which is carried as part of `Self`.
> This data isn't encoded as CVars in the circuit, since the data may be large (e.g. a sparse merkle tree), or may only be used by witness computations / for debugging.

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// The function does this by adding constraints to your constraint system.
+    fn check(&self, cs: &mut RunState<F>);
+
+    /// Deserializes this type into the out-of-circuit type (along with some auxiliary data).
+    fn deserialize(&self) -> (Self::OutOfCircuit, Self::Auxiliary);
+
+    /// Serializes the out-of-circuit type (along with some auxiliary data) into this type.
+    fn serialize(out_of_circuit: Self::OutOfCircuit, aux: Self::Auxiliary) -> Self;
+
+    /// Returns some auxiliary data (TODO: more doc).
```
mrmr1993, line 46: The 'default' value of `Self::Auxiliary`. This is passed to `of_cvars_unsafe` when we are not generating a witness, since we have no candidate value to get the `Auxiliary` from.

Note that we use an explicit value here rather than `Auxiliary: Default`, since the default value for the type may not match the default value we actually want to pass!

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    fn value_of_field_elements(x: (Vec<F>, Self::Auxiliary)) -> Self::OutOfCircuit;
+
+    //
+    // new functions that might help us with generics?
+    //
+
+    fn compute<FUNC>(
+        cs: &mut RunState<F>,
+        check: TypeCreation,
```
mrmr1993, line 60: It would be nice to call this `compute_generic` and have `compute`/`compute_unsafe` that specialize it (or equivalently remove this and have just those 2 functions). This API seems a little cumbersome for users.

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+        let values = cvars.iter().map(|cvar| g.read_var(cvar)).collect();
+        Self::value_of_field_elements((values, aux))
+    }
+}
+
+//
+// Auto traits
+//
+
+impl<F, T> SnarkyType<F> for (T, T)
```
mrmr1993, line 84: Should this implementation be `impl<F, T1, T2> SnarkyType<F> for (T1, T2)`?

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn deserialize(&self) -> (Self::OutOfCircuit, Self::Auxiliary) {
+        todo!()
+    }
+
+    fn serialize(out_of_circuit: Self::OutOfCircuit, aux: Self::Auxiliary) -> Self {
+        todo!()
+    }
+
+    fn constraint_system_auxiliary() -> Self::Auxiliary {
+        todo!()
```
mrmr1993, line 126: `(T1::constraint_system_auxiliary(), T2::constraint_system_auxiliary())`

kimchi/src/snarky/cvar.rs
```diff
@@ -0,0 +1,516 @@
...
+    fn deserialize(&self) -> (Self::OutOfCircuit, Self::Auxiliary) {
+        todo!()
+    }
+
+    fn serialize(out_of_circuit: Self::OutOfCircuit, aux: Self::Auxiliary) -> Self {
+        todo!()
+    }
+
+    fn constraint_system_auxiliary() -> Self::Auxiliary {
+        todo!()
```
mrmr1993, line 417: `()`

kimchi/src/snarky/cvar.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    fn value_to_field_elements(x: &Self::OutOfCircuit) -> (Vec<F>, Self::Auxiliary) {
+        todo!()
```
mrmr1993, line 421: `(x.clone(), ())`

kimchi/src/snarky/cvar.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    fn value_of_field_elements(x: (Vec<F>, Self::Auxiliary)) -> Self::OutOfCircuit {
+        todo!()
```
mrmr1993, line 425: `x.0`

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn deserialize(&self) -> (Self::OutOfCircuit, Self::Auxiliary) {
+        todo!()
+    }
+
+    fn serialize(out_of_circuit: Self::OutOfCircuit, aux: Self::Auxiliary) -> Self {
+        todo!()
+    }
+
+    fn constraint_system_auxiliary() -> Self::Auxiliary {
+        todo!()
```
mrmr1993, line 61: `()`

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    fn value_to_field_elements(x: &Self::OutOfCircuit) -> (Vec<F>, Self::Auxiliary) {
+        todo!()
```
mrmr1993, line 65:
```rust
if x {
  (vec![F::one()], ())
} else {
  (vec![F::zero()], ())
}
```

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    fn value_of_field_elements(x: (Vec<F>, Self::Auxiliary)) -> Self::OutOfCircuit {
+        todo!()
```
mrmr1993, line 69: x.0 != F::zero()

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    pub fn false_() -> Self {
+        Self(CVar::Constant(F::zero()))
+    }
+
+    pub fn not(&self) -> Self {
+        Self(Self::true_().0 - &self.0)
+    }
+
+    pub fn if_(&self, then_: Self, else_: Self) -> Self {
```
mrmr1993, line 89: This will need a reference to the constraint system (equv. run state)

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+        todo!()
```
mrmr1993, line 90: It would be nice to remove this and use the 'generic' `if_`. Something like:
```
pub fn if_<F: PrimeField, T: SnarkyType<F>>(cs: RunState<F>, b: Boolean<F>, then_: T, else_: T) -> T
```
which should be easy-ish to implement using `to_cvars` and `of_cvars_unsafe` (I believe there's already a version of this in the snarky repo.)

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    pub fn or(&self, other: &Self, cs: &mut RunState<F>) -> Self {
+        let both_false = self.not().and(&other.not(), cs);
+        both_false.not()
+    }
+
+    pub fn any(xs: &[&Self], cs: &mut RunState<F>) -> Self {
+        if xs.is_empty() {
+            return Self::false_(); // TODO: shouldn't we panic instead?
```
mrmr1993, line 104: Probably not if we don't want to crash the WASM interpreter.

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+        } else if xs.len() == 1 {
+            return xs[0].clone();
+        } else if xs.len() == 2 {
+            return xs[0].or(&xs[1], cs); // TODO: is this better than below?
```
mrmr1993, line 108: Usually not.

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+        let xs: Vec<_> = xs.into_iter().map(|x| &x.0).collect();
+        let sum = CVar::sum(&xs);
+        let all_zero = sum.equal(cs, &zero);
+
+        all_zero.not()
+    }
+
+    pub fn all(xs: &[Self], cs: &mut RunState<F>) -> Self {
+        if xs.is_empty() {
+            return Self::true_(); // TODO: shouldn't we panic instead?
```
mrmr1993, line 122: As above.

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+        } else if xs.len() == 1 {
+            return xs[0].clone();
+        } else if xs.len() == 2 {
+            return xs[0].and(&xs[1], cs); // TODO: is this better than below?
```
mrmr1993, line 126: Usually not.

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+            CVar::Constant(x) => Some(x == F::one()),
+            _ => None,
+        }
+    }
+
+    pub fn xor(&self, other: &Self, state: &mut RunState<F>) -> Self {
+        match (self.to_constant(), other.to_constant()) {
+            (Some(x), Some(y)) => {
+                // (var_of_value (Caml.not (Bool.equal b1 b2)))
+                todo!()
```
mrmr1993, line 147: You can remove this case and let it fall through to `(Some(true), _)` or `(Some(false), _)`, if that makes it easier.

kimchi/src/snarky/boolean.rs
```diff
@@ -0,0 +1,192 @@
...
+
+                    todo!()
+                });
+
+                let x = &self.0 + &self.0;
+                let y = &other.0;
+                let z = &self.0 + &other.0 - &res.0;
+
+                // TODO: annotation?
+                state.assert_r1cs(None, x, y.clone(), z);
```
mrmr1993, line 186: Nit: unnecessary clone

kimchi/src/snarky/checked_runner.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// This is done by calling [SnarkyType::check] on a snarky type.
+    #[default]
+    Checked,
+
+    /// This will create a new type without any constraints.
+    /// This should only be used internally when we know the type is already constrained to be valid.
+    Unsafe,
+}
+
+/// The mode in which the [RunState] is running.
```
mrmr1993, line 61: IIRC, the design of the OCaml snarky allows you to do both at once -- that is, to generate a witness while you're also building the constraint system. What do you think about enabling this here too? (Doesn't need to be now / in this PR.)

kimchi/src/snarky/checked_runner.rs
```diff
@@ -0,0 +1,424 @@
...
+    WitnessGeneration,
+}
+
+/// The state used when compiling a circuit in snarky, or used in witness generation as well.
+#[derive(Default)]
+pub struct RunState<F>
+where
+    F: PrimeField,
+{
+    /// The constraint system used to build the circuit. Why is this an option?
```
mrmr1993, line 78: > Why is this an option?
You're only building the constraint system when you're in 'circuit generation' mode.

kimchi/src/snarky/checked_runner.rs
```diff
@@ -0,0 +1,424 @@
...
+    system: Option<SnarkyConstraintSystem<F>>,
+
+    /// The public input of the circuit used in witness generation.
+    public_input: Vec<F>,
+
+    /// The private input of the circuit used in witness generation. Still not sure what that is, or why we care about this.
+    private_input: Vec<F>,
+
+    /// If this is set to false, constraints won't be added to the system. Why???
```
mrmr1993, line 87: Sometimes you want to run a circuit in a lightweight 'run this snarky code and just give me the result' form, so that you can reuse it as if it's native code.
This is most useful for testing, where you want to test that the in-circuit implementation is consistent with some native one.

kimchi/src/snarky/checked_runner.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// The number of public inputs.
+    num_public_inputs: usize,
+
+    /// A counter used to track private inputs as they're being created.
+    next_private_input: usize,
+
+    /// Indication that we're running the witness generation (as opposed to the circuit creation).
+    mode: Mode,
+
+    /// seems useless?
```
mrmr1993, line 99: Per our conversation, we may still need this for SnarkyJS, because of hacks they have there.

kimchi/src/snarky/cvar.rs
```diff
@@ -0,0 +1,516 @@
...
+    pub fn sum(vs: &[&Self]) -> Self {
+        let terms: Vec<_> = vs.into_iter().map(|v| (F::one(), (*v).clone())).collect();
+        Self::linear_combination(&terms)
+    }
+
+    pub fn mul(&self, other: &Self, label: Option<&'static str>, cs: &mut RunState<F>) -> Self {
+        match (self, other) {
+            (CVar::Constant(x), CVar::Constant(y)) => CVar::Constant(*x * y),
+
+            // TODO: this was not in the original ocaml code, but seems correct to me
```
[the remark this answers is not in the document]
mimoo: if we end up removing the `scale` thing which seems to be a remnant of R1CS it might be good to keep this for now, doesn't seem to hurt, wdyt?

kimchi/src/snarky/cvar.rs
mimoo: shouldn't that belong in the ocaml bindings?

kimchi/src/snarky/traits.rs
mimoo: thanks!

mimoo: I still don't get how "custom" data can be compatible with the API. For example, `constraint_system_auxiliary` seems to produce the data on demand.

kimchi/src/snarky/traits.rs
mimoo: oh I see, wouldn't it make sense to have two sets of functions, functions that take an auxiliary as param, and functions that don't?

kimchi/src/snarky/boolean.rs
mimoo: or at least return an error

kimchi/src/snarky/boolean.rs
mimoo: (this is taken from the OCaml code btw)

kimchi/src/snarky/boolean.rs
mimoo: necessary because we have a reference here (btw clippy warns us when clones are unnecessary :o)

kimchi/src/snarky/checked_runner.rs
mimoo: tbh I don't like all the hidden modes and ways to run snarky, I think the API should remain pretty basic: compile, and then construct witness

kimchi/src/snarky/checked_runner.rs
mimoo: thanks!

I think we should have an enum that lists all of the mode in which we can run

mimoo: btw this is actually to test the witness :o

kimchi/src/snarky/checked_runner.rs
mimoo: I'll remove it for now so I can reintroduce it more cleanly

kimchi/src/circuits/gate.rs
```diff
@@ -171,6 +171,28 @@ pub struct CircuitGate<F: PrimeField> {
     pub coeffs: Vec<F>,
 }
 
+/*
```
imeckler, line 174: Delete?

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,161 @@
...
+
+//
+// Auto traits
+//
+
+impl<F> SnarkyType<F> for ()
+where
+    F: PrimeField,
+{
+    type Auxiliary = ();
```
imeckler, line 82: Can we provide a default for this associated type? Or is that not supported

kimchi/src/snarky/traits.rs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        let (cvars, aux) = self.to_cvars();
+        let values = cvars.iter().map(|cvar| g.read_var(cvar)).collect();
+        Self::value_of_field_elements(values, aux)
+    }
+}
+
+//
+// Auto traits
+//
```
imeckler, line 76: Could be nice to give one for

```rust
impl<F, T, const N: usize> SnarkyType<F> for [T; N]
where
F: PrimeField, T: SnarkyType<F>
```

mimoo: yeah I was just experimenting here, I think we should provide default implementations for all of these!

kimchi/src/snarky/traits.rs
mimoo: oh yeah I think we can do this!

mimoo:
created smaller PRs

PR #758, opened 2022-09-21
Implement fmt::Display for ExprOps types

jspada:
* This enables direct printing of the constraints or values therein (e.g. for debugging)
* Depending on the `ArgumentEnv`, displays either text for `Expr<ConstantExpr<F>>` or actual values for `Field`
* Remove range-check use of `Wire::create` (by the way improvement)

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -8,14 +8,15 @@ use crate::{
     },
     proof::ProofEvaluations,
 };
-use ark_ff::{FftField, Field, One, PrimeField, Zero};
+use ark_ff::{FftField, Field, One, Zero};
 use ark_poly::{
     univariate::DensePolynomial, EvaluationDomain, Evaluations, Radix2EvaluationDomain as D,
 };
 use itertools::Itertools;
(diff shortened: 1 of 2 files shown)


kimchi/src/circuits/expr.rs
```diff
@@ -197,6 +198,23 @@ impl Column {
             Column::Coefficient(i) => format!("c_{{{}}}", i),
         }
     }
+
+    fn text(&self) -> String {
```
querolita, line 202: Practice will tell us if we need a more verbose version of these texts or if single letters is meaningful enough :)

querolita (approved):
Useful for debugging!

kimchi/src/circuits/expr.rs
jspada: Yeah, I just tried to keep it similar to the other versions of printing, which are language specific.  There's a tradeoff of compactness vs verbosity regarding which is clearer for debugging.

fabrizio-m (approved):
Looks good, you may want to consider padding in some cases so that the numbers will be aligned even if they have different sizes.

PR #765, opened 2022-09-26
[curves] introduce wrappers for Fp and Fq

mimoo:
This introduces wrapper types for `ark_ff::BigInteger256` as well as for our own `Fp` and `Fq`. These are supposed to add no real overhead as most functions are inlined:

```
Proof creation/proof creation (SRS size 2^15)                                                                          
                        time:   [3.3081 s 3.3659 s 3.4263 s]
                        change: [-0.7666% +2.6613% +5.8995%] (p = 0.15 > 0.05)
```

(compare this with https://github.com/o1-labs/proof-systems/pull/750)

The benefit is that we can implement ocaml traits on these.

I also implemented serde serialize/deserialize on them, but I didn't manage to remove the serde_as/serde_with stuff elsewhere. I think it's doable though.


mimoo:
looks like no overhead detected:

```
Proof creation/proof creation (SRS size 2^15)                                                                          
                        time:   [3.3081 s 3.3659 s 3.4263 s]
                        change: [-0.7666% +2.6613% +5.8995%] (p = 0.15 > 0.05)
```

(compare with https://github.com/o1-labs/proof-systems/pull/750)

mrmr1993 (approved):
Really nice!
As a sanity check, it would be good to make a Mina PR and make sure that CI is happy there

mimoo:
playing with it here: https://github.com/MinaProtocol/mina/pull/11906

it seems to compile the stubs. Not sure if mina compiles yet, and haven't touched the wasm part.

curves/src/pasta/arkworks/fields.rs
```diff
@@ -0,0 +1,329 @@
...
+                    }
+                }
+
+                //
+                // Helpers
+                //
+
+                #[ocaml_gen::func]
+                #[ocaml::func]
+                pub fn [<$name:snake _size_in_bits>]() -> ocaml::Int {
```
mrmr1993, line 84: Per my message OOB: I don't think the bindings for Mina should live here. The behaviour and API are mina-specific, but this library is supposed to be general-purpose.

mrmr1993 (changes requested):
See comment. Can you rework this so that it's only moving the OCaml types into this repo, and leave the rest on the Mina side?

mimoo:
probably we should try this approach first: https://github.com/arkworks-rs/algebra/pull/506

mimoo:
ping @mrmr1993

mimoo:
urg actually I don't think this will solve all of our problems for serde at least... Because things like `G::Scalar` are not bound by `Serialize/Deserialize`. We might have to do the same for `AffineCurve`

Issue #759, opened 2022-09-23
implement strong types for chunked evaluations


mimoo:
In https://github.com/o1-labs/proof-systems/pull/680 we removed code related to chunked evaluations: when a polynomial is too big so we have to split it into "chunks" and evaluate these chunks separately.

We used to have that due to one polynomial (the quotient polynomial) but we actually do not evaluate it anymore (thanks to [Maller's optimization](https://o1-labs.github.io/proof-systems/plonk/maller_15.html))

We decided against going with #680 because we might want to actually support a different path going forward: a smaller URS which would mean that everything is chunked.

The goal of this PR is to clean these chunked evaluations that #680 tried to remove:

* in `proof.rs::ProofEvaluations` and other similar types, `F` should become `ChunkedEvaluations<F>`.
* the code related to these chunked evaluations (I think in prover.rs or verifier.rs should be moved to methods on this ChunkedEvaluations
* there's already some Chunked types in the utils crate, so it probably is a good place to create that new type
* as long as that new type has a nice API that could be used outside of kimchi by other people, it'll be a good thing!

protip: just look at #680 to find places that are impacted by this issue

mimoo:
some more information on what would be in such a type. I think it should closely look like [ChunkedPolynomial](https://github.com/o1-labs/proof-systems/blob/master/utils/src/chunked_polynomial.rs) (in that it also tracks the size of each chunk).

we should only be able to create it by evaluating a polynomial, ideally the [`evaluate_chunks`](https://github.com/o1-labs/proof-systems/blob/master/utils/src/chunked_polynomial.rs#L19) function should return a `ChunkedPolynomial`.

I think introducing this type first would be a good idea. Then two things can be done (I can help provide more support there as well):

first: simplify the handling of chunks

<img width="601" alt="Screen Shot 2022-09-28 at 7 07 00 PM" src="https://user-images.githubusercontent.com/1316043/192922259-9962cc8f-d597-45fd-b6c0-d0f2aca89ab5.png">

second: replace all the vectors in `ProofEvaluations` and `LookupEvaluations` with this new type

mimoo:
> I tried to update the Vec<F> inside ProofEvaluations and LookupEvaluations, but not much successful.

it probably would be easier to start like this:

1. create a type called `ChunkedEvaluations { chunks: Vec<F> }` in `utils/src/chunked_evaluations.rs`
2. replace the `Vec<F>` in the structures you mentioned with that wrapper structure

[...]

PR #767, opened 2022-09-28
Update the process of absorbing `ProofEvaluations` in `FrSponge`

duguorong009:
**What's been done:**
Update the `absorb_evaluations` of `FrSponge` trait, so that `sponge` can absorb all the evaluations from proof.

**Reference:** #221


kimchi/src/plonk_sponge.rs
```diff
@@ -59,36 +59,25 @@ impl<Fr: PrimeField> FrSponge<Fr> for DefaultFrSponge<Fr, SC> {
...
-
-        if let Some(l) = e.lookup.as_ref() {
+        let ProofEvaluations {
+            w,
+            z,
+            s,
+            lookup,
+            generic_selector,
+            poseidon_selector,
+        } = ProofEvaluations::transpose(e);
```
mimoo, line 69: I think it's a better approach to use the code here https://github.com/o1-labs/proof-systems/issues/221#issuecomment-1246346523 so that we can remove this `transpose` function (which is error prone and not really elegant imo)

mimoo: btw we don't have to create an `Absorbable` trait, we can just put the logic here in this function:

```
let [zeta_evals, zeta_omega_evals] = &evals;

        // we interleave points from each evaluations,
        // it makes it easier to absorb them in the verifier circuit
        let points: Vec<F> = zeta_evals
            .iter()
            .interleave(zeta_omega_evals.iter())
            .collect();

        sponge.absorb_multiple(&points);
```

duguorong009: @mimoo 
I think this also needs the `ChunkEvaluations` implementation first.
The reason is that I think there is mismatch between the `ProofEvaluations` we think.
For example, the `ProofEvaluations` I mean is like
```

/// Polynomial evaluations contained in a `ProverProof`.
/// - **Chunked evaluations** `Field` is instantiated with vectors with a length that equals the length of the chunk
/// - **Non chunked evaluations** `Field` is instantiated with a field, so they are single-sized#[serde_as]
#[serde_as]
#[serde(bound(
    serialize = "Vec<o1_utils::serialization::SerdeAs>: serde_with::SerializeAs<Field>",
    deserialize = "Vec<o1_utils::serialization::SerdeAs>: serde_with::DeserializeAs<'de, Field>"
))]
pub struct ProofEvaluations<Field> {
    /// witness polynomials
    #[serde_as(as = "[Vec<o1_utils::serialization::SerdeAs>; COLUMNS]")]
    pub w: [Field; COLUMNS],
    /// permutation polynomial
    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
    pub z: Field,
    /// permutation polynomials
    /// (PERMUTS-1 evaluations because the last permutation is only used in commitment form)
    #[serde_as(as = "[Vec<o1_utils::serialization::SerdeAs>; PERMUTS - 1]")]
    pub s: [Field; PERMUTS - 1],
    /// lookup-related evaluations
    pub lookup: Option<LookupEvaluations<Field>>,
    /// evaluation of the generic selector polynomial
    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
    pub generic_selector: Field,
    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
    pub poseidon_selector: Field,
}
```

While yours is like
```
/// Polynomial evaluations contained in a `ProverProof`.
#[serde_as]
pub struct ProofEvaluations<F>
where
    F: Field,
{
    /// witness polynomials
    #[serde_as(as = "[o1_utils::serialization::SerdeAs; COLUMNS]")]
    pub w: [F; COLUMNS],

    /// permutation polynomial
    pub z: F,

    /// permutation polynomials
    /// (PERMUTS-1 evaluations because the last permutation is only used in commitment form)
    #[serde_as(as = "[o1_utils::serialization::SerdeAs; PERMUTS - 1]")]
    pub s: [F; PERMUTS - 1],

    /// lookup-related evaluations
    #[serde(bound = "LookupEvaluations<F>: Serialize + DeserializeOwned")]
    pub lookup: Option<LookupEvaluations<F>>,

    /// evaluation of the generic selector polynomial
    pub generic_selector: F,

    pub poseidon_selector: F,
}
```
(https://github.com/o1-labs/proof-systems/blob/501616da46b26d71114f92dc4524dbf27ee97a2d/kimchi/src/proof.rs#L44-L75)
Simply, I think the `transpose` is needed current `ProofEvaluations`, and we can introduce `iter` when the `ProofEvaluations` is updated as yours(I mean `ChunkEvaluations` introduced).
Right?

mimoo: perhaps it would be better to have a `ChunkedEvaluation` type before that change to make it simpler though :o

duguorong009: Yes, I also think that way, @mimoo .
Only when we introduce the `ChunkedEvaluations` type to the system, I can tackle the 2 issues I am working on, in best way.

duguorong009: One request, @mimoo 
Can you make time to create the PR of `ChunkedEvaluations` type? or at least the prototype that I can reference and improve?
The reason is that I am still not much familiar with whole system.
(Even though I have been reading several books(mina book, halo2 book, real-world cryptography ...), the implementation needs some more practice for me. 😄 )

mimoo: I'll try to add some more information on https://github.com/o1-labs/proof-systems/issues/759

duguorong009: Thanks so much for detailed guide! @mimoo 
I will try to take more look at the system & create the PR myself.

mimoo (approved):
nice! this looks great : )

PR #775, opened 2022-10-05
[Example] Implement bitwise 'unpacking' gate

mrmr1993:


jspada (approved):
There's one thing to fix, otherwise looks good.

PR #791, opened 2022-10-11
back to previous format

jspada:


jspada:
> wasn't this already fixed in #766 ?

It got dragged back in by another merge

PR #792, opened 2022-10-11
[easy][gate] create constructor for gate

mimoo:
This will facilitate refactors on the gate, for example https://github.com/o1-labs/proof-systems/pull/793


jspada (approved):
Looks good.   Nice idea

querolita:
Less verbose, good idea

PR #796, opened 2022-10-11
[SRS] encode bound on basefield of commitmentcurve in trait

mimoo:
Ideally we shouldn't need to add additional bounds when we use a `KimchiCurve` type. It looks like I can't write `trait KimchiCurve : CommitmentCurve<BaseField: PrimeField>` so this is another way to encode that in the trait. It seems to work.

### DIFF
--- a/circuit-construction/src/prover.rs
+++ b/circuit-construction/src/prover.rs
@@ -1,6 +1,6 @@
 use crate::writer::{Cs, GateSpec, System, Var, WitnessGenerator};
-use ark_ec::{AffineCurve, ProjectiveCurve};
-use ark_ff::{FftField, One, PrimeField, SquareRootField, Zero};
+use ark_ec::AffineCurve;
+use ark_ff::{One, PrimeField, Zero};
 use commitment_dlog::{
     commitment::{CommitmentCurve, PolyComm},
     srs::{endos, SRS},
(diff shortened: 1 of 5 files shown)


circuit-construction/src/prover.rs
```diff
@@ -12,94 +12,9 @@ use kimchi::{
...
-    type OuterMap = <Other as CommitmentCurve>::Map;
-
-    type InnerField = Fq;
-    type OuterField = Fp;
-    type Inner = Vesta;
-    type Outer = Other;
-    type InnerProj = <Vesta as AffineCurve>::Projective;
-    type OuterProj = <Other as AffineCurve>::Projective;
-}
-
```
mimoo, line 102: All of this was replaced by `KimchiCurve`, also we should create a PR to deprecate this

poly-commitment/src/commitment.rs
```diff
@@ -321,7 +321,8 @@ pub fn squeeze_challenge<
     squeeze_prechallenge(sponge).to_field(endo_r)
 }
 
-pub trait CommitmentCurve: AffineCurve {
+pub trait CommitmentCurve: AffineCurve<BaseField = Self::CommitmentField> {
+    type CommitmentField: PrimeField;
```
mimoo, line 325: this is the trick: add a new associated type that encodes the bound

poly-commitment/src/srs.rs
```diff
@@ -71,10 +67,7 @@ where
     G::of_coordinates(x, y)
 }
 
-impl<G: CommitmentCurve> SRS<G>
-where
-    G::BaseField: PrimeField,
-{
```
mimoo, line 77: now we can remove these kinds of bounds

mrmr1993 (approved):
Nice

PR #813, opened 2022-10-17
[snarky-rs] refactor compute_witness to produce witness format

mimoo:
needed for compatibility with kimchi


mrmr1993 (approved):

mrmr1993 (changes requested):
While we're re-examining the path forward for the snarky-rs project, I would prefer that we don't merge this

mimoo:
@mrmr1993 can you rereview this?

PR #815, opened 2022-10-17
[snarky-rs] allow disabling double generic gate for better debugging

mimoo:


mrmr1993 (approved):

mrmr1993 (changes requested):
While we're re-examining the path forward for the snarky-rs project, I would prefer that we don't merge this

PR #819, opened 2022-10-19
snarky poseidon

fabrizio-m:


fabrizio-m:
I have created a `Poseidon2` constraint to no break Ocaml use, but only one will be needed after refactoring that.
There is a mock function providing the sponge params util it is defined and implemented how to access them.

kimchi/src/snarky/poseidon.rs
```diff
@@ -0,0 +1,74 @@
...
+use ark_ff::PrimeField;
+use itertools::Itertools;
+use oracle::{
+    constants::PlonkSpongeConstantsKimchi, permutation::full_round,
+    poseidon::ArithmeticSpongeParams,
+};
+use std::iter::successors;
+
+const ROUNDS: usize = 55;
+const ROUNDS_PER_ROW: usize = 5;
```
fabrizio-m, line 16: I just hardcoded the constants for no particular reasons, any ideas about how to handle them?

mimoo: Screen Shot 2022-10-24 at 10 27 56 AM

polynomials/poseidon.rs should have these already, otherwise maybe a good idea to place them there?

kimchi/src/snarky/checked_runner.rs
```diff
@@ -404,4 +410,7 @@ where
             panic!("woot");
         }
     }
+    pub fn poseidon(&mut self, preimage: (CVar<F>, CVar<F>)) -> (CVar<F>, CVar<F>) {
+        super::poseidon::poseidon(self, preimage)
+    }
```
mimoo, line 415: looks good, but please add a space between functions, and documentation as well :o

mimoo: I'm guessing `cargo fmt` is going to yell at you anyway?

kimchi/src/snarky/traits.rs
```diff
@@ -159,3 +159,31 @@ where
         (out1, out2)
     }
 }
+
+impl<F: PrimeField, const T: usize> SnarkyType<F> for [CVar<F>; T] {
```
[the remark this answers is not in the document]
mimoo: I'm wondering if we can write something like this as well:

```rust
impl<F, T, const SIZE> SnarkyType<F> for [T; SIZE] 
where
  F: PrimeField,
  T: SnarkyType,
  SIZE: usize
```

kimchi/src/snarky/constraint_system.rs
```diff
@@ -1190,6 +1196,34 @@ impl<Field: PrimeField> SnarkyConstraintSystem<Field> {
...
                 self.add_row(vars, GateType::Zero, vec![]);
             }
+            KimchiConstraint::Poseidon2 { states, last } => {
+                states
+                    .map(|round| round.map(|x| self.reduce_to_var(x)))
+                    .into_iter()
+                    .zip({ self.constants.poseidon.round_constants.clone() }.into_iter())
+                    .chunks(5)
+                    .into_iter()
+                    .for_each(|rounds| {
```
mimoo, line 1206: I think this is pretty dense and hard to read, may I suggest spreading the logic over more lines of code to make it more readable:

```rust
let states = states.map(|round| round.map(|v| self.reduce_to_var(v));

let states_and_round_constants = states.zip(&self.constants.poseidon.round_constants);

for (round_state, round_constant) in states_and_round_constants.chunks(5) {
  // ...
}
```

```

File: kimchi/src/snarky/poseidon.rs [hunk 5]
```diff
@@ -0,0 +1,74 @@
...
+    runner.add_constraint(constraint, Some("Poseidon"));
+    hash
+}
+
+fn round<F: PrimeField>(
+    elements: &[CVar<F>; 3],
+    runner: &mut RunState<F>,
+    round: usize,
+) -> [CVar<F>; 3] {
+    let params = params();
```
mimoo, line 61: I think you should pass the params as argument to `round()`, since `poseidon()` has access to them

kimchi/src/snarky/constraint_system.rs
```diff
@@ -1190,6 +1196,34 @@ impl<Field: PrimeField> SnarkyConstraintSystem<Field> {
                 ];
                 self.add_row(vars, GateType::Zero, vec![]);
             }
+            KimchiConstraint::Poseidon2 { states, last } => {
```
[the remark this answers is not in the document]
fabrizio-m: Just temporal until we replace the uses, then we can delete and rename.

kimchi/src/snarky/traits.rs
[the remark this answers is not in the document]
fabrizio-m: I realized it just after finishing the implementation, it can be done, I'll do it later in its own PR.

kimchi/src/snarky/poseidon.rs
```diff
@@ -0,0 +1,74 @@
...
+    runner.compute(loc!(), |env| {
```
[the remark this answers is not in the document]
fabrizio-m: Yeah, I completely forgot to propagate it.

kimchi/src/snarky/poseidon.rs
fabrizio-m: Do you mean through `RunState`? I will add some method to expose the constants then.

kimchi/src/snarky/poseidon.rs
```diff
@@ -0,0 +1,74 @@
...
+        let state = elements.clone().map(|var| env.read_var(&var));
+        //remove
+        let mut state = state.to_vec();
+        full_round::<F, PlonkSpongeConstantsKimchi>(&params, &mut state, round);
+        state.try_into().unwrap()
+    })
+}
+
+//to be replaced
+fn params<F: PrimeField>() -> ArithmeticSpongeParams<F> {
```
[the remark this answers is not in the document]
fabrizio-m: It provides the params to make it compile, but I had not yet decided from where to get the params from.

kimchi/src/snarky/checked_runner.rs
fabrizio-m: I thought that too, but it seems that `cargo fmt` accepts either 0 or 1 spaces. I'll add it.

kimchi/src/snarky/poseidon.rs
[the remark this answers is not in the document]
fabrizio-m: Propagated

kimchi/src/snarky/poseidon.rs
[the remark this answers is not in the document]
fabrizio-m: removed

kimchi/src/snarky/poseidon.rs
fabrizio-m: done

kimchi/src/snarky/constraint_system.rs
fabrizio-m: I thought that this approach would be easier to read than the manual indexing one, readability may be more subjective than I thought.
I'm not sure I understand the suggestion, you say to spread it over more lines to make it less dense, but your example has less lines, do you mean having more statements and intermediate variables?

kimchi/src/snarky/poseidon.rs
fabrizio-m: I was no sure about using those, I see that there are some other params with for example more rounds, do we have to support those?
I think I will use those from `PlonkSpongeConstantsKimchi` for now and later we can change it or use const generics.

kimchi/src/snarky/constraint_system.rs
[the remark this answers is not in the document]
mimoo: I guess my question was more, how do they differ? I'll give it a better read later, but from a glance it looks like they were the same (except that last is given separately)

kimchi/src/snarky/constraint_system.rs
mimoo: I find that style easier to read yeah, but I agree that it's suggestive. If you don't like my suggestion feel free to ignore it ;)

kimchi/src/snarky/constraint_system.rs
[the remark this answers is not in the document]
fabrizio-m: `Poseidon` takes the rounds in the order they were created, makes groups of 5 and rotates the last 4 of each 5. `Poseidon2` takes the rounds already in the order that the constraint system expects. Also some the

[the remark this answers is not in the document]
mimoo: cool thanks :) let me know when this is not a draft anymore and I'll take a final look!

kimchi/src/snarky/constraint_system.rs
fabrizio-m: I think it will be better to discuss style in a more comprehensive way and including more people, and we can maybe make some guidelines.
I will remove draft and we can revisit and change it later.

kimchi/src/snarky/constraint_system.rs
```diff
@@ -158,6 +160,10 @@ pub enum KimchiConstraint<Var, Field> {
     Poseidon {
         state: Vec<Vec<Var>>,
     },
+    Poseidon2 {
+        states: [[Var; 3]; 55],
+        last: [Var; 3],
```
mimoo, line 165: these should be constants hardcoded somewhere : o

I see that they're in a generic trait...

```rust
impl SpongeConstants for PlonkSpongeConstantsKimchi {
    const SPONGE_RATE: usize = 2;
    const PERM_ROUNDS_FULL: usize = 55;
    const PERM_ROUNDS_PARTIAL: usize = 0;
    const PERM_HALF_ROUNDS_FULL: usize = 0;
    const PERM_SBOX: u32 = 7;
    const PERM_FULL_MDS: bool = true;
    const PERM_INITIAL_ARK: bool = false;
}
```

maybe it's a good reason to remove `SpongeConstants` and hardcode everything. For things that still need to change for the legacy sponge, we can pass these values as arguments when we initialize the sponge maybe? But it seems overkill to have a trait for that now if we're only going to implement one thing in snarky

kimchi/src/snarky/poseidon.rs
```diff
@@ -0,0 +1,72 @@
...
+        let constraint = Constraint::KimchiConstraint(KimchiConstraint::Poseidon2 { states, last });
+        (constraint, hash)
+    };
+    runner.add_constraint(constraint, Some("Poseidon"));
+    hash
+}
+
+fn round<F: PrimeField>(
+    loc: String,
+    elements: &[CVar<F>; 3],
```
mimoo, line 60: it'd be nice if all these constants would be hardcoded somewhere : o

kimchi/src/snarky/poseidon.rs
```diff
@@ -0,0 +1,72 @@
...
+    runner: &mut RunState<F>,
+    round: usize,
+    params: &ArithmeticSpongeParams<F>,
+) -> [CVar<F>; 3] {
+    runner.compute(loc, |env| {
+        let state = elements.clone().map(|var| env.read_var(&var));
+        //remove
```
mimoo, line 67: ?

kimchi/src/snarky/constraint_system.rs
```diff
@@ -1190,6 +1196,34 @@ impl<Field: PrimeField> SnarkyConstraintSystem<Field> {
...
+                            })
+                            .unzip();
+                        self.add_row(vars, GateType::Poseidon, coeffs);
+                    });
+                let last = last
+                    .map(|x| self.reduce_to_var(x))
+                    .map(Some)
+                    .into_iter()
+                    .chain(repeat(None))
+                    .take(15)
```
mimoo, line 1223: IIRC you don't have to do this, you can just have a vec of size 3

mimoo:
Nice! This looks great :) I left you a few more comments. But I have a bigger question: should this be merged on top of snarky or the snarky2 branch? I'm trending towards [snarky2](https://github.com/o1-labs/proof-systems/compare/mimoo/snarky...mimoo/snarky2) which contains the latest including the public output. I think that branch has some issues, for example I think you can't actually change the values of the private input after compilation... but at least we'll be able to write a test to see if the poseidon gate works as intended :o also if we merge this on snarky, then snarky2 branch doesn't extend snarky anymore

mimoo (approved):
(approving in case you disagree with my suggestion and want to merge this)

kimchi/src/snarky/constraint_system.rs
mrmr1993: > maybe it's a good reason to remove `SpongeConstants` and hardcode everything.

I would prefer if we didn't. Is it so problematic to have
```rust
    Poseidon2 {
        states:
            [
                [Var; PlonkSpongeConstantsKimchi::SPONGE_WIDTH];
                PlonkSpongeConstantsKimchi::PERM_ROUNDS_FULL
            ],
        last: [Var; PlonkSpongeConstantsKimchi::SPONGE_WIDTH],
    }
```
or similar?

mimoo: it's not a problem but then you need to carry that type parameter in a bunch of places, which isn't ideal IMO (I have low tolerance for generics I guess).

BTW I was suggesting this for a future refactor, not for this PR :o

kimchi/src/snarky/constraint_system.rs
fabrizio-m: Indeed.

kimchi/src/snarky/constraint_system.rs
fabrizio-m: I was doing as Matthew's example, I just forgot to replace there. I'll change it now.
For the refactor, in the future we may be able to change the trait to a constant like:
```rust
pub struct SpongeConstants2 {
    sponge_capacity: usize,
    sponge_width: usize,
    sponge_rate: usize,
    perm_rounds_full: usize,
    perm_rounds_partial: usize,
    perm_half_rounds_full: usize,
    perm_sbox: u32,
    perm_full_mds: bool,
    perm_initial_ark: bool,
}
const PlonkSpongeConstantsKimchi: SpongeConstants2 = SpongeConstants2 {
    sponge_capacity: 1,
    sponge_width: 3,
    sponge_rate: 2,
    perm_rounds_full: 55,
    perm_rounds_partial: 0,
    perm_half_rounds_full: 0,
    perm_sbox: 7,
    perm_full_mds: true,
    perm_initial_ark: false,
};
```

kimchi/src/snarky/poseidon.rs
fabrizio-m: I'll change to SPONGE_WIDTH

kimchi/src/snarky/poseidon.rs
fabrizio-m: I had to do those conversions because `full_round` takes only a `&mut Vec` and was unsure to let it that way or changing that.
I added `full_round2`, you can check if it's worth it.

oracle/src/permutation.rs
```diff
@@ -41,6 +41,22 @@ pub fn full_round<F: Field, SC: SpongeConstants>(
         state[i].add_assign(x);
     }
 }
+//same as full round but works better with arrays
```
mimoo, line 44: nit: spacing and `///`

fabrizio-m:
I think it should be ready now, I will try to merge into snarky2.

PR #820, opened 2022-10-20
Evaluate coefficients columns

mrmr1993:
This PR is an experiment for evaluating coefficients columns while proving.


kimchi/src/prover.rs
```diff
@@ -1197,6 +1210,14 @@ where
                 .map(|(w, c)| (w, None, c.blinders.clone()))
                 .collect::<Vec<_>>(),
         );
+        polynomials.extend(
+            index
+                .cs
+                .coefficientsm
+                .iter()
+                .map(|coefficientm| (coefficientm, None, non_hiding(1)))
```
mimoo, line 1218: why not `fixed_hiding(1)`?

kimchi/src/verifier.rs
```diff
@@ -388,6 +388,19 @@ where
                     })
                     .collect::<Vec<_>>(),
             );
+            es.extend(
+                (0..COLUMNS)
```
mimoo, line 392: maybe we should have a `COEFFICIENTS`, since in theory we don't need one coefficient per register/column

mimoo (approved):

kimchi/src/verifier.rs
mrmr1993: I agree. Lets open an issue for it, seems easy to pick up :)

jspada (approved):
Looks good to me.  Can't wait to try this out.

kimchi/src/circuits/polynomials/chacha.rs
```diff
@@ -549,6 +549,7 @@ mod tests {
             w: array::from_fn(|_| F::rand(rng)),
             z: F::rand(rng),
             s: array::from_fn(|_| F::rand(rng)),
+            coefficients: array::from_fn(|_| F::rand(rng)),
```
querolita, line 552: Just a stupid question, why do we fill them up with random values when they are not used instead of zeros (or ones, or other value)?

querolita: Oh, I am looking at the definition of coefficients inside ProofEvaluations now. Perhaps this is for zero knowledge, so that the proof is simulateable?

kimchi/src/proof.rs
```diff
@@ -249,6 +254,9 @@ impl<F: FftField> ProofEvaluations<Vec<F>> {
     pub fn combine(&self, pt: F) -> ProofEvaluations<F> {
         ProofEvaluations::<F> {
             s: array::from_fn(|i| DensePolynomial::eval_polynomial(&self.s[i], pt)),
+            coefficients: array::from_fn(|i| {
```
querolita, line 257: I thought `coefficients` were already an evaluation of the 15 coefficient polynomials at a challenge point. Thus, I am not sure I understand why `combine` uses those field elements to compute `c0·pt^0 + c1·pt + c2·pt^2 + ... + c14·pt^14`. Then the question is more like: what do these 15 field elements represent?

kimchi/src/prover.rs
```diff
@@ -940,6 +940,11 @@ where
                         .to_chunked_polynomial(index.max_poly_size)
                         .evaluate_chunks(zeta)
                 }),
+                coefficients: array::from_fn(|i| {
```
querolita, line 943: Ah ok so indeed the coefficients are the evaluation on zeta (rows) for the monomial-formed polynomial representing the coefficients (meaning, the `coefficientsm` field of the constraint system). So basically the `coefficients` field inside the ProofEvaluations refers to a "doubly" evaluated version of the coefficients, over the zeta first, and then the outcome over the `pt` above, right?

kimchi/src/prover.rs
```diff
@@ -970,6 +975,11 @@ where
                         .to_chunked_polynomial(index.max_poly_size)
                         .evaluate_chunks(zeta_omega)
                 }),
+                coefficients: array::from_fn(|i| {
+                    index.cs.coefficientsm[i]
+                        .to_chunked_polynomial(index.max_poly_size)
```
querolita, line 980: Ah good, and then the same but for zeta_omega (next row)

kimchi/src/circuits/polynomials/chacha.rs
mimoo: my guess: because we do this for the other fields. We don't have a great strategy for testing anyway, we should implement a trait like this https://docs.rs/proptest/latest/proptest/arbitrary/trait.Arbitrary.html when we want random values

kimchi/src/proof.rs
mimoo: (this is for when we split polynomials if I remember correctly, which we don't use atm)

check "coefficients" under the spec here: https://o1-labs.github.io/proof-systems/specs/kimchi.html#tables-used-to-describe-a-circuit

PR #821, opened 2022-10-21
Keccak rotation gate

querolita:
This work in progress pull request is addressing 64-bit word rotation for Keccak. Multiple paths have been tried out:

 * [x] Vanishree's design of 4 gate types: rot by 1 bit, by 2, by 3, and by multiple of 4
         - Issue: one single rotation implied 6 rows, and the weights for the decomposition needed a number of generic gates at the beginning to check correctness.
 * [x] Optimization of Vanishree's design with 2 gate types: rot by 1 or 2 or 3 bits, and by multiple of 4
         - Same issues but 2 less circuit gates
 * [x] Decomposition into top and bottom parts:
         - This one required only one gate with one row, but I didn't know how to range check the lengths of each side.
 * [x] RangeCheck0 layout:
         - Followed the approach above but splitting top and bottom into 4 limbs of 12 bits and 8 crumbs of 2 bits. But then I needed many constants to have access to, basically to correctly use the exponents for decomposition. 
          * The first workaround needed the next row to store those parameters. But I was not able to use them as a power of two because we cannot do powers of `T` type. Also, these witness values couldn't be used as indices for loops for the same reason. So I had the problem of generalising a gate with publicly checkable parameters, but customizable depending on their values. Some of these parameters could be checked thanks to CNF, checking the correctness of the lengths of each fragment, the number of limbs and crumbs in each half, etc. But still, the loop problem remained.
          * If we stored these parameters directly as powers in the coefficients, we could have used them for the decomposition for the exponents, but still not to index in loops or sums.
          * Another alternative for these parameters instead of coefficients is [Matthew's idea](https://github.com/o1-labs/proof-systems/pull/820) to have access to constant polynomials that would be interpolations (as a substitute of lookup tables, because in that row we would already have had looked up 4 times to check the limbs). Nonetheless, the indexing problem remains.
          * The last route I took for this approach had to do with flags. Instead of rotation value, we would store in the coefficients flags to turn on and off certain limbs and crumbs from the most significant half (0, 1) and if any was broken then it would have a negative value between -1 and -11. The problem of this approach was that the degree of the constraints were blown up, like 12 or more. So no good.
 * [x] After meeting with Joseph on 20th Oct, he suggested a swap option with xors. 
          - The problem was that it seemed to need more rows than we wanted, or require too many plookups per row.
 * [x] The (almost) final idea, which I think is the best so far and also has a cool name `Kemuro`, derived from Joseph's proposal on 21st Oct, revisits the approach of the decomposition into top and bottom parts:
          - The idea is that we had to decompose the word into a half that exceeds and goes back to the beginning, and a part that stays and is shifted to the left. 
          - So far, the same as in the other idea. The cool thing about this approach is that it shows that it is not necessary no have limbs and crumbs to check the lengths of the components. How so? Inspiration from FFMul:
          * Rotation can be decomposed into two operations: a shift to the left, and addition of the excess.
          * Thus, we just have to express the word as in FFMul but with no limbs, because these are small elements, by providing witnesses for the quotient and remainder as follows: `word * 2^rot = excess * 2^64 + shifted`, and then the rotated word is nothing but `rotated = shifted + excess`. 
          * The good thing is this approach can fit up to 3 rotations per row, provided that we have range checks for word and rotated.
          * Since Keccak rotations are performed on 25 values (and one of them is a rotation by 0 bits and we can skip it), then we can represent the whole rotation gadget with just 8 KeccakRot rows and 24 RangeCheck0 rows, meaning 32 rows 😄 
          * And why this has no problems with loops nor exponents? Because the maths do it for you, no need to worry manually! all you need is 2^rot for each of the 3 words, stored as coefficients of the gate
 * [x] On Monday 24th and after talking to Matthew, we realised that the previous approach was very efficient but not sound. This is because a malicious prover could create bad witnesses for the excess and shifted parts whose addition is a 64-bit word but wrong one. Thus, we need to range constrain `excess < 2^rot` and `shifted < 2^64`. The resulting (and current) design for rotation of one word consists on one `KeccakRot` gate followed by one `RangeCheck0` gate. 
  * [ ] It now works with coefficients carrying the `2^rot` value as part of the circuit.


kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -130,6 +120,145 @@ where
...
+        let shifted = env.witness_next(0);
+        let two_to_rot = env.coeff(0);
+
+        // Obtains the following checks:
+        // word * 2^{rot} = (excess * 2^64 + shifted)
+        // rotated = shifted + excess
+        constraints.push(
+            word * two_to_rot.clone() - (excess.clone() * two_to_64.clone() + shifted.clone()),
+        );
+        constraints.push(rotated - (shifted + excess.clone()));
```
mrmr1993, line 237: Shouldn't this be `rotated = shifted * 2^{64-rot} + excess`?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -130,6 +120,145 @@ where
...
+
+        // Compute the bound from the crumbs and limbs
+        let mut power_of_2 = T::one();
+        let mut bound = T::zero();
+
+        // Sum 2-bit limbs
+        for i in (7..COLUMNS).rev() {
+            bound += power_of_2.clone() * env.witness_curr(i);
```
mrmr1993, line 245: Don't we need to check that these limbs are 2 bits in the constraint?

querolita: sure, that's done in rows 219-221, meaning -> `crumb(&env.witness_curr(i))`, these corresponds to the first 8 constraints of the gate

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: It could, but the way this shifted is defined is to be such that: `word * 2^{rot} = excess * 2^{64} + shifted`, sort of like in foreign field multiplication with quotient and remainders, meaning that this shifted already contains the `2^{64-rot}` term.

querolita: We could do the other way round if we change the witness generation code accordingly. But since the original idea of this variant of the gate was precisely FFMul, we kept the _quotient & remainder_ apprach.

book/src/specs/kimchi.md
```diff
@@ -1398,6 +1388,86 @@ Kimchi framework that dictated those approaches.
 * only 4 lookups per row
 * only first 7 columns are available to the permutation polynomial
 
+##### `KeccakRot` - Constraints for rotation of 64-bit words
```
mrmr1993, line 1391: Since this gate is a general-purpose rotation gate, should we call it `Rot` instead? It will be useful for SHA256 and other hash functions, not just keccak!

book/src/specs/kimchi.md
```diff
@@ -1661,6 +1731,10 @@ pub struct VerifierIndex<G: KimchiCurve> {
     /// Foreign field modulus
     pub foreign_field_modulus: Option<BigUint>,
 
+    /// Keccak rotation table
+    #[serde_as(as = "Option<[[o1_utils::serialization::SerdeAs; 5]; 5]>")]
+    pub keccak_rotation_table: Option<[[G::ScalarField; 5]; 5]>,
```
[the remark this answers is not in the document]
querolita: It is indeed something related to the keccak gadget only. But since it is part of the definition of the algorithm, I was not sure if this needed to be somewhere else more "accessible" for the verifier to check or something.

book/src/specs/kimchi.md
querolita: Ok, so basically the same as in XOR last time. Nice, I will have it as an independent gate and inside the Keccak module only the strictly Keccak operations -> meaning the sequence of gates to build the gadget etcetera

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,590 @@
...
+/// Get combined constraints for a given rotation circuit gate type
+pub fn circuit_gate_constraints<F: PrimeField>(typ: GateType, alphas: &Alphas<F>) -> E<F> {
+    match typ {
+        GateType::Rot64 => Rot64::combined_constraints(alphas),
+        _ => panic!("invalid gate type"),
+    }
+}
+
+/// Get the combined constraints for all rotation circuit gate types
+pub fn combined_constraints<F: PrimeField>(alphas: &Alphas<F>) -> E<F> {
```
mrmr1993, line 356: Nit: this seems rather redundant, given that we have `Rot64::combined_constraints` already

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,590 @@
...
+    })
+}
+
+/// Get vector of rotation circuit gate types
+pub fn circuit_gates() -> [GateType; GATE_COUNT] {
+    [GateType::Rot64]
+}
+
+/// Number of constraints for a given rotation gate type
+pub fn circuit_gate_constraint_count<F: PrimeField>(typ: GateType) -> u32 {
```
mrmr1993, line 340: Likewise

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,590 @@
...
+    match typ {
+        GateType::Rot64 => Rot64::<F>::CONSTRAINTS,
+        _ => panic!("invalid gate type"),
+    }
+}
+
+/// Get combined constraints for a given rotation circuit gate type
+pub fn circuit_gate_constraints<F: PrimeField>(typ: GateType, alphas: &Alphas<F>) -> E<F> {
```
mrmr1993, line 348: Likewise

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,590 @@
...
+        // TODO:
+        // If we ever want to make this gate more generic, the power of two for the length
+        // could be a coefficient of the gate instead of a fixed value in the constraints.
+        let two_to_64 = T::from(2u64).pow(64);
+
+        let word = env.witness_curr(0);
+        let rotated = env.witness_curr(1);
+        let excess = env.witness_curr(2);
+        let shifted = env.witness_next(0);
+        let two_to_rot = env.coeff(0);
```
mrmr1993, line 469: Does this mean that we need to merge https://github.com/o1-labs/proof-systems/pull/820 first?

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,590 @@
...
+    let aggreg8 = aggreg_coeffs.evaluate_over_domain_by_ref(cs.domain.d8);
+
+    Ok(LookupEnvironmentData {
+        aggreg8,
+        sorted8,
+        joint_lookup_table_d8,
+    })
+}
+
+/// Get vector of rotation circuit gate types
```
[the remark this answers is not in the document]
querolita: Agree

kimchi/src/circuits/polynomials/rot.rs
querolita: I was just assuming that the coefficient of the gate carries 2^rot already as a known parameter. In that sense, I don't think 820 is needed for this, but when it is merged we will be able to use it instead.

mrmr1993: Won't this fail the linearization check if `coeffs(0)` isn't evaluated?

querolita: OOOH!? Ha, I was turning mad looking for the problem with the linearization! I thought doing this was enough: 
```    
pub fn create_rot64(new_row: usize, rot: u32) -> (usize, Vec<Self>) {
        let gates = vec![
            CircuitGate {
                typ: GateType::Rot64,
                wires: Wire::new(new_row),
                coeffs: vec![F::from(2u64.pow(rot))],
            },
            CircuitGate {
                typ: GateType::RangeCheck0,
                wires: Wire::new(new_row + 1),
                coeffs: vec![],
            },
        ];
        (new_row + gates.len(), gates)
    }
```

querolita: meaning, that the Rot64 gate does get initialized coefficient at the time of creation...

querolita: How does the generic gate or poseidon deal with coefficients then?

kimchi/src/prover.rs
```diff
@@ -633,7 +635,7 @@ where
 
             if let Some(selector) = index.cs.foreign_field_add_selector_poly.as_ref() {
                 index_evals.extend(
-                    foreign_field_add::gadget::circuit_gates()
+                    [GateType::ForeignFieldAdd]
```
mrmr1993, line 638: Replace with a `push` (although really, this PR shouldn't be touching any other gates at all..)

kimchi/src/prover_index.rs
```diff
@@ -84,8 +84,10 @@ impl<G: KimchiCurve> ProverIndex<G> {
                 .map(|lcs| &lcs.configuration),
             cs.foreign_field_add_selector_poly.is_some(),
             cs.xor_selector_poly.is_some(),
+            cs.rot_selector_poly.is_some(),
             true,
         );
+        println!("expr linearization");
```
mrmr1993, line 90: Delete

kimchi/src/tests/rot.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    // Temporary workaround for lookup-table/domain-size issue
+    for _ in 0..(1 << 13) {
+        gates.push(CircuitGate::zero(Wire::for_row(next_row)));
+        next_row += 1;
+    }
+
+    ConstraintSystem::create(gates).build().unwrap()
+}
+/* TODO: STILL DOES NOT WORK WITH COEFFICIENTS
```
mrmr1993, line 25: Merge `master`, this should now work.

kimchi/src/circuits/expr.rs
```diff
@@ -1832,6 +1832,36 @@ impl<F: Neg<Output = F> + Clone + One + Zero + PartialEq> Expr<F> {
                         // without calling remove.
                     }
                 }
+            } else if unevaluated.len() == 2
```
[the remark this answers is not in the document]
querolita: Oh, I hadn't realised that was still there. That was just an experiment before the coefficients update, but forgot to remove it

kimchi/src/prover.rs
querolita: I guess I started modifying these because I received some comments suggesting to remove these methods when gadgets had one single gate type (as opposed to rangecheck), and then included the changes in the same branch.

kimchi/src/prover.rs
```diff
@@ -786,13 +787,23 @@ where
...
                     check_constraint!(index, xor);
                 }
             }
 
+            // rot
+            {
+                if index.cs.rot_selector_poly.is_some() {
+                    let rot = Rot64::combined_constraints(&all_alphas).evaluations(&env);
+                    assert_eq!(rot.domain().size, t4.domain().size);
+                    t4 += &rot;
```
mrmr1993, line 802: This should be  `t8`, and the assertion above is incorrect.

querolita: Oh thanks, I was not really sure what was the domain being used for each gate... For example chacha uses t4 and range-check uses t8. The one for `xor` is fine though? meaning t4. Is this something that has something to do with the number of rows each type requires?

mrmr1993: It has to do with the degree of the polynomial. This has degree > 4, so it needs to be evaluated over the `d8` domain instead of the `d4` one.

querolita: Oh wow... You mean including the selector polynomial multiplication, right? Thanks for this explanation, it will be the most useful!!

kimchi/src/circuits/witness/sum_copy_bits_cell.rs
```diff
@@ -0,0 +1,40 @@
+use ark_ff::Field;
+use o1_utils::FieldHelpers;
+
+use super::{variables::Variables, WitnessCell};
+use crate::circuits::polynomial::COLUMNS;
+
+/// Witness cell copied from bits of another witness cell
```
jspada, line 7: This comment is not true.  Suggest renaming to `ConstantBits` cell and updating comments & doc accordingly

kimchi/src/circuits/witness/sum_copy_bits_cell.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct SumCopyBitsCell<F: Field> {
+    row: usize,
+    col: usize,
+    start: usize, // inclusive
+    end: usize,   // exclusive
+    sum: F,       // value added to the (row,col) cell
+}
+
+impl<F: Field> SumCopyBitsCell<F> {
+    /// Create witness cell copied from bits [start, end) of the witness cell at position (row, col)
```
jspada, line 17: Same as above.

kimchi/src/linearization.rs
```diff
@@ -68,6 +68,10 @@ pub fn constraints_expr<F: PrimeField + SquareRootField>(
         expr += xor::Xor16::combined_constraints(&powers_of_alpha);
     }
 
+    if rot {
+        expr += rot::Rot64::combined_constraints(&powers_of_alpha);
```
jspada, line 72: I would stick with "common interface pattern" like in range-check and foreign_field_mul, so that one day we can more easily update the code to use a Gadget/gate interface and all this code could be generic, i.e. do the same thing for each switched on gate.  There's no reason to hard code and duplicate all this code really.

kimchi/src/linearization.rs
```diff
@@ -117,6 +121,11 @@ pub fn linearization_columns<F: FftField + SquareRootField>(
         h.insert(Witness(i));
     }
 
+    // the coefficient polynomials
+    for i in 0..COLUMNS {
```
jspada, line 125: Cool.  Part of this PR?

kimchi/src/prover.rs
```diff
@@ -632,18 +634,17 @@ where
             }
 
             if let Some(selector) = index.cs.foreign_field_add_selector_poly.as_ref() {
-                index_evals.extend(
-                    foreign_field_add::gadget::circuit_gates()
-                        .iter()
-                        .enumerate()
-                        .map(|(_, gate_type)| (*gate_type, &selector.eval8)),
-                );
+                index_evals.insert(GateType::ForeignFieldAdd, &selector.eval8);
```
jspada, line 637: Same ... common interface idea.

kimchi/src/prover.rs
```diff
@@ -775,8 +776,8 @@ where
             // foreign field addition
             {
                 if index.cs.foreign_field_add_selector_poly.is_some() {
-                    let ffadd = foreign_field_add::gadget::combined_constraints(&all_alphas)
-                        .evaluations(&env);
+                    let ffadd =
+                        ForeignFieldAdd::combined_constraints(&all_alphas).evaluations(&env);
```
jspada, line 780: same common interface idea...

kimchi/src/tests/keccak.rs
```diff
@@ -0,0 +1,60 @@
...
+    gate::CircuitGate,
+    polynomials::keccak::{self, ROT_TAB},
+    wires::Wire,
+};
+use ark_ec::AffineCurve;
+use mina_curves::pasta::{Fp, Pallas, Vesta};
+use rand::Rng;
+
+//use super::framework::TestFramework;
+type PallasField = <Pallas as AffineCurve>::BaseField;
```
jspada, line 14: Please run tests on both curves 👍🏼

kimchi/src/tests/keccak.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn create_test_constraint_system() -> ConstraintSystem<Fp> {
+    let (mut next_row, mut gates) = { CircuitGate::<Fp>::create_keccak(0) };
+
+    // Temporary workaround for lookup-table/domain-size issue
```
jspada, line 19: I wish we don't need this anymore.

kimchi/src/tests/keccak.rs
```diff
@@ -0,0 +1,60 @@
...
+        gates.push(CircuitGate::zero(Wire::for_row(next_row)));
+        next_row += 1;
+    }
+
+    ConstraintSystem::create(gates).build().unwrap()
+}
+
+#[test]
+// Test that all of the offsets in the rotation table work fine
+fn test_keccak_table() {
```
jspada, line 30: Please add negative tests so we know the constraints actually trigger

kimchi/src/tests/rot.rs
```diff
@@ -0,0 +1,86 @@
...
+    polynomials::rot::{self, LEFT, RIGHT},
+    wires::Wire,
+};
+use ark_ec::AffineCurve;
+use mina_curves::pasta::{Fp, Pallas, Vesta};
+use rand::Rng;
+
+use super::framework::TestFramework;
+
+type PallasField = <Pallas as AffineCurve>::BaseField;
```
jspada, line 13: Please add tests on Vesta too.

kimchi/src/tests/rot.rs
```diff
@@ -0,0 +1,86 @@
...
+    let (mut next_row, mut gates) = CircuitGate::<Fp>::create_rot(0, rot, LEFT);
+
+    // Temporary workaround for lookup-table/domain-size issue
+    for _ in 0..(1 << 13) {
+        gates.push(CircuitGate::zero(Wire::for_row(next_row)));
+        next_row += 1;
+    }
+
+    // Create input
+    let word = rand::thread_rng().gen_range(0..2u128.pow(64)) as u64;
```
jspada, line 40: Make deterministic

kimchi/src/tests/rot.rs
```diff
@@ -0,0 +1,86 @@
...
+    TestFramework::default()
+        .gates(gates)
+        .witness(witness)
+        .setup()
+        .prove_and_verify();
+}
+
+#[test]
+// End-to-end test
+fn test_prove_and_verify() {
```
jspada, line 54: Please make this work on both curves

kimchi/src/tests/rot.rs
```diff
@@ -0,0 +1,86 @@
...
+    prove_and_verify();
+}
+
+fn test_rot(word: u64, rot: u32, side: bool) {
```
jspada, line 58: Please add some negative tests to check the implementation matches the theory-- that each constraint fails on bad data.

kimchi/src/tests/rot.rs
```diff
@@ -0,0 +1,86 @@
...
+        assert_eq!(PallasField::from(word.rotate_left(rot)), witness[1][1]);
+    } else {
+        assert_eq!(PallasField::from(word.rotate_right(rot)), witness[1][1]);
+    }
+}
+
+#[test]
+// Test that a random offset between 1 and 63 work as expected, both left and right
+fn test_rot_random() {
+    let rot = rand::thread_rng().gen_range(1..=63);
```
jspada, line 82: Please make deterministic

utils/src/foreign_field.rs
```diff
@@ -178,21 +178,21 @@ mod tests {
         let bytes = SECP256K1_MOD;
         let big = BigUint::from_bytes_be(bytes);
         let fe = ForeignElement::<BaseField, 3>::from_be(bytes);
-        assert_eq!(fe.to_big(), big);
+        assert_eq!(fe.to_biguint(), big);
```
jspada, line 181: I guess this is not part of this PR

utils/src/foreign_field.rs
```diff
@@ -178,21 +178,21 @@ mod tests {
...
     }
 
     #[test]
     fn test_from_biguint() {
         let one = ForeignElement::<BaseField, 3>::from_be(&[0x01]);
         assert_eq!(
-            BaseField::from_biguint(one.to_big()).unwrap(),
+            BaseField::from_biguint(one.to_biguint()).unwrap(),
```
jspada, line 188: Need to update against master.

jspada (commented):
Really awesome gate -- love the design.  Checked constraints carefully - they look simple, elegant and correct.

Since this branch needs to do quite a few updates against master I will need to make another pass afterwords.

Nevertheless, I've left tons of feedback and suggestions for now.  Hope it's helpful!

Nice work!

kimchi/src/circuits/witness/sum_copy_bits_cell.rs
jspada: Actually, I think the way you are using it is .. `VariableBitsCell`

kimchi/src/circuits/constraints.rs
```diff
@@ -590,7 +594,7 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
         };
 
         // Foreign field addition constraint selector polynomial
-        let ffadd_gates = foreign_field_add::gadget::circuit_gates();
+        let ffadd_gates = [GateType::ForeignFieldAdd];
```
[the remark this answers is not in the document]
querolita: Again, deleting because Matthew asked me to

kimchi/src/circuits/constraints.rs
```diff
@@ -604,7 +608,17 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
             if circuit_gates_used.is_disjoint(&xor_gate.into_iter().collect()) {
                 None
             } else {
-                Some(selector_polynomial(GateType::Xor16, &gates, &domain))
+                Some(selector_polynomial(xor_gate[0], &gates, &domain))
```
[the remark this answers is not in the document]
querolita: Leaving open for the future

kimchi/src/circuits/polynomials/keccak.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Creates the 5x5 table of rotation bits for Keccak modulo 64
+/// | x \ y |  0 |  1 |  2 |  3 |  4 |
+/// | ----- | -- | -- | -- | -- | -- |
+/// | 0     |  0 | 36 |  3 | 41 | 18 |
+/// | 1     |  1 | 44 | 10 | 45 |  2 |
+/// | 2     | 62 |  6 | 43 | 15 | 61 |
+/// | 3     | 28 | 55 | 25 | 21 | 56 |
+/// | 4     | 27 | 20 | 39 |  8 | 14 |
+pub const ROT_TAB: [[u32; 5]; 5] = [
```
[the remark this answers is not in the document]
querolita: because it reads better in rust when defining the table

kimchi/src/circuits/polynomials/keccak.rs
```diff
@@ -0,0 +1,81 @@
...
+    [62, 6, 43, 15, 61],
+    [28, 55, 25, 21, 56],
+    [27, 20, 39, 8, 14],
+];
+
+impl<F: PrimeField> CircuitGate<F> {
+    /// Creates Keccak gadget.
+    /// Right now it only creates an initial generic gate with all zeros starting on `new_row` and then
+    /// calls the Keccak rotation gadget
+    pub fn create_keccak(new_row: usize) -> (usize, Vec<Self>) {
```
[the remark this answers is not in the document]
querolita: This is just a toy function, it does not create a keccak gadget whatsoever, just does what the comment above says. What I am saying is that this module has limited functionality as it is super incomplete at this moment, but is useful to try out the Keccak rotation table. It will be an iteratively built gadget in the future. It feels bit unnecessary at this point to have an extend version of it, specially because Perhaps a full keccak may not even fit in one single circuit. Nonetheless, as we keep writing these gadgets and see how many rows they need, then yes we will have an extend version for it.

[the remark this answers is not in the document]
querolita: Also, this is the rotation PR, for that reason I am just addressing ROT-related stuff

kimchi/src/circuits/polynomials/keccak.rs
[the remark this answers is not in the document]
querolita: The keccak official spec writes the table transposed, and for that reason I had it transposed in the rot spec. But at this point, I think I will just transpose it there as well for the sake of uniformallity.

kimchi/src/circuits/polynomials/keccak.rs
```diff
@@ -0,0 +1,81 @@
...
+                gates.connect_64bit(zero_row, rot_row - 1);
+            }
+        }
+        (rot_row, gates.to_vec())
+    }
+}
+
+/// Create a Keccak rotation (whole table)
+/// Input: state (5x5) array of words to be rotated
+pub fn create_witness_keccak_rot<F: PrimeField>(state: [[u64; 5]; 5]) -> [Vec<F>; COLUMNS] {
```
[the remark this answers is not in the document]
querolita: Not sure actually, depends on snarky rs and all of that... For now it is just a file because I am only addressing rotation. In the future though, if we keep writing keccak here, then perhaps it may need a directory. Or maybe not, because the big stuff is already split in rot, xor, and, not... And the Keccak module will only make calls to them.

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,575 @@
...
+};
+use ark_ff::{PrimeField, Zero};
+use ark_poly::{
+    univariate::DensePolynomial, EvaluationDomain, Evaluations, Radix2EvaluationDomain as D,
+};
+use rand::{rngs::StdRng, SeedableRng};
+use std::marker::PhantomData;
+use std::{array, collections::HashMap};
+
+pub const LEFT: bool = true;
```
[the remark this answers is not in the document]
querolita: Agree thanks

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,575 @@
...
+            Wire::for_row(new_row),
+            GenericGateSpec::Pub,
+            None,
+        )];
+
+        let rot_row = zero_row + 1;
+        let mut rot64_gates = if side == LEFT {
+            Self::create_rot64(rot_row, rot)
+        } else {
+            Self::create_rot64(rot_row, 64 - rot)
```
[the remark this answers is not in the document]
querolita: 😊

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,575 @@
...
+        constraints.push(
+            word * two_to_rot.clone() - (excess.clone() * two_to_64.clone() + shifted.clone()),
+        );
+        constraints.push(rotated - (shifted + excess.clone()));
+
+        // Compute the bound from the crumbs and limbs
+        let mut power_of_2 = T::one();
+        let mut bound = T::zero();
+
+        // Sum 2-bit limbs
```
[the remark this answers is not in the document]
querolita: Yeah, agree. There is no proper RFC for the rotation gate atm, but I will remember this to include it when I write it (Jo agreed that I could skip the writeup until the finalisation logic is done).

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,575 @@
...
+        }
+
+        // Check that bound = excess - 2^rot + 2^64 so as to prove that excess < 2^64
+        constraints.push(bound - (excess - two_to_rot + two_to_64));
+
+        constraints
+    }
+}
+
+// ROTATION WITNESS COMPUTATION
```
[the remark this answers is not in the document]
querolita: There is no witness.rs for rotation

kimchi/src/circuits/polynomials/xor.rs
```diff
@@ -6,7 +6,7 @@ use crate::{
     circuits::{
         argument::{Argument, ArgumentEnv, ArgumentType},
         constraints::ConstraintSystem,
-        expr::{self, constraints::ExprOps, l0_1, Environment, LookupEnvironment, E},
+        expr::{self, constraints::ExprOps, l0_1, Environment, LookupEnvironment},
```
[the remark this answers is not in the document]
querolita: Right, it comes from the modifications for these gates, included in the XOR pr

kimchi/src/linearization.rs
querolita: No idea why this is marked as modified, I did not introduce that, it is just part of the coefficient PR

kimchi/src/tests/keccak.rs
querolita: desire or affirmation?

utils/src/foreign_field.rs
querolita: no it is not, it is part of the XOR update one but I needed some of those modifications to make the code more readable. In any case XOR update needs to be merged before, and then this PR will look cleaner

utils/src/foreign_field.rs
querolita: Yup

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,575 @@
...
+    ]
+}
+
+fn rot_row<F: PrimeField>(sum: F, curr_row: usize) -> [Box<dyn WitnessCell<F>>; COLUMNS] {
+    [
+        VariableCell::create("word"),
+        VariableCell::create("rotated"),
+        VariableCell::create("excess"),
+        /* 12-bit plookups */
+        SumCopyBitsCell::create(curr_row, 2, 52, 64, sum),
```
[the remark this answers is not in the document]
querolita: I guess I can replace this logic within the variable map though, I will give it a try and check if it is readable.

kimchi/src/circuits/witness/sum_copy_bits_cell.rs
querolita: Changed it to VariableBitsCell with a "modified" input at calling time instead of in the layout level

kimchi/src/tests/rot.rs
querolita: How did that get in there? I thought I had made all tests deterministic last week 😞

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,575 @@
...
+        // Append them to the full gates vector
+        gates.append(&mut rot64_gates);
+        // Check that 2 most significant limbs of shifted are zero
+        gates.connect_64bit(zero_row, rot_row + 1);
+
+        (new_row + gates.len(), gates)
+    }
+
+    /// Verifies the rotation gate
+    pub fn verify_rot<G: KimchiCurve<ScalarField = F>>(
```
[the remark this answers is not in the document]
jspada: Sure, that sounds fine too.

book/src/rfcs/keccak.md
```diff
@@ -32,3 +42,126 @@ Before we describe the specific gate design approaches, below are some constrain
...
+\end{align*}
+$$
+means that $a \cdot 2^{64} + b = a'\cdot 2^{64} + b'$. Moving terms to the left side, we have an equivalent equality: $2^{64}(a-a') + (b-b')=0 \text{ mod } n$. There are three cases to consider:
+
+- $a = a'$ and $b \neq b'$: then $(b - b') \equiv_n 0$ and this can only happen if $b' = b + n$. But since $n > 2^{64}$, then $b'$ cannot be smaller than $2^{64}$ as it was assumed. CONTRADICTION.
+  
+- $b = b'$ and $a \neq a'$: then $2^{64}(a - a') \equiv_n 0$ and this can only happen if $a' = a + n$. But since $n > 2^r$, then $a'$ cannot be smaller than $2^r$ as it was assumed. CONTRADICTION.
+
+- $a\neq a'$ and $b \neq b'$: then we have something like $2^{64} \alpha + \beta \equiv_n 0$. We choose to express $\alpha$:
+  - $$2^{64} \alpha = n - \beta = n - b + b' > n - 2^{64} + 1$$
```
jspada, line 157: Something's up with the bullets here too.  Maybe either remove bullets or wrap equations with single `$` instead

jspada, line 157: Maybe good to mention $-\beta = n - \beta$, i.e. negation in $\mathbb{F_{64}}$

book/src/rfcs/keccak.md
```diff
@@ -32,3 +42,126 @@ Before we describe the specific gate design approaches, below are some constrain
 Kimchi framework that dictated those approaches.
 * only 4 lookups per row
 * only first 7 columns are available to the permutation polynomial
+<<<<<<< HEAD
+
+## Rot64
+
+It is clear from the definition of the rotation gate that its constraints are complete
```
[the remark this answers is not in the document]
jspada: I feel like this is right thing to do: whenever we do something **new** or somehow **different** the onus of proof of correctness, completeness, soundness should be on us and planned for in timelines.

book/src/rfcs/keccak.md
```diff
@@ -32,3 +42,125 @@ Before we describe the specific gate design approaches, below are some constrain
...
+  - According to the assumption, both $0\leq a<2^r$ and $0\leq a'<2^r$. This means, the difference $\alpha:=(a - a')$ lies anywhere in between the following interval: 
+  - $$1 - 2^r \leq \alpha \leq 2^r - 1$$
+  - We plug in this interval to the above equation to obtain the following interval for $\beta$:
+  - $$k\cdot n - 2^{64}(1-2^r)\leq \beta \leq k\cdot n - 2^{64}(2^r - 1) $$
+  - We look at this interval from both sides of the inequality: $\beta \geq kn - 2^{64} + 2^{64+r}$ and $\beta \leq kn + 2^{64} - 2^{64+r}$ and we wonder if $kn - 2^{64} + 2^{64+r} \leq kn + 2^{64} - 2^{64+r}$ is at all possible. We rewrite as follows:
+  - $$2^{64+r} - 2^{64} \leq 2^{64} - 2^{64+r}$$
+  - $$2\cdot2^{64+r} \leq 2\cdot2^{64} $$
+  - $$2^{64+r} \leq 2^{64} $$
+  - But this can only happen if $r\leq 0$, which is impossible since we assume $0 < r < 64$.CONTRADICTION.
+- EOP.
```
jspada, line 166: Nice proof!

kimchi/src/circuits/expr.rs
```diff
@@ -508,6 +508,7 @@ pub enum FeatureFlag {
     ForeignFieldAdd,
     ForeignFieldMul,
     Xor,
+    Rot,
```
jspada, line 511: For simplicity and consistency, I feel like these should match the `GateType` names

kimchi/src/circuits/lookup/index.rs
```diff
@@ -45,7 +45,7 @@ pub struct LookupSelectors<T> {
     /// LookupGate pattern lookup selector
     pub lookup_gate: Option<T>,
     /// RangeCheckGate pattern lookup selector
-    pub range_check_gate: Option<T>,
+    pub range_check: Option<T>,
```
jspada, line 48: For consistency, make the name match the others ending with `_gate`

kimchi/src/circuits/lookup/index.rs
```diff
@@ -60,7 +60,7 @@ struct LookupSelectorsSerdeAs<F: FftField> {
     #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
     pub lookup_gate: Option<E<F, D<F>>>,
     #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
-    pub range_check_gate: Option<E<F, D<F>>>,
+    pub range_check: Option<E<F, D<F>>>,
```
jspada, line 63: Same comment as above.  To keep the diff smaller and keep naming consistent, there's no reason to rename these variables.

kimchi/src/circuits/lookup/index.rs
```diff
@@ -150,7 +150,7 @@ impl<T> LookupSelectors<T> {
             xor: xor.map(f),
             chacha_final: chacha_final.map(f),
             lookup_gate: lookup_gate.map(f),
-            range_check_gate: range_check_gate.map(f),
+            range_check: range_check.map(f),
```
jspada, line 153: Yeah, definitely revert all these changes.  There's no reason to make them and doesn't have anything to do with this PR.

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -286,15 +286,15 @@ pub enum LookupPattern {
     Xor,
     ChaChaFinal,
     LookupGate,
-    RangeCheckGate,
+    RangeCheck,
```
jspada, line 289: Revert these changes.

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -286,15 +286,15 @@ pub enum LookupPattern {
...
     ForeignFieldMulGate,
 }
 
 impl LookupPattern {
     /// Returns the maximum number of lookups per row that are used by the pattern.
     pub fn max_lookups_per_row(&self) -> usize {
         match self {
-            LookupPattern::Xor | LookupPattern::ChaChaFinal | LookupPattern::RangeCheckGate => 4,
+            LookupPattern::Xor | LookupPattern::ChaChaFinal | LookupPattern::RangeCheck => 4,
```
jspada, line 297: Same

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -418,7 +418,7 @@ impl LookupPattern {
         match self {
             LookupPattern::Xor | LookupPattern::ChaChaFinal => Some(GateLookupTable::Xor),
             LookupPattern::LookupGate => None,
-            LookupPattern::RangeCheckGate => Some(GateLookupTable::RangeCheck),
+            LookupPattern::RangeCheck => Some(GateLookupTable::RangeCheck),
```
jspada, line 421: Revert

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -431,7 +431,9 @@ impl LookupPattern {
             (ChaCha0 | ChaCha1 | ChaCha2, Curr | Next) => Some(LookupPattern::Xor),
             (ChaChaFinal, Curr | Next) => Some(LookupPattern::ChaChaFinal),
             (Lookup, Curr) => Some(LookupPattern::LookupGate),
-            (RangeCheck0, Curr) | (RangeCheck1, Curr | Next) => Some(LookupPattern::RangeCheckGate),
+            (RangeCheck0, Curr) | (RangeCheck1, Curr | Next) | (Rot64, Curr) => {
```
jspada, line 434: Revert RangeCheck back to RangeCheckgate and leave the rest

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,608 @@
...
+    const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::Rot64);
+    const CONSTRAINTS: u32 = 11;
+
+    // Constraints for rotation of three 64-bit words by any three number of bits modulo 64
+    // (stored in coefficient as a power-of-two form)
+    //   * Operates on Curr row
+    //   * Shifts the words by `rot` bits and then adds the excess to obtain the rotated word.
+    fn constraint_checks<T: ExprOps<F>>(env: &ArgumentEnv<F, T>) -> Vec<T> {
+        // Check that the last 8 columns are 2-bit crumbs
+        // C0..C7: x * (x - 1) * (x - 2) * (x - 3) = 0
```
jspada, line 453: Awesome!

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,608 @@
...
+        let mut constraints = (7..COLUMNS)
+            .map(|i| crumb(&env.witness_curr(i)))
+            .collect::<Vec<T>>();
+
+        // TODO:
```
jspada, line 458: Maybe this is a `NOTE:` instead

kimchi/src/circuits/polynomials/rot.rs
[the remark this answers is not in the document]
jspada: I see you put them in the comments already - great!

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,608 @@
...
+    for col in 0..COLUMNS {
+        witness[col].extend(rot_witness[col].iter());
+    }
+    init_rot64(witness, rot_row, word, rotated, excess, shifted, bound);
+}
+
+/// Create a rotation witness
+/// Input: word to be rotated, rotation offset, and side of rotation.
+/// Output: witness for rotation word and initial row with all zeros
+pub fn create_witness<F: PrimeField>(word: u64, rot: u32, side: RotMode) -> [Vec<F>; COLUMNS] {
```
jspada, line 567: Do you allow `rot = 0` or  `rot > 64` for this pub interface?  If not, maybe check it or assert like you did for the other interface.

kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -0,0 +1,608 @@
...
+            power_of_2 *= T::from(4u64); // 2 bits
+        }
+
+        // Sum 12-bit limbs
+        for i in (3..=6).rev() {
+            bound += power_of_2.clone() * env.witness_curr(i);
+            power_of_2 *= 4096u64.into(); // 12 bits
+        }
+
+        // Check that excess < 2^rot by checking that bound < 2^64
```
jspada, line 493: Maybe you also want to refer to your proof in the RFC here too?

kimchi/src/circuits/witness/sum_copy_bits_cell.rs
```diff
@@ -0,0 +1,40 @@
+use ark_ff::Field;
```
jspada, line 1: Reminder.  This is part of this PR or from XOR and not used anymore, so should be removed.

kimchi/src/tests/keccak.rs
jspada: You are probably right.  Pass on this gate

kimchi/src/tests/keccak.rs
jspada: desire

jspada: I wish proof-system automatically took care of this optimally

kimchi/src/tests/rot.rs
jspada: I guess this gate doesn't require this

jspada (commented):
Really good progress - noticed a few more things dragged in by merges that need to be removed.  Also, left some feedback and suggestions.

book/src/rfcs/keccak.md
```diff
@@ -19,8 +19,18 @@ Fp[0,0] &= F[0,0] \oplus RC
 $$
 
 FOR $0\leq x, y \leq 4$ and $\rho[x,y]$ is the rotation offset defined for Keccak.
+The values are in the table below extracted from the Keccak reference
```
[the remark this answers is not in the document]
querolita: It is pending writing the RFCs, this will happen after pickles

book/src/rfcs/keccak.md
```diff
@@ -32,3 +42,126 @@ Before we describe the specific gate design approaches, below are some constrain
...
+$$
+
+We chose these word and field lengths to better understand the behaviour of the solution. Here, we can see two "classes" of evaluations.
+
+- If we choose an even $k$, then $y$ will have the following shape: 
+  - $$26 - 16 \cdot (2 \cdot n) \iff 26 - 32n \equiv_{32} 26 $$
+  - Meaning, if $x = 2n$ then $y = 26$.
+
+- If on the other hand, we chose an odd $k$, then $y$ will have the following shape instead:
+  - $$26 - 16 \cdot (2 \cdot n + 1) \iff 26 - 32n - 16 \equiv_{32} 26 - 16 = 10$$
```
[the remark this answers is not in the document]
querolita: I wanted it centered

book/src/rfcs/keccak.md
```diff
@@ -32,3 +42,126 @@ Before we describe the specific gate design approaches, below are some constrain
...
+  - Meaning, if $x = 2n+1$ then $y = 10$. 
+
+Thus, possible solutions to the diophantic equation are:
+
+$$
```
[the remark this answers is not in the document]
querolita: thanks! just playing with numbers a bit

book/src/rfcs/keccak.md
```diff
@@ -32,3 +42,126 @@ Before we describe the specific gate design approaches, below are some constrain
...
+assignment satisfies the constraints. Meaning, that there is no dishonest instance that
+can make the constraints pass. We will also give an example where one could find wrong
+rotation witnesses that would satisfy the constraints if we did not check the range.
+
+### Necessity of range checks
+
+First of all, we will illustrate the necessity of range-checks with a simple example. 
+For the sake of readability, we will use some toy field lengths. In particular, let us 
+assume that our words have 4 bits, meaning all of the elements between `0x0` and `0xF`. 
+Next, we will be using the native field $\mathbb{F}_{2^5}$.
```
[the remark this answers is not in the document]
querolita: okay

book/src/rfcs/keccak.md
querolita: it compiles nice with mdbook though

kimchi/src/circuits/expr.rs
querolita: well, rangecheck does not come with a 0 or a 1, nor chacha has the 0 1 final, I think FeatureFlag just needs to be readable for us, no need to be so specific on which exactly. If we ever have a rot128, we would still want the feature flag to be general for rotation, and not tied to 64 or whatever

kimchi/src/circuits/lookup/index.rs
querolita: I have made them uniform. For example, Xor not only names the Xor16 gate but also ChaCha0 and ChaCha1, but naming it XorGate could be misleading. I think the same type of situation may arise for other patterns as well.

kimchi/src/circuits/lookup/index.rs
querolita: I believed a change could be nice to be uniform with respect to the Xor pattern

kimchi/src/tests/keccak.rs
querolita: yeah, i will add those in the rot test file

kimchi/src/circuits/lookup/index.rs
jspada: Yeah, disregard my comment.  For some reason I thought this was not the patterns.  Definitely we do not want a Gate suffix for the patterns :-)

kimchi/src/circuits/lookup/index.rs
jspada: I prefer it follow the mnemonics of `lookup_gate` et al also it's not related to this PR, so should not be done here.

book/src/rfcs/keccak.md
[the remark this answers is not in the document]
jspada: Oh, maybe then remove the bullet?  Looks strange to me, but you know more about this stuff so go with what you think,

kimchi/src/circuits/lookup/index.rs
```diff
@@ -114,9 +114,9 @@ impl<T> std::ops::Index<LookupPattern> for LookupSelectors<T> {
         match index {
             LookupPattern::Xor => &self.xor,
             LookupPattern::ChaChaFinal => &self.chacha_final,
-            LookupPattern::LookupGate => &self.lookup_gate,
-            LookupPattern::RangeCheckGate => &self.range_check_gate,
-            LookupPattern::ForeignFieldMulGate => &self.ffmul_gate,
+            LookupPattern::Lookup => &self.lookup,
+            LookupPattern::RangeCheck => &self.range_check,
```
jspada, line 118: Good fix.

jspada (approved):
Looks great - approved!

PR #822, opened 2022-10-24
Implement generic witness layout for custom gates

jspada:
Extract common witness computation code from range-check and foreign-field-add gates into reusable generic kit for foreign field multiplication and other custom gates

### DIFF
--- a/kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
+++ b/kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
@@ -1,13 +1,15 @@
 //! This module computes the witness of a foreign field addition circuit.
 
-use crate::circuits::{
-    polynomial::COLUMNS,
-    polynomials::range_check::{
-        self,
-        witness::{extend_witness, handle_standard_witness_cell, CopyWitnessCell, ZeroWitnessCell},
+use crate::circuits::witness::Variables;
+use crate::{
+    circuits::{
+        polynomial::COLUMNS,
+        polynomials::range_check,
+        witness::{self, ConstantCell, CopyCell, VariableCell, WitnessCell},
     },
+    variable_map,
 };
-use ark_ff::{Field, PrimeField};
+use ark_ff::PrimeField;
 use num_bigint::BigUint;
 use o1_utils::foreign_field::{ForeignElement, HI, LO, MI, TWO_TO_LIMB};
 use std::array;
@@ -108,7 +110,7 @@ fn compute_subadd_values<F: PrimeField>(
 /// inputs: list of all inputs to the chain of additions/subtractions
 /// opcode: true for addition, false for subtraction
 /// modulus: modulus of the foreign field
-pub fn create_witness<F: PrimeField>(
+pub fn create<F: PrimeField>(
     inputs: &Vec<BigUint>,
     opcodes: &Vec<FFOps>,
     modulus: BigUint,
@@ -127,7 +129,7 @@ pub fn create_witness<F: PrimeField>(
 
     // Create multi-range-check witness for first left input
     let mut left = ForeignElement::from_biguint(inputs[LO].clone());
-    extend_witness(&mut witness, left.clone());
+    range_check::witness::extend(&mut witness, left.clone());
     let mut add_values: Vec<(F, F, F, F)> = vec![];
     for i in 0..num {
         let right = ForeignElement::from_biguint(inputs[i + 1].clone());
@@ -136,8 +138,8 @@ pub fn create_witness<F: PrimeField>(
         // Create multi-range-check witness for right_input (left_input was done in previous iteration) and output
         // We only obtain the 3 lower limbs of right because the range check takes only 264 bits now
         let right_3_limb = ForeignElement::new([right[LO], right[MI], right[HI]]);
-        extend_witness(&mut witness, right_3_limb);
-        extend_witness(&mut witness, output.clone());
+        range_check::witness::extend(&mut witness, right_3_limb);
+        range_check::witness::extend(&mut witness, output.clone());
 
         add_values.append(&mut vec![(sign, overflow, carry_lo, carry_mi)]);
         left = output; // output
@@ -153,7 +155,7 @@ pub fn create_witness<F: PrimeField>(
     assert_eq!(overflow, F::one());
 
     // Final RangeCheck for bound
-    extend_witness(&mut witness, bound);
+    range_check::witness::extend(&mut witness, bound);
     let mut offset = witness[LO].len(); // number of witness rows of the gadget before the first row of the addition gate
 
     // Include FFAdds gates for operations and final bound check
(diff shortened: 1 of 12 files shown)


kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -231,32 +198,33 @@ fn init_foreign_field_add_rows<F: PrimeField>(
...
-    for (row, wit) in witness_shape.iter().enumerate() {
-        for (col, cell) in wit.iter().enumerate() {
-            handle_ffadd_rows(witness, cell, (row, col), offset, sign, overflow, carry);
-        }
-    }
+    witness::init(
+        witness,
+        offset,
+        &witness_shape,
+        &variable_map!["sign" => sign, "overflow" => overflow, "carry_lo" => carry[0], "carry_mi" => carry[1]],
```
querolita, line 226: Can't the variable map work over enums? or if you do then you have to explicitly address all of them (both the ones you use in this gate, and the ones you don't)?

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -267,90 +235,49 @@ fn init_foreign_field_fin_rows<F: PrimeField>(
...
-            }
-        }
-        WitnessCell::Constant(field_elem) => witness[col][offset + row] = *field_elem,
-        WitnessCell::Ignore => (),
-    }
+    witness::init(
+        witness,
+        offset,
+        &witness_shape,
+        &variable_map![ "carry_lo" => carry[0], "carry_mi" => carry[1]],
```
querolita, line 281: Interesting that you do not need to explain the rest of variables... It was annoying to set the other terms to zero in this section.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -267,90 +235,49 @@ fn init_foreign_field_fin_rows<F: PrimeField>(
 ) {
     let out_row = 8 * num; // row where the final result is stored in RC
     let bound_row = 8 * num + 4; // row where the final bound is stored in RC
-    let witness_shape: [[WitnessCell<F>; COLUMNS]; 2] = [
+    let witness_shape: [[Box<dyn WitnessCell<F>>; COLUMNS]; 2] = [
```
mrmr1993, line 238: Nit: these can probably be `&dyn WitnessCell<F>` to avoid the heap allocations.

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -79,152 +24,118 @@ impl ZeroWitnessCell {
...
-        ValueWitnessCell::create(),
+        VariableCell::create(limb_name),
         /* 12-bit copies */
         // Copy cells are required because we have a limit
         // of 4 lookups per row.  These two lookups are moved to
         // the 4th row, which is a Zero circuit gate, and the
         // RangeCheck1 circuit gate triggers the lookup constraints.
-        LimbWitnessCell::create(row, 0, 76, 88),
-        LimbWitnessCell::create(row, 0, 64, 76),
+        CopyBitsCell::create(row, 0, 76, 88),
```
mrmr1993, line 92: This would be a bit more readable (for me) as `Box::create(CopyBitsCell {row, col: 0, start:76, end:88})`. Likewise elsewhere.

kimchi/src/circuits/polynomials/range_check/witness.rs
```diff
@@ -79,152 +24,118 @@ impl ZeroWitnessCell {
...
-        LimbWitnessCell::create(2, 0, 10, 12),
-        LimbWitnessCell::create(2, 0, 8, 10),
-        LimbWitnessCell::create(2, 0, 6, 8),
-        LimbWitnessCell::create(2, 0, 4, 6),
-        LimbWitnessCell::create(2, 0, 2, 4),
-        LimbWitnessCell::create(2, 0, 0, 2),
-    ],
-];
+        /* row 1, RangeCheck0 row */
+        range_check_0_row("v0", 0),
```
mrmr1993, line 30: This seems a tad brittle, depending on the variable name `v0` used elsewhere.

mrmr1993 (approved):
I think we could refine this interface some -- currently it seems rather brittle -- but this does look like it makes it easier to construct witness generation code. Good work.

querolita (approved):
We needed a witness layout and this is a good candidate specially for complex gates that are being created lately. Well done!

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
jspada: Hm.. Yeah -- I agree. I originally had it the way you suggest b/c it was my preferred way, but it was not happy to allocate them and was complaining about static lifetimes.  So, in interest of time, I gave up.  I will try to fix in a subsequent PR.

kimchi/src/circuits/polynomials/range_check/witness.rs
jspada: There's also a way to map them explicitly with `variable_map!["v0" => blar_dee_blar]`, if that helps.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -231,32 +198,33 @@ fn init_foreign_field_add_rows<F: PrimeField>(
...
-            WitnessCell::Standard(ZeroWitnessCell::create()),
-            WitnessCell::Standard(ZeroWitnessCell::create()),
-            WitnessCell::Standard(ZeroWitnessCell::create()),
+            CopyCell::create(left_row, 0),      // left_input_lo
+            CopyCell::create(left_row + 1, 0),  // left_input_mi
+            CopyCell::create(left_row + 2, 0),  // left_input_hi
+            CopyCell::create(right_row, 0),     // right_input_lo
+            CopyCell::create(right_row + 1, 0), // right_input_mi
+            CopyCell::create(right_row + 2, 0), // right_input_hi
+            VariableCell::create("sign"),
```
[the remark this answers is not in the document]
jspada: I couldn't I think of any other way to do it.  And initially I was not a fan either, but in the end I think it's actually fine.  

Think of the layout as a "template", then it makes sense to give parts of it names, which of course must be strings.

Basically, you're just saying this cell is called "blah"

Later on when you generate the witness, you pass the `init_witness` function the layout and the mapping of variables, i.e., `variable_map!["blah" => F::one(), "bleh" => f_max]`

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
jspada: The old enums way was a lot more complex and needed a lot more code to do the same thing.  Also, I couldn't think of a way to make it generic, so you don't need to redefine types every time you want to do the same thing just for a new variable.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
jspada: Well, you only need to map whatever variables are used by the layout template.  In this case only two are used, so those are the only variables in the witness to set.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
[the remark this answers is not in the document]
querolita: yeah I saw that, but the strings seemed a bit error prone perhaps. Anyway, it is perfect for now

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
querolita: Aha yeah that is very nice

PR #825, opened 2022-10-27
snarky EC

fabrizio-m:


mimoo:
Ouch, that's actually really verbose. The more I'm thinking about that, the more I'm thinking that semantically it doesn't make sense either does it? A circuit is something that's written over a field, not over a curve. I think it's fine to have the KimchiCurve be passed in the `new()` function of RunState, and we can take the constants and all we want from there then (which is what we do already no?), but passing the curve everywhere seems counter productive no?

fabrizio-m:
I was thinking about that, and I also noticed that even if I make the snarky gates generic, the kimchi gates are not generic and only work with pasta or curves with a very similar law.
My main interest to access the curve is to avoid reimplementing it inside `compute()`, what about this:
```rust
pub trait SnarkyField: PrimeField {
    type Curve: AffineCurve<BaseField = Self>;
}
``` 
We replace `PrimeField` with a subtrait that offers access to the curve, the code will the be same as before.

mimoo:
> My main interest to access the curve is to avoid reimplementing it inside compute(), what about this:

Can't you just use the closure to capture the values (which are IIUC the round constants)?

fabrizio-m:
Round constants are in poseidon, here I need the implementation of elliptic curve operations that live in `AffineCurve` and `GroupAffine`, I'll soon add a commit with an example of it would work with the trait.

mimoo:
is it to verify that the operations are correctly done? Just looking at izaak's circuit-construction he doesn't use the group operations, at least for the ec_add gate (https://github.com/o1-labs/proof-systems/blob/mimoo%2Fsnarky/circuit-construction/src/writer.rs#L263)

fabrizio-m:
The kimchi gate basically takes 3 points and some other info and enforces that the third is the sum of the first 2 but starting from two points I have to compute the third and the other things.
Izaak is reimplementing the group operation in that function, I can do the same, it just seems to me more maintainable to reuse ark-ec.

fabrizio-m:
It seems that I will have to reimplement some parts anyway to access some of the intermediate values, so I can just implement the whole operation.

mimoo:
btw something else we can do in Rust, if this is something we think we need down the line, is to preserve the `KimchiCurve` we pass in `new()` in the runstate (we can do that using `PhantomData`)

PR #828, opened 2022-11-01
[easy] fix specs

querolita:
Fixing some tables that were broken in the specs


kimchi/src/circuits/polynomials/range_check/circuitgates.rs
```diff
@@ -111,7 +111,8 @@ use crate::circuits::{
 };
 use ark_ff::PrimeField;
 
-//~ ##### `RangeCheck0` - Range check constraints
+//~
+//~ **`RangeCheck0` - Range check constraints**
```
mimoo, line 115: yeah I think that's the way!

mimoo (approved):

PR #831, opened 2022-11-05
introduce an ASM-like language for debugging

mimoo:
This PR adds an ASM-like language that's useful to inspect circuits and debug. (It also adds the public input size to the `Circuit` type, and adds some convenience functions, to make this work.)

It looks like this:

```
row0.pub.Generic<1,0,0,0,0>
.l1 -> row4.l1

row1.pub.Generic<1,0,0,0,0>
.l1 -> row2.l1

row2.Generic<-1,0,0,1,0><-1,0,0,1,0>
.l1 -> row4.r1, .r1 -> row1.l1
.l2 -> row0.l1, .r2 -> row2.l2

row3.Generic<-1,0,0,1,0><-1,0,0,1,0>
.l1 -> row4.r2, .r1 -> row3.l1
.l2 -> row4.l2, .r2 -> row3.l2

row4.Generic<0,0,1,-1,0><0,0,1,-1,0>
.l1 -> row2.r2, .r1 -> row2.r1, .o1 -> row5.l1
.l2 -> row3.r2, .r2 -> row3.r1, .o2 -> row4.o1

row5.Generic<1,0,0,0,-1>
.l1 -> row4.o2
```

See https://github.com/MinaProtocol/mina/pull/12125 to see it used in tests

### DIFF
--- a/kimchi/src/snarky/asm.rs
+++ b/kimchi/src/snarky/asm.rs
@@ -67,7 +67,6 @@ where
 
             // wires
             {
-                let mut wires_str = vec![];
                 for (
                     col,
                     Wire {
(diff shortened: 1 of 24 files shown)


kimchi/src/snarky/asm.rs
```diff
@@ -0,0 +1,257 @@
...
+                } else {
+                    let var_idx = vars.pos(x);
+                    format!("c{var_idx}")
+                }
+            })
+            .collect()
+    }
+}
+
+/// Very dumb way to write an ordered hash set.
```
jspada, line 167: Maybe this would work?
https://crates.io/crates/linked_hash_set

jspada (approved):
This is a really nice feature.  Cool idea.

mrmr1993:
I'm very much not a fan of this format, and would prefer to revert this.

For the [snarky tests PR](https://github.com/MinaProtocol/mina/pull/12125), can you use a JSON representation instead, like we discussed?

PR #833, opened 2022-11-05
Add a `PointEvaluations` type, replace pairs of evaluations

mrmr1993:
This PR replaces `[ProofEvaluations<F>; 2]` everywhere with a `ProofEvaluations<PointEvaluations<F>>`.

The majority of changes are because of the transposition. This is a no-op.

Edit: the Mina side is [in this branch](https://github.com/MinaProtocol/mina/compare/feature/transpose-proof-systems-evals?expand=1).


kimchi/src/proof.rs
```diff
@@ -44,29 +56,20 @@ pub struct LookupEvaluations<Field> {
...
-#[serde(bound(
-    serialize = "Vec<o1_utils::serialization::SerdeAs>: serde_with::SerializeAs<Field>",
-    deserialize = "Vec<o1_utils::serialization::SerdeAs>: serde_with::DeserializeAs<'de, Field>"
-))]
-pub struct ProofEvaluations<Field> {
+pub struct ProofEvaluations<Eval> {
     /// witness polynomials
-    #[serde_as(as = "[Vec<o1_utils::serialization::SerdeAs>; COLUMNS]")]
-    pub w: [Field; COLUMNS],
+    pub w: [Eval; COLUMNS],
```
mimoo, line 61: why not just have `PointEvaluations` instead of `Eval`?

mrmr1993: So that, later, we can use this for
* the polynomials
* the commitments
* the evaluations
rather than having a separate structure for each. Specialising it now just means that I'll have to un-specialise it later.

mimoo: aren't these also going to use `PointEvaluations` since it is generic as well now?

mrmr1993: The commitments and the polynomials don't have anything to do with `zeta` or `zeta_omega`, they're computed before we know what they are. Feels very odd to try to shoehorn them into this type.

book/src/specs/kimchi.md
```diff
@@ -1720,28 +1720,40 @@ You can find these operations under the [proof creation](#proof-creation) and [p
...
+    serialize = "Vec<o1_utils::serialization::SerdeAs>: serde_with::SerializeAs<Evals>",
+    deserialize = "Vec<o1_utils::serialization::SerdeAs>: serde_with::DeserializeAs<'de, Evals>"
 ))]
-pub struct LookupEvaluations<Field> {
+pub struct PointEvaluations<Evals> {
+    /// Evaluation at the challenge point zeta.
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    pub zeta: Evals,
+    /// Evaluation at `zeta . omega`, the product of the challenge point and the group generator.
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
```
mimoo, line 1739: I don't understand the change of `Vec<E>` to `E`, what is the rationale here? This makes the code harder to follow imo

mimoo (approved):
This is a good PR, it simplifies a lot of the code, but I do think that it adds too much genericity and it is making the code harder to parse for humans. Perhaps with good documentation we can make it easier for people to understand how the different parameters are supposed to be instantiated, but at this point I'm failing to see why some of these structs really need to be generic. Maybe the next sets of PRs will make that clearer.

kimchi/src/proof.rs
mimoo: when you say later you mean in the other PRs in this series? Otherwise I think this is premature optimization

PR #835, opened 2022-11-05
Add a verifier context, accessors for commitments

mrmr1993:
This PR
* adds a `verifier::Context` type
* adds a `Context::get_column` helper
* uses the helper to replace the match statement in the verifier


kimchi/src/error.rs
```diff
@@ -67,6 +67,9 @@ pub enum VerifyError {
 
     #[error("the evaluation for {0:?} is missing")]
     MissingEvaluation(crate::circuits::expr::Column),
+
+    #[error("the commitment for {0:?} is missing")]
```
mimoo, line 71: I think we should have the error also answer "missing from what?" (same comment for the error above)

mimoo (approved):
beautiful! really like where this is going : )

PR #836, opened 2022-11-05
Use iterators over the standard evaluated columns in the verifier

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/835, using the changes there to simplify the code for selecting commitments and their evaluations into a loop over an iterator.

PR #837, opened 2022-11-05
Add lookup_info to the verifier index, use it to iterate over some evaluations

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/836.

This
* adds the `lookup_info` from the prover index into the verifier index
* removes the `max_joint_size` subfield from the verifier index
* uses iterators for the sorted and aggregation columns of the lookup argument.


kimchi/src/verifier.rs
```diff
@@ -646,7 +645,21 @@ where
...
     .chain((0..PERMUTS - 1).map(Column::Permutation))
-    {
+    //~~ - lookup commitments
+    .chain(
+        index
+            .lookup_index
+            .as_ref()
+            .map(|li| {
+                // add evaluations of sorted polynomials
+                (0..li.lookup_info.max_per_row + 1)
```
mimoo, line 655: ah interesting, so did we have a bug before since we didn't enforce the length of the sorted stuff? 
I'm guessing that no because if we didn't have enough the creation to the commitment to the composition polynomial will fail later here as we will request a sorted commitment that doesn't exist.
The prover can still put more commitments or evals than necessary though

mimoo: in https://o1-labs.github.io/proof-systems/rfcs/3-lookup.html#the-sorted-vector-s it says the length of the sorted vector (before being chunked) is `n * queries + len(lookup table)`. 

Once we chunk, we do the snake trick and so there's some overhead as some elements are repeated. So it sounds to me like we're simplifying the chunks to be `queries + 1` when it should really be `queries + ceil_div(len(lookup_table) + overhead, n)`. I believe we do have some checks to make sure that the lookup table is not larger than the domain, but what about the overhead?

mimoo (approved):

kimchi/src/verifier.rs
mrmr1993: > It sounds to me like we could have the chunks actually be `queries + 2`.

This never happens, by design. We could change that if we want tables larger than the domain size (see e.g. [this unreviewed PR](https://github.com/o1-labs/proof-systems/pull/235) from over a year ago, where I did exactly that).

PR #839, opened 2022-11-05
Iterate over gates in prover instead of calling them all manually

mrmr1993:
This PR is built upon #838.

This PR adds a new trait `DynArgument` to work around an incompatibility between `dyn` trait implementations and the `Argument` trait. It then uses this trait to remove the boilerplate for adding gates from the prover, replacing it with a loop over an iterator.

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     circuits::{
-        argument::{Argument, ArgumentType},
+        argument::ArgumentType,
         expr::{l0_1, Constants, Environment, LookupEnvironment},
         gate::GateType,
         lookup::{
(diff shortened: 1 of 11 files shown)


kimchi/src/prover.rs
```diff
@@ -703,94 +697,52 @@ where
...
+                .filter_map(|(gate, is_enabled)| if is_enabled { Some(gate) } else { None })
+                {
+                    let constraint = gate.combined_constraints(&all_alphas);
+                    let eval = constraint.evaluations(&env);
+                    if eval.domain().size == t4.domain().size {
+                        t4 += &eval;
+                    } else if eval.domain().size == t8.domain().size {
+                        t8 += &eval;
+                    } else {
+                        panic!("Bad evaluation")
```
mimoo, line 741: a better error would be nice here. Or `unimplemented!()`

mimoo (approved):

PR #840, opened 2022-11-08
Add the `ChunkedEvaluations` type

duguorong009:
**What's been done:**
- Add a new type `ChunkedEvaluations` for the `ProofEvaluations` refactor

Reference: #759

### DIFF
--- a/utils/src/chunked_evaluations.rs
+++ b/utils/src/chunked_evaluations.rs
@@ -1,13 +1,16 @@
 //! This module contains a type [ChunkedEvaluations],
 
+use ark_ff::PrimeField;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
 /// This struct contains multiple chunk evaluations.
 #[serde_as]
 #[derive(Clone, Serialize, Deserialize)]
-#[serde(bound = "F: ark_serialize::CanonicalDeserialize + ark_serialize::CanonicalSerialize")]
-pub struct ChunkedEvaluations<F> {
+pub struct ChunkedEvaluations<F>
+where
+    F: PrimeField,
+{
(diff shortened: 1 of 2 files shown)


duguorong009:
@mimoo 
Can you please review the PR?
Thanks in advance!

utils/src/chunked_evaluations.rs
```diff
@@ -0,0 +1,17 @@
+//! This module contains a type [ChunkedEvaluations],
+
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+/// This struct contains multiple chunk evaluations.
+#[serde_as]
+#[derive(Clone, Serialize, Deserialize)]
+#[serde(bound = "F: ark_serialize::CanonicalDeserialize + ark_serialize::CanonicalSerialize")]
```
mimoo, line 9: I don't think you need this bound no?

duguorong009: Um, I think that bound restricts the possible types of `F`.
My first attempt was like 
```
pub struct ChunkedEvaluations<F: ark_ff::Field> {
...
```
I think we can restrict the `F` type bound by using above `serde(bound)`

mimoo: oh right, I think you should just add `pub struct ChunkedEvaluations<F> where F: PrimeField { ...` which is simpler : o

duguorong009: Done https://github.com/o1-labs/proof-systems/pull/840/commits/1de06fb10d51ac2bfffddd1a622938ab704079b4

mimoo (approved):
nice! thanks a lot :)

PR #846, opened 2022-11-13
Use the `Argument` trait to implement the generic gate

mrmr1993:
This PR replaces the old generic gate implementation with a vastly simplified version using the `expr` framework.

This update maintains compatibility with Mina by providing a `generic: bool` argument to the linearisation functions: this allows pickles to pass `false` and still continue as if this change was not made.

Proofs made before and after this change are entirely compatible; this is a no-op.


mimoo (approved):
beautiful!

PR #847, opened 2022-11-13
Use coefficient evaluations for the generic and poseidon gates

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/846.

This marks the coefficients columns as 'evaluated', so that the evaluations are used in the linearisation instead of the commitments.

This allows us to reduce the number of constraints used by the pickles verifier circuit during recursion.


mimoo (approved):
nice!

PR #851, opened 2022-11-15
[docs] Fix reference to `kimchi.png` in `arguments.md`

tizoc:
It is not being rendered correctly in https://o1-labs.github.io/proof-systems/kimchi/arguments.html


mimoo (approved):
Awesome thanks a lot for the fix :)

PR #859, opened 2022-11-21
add a Circuit type

mimoo:
just some cleaning as part of converting a circuit in JSON. Let's enforce semantically what a circuit is once and for all.


mrmr1993 (changes requested):
I don't think we should inject another layer. The `ConstraintSystem` is the `Circuit`. See https://github.com/o1-labs/proof-systems/pull/869 for moving the prover precomputations into the prover index, which is probably what you intended.

mimoo:
looks like I should probably wait for this one to land: https://github.com/o1-labs/proof-systems/pull/866/files

PR #862, opened 2022-11-22
FFAdd constraint and witness optimization

querolita:
This PR brings an optimization to FFAdd reducing the number of constraints by 2 and reducing by 1 the number of witness cells needed. The idea is to condense claims about the low and high limbs in a longer 176-bit super-limb. This gets rid of the low carry flag (so only the old middle carry flag is in use now, together with its carry check constraint) and the three old result checks now become just 2.

I checked the tests and all works as it used to before. I added another helper to write FFAdd gadgets without leading rangecheck gates to allow for better chainability when we use this gate inside the ECDSA algorithm in combination with the others.

### DIFF
--- a/book/src/rfcs/ffadd.md
+++ b/book/src/rfcs/ffadd.md
@@ -148,6 +148,19 @@ Mathematically speaking, a subtraction within a field is no more than an additio
 
 Instead, our gate encodes subtractions and additions directly within the sign term that is multiplying the right input. This way, there is no need to check that the negated value is performed correctly (which would require an additional row for a potential `FFNeg` gate). 
 
+### Optimization
+
+Inspired by the halving approach in foreign field multiplication, an optimized version of the above gate results in a reduction by 2 in the number of constraints and by 1 in the number of witness cells needed. The main idea is to condense the claims about the low and middle limbs in one single larger limb of 176 bits, which fit in our native field. This way, we can get rid of the low carry flag, its corresponding carry check, and the three result checks become just two. These are the new equations:
+
+$$
+\begin{aligned}
+r_{bot} &= (a_0 + 2^{88}\cdot a_1) + s \cdot (b_0 + 2^88) - q \cdot (f_0 + 2^88 \cdot f_1) - c \cdot 2^{176} \\
+r_{top} &= a_2 + s \cdot b_2 - q \cdot f_2 + c 
+\end{aligned}
+$$
+
+with `r_top = r_2` and `c = c_1`. 
+
 ## Gadget
 
 The foreign field gadget will be composed by 4 sets of `RangeCheck` gadgets for witnesses $a, b, r, u$ accounting for 16 rows in total; followed by one row with `ForeignFieldAdd` gate type; a final `ForeignFieldAdd` with a `Zero` row. The first four rows constrain the range of the left input. The following four constrain the range of the right input. The next range check is for the result of the addition of left and right. Next, a final range check for the upper bound therm. Then the foreign field addition gate performs the actual addition. And the final foreign field addition gate (followed by a zero gate) takes care of the final upper bound operation. A total of 19 rows with 15 columns in Kimchi.
--- a/kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
+++ b/kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
@@ -83,6 +83,11 @@ use std::{array, marker::PhantomData};
 //~ final bound check is to make sure that the final result (`min_result`) is indeed the minimum one
 //~ (meaning less than the modulus).
 //~
+//~ A more optimized version of these constraints is able to reduce by 2 the number of constraints and
+//~ by 1 the number of witness cells needed. The idea is to condense the low and middle limbs in one longer
+//~ limb of 176 bits (which fits inside our native field) and getting rid of the low carry flag.
+//~ With this idea in mind, the sole carry flag we need is the one located between the middle and the high limbs.
+//~
 //~ You could lay this out as a double-width gate for chained foreign additions and a final row, e.g.:
 //~
 //~ | col | `ForeignFieldAdd`       | chain `ForeignFieldAdd` | final `ForeignFieldAdd` | final `Zero`      |
@@ -95,8 +100,8 @@ use std::{array, marker::PhantomData};
 //~ |   5 | `right_input_hi` (copy) |                         |  2^88           (check) |                   |
 //~ |   6 | `sign`           (copy) |                         |  1              (check) |                   |
 //~ |   7 | `field_overflow`        |                         |  1              (check) |                   |
-//~ |   8 | `carry_lo`              |                         | `bound_carry_lo`        |                   |
-//~ |   9 | `carry_mi`              |                         | `bound_carry_mi`        |                   |
+//~ |   8 | `carry`                 |                         | `bound_carry`           |                   |
+//~ |   9 |                         |                         |                         |                   |
 //~ |  10 |                         |                         |                         |                   |
 //~ |  11 |                         |                         |                         |                   |
 //~ |  12 |                         |                         |                         |                   |
@@ -118,7 +123,7 @@ where
     F: PrimeField,
 {
     const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::ForeignFieldAdd);
-    const CONSTRAINTS: u32 = 7;
+    const CONSTRAINTS: u32 = 5;
 
     fn constraint_checks<T: ExprOps<F>>(env: &ArgumentEnv<F, T>) -> Vec<T> {
         let foreign_modulus: [T; LIMB_COUNT] = array::from_fn(|i| env.foreign_modulus(i));
@@ -138,8 +143,7 @@ where
         let field_overflow = env.witness_curr(7);
 
         // Result carry bits for limb overflows / underflows.
-        let carry_lo = env.witness_curr(8);
-        let carry_mi = env.witness_curr(9);
+        let carry = env.witness_curr(8);
 
         let result_lo = env.witness_next(0);
         let result_mi = env.witness_next(1);
@@ -153,31 +157,30 @@ where
             (sign.clone() + T::one()) * (sign.clone() - T::one()),
         ];
 
-        // Constraints to check the carry flags are -1, 0, or 1.
-        checks.push(carry(&carry_lo));
-        checks.push(carry(&carry_mi));
+        // Constraints to check the carry flag is -1, 0, or 1.
+        checks.push(is_carry(&carry));
+        //checks.push(carry(&carry_mi));
 
         // Auxiliary inline function to obtain the constraints of a foreign field addition result
         let mut result = {
-            // r_0 = a_0 + s * b_0 - q * f_0 - 2^88 * c_0
-            let result_computed_lo = left_input_lo + sign.clone() * right_input_lo
-                - field_overflow.clone() * foreign_modulus[0].clone()
-                - carry_lo.clone() * two_to_limb.clone();
-            // r_1 = a_1 + s * b_1 - q * f_1 - 2^88 * c_1 + c_0
-            let result_computed_mi = left_input_mi + sign.clone() * right_input_mi
-                - field_overflow.clone() * foreign_modulus[1].clone()
-                - carry_mi.clone() * two_to_limb
-                + carry_lo;
-            // r_2 = a_2 + s * b_2 - q * f_2 + c_1
-            let result_computed_hi = left_input_hi + sign * right_input_hi
+            // a_bot = a_0 + a_1 * 2^88
+            // b_bot = b_0 + b_0 * 2^88
+            // f_bot = f_0 + f_1 * 2^88
+            // r_bot = a_bot + s * b_bot - q * f_bot - c * 2^176
+            let result_bot = two_limb(&left_input_lo, &left_input_mi)
+                + sign.clone() * two_limb(&right_input_lo, &right_input_mi)
+                - field_overflow.clone() * two_limb(&foreign_modulus[0], &foreign_modulus[1])
+                - carry.clone() * two_to_limb.clone() * two_to_limb.clone();
+            // r_top = a_2 + s * b_2 - q * f_2 + c
+            let result_top = left_input_hi + sign * right_input_hi
                 - field_overflow * foreign_modulus[2].clone()
-                + carry_mi;
-
+                + carry;
             // Result values match
+            // r_bot = r_0 + 2^88 * r_1
+            // r_top = r_2
             vec![
-                result_lo - result_computed_lo,
-                result_mi - result_computed_mi,
-                result_hi - result_computed_hi,
+                result_bot - two_limb(&result_lo, &result_mi),
+                result_top - result_hi,
             ]
         };
 
@@ -189,7 +192,12 @@ where
 }
 
 // Auxiliary function to obtain the constraints to check a carry flag
-fn carry<F: PrimeField, T: ExprOps<F>>(flag: &T) -> T {
+fn is_carry<F: PrimeField, T: ExprOps<F>>(flag: &T) -> T {
     // Carry bits are -1, 0, or 1.
     flag.clone() * (flag.clone() - T::one()) * (flag.clone() + T::one())
+}
+
+// Auxiliary function to obtain the composed value of the two lowest limbs of an element
+fn two_limb<F: PrimeField, T: ExprOps<F>>(lo_limb: &T, mi_limb: &T) -> T {
+    lo_limb.clone() + mi_limb.clone() * T::literal(F::from(TWO_TO_LIMB))
 }
--- a/kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
+++ b/kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
@@ -14,12 +14,11 @@ use crate::{
     circuits::{
         argument::{Argument, ArgumentType},
         constraints::ConstraintSystem,
-        expr::{self, l0_1, Environment, LookupEnvironment, E},
+        expr::{self, l0_1, Environment, LookupEnvironment},
         gate::{CircuitGate, CircuitGateError, CircuitGateResult, Connect, GateType},
         lookup::{
             self,
             lookups::{LookupInfo, LookupsUsed},
-            tables::{GateLookupTable, LookupTable},
         },
         polynomial::COLUMNS,
         wires::Wire,
@@ -28,9 +27,6 @@ use crate::{
 };
 
 use super::circuitgates::ForeignFieldAdd;
-
-/// Number of gates used by the foreign field addition gadget
-pub const GATE_COUNT: usize = 1;
 
 impl<F: PrimeField> CircuitGate<F> {
     /// Outputs next row
@@ -41,7 +37,7 @@ impl<F: PrimeField> CircuitGate<F> {
         *next_row = subsequent_row;
     }
 
-    /// Create foreign field addition gate
+    /// Create foreign field addition gate chain including optional range checks
     ///     Inputs
     ///         starting row
     ///         number of addition gates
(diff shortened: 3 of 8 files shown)


kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
```diff
@@ -64,20 +60,30 @@ impl<F: PrimeField> CircuitGate<F> {
...
-        // ----------------------------------------------
-        let mut circuit_gates = vec![];
-        let mut next_row = start_row;
-
-        CircuitGate::append_multi_range_check_rows(&mut next_row, &mut circuit_gates); // left input
-        for _ in 0..num {
-            for _ in 0..2 {
-                // right input and result
-                CircuitGate::append_multi_range_check_rows(&mut next_row, &mut circuit_gates);
+    pub fn create_ffadd_chain(
```
jspada, line 63: Nice, but I think you may just want to leave the range-checks always separate since we either have already range checked `a` and `b` and simply want to range check `r`.  Or don't want to range check `r`, but only want to range check `a` and `b`.

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
```diff
@@ -399,34 +406,3 @@ fn set_up_lookup_env_data<F: PrimeField>(
         joint_lookup_table_d8,
     })
 }
-
-/// Get array of foreign field addition circuit gate types
-pub fn circuit_gates() -> [GateType; GATE_COUNT] {
```
jspada, line 404: I think you should probably leave these interfaces and also use them from proof systems.  If we have all gates use them, then that means all gates/gadgets have the same interface.  This means that the proof system code is always be the same, which means that perhaps one day we can make the proof system hook up to all the gates in a generic way, rather than each differently.  I was sort of moving things in this direction by abstracting things in this way and then using them from proof systems.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -96,24 +107,46 @@ fn compute_subadd_values<F: PrimeField>(
...
+        output: result,
+        sign,
+        ovf: field_overflow,
+        carry: carry_bot,
+    }
+}
+
+// Returns the bottom composition of an element from its low and middle limbs
+fn bottom<F: PrimeField>(lo: &F, mi: &F) -> F {
+    *lo + *mi * F::from(TWO_TO_LIMB)
```
jspada, line 138: Once updated against master you should use `F::two_to_limb()`

Also, you should rename this to `two_limb` or `compact_limb` for consistency with constraints code.

Even better, you don't need to re-implement this function here. Thanks to `ExprOps`, you can just reuse `cirguitgates::two_limb()` everywhere.

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -189,7 +192,12 @@ where
...
 
 // Auxiliary function to obtain the constraints to check a carry flag
-fn carry<F: PrimeField, T: ExprOps<F>>(flag: &T) -> T {
+fn is_carry<F: PrimeField, T: ExprOps<F>>(flag: &T) -> T {
     // Carry bits are -1, 0, or 1.
     flag.clone() * (flag.clone() - T::one()) * (flag.clone() + T::one())
 }
+
+// Auxiliary function to obtain the composed value of the two lowest limbs of an element
+fn two_limb<F: PrimeField, T: ExprOps<F>>(lo_limb: &T, mi_limb: &T) -> T {
```
jspada, line 201: This might be better to call `compact_limb` it can also be reused in your witness code.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -127,74 +160,134 @@ pub fn create<F: PrimeField>(
...
-            compute_subadd_values(&left, &right, opcodes[i], &foreign_modulus);
-        // Create multi-range-check witness for right_input (left_input was done in previous iteration) and output
+        let add_values = compute_ffadd_values(&left, &right, opcodes[i], &foreign_modulus);
         // We only obtain the 3 lower limbs of right because the range check takes only 264 bits now
         let right_3_limb = ForeignElement::new([right[LO], right[MI], right[HI]]);
-        range_check::witness::extend(&mut witness, right_3_limb);
-        range_check::witness::extend(&mut witness, output.clone());
+        // Create multi-range-check witness for right_input (left_input was done in previous iteration) and output
+        if range_checks {
+            range_check::witness::extend(&mut witness, right_3_limb);
```
jspada, line 176: Once updated against master, you can just do `range_check::witness::extend(&mut witness, right.limbs);` and avoid the extra code of `let right_3_limb = ForeignElement::new([right[LO], right[MI], right[HI]]);`

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -127,74 +160,134 @@ pub fn create<F: PrimeField>(
...
     let right = ForeignElement::<F, 4>::from_biguint(BigUint::from(TWO_TO_LIMB).pow(3));
 
-    let (bound, sign, overflow, bound_carry_lo, bound_carry_mi) =
-        compute_subadd_values(&left, &right, FFOps::Add, &foreign_modulus);
+    let bound_values = compute_ffadd_values(&left, &right, FFOps::Add, &foreign_modulus);
     // Make sure they have the right value
-    assert_eq!(sign, F::one());
-    assert_eq!(overflow, F::one());
+    assert_eq!(bound_values.sign, F::one());
+    assert_eq!(bound_values.ovf, F::one());
```
jspada, line 190: Would be cool to assert the other possible carries that we think are zero actually are.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -127,74 +160,134 @@ pub fn create<F: PrimeField>(
...
-        );
+        if range_checks {
+            init_ff_add_rows_rc(&mut witness, offset, i, value.sign, value.ovf, value.carry);
+        } else {
+            let right = ForeignElement::new([value.right[LO], value.right[MI], value.right[HI]]);
+            init_ff_add_rows(
+                &mut witness,
+                offset,
+                value.left.clone(),
+                right,
```
jspada, line 219: Same as above,  now you can just do `value.right.limbs` and completely avoid the extra `let right = ForeignElement::new([value.right[LO], value.right[MI], value.right[HI]]);` line.

kimchi/src/prover.rs
```diff
@@ -632,12 +633,7 @@ where
             }
 
             if let Some(selector) = index.cs.foreign_field_add_selector_poly.as_ref() {
-                index_evals.extend(
-                    foreign_field_add::gadget::circuit_gates()
-                        .iter()
-                        .enumerate()
-                        .map(|(_, gate_type)| (*gate_type, &selector.eval8)),
-                );
+                index_evals.insert(GateType::ForeignFieldAdd, &selector.eval8);
```
jspada, line 636: Suggest to switch back to the "common interface" code, so one day this prover code can be made generic.

kimchi/src/prover.rs
```diff
@@ -775,8 +771,8 @@ where
             // foreign field addition
             {
                 if index.cs.foreign_field_add_selector_poly.is_some() {
-                    let ffadd = foreign_field_add::gadget::combined_constraints(&all_alphas)
-                        .evaluations(&env);
+                    let ffadd =
+                        ForeignFieldAdd::combined_constraints(&all_alphas).evaluations(&env);
```
jspada, line 775: Same.

kimchi/src/tests/foreign_field_add.rs
```diff
@@ -162,7 +160,7 @@ fn test_ffadd(
         .iter()
         .map(|x| BigUint::from_bytes_be(x))
         .collect::<Vec<BigUint>>();
-    let witness = witness::create(&inputs, ops, foreign_modulus);
+    let witness = witness::create_ffadd_chain_witness(&inputs, ops, foreign_modulus, true);
```
jspada, line 163: For reasons I described earlier, I would make this stuff with the test multi-range-checks part of the test code, rather than the actual gate's witness code.

kimchi/src/tests/foreign_field_add.rs
```diff
@@ -844,6 +842,38 @@ fn test_random_small_sub() {
...
+    let (mut witness, cs) = test_ffadd(
+        SECP256K1_MOD,
+        vec![left_input.clone(), right_input.clone()],
+        &vec![FFOps::Add],
+    );
+    // Modify sign of bound
+    // It should be constrained that sign needs to be 1
+    witness[6][17] = -PallasField::one();
+    assert_eq!(
+        cs.gates[17].verify_foreign_field_add::<Vesta>(0, &witness, &cs),
```
jspada, line 860: You should use `gate.verify_witness::<G>(..)` instead. 

Also this means you can verify which constraint it failed on.  Also, it's way faster to run!

kimchi/src/tests/foreign_field_add.rs
```diff
@@ -975,36 +996,55 @@ fn prove_and_verify_6() {
     prove_and_verify(6);
 }
 
-/*
+// Test with FFAdd gates without range checks
```
jspada, line 999: Similar to Matthew's PR review request for ffMul, I think we should test ffAdd on both of curves: `Pallas` and `Vesta`.

This can now be done using the `TestFramework`, so I'd just update those tests that you have using `TestFamework`, rather than all of them to run on both curves.  You can see the ffMul gate to see how it's done.

book/src/rfcs/ffadd.md
```diff
@@ -148,6 +148,19 @@ Mathematically speaking, a subtraction within a field is no more than an additio
 
 Instead, our gate encodes subtractions and additions directly within the sign term that is multiplying the right input. This way, there is no need to check that the negated value is performed correctly (which would require an additional row for a potential `FFNeg` gate). 
 
+### Optimization
+
+Inspired by the halving approach in foreign field multiplication, an optimized version of the above gate results in a reduction by 2 in the number of constraints and by 1 in the number of witness cells needed. The main idea is to condense the claims about the low and middle limbs in one single larger limb of 176 bits, which fit in our native field. This way, we can get rid of the low carry flag, its corresponding carry check, and the three result checks become just two. These are the new equations:
+
```
jspada, line 154: It would make things clearer to add a diagram like in ffMul's bound addition.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -96,24 +107,46 @@ fn compute_subadd_values<F: PrimeField>(
...
-    let carry_mi =
+    // c = [ (a1 * 2^88 + a0) + s * (b1 * 2^88 + b0) - q * (f1 * 2^88 + f0) - (r1 * 2^88 + r0) ] / 2^176
+    //  <=>
+    // c = r2 - a2 - s*b2 + q*f2
+
+    let carry_bot: F = (bottom(&left_input[LO], &left_input[MI])
+        + bottom(&right_input[LO], &right_input[MI]) * sign
+        - bottom(&foreign_modulus[LO], &foreign_modulus[MI]) * field_overflow
+        - bottom(&result[LO], &result[MI]))
+        / (two_to_limb * two_to_limb);
```
jspada, line 118: Once you update against master, for simplicity and `ExprOps` reusability you should replace all of these `two_to_limb` with T::two_to_limb()`.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -25,18 +25,29 @@ pub enum FFOps {
...
+fn compute_ffadd_values<F: PrimeField>(
     left_input: &ForeignElement<F, 3>,
     right_input: &ForeignElement<F, 4>,
     opcode: FFOps,
     foreign_modulus: &ForeignElement<F, 3>,
-) -> (ForeignElement<F, 3>, F, F, F, F) {
+) -> FFAddValues<F> {
     assert_ne!(opcode, FFOps::Mul);
 
     let two_to_limb = F::from(TWO_TO_LIMB);
```
jspada, line 53: No need for this anymore.

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -153,31 +157,30 @@ where
...
-            // r_2 = a_2 + s * b_2 - q * f_2 + c_1
-            let result_computed_hi = left_input_hi + sign * right_input_hi
+            // a_bot = a_0 + a_1 * 2^88
+            // b_bot = b_0 + b_0 * 2^88
+            // f_bot = f_0 + f_1 * 2^88
+            // r_bot = a_bot + s * b_bot - q * f_bot - c * 2^176
+            let result_bot = two_limb(&left_input_lo, &left_input_mi)
+                + sign.clone() * two_limb(&right_input_lo, &right_input_mi)
+                - field_overflow.clone() * two_limb(&foreign_modulus[0], &foreign_modulus[1])
+                - carry.clone() * two_to_limb.clone() * two_to_limb;
```
jspada, line 173: Instead of `two_to_limb.clone() * two_to_limb` you can now just do

`T::two_to_2limb()`

jspada (commented):
Really nice optimisation added a bunch of feedback.  Needs update against master before another pass.

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
jspada: or maybe just `a` or `b`

book/src/rfcs/ffadd.md
```diff
@@ -148,6 +148,19 @@ Mathematically speaking, a subtraction within a field is no more than an additio
...
+$$
+\begin{aligned}
+r_{bot} &= (a_0 + 2^{88}\cdot a_1) + s \cdot (b_0 + 2^88) - q \cdot (f_0 + 2^88 \cdot f_1) - c \cdot 2^{176} \\
```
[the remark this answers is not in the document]
querolita: and lacking b_1, good catch

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -189,7 +192,12 @@ where
...
 // Auxiliary function to obtain the constraints to check a carry flag
-fn carry<F: PrimeField, T: ExprOps<F>>(flag: &T) -> T {
+fn is_carry<F: PrimeField, T: ExprOps<F>>(flag: &T) -> T {
     // Carry bits are -1, 0, or 1.
     flag.clone() * (flag.clone() - T::one()) * (flag.clone() + T::one())
 }
+
+// Auxiliary function to obtain the composed value of the two lowest limbs of an element
+fn two_limb<F: PrimeField, T: ExprOps<F>>(lo_limb: &T, mi_limb: &T) -> T {
+    lo_limb.clone() + mi_limb.clone() * T::literal(F::from(TWO_TO_LIMB))
```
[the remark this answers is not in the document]
querolita: thanks!

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
querolita: I guess I mostly created this gadget for test purposes, but I presume we won't be using it a lot in a real use case. For that reason we also have the `create_single_ffadd` gadget, which will be much more convenient.

querolita: I guess you mean to have a completely separate gadget

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
querolita: I understand, but I am removing after Matthew's suggestion to delete this for single-gate gadgets

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
querolita: Ha! indeed, thanks!

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
querolita: do you mean if I change the input type of the init_ffadd_row function?

kimchi/src/prover.rs
querolita: As above, changed it after Matthew asked me to do it for single gate gadgets

kimchi/src/tests/foreign_field_add.rs
querolita: I see!

querolita: in this case the code duplication is so huge that i don't feel it is any elegant..

kimchi/src/tests/foreign_field_add.rs
querolita: Ah, nice then!

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -118,7 +123,7 @@ where
     F: PrimeField,
 {
     const ARGUMENT_TYPE: ArgumentType = ArgumentType::Gate(GateType::ForeignFieldAdd);
-    const CONSTRAINTS: u32 = 7;
+    const CONSTRAINTS: u32 = 5;
```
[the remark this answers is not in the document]
querolita: I am unsure about how to do this without duplicating the circuit gate definition, and in any case seems like a small theoretical optimization that seems nice to have but not super relevant in practice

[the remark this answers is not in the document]
jspada: I would just stash you changes.  Run the TestFramework test -> I think it outputs the runtime.  Record it.
Unstash your changes (i.e. get back to this opt) -> Run the TestFramework test -> Record it.

Paste the results in the description of this PR!

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
jspada: I just mean have your gadget/gate do the simplest thing.. just the ffAdd gate and witness alone, leaving the range checks out.  This will be simpler code for you and leave maximum flexibility to the user/caller of this gate.  I still propose that you have one test that tests the gate with the (external) range checks, but that's just test code.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
jspada: yup

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
querolita: Done! Now there's a simple chain function and external expansion to include rangechecks for tests

kimchi/src/tests/foreign_field_add.rs
querolita: Did some refactor so there is no code duplication and became modular

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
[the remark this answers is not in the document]
querolita: Oh only that? I thought you meant like a whole complex benchmark. I have just updated the description of this PR with those numbers.

kimchi/src/tests/foreign_field_add.rs
querolita: Done!

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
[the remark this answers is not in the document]
jspada: Awesome!  
Nah, I just meant simplest thing.

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
jspada: Ah, ok - he knows better! 👍🏼

kimchi/src/prover.rs
jspada: Let's go with what Matthew says.

kimchi/src/tests/foreign_field_add.rs
jspada: Great! 👍🏼

book/src/rfcs/foreign_field_add.md
```diff
@@ -148,6 +148,57 @@ Mathematically speaking, a subtraction within a field is no more than an additio
 
 Instead, our gate encodes subtractions and additions directly within the sign term that is multiplying the right input. This way, there is no need to check that the negated value is performed correctly (which would require an additional row for a potential `FFNeg` gate).
 
+### Optimization
```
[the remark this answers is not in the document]
querolita: mmm... I am not sure I know what out-of-sync-ness you're referring to...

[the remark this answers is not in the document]
jspada: the layout was not matching the code.  For example sign was in column 9 in RFC and code had it in 6 (IIRRC)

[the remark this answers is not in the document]
jspada: seems ok now though

PR #866, opened 2022-11-24
Move polynomials out of ConstraintSystem

mrmr1993:
This PR moves the polynomials out of the `ConstraintSystem` structure, and into sub-structures.

This is part of a longer term plan to remove them from `ConstraintSystem` entirely, and to place them in `ProverIndex`. The constraint system should define the layout of the circuit; these precomputations don't contribute to that goal.


mimoo (approved):
this is great

PR #870, opened 2022-11-25
Compute 'chunked' Lagrange commitments

mrmr1993:
This PR tweaks the SRS precomputations to calculate a 'chunked' representation of the Lagrange commitments.

The lack of this prevented the 'chunked' approach from working with public inputs. This brings us closer to a working version of the chunked prover.

### DIFF
--- a/kimchi/src/verifier.rs
+++ b/kimchi/src/verifier.rs
@@ -471,6 +471,9 @@ where
             proof.prev_challenges.len(),
         ));
     }
+    if proof.public.len() != index.public {
+        return Err(VerifyError::IncorrectPubicInputLength(index.public));
+    }
 
     //~ 1. Commit to the negated public input polynomial.
     let lgr_comm = index
@@ -478,20 +481,9 @@ where
         .lagrange_bases
         .get(&index.domain.size())
         .expect("pre-computed committed lagrange bases not found");
-    let com: Vec<_> = lgr_comm
-        .iter()
-        .map(|c| PolyComm {
-            unshifted: vec![*c],
-            shifted: None,
-        })
-        .take(index.public)
-        .collect();
-    let com_ref: Vec<_> = com.iter().collect();
-    if proof.public.len() != index.public {
-        return Err(VerifyError::IncorrectPubicInputLength(index.public));
-    }
+    let com: Vec<_> = lgr_comm.iter().take(index.public).collect();
     let elm: Vec<_> = proof.public.iter().map(|s| -*s).collect();
-    let public_comm = PolyComm::<G>::multi_scalar_mul(&com_ref, &elm);
+    let public_comm = PolyComm::<G>::multi_scalar_mul(&com, &elm);
     let public_comm = {
         index
             .srs()
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -33,11 +33,9 @@ use super::evaluation_proof::*;
 
 /// A polynomial commitment.
 #[serde_as]
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct PolyComm<C>
-where
-    C: CanonicalDeserialize + CanonicalSerialize,
-{
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(bound = "C: CanonicalDeserialize + CanonicalSerialize")]
+pub struct PolyComm<C> {
     #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
     pub unshifted: Vec<C>,
     #[serde_as(as = "Option<o1_utils::serialization::SerdeAs>")]
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -911,6 +911,58 @@ mod tests {
     }
 
     #[test]
+    fn test_chunked_lagrange_commitments() {
+        let n = 64;
+        let domain = D::<Fp>::new(n).unwrap();
+
+        let mut srs = SRS::<VestaG>::create(n / 2);
+        srs.add_lagrange_basis(domain);
+
+        let expected_lagrange_commitments: Vec<_> = (0..n)
+            .map(|i| {
+                let mut e = vec![Fp::zero(); n];
+                e[i] = Fp::one();
+                let p = Evaluations::<Fp, D<Fp>>::from_vec_and_domain(e, domain).interpolate();
+                srs.commit_non_hiding(&p, None)
+            })
+            .collect();
+
+        let computed_lagrange_commitments = srs.lagrange_bases.get(&domain.size()).unwrap();
+        for i in 0..n {
+            assert_eq!(
+                computed_lagrange_commitments[i],
+                expected_lagrange_commitments[i],
+            );
+        }
+    }
+
+    #[test]
+    fn test_offset_chunked_lagrange_commitments() {
+        let n = 64;
+        let domain = D::<Fp>::new(n).unwrap();
+
+        let mut srs = SRS::<VestaG>::create(n / 2 + 1);
+        srs.add_lagrange_basis(domain);
+
+        let expected_lagrange_commitments: Vec<_> = (0..n)
+            .map(|i| {
+                let mut e = vec![Fp::zero(); n];
+                e[i] = Fp::one();
+                let p = Evaluations::<Fp, D<Fp>>::from_vec_and_domain(e, domain).interpolate();
+                srs.commit_non_hiding(&p, Some(64))
+            })
+            .collect();
+
+        let computed_lagrange_commitments = srs.lagrange_bases.get(&domain.size()).unwrap();
+        for i in 0..n {
+            assert_eq!(
+                computed_lagrange_commitments[i],
+                expected_lagrange_commitments[i],
+            );
+        }
+    }
+
+    #[test]
     fn test_opening_proof() {
         // create two polynomials
         let coeffs: [Fp; 10] = array::from_fn(|i| Fp::from(i as u32));
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -53,7 +53,7 @@ where
     pub blinders: PolyComm<G::ScalarField>,
 }
 
-impl<A: Copy> PolyComm<A>
+impl<A: Clone> PolyComm<A>
 where
     A: CanonicalDeserialize + CanonicalSerialize,
 {
@@ -62,8 +62,8 @@ where
         F: FnMut(A) -> B,
         B: CanonicalDeserialize + CanonicalSerialize,
     {
-        let unshifted = self.unshifted.iter().map(|x| f(*x)).collect();
-        let shifted = self.shifted.map(f);
+        let unshifted = self.unshifted.iter().map(|x| f(x.clone())).collect();
+        let shifted = self.shifted.as_ref().map(|x| f(x.clone()));
         PolyComm { unshifted, shifted }
     }
 
@@ -607,22 +607,22 @@ impl<G: CommitmentCurve> SRS<G> {
         &self,
         domain: D<G::ScalarField>,
         plnm: &Evaluations<G::ScalarField, D<G::ScalarField>>,
-        max: Option<usize>,
     ) -> PolyComm<G> {
-        let is_zero = plnm.evals.par_iter().all(|x| x.is_zero());
-        let basis = match self.lagrange_bases.get(&domain.size()) {
+        let lagrange_basis = self.lagrange_bases.get(&domain.size());
+        let basis = match lagrange_basis.as_ref() {
             None => panic!("lagrange bases for size {} not found", domain.size()),
-            Some(v) => &v[..],
+            Some(v) => v,
+        };
+        let commit_evaluations = |evals: &Vec<G::ScalarField>, basis: &Vec<PolyComm<G>>| {
+            PolyComm::<G>::multi_scalar_mul(&basis.iter().collect::<Vec<_>>()[..], &evals[..])
         };
         match domain.size.cmp(&plnm.domain().size) {
             std::cmp::Ordering::Less => {
                 let s = (plnm.domain().size / domain.size) as usize;
                 let v: Vec<_> = (0..(domain.size())).map(|i| plnm.evals[s * i]).collect();
-                Self::commit_helper(&v[..], basis, None, is_zero, max)
+                commit_evaluations(&v, basis)
             }
-            std::cmp::Ordering::Equal => {
-                Self::commit_helper(&plnm.evals[..], basis, None, is_zero, max)
-            }
+            std::cmp::Ordering::Equal => commit_evaluations(&plnm.evals, basis),
             std::cmp::Ordering::Greater => {
                 panic!("desired commitment domain size greater than evaluations' domain size")
             }
@@ -633,10 +633,9 @@ impl<G: CommitmentCurve> SRS<G> {
         &self,
         domain: D<G::ScalarField>,
         plnm: &Evaluations<G::ScalarField, D<G::ScalarField>>,
-        max: Option<usize>,
         rng: &mut (impl RngCore + CryptoRng),
     ) -> BlindedCommitment<G> {
-        self.mask(self.commit_evaluations_non_hiding(domain, plnm, max), rng)
+        self.mask(self.commit_evaluations_non_hiding(domain, plnm), rng)
     }
 
     /// This function verifies batch of batched polynomial commitment opening proofs
@@ -900,10 +899,7 @@ mod tests {
                 let mut e = vec![Fp::zero(); n];
                 e[i] = Fp::one();
                 let p = Evaluations::<Fp, D<Fp>>::from_vec_and_domain(e, domain).interpolate();
-                let c = srs.commit_non_hiding(&p, None);
-                assert!(c.shifted.is_none());
-                assert_eq!(c.unshifted.len(), 1);
-                c.unshifted[0]
+                srs.commit_non_hiding(&p, None)
             })
             .collect();
 
@@ -911,7 +907,7 @@ mod tests {
         for i in 0..n {
             assert_eq!(
                 computed_lagrange_commitments[i],
-                expected_lagrange_commitments[i]
+                expected_lagrange_commitments[i],
             );
         }
     }
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -606,11 +606,10 @@ impl<G: CommitmentCurve> SRS<G> {
         domain: D<G::ScalarField>,
         plnm: &Evaluations<G::ScalarField, D<G::ScalarField>>,
     ) -> PolyComm<G> {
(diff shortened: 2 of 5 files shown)


mimoo:
> The lack of this prevented the 'chunked' approach from working with public inputs

what's the problem there?

poly-commitment/src/commitment.rs
```diff
@@ -62,8 +62,8 @@ where
         F: FnMut(A) -> B,
         B: CanonicalDeserialize + CanonicalSerialize,
     {
-        let unshifted = self.unshifted.iter().map(|x| f(*x)).collect();
-        let shifted = self.shifted.map(f);
+        let unshifted = self.unshifted.iter().map(|x| f(x.clone())).collect();
+        let shifted = self.shifted.as_ref().map(|x| f(x.clone()));
```
mimoo, line 64: I think you can just write `self.shifted.cloned().map(f)`

poly-commitment/src/commitment.rs
```diff
@@ -62,8 +62,8 @@ where
         F: FnMut(A) -> B,
         B: CanonicalDeserialize + CanonicalSerialize,
     {
-        let unshifted = self.unshifted.iter().map(|x| f(*x)).collect();
-        let shifted = self.shifted.map(f);
+        let unshifted = self.unshifted.iter().map(|x| f(x.clone())).collect();
```
mimoo, line 63: `self.unshifted.iter().cloned().map(f).collect()`?

mrmr1993:
> what's the problem there?

There wasn't a defined way to generate the chunked commitment by lagrange interpolation before, so attempting to generate the public input polynomial would use the 'unchunked' version. This would generate a polynomial of the incorrect degree: the lagrange polynomial should be `(x^n - 1)/(x - w^i)` (scaled by a constant when normalised), but instead we'd only be getting the polynomial terms from the first chunk.

Naturally, this will cause the public polynomial to modify the values at other entries than the desired `i`, causing verification to fail.

kimchi/src/verifier.rs
```diff
@@ -471,27 +471,19 @@ where
...
-        .map(|c| PolyComm {
-            unshifted: vec![*c],
-            shifted: None,
-        })
-        .take(index.public)
-        .collect();
-    let com_ref: Vec<_> = com.iter().collect();
-    if proof.public.len() != index.public {
-        return Err(VerifyError::IncorrectPubicInputLength(index.public));
-    }
```
mimoo, line 492: at this point I'm not sure I understand why this check is removed

mrmr1993: I moved it up, so that we don't waste time computing the commitments if the proof is malformed.

poly-commitment/src/commitment.rs
```diff
@@ -33,11 +33,9 @@ use super::evaluation_proof::*;
 
 /// A polynomial commitment.
 #[serde_as]
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct PolyComm<C>
-where
-    C: CanonicalDeserialize + CanonicalSerialize,
-{
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(bound = "C: CanonicalDeserialize + CanonicalSerialize")]
```
mimoo, line 37: ah interesting, IIUC this will do something like

```rust
impl<C> serde::Serialize for PolyComm<C> where C: CanonicalSerialize + CanonicalDeserialize { ... }
```

and as such as long as we don't (de)serialize the struct the bound won't matter. I guess we should do that for all of our structs then : o (we often just add `F: (Prime)Field` to satisfy serde).
```

File: poly-commitment/src/commitment.rs [hunk 5]
```diff
@@ -607,22 +607,22 @@ impl<G: CommitmentCurve> SRS<G> {
...
-        max: Option<usize>,
     ) -> PolyComm<G> {
-        let is_zero = plnm.evals.par_iter().all(|x| x.is_zero());
-        let basis = match self.lagrange_bases.get(&domain.size()) {
+        let lagrange_basis = self.lagrange_bases.get(&domain.size());
+        let basis = match lagrange_basis.as_ref() {
             None => panic!("lagrange bases for size {} not found", domain.size()),
-            Some(v) => &v[..],
+            Some(v) => v,
+        };
```
mimoo, line 613:
```
let basis = self.lagrange_bases.get(&domain.size()).unwrap_or_else(|| panic!("lagrange bases for size {} not found", domain.size()));
```

poly-commitment/src/commitment.rs
```diff
@@ -607,22 +607,22 @@ impl<G: CommitmentCurve> SRS<G> {
         &self,
         domain: D<G::ScalarField>,
         plnm: &Evaluations<G::ScalarField, D<G::ScalarField>>,
-        max: Option<usize>,
```
mimoo, line 610: at this point I'm not sure why this is removed. I know we don't use it but it seems like it's a worthy API to keep around

poly-commitment/src/commitment.rs
```diff
@@ -607,22 +607,22 @@ impl<G: CommitmentCurve> SRS<G> {
...
             std::cmp::Ordering::Less => {
                 let s = (plnm.domain().size / domain.size) as usize;
                 let v: Vec<_> = (0..(domain.size())).map(|i| plnm.evals[s * i]).collect();
-                Self::commit_helper(&v[..], basis, None, is_zero, max)
-            }
-            std::cmp::Ordering::Equal => {
-                Self::commit_helper(&plnm.evals[..], basis, None, is_zero, max)
+                commit_evaluations(&v, basis)
             }
+            std::cmp::Ordering::Equal => commit_evaluations(&plnm.evals, basis),
```
mimoo, line 623: if we remove `commit_helper` here I think it's only used in a single place, so might as well inline it

kimchi/src/verifier.rs
```diff
@@ -471,27 +471,19 @@ where
             proof.prev_challenges.len(),
         ));
     }
+    if proof.public.len() != index.public {
+        return Err(VerifyError::IncorrectPubicInputLength(index.public));
+    }
```
mimoo, line 476: oh right I see it now, we're missing a spec comment here but it's unrelated to this PR

poly-commitment/src/commitment.rs
mrmr1993: This is for evaluations. The coefficients version is still around, but the evaluations form never made any sense.

In my experience, reviewing commit by commit is nearly always unwise, unless the author explicitly recommends it. The whole point of this PR is the [final commit](https://github.com/o1-labs/proof-systems/pull/870/commits/9c82400b649340bbdc9c25deb2148a94062e15ba), things were only split this way so that I could test as I went.

mimoo: why wouldn't you want to enforce an upperbound on a polynomial because it's in evaluation form? Not sure I get it

mimoo: also now at this point I'm wondering why moving from `commit_helper` to `multi_scalar_mul` which doesn't do any chunking. What if we're trying to commit to a polynomial larger than the SRS?

poly-commitment/src/srs.rs
```diff
@@ -150,13 +145,75 @@ where
...
+        for i in 0..num_unshifteds {
+            // Initialize the vector with zero curve points
+            let mut lg: Vec<<G as AffineCurve>::Projective> =
+                vec![<G as AffineCurve>::Projective::zero(); n];
+            // Overwrite the terms corresponding to that chunk with the SRS curve points
+            let start_offset = i * srs_size;
+            let num_terms = min((i + 1) * srs_size, n) - start_offset;
+            for j in 0..num_terms {
+                lg[start_offset + j] = self.g[j].into_projective()
+            }
```
mimoo, line 180:
```rust
let range = start_offset..(start_offset + num_terms);
for (el, g) in lg[range].iter_mut().zip(&self.g) {
  *el = g.into_project()
}
```

mimoo: or can you just do something like this?

```rust
let gs = self.g.iter().map(|p| p.into_projective());
for (chunk, chunk_of_g) in lg.iter_mut().zip(gs).chunks(srs_size) {
  chunk.copy_from_slice(chunk_of_g[..chunk.len()]);
}
```

poly-commitment/src/commitment.rs
```diff
@@ -900,18 +897,67 @@ mod tests {
...
+                let p = Evaluations::<Fp, D<Fp>>::from_vec_and_domain(e, domain).interpolate();
+                srs.commit_non_hiding(&p, None)
+            })
+            .collect();
+
+        let computed_lagrange_commitments = srs.lagrange_bases.get(&domain.size()).unwrap();
+        for i in 0..n {
+            assert_eq!(
+                computed_lagrange_commitments[i],
+                expected_lagrange_commitments[i],
```
mimoo, line 934: BTW I'm still trying to understand why the redundancy in the lagrange basis. Here your lagrange base will have the same basis written twice (each half being the same):

```
computed_lagrange_commitments[n / 2] == computed_lagrange_commitments[0]
```

why do this instead of modifying the algorithm? As this is a waste of space no?

poly-commitment/src/srs.rs
mimoo: and you could perhaps even start by taking the last chunk and only using it if it's a multiple, otherwise use it for the shifted part

poly-commitment/src/srs.rs
```diff
@@ -150,13 +145,75 @@ where
...
+            <G as AffineCurve>::Projective::batch_normalization(lg.as_mut_slice());
+            // Append the 'partial Langrange polynomials' to the vector of unshifted chunks
+            unshifted.push(lg)
+        }
 
-        <G as AffineCurve>::Projective::batch_normalization(lg.as_mut_slice());
-        self.lagrange_bases
-            .insert(n, lg.iter().map(|g| g.into_affine()).collect());
+        // If the srs size does not exactly divide the domain size
+        let shifted: Option<Vec<<G as AffineCurve>::Projective>> =
```
mimoo, line 189: why shift the last chunk though?

mimoo (commented):
LGTM besides:

- It looks inefficient to me to repeat the basis
- I don't understand the shifted part

poly-commitment/src/commitment.rs
mrmr1993: The `multi_scalar_mul` is against the lagrange basis, which is now a `PolyComm`. That is to say: the lagrange basis encodes chunked commitments of size `domain.d1` and chunk size of `srs.len()`.

It's also worth mentioning: the old code was always broken. Attempting to 'chunk' some evaluations would *chunk the evaluations*, rather than the underlying polynomial, which is never what you want. Consider modifying an evaluation by `x` at the index `i`. The contribution of this evaluation to the entire polynomial is `x * k_i * (x^n - 1) / (x - w^i)`, where `k_i` is the Lagrange polynomial normalisation factor.

So, when chunking correctly, every evaluation will have an effect on every chunk, due to the `x^(n-1)` terms in that expression above. However, chunking the evaluations results in each evaluation only affecting the chunk matching its index, resulting in a junk commitment.

poly-commitment/src/commitment.rs
mrmr1993: I'm not able to do this. If I write it this way, `f` gets moved, and so I can't use it for `shifted`.

poly-commitment/src/commitment.rs
mrmr1993: I'm not able to do this. If I write it this way, `f` gets moved, and so I can't use it for `shifted`.

poly-commitment/src/commitment.rs
mrmr1993: This assertion is untrue. You can confirm yourself by injecting it into the test
```rust
assert_eq!(computed_lagrange_commitments[n / 2], computed_lagrange_commitments[0]);
```

poly-commitment/src/commitment.rs
mrmr1993: In case it's helpful, lets work through an example with a domain size of 4 and a srs size of 2. I'll use complex numbers (1, i, -1, -i) so that I don't need to remember what sqrt(-1) is in the pasta curves.

Suppose we have the degree 4 polynomial
```
f(x) = 1 + 2 x + 3 x^2 + 4 x^3
```
and thus evaluations
```
[f(1), f(i), f(-1), f(-i)]
```
which we can compute directly
```
[10, -2 - 2i, -2, -2 + 2i]
```

The chunked version of `f` has 2 parts:
```
f_low(x) = 1 + 2 x
f_high(x) = 3 + 4 x
```

We can also compute the Lagrange normalisation factors for each term, those being `1/l_i(w^i)` where `l_i(x) = (x^4 - 1)/(x - w^i)`. Recall that `x^4 - 1 = (x - w)(x - w^2)(x - w^3)(x - w^4)`.

```
l_0(1) = (1 - i)(1 + i)(1 + 1) = 2(2 + 1) = 6
l_1(i) = (i - 1)(i + i)(i + 1) = 2i(-1 - 1) = -4i
l_2(-1) = (-1 - 1)(-1 - i)(-1 + i) = -2(1 + 1) = -4
l_3(-i) = (-i - 1)(-i - i)(-i + 1) = -2i(-1 - 1) = 4i
```

Correspondingly, if we change the first evaluation, e.g. by increasing it by 6, we modify the original polynomial by `6/l_0(w^0) * (x^4 - 1)/(x - w^0)`, which expands to
```
  1/1 * (x^4                                           - 1)     / (x - 1)
=       (x^4 - x^3   + x^3 - x^2   + x^2 - x^1   + x^1 - x^0)   / (x - 1)
=       (x^3 (x - 1) + x^2 (x - 1) + x^1 (x - 1) + x^0 (x - 1)) / (x - 1)
=       (x^3         + x^2         + x^1         + x^0)
```

We can check this by adding this into the original polynomial:
```
f(x) = 1 + 2 x + 3 x^2 + 4 x^3
g(x) = 2 + 3 x + 4 x^2 + 5 x^3
```
and evaluating `g` at the domain
```
[g(1), g(i), g(-1), g(-i)] = [16, -2 - 2i, -2, -2 + 2i]
```

------------------------

Chunking `g(x)` gives
```
g_low(x) = 2 + 3 x
g_high(x) = 4 + 5 x
```

So we have that
* the lagrange polynomial `L_1(x) = 1/6 (x^3 + x^2 + x^1 + x^0)`
* the chunked lagrange polynomials are `L_1_low(x) = 1/6 (1 + x)`, `L_1_high(x) = 1/6 (1 + x)`.

The same applies for each of the other evaluations: we need to consider their impact *on each chunk of the polynomial* in order to capture their effect on the entire polynomial faithfully. We then use the ifft trick to compute these contributions efficiently as commitments.

Since commitments are additively homomorphic (equiv. a vector field representing the polynomials), we can do a multi-scalar-multiplication to generate a commitment to the entire polynomial from the evaluations and the commitments to the corresponding chunked lagrange polynomials.

poly-commitment/src/srs.rs
mrmr1993: Our chunk size is always `let k = srs.g.len()`. If that doesn't divide the domain size, we want to ensure that the prover hasn't committed to a polynomial larger than the domain. Thus, we need to do an 'upper bound check' by shifting all the way to the right.

Consider a chunked polynomial commitment
```
PolyComm { shifted: Some(s_comm), unshifted: [a_comm, b_comm, c_comm, d_comm] }
```
which corresponds to some evaluations at a point `p` of
```
[a_p, b_p, c_p, d_p]
```

So, we can check using an opening proof that
* `a_comm` opens at `a`,
* `b_comm` opens at `b`,
* `c_comm` opens at `c`,
* `d_comm` opens at `d`

to confirm that the user has committed to some polynomial of maximum degree `4 * k` which evaluates at `p` to `a + p^k * b + p^(2k) * c + p^(3k) * d`. However, if `k` doesn't divide the domain size, the prover's polynomial may be too large, which may allow them to create an invalid proof by cheating the vanishing polynomial division check.

To circumvent this, we make sure that the maximum degree of `d_comm` is `n%k` by providing `s_comm` that opens at `p` to `d * p^{k - n%k}`.

Since `p` is sampled randomly after `d_comm` and `s_comm` are committed to, we know that, with overwhelming probability, for that relationship to hold true we must have
```
d_comm = comm(d(X))
s_comm = comm(X^{k - n%k} * d(X))
```
and thus the prover could only have constructed both commitments if there are commitments to the `deg(d(X)) + k - n%k` powers of `X` available in the srs. Equivalently, `deg(d(X)) + k - n%k <= k`, and so `deg(d(X) <= n%k`.

Putting it all together, we have a polynomial of degree `3*k + n%k = n`, as desired.

poly-commitment/src/srs.rs
mrmr1993: I see you've done it [here](https://github.com/o1-labs/proof-systems/pull/887). Thanks!

poly-commitment/src/commitment.rs
mimoo: yeah it does :o https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned-1

poly-commitment/src/srs.rs
mimoo: did it there for a different function :D but I can take this one after this lands

poly-commitment/src/commitment.rs
mrmr1993: Wouldn't compile for me locally :thinking: Guess it's a recent addition, don't really want to deal with the rust bump in this PR.

mimoo: oh yeah that's a nit so no worries

poly-commitment/src/srs.rs
```diff
@@ -150,13 +145,75 @@ where
...
+            if num_unshifteds * self.g.len() == n {
+                None
+            } else {
+                // Initialize the vector to zero
```
jspada, line 193: Just to make sure the code obeys the theory, are there any test cases where `k` doesn't divide `n`?

In particular the edge cases of `k - n%k = 1` and `k - n%k = k - 1`

jspada (approved):
Firstly, this is pretty amazing work.  Recognising that you are able to get the commitments to the normalized lagrange polynomials by applying the IFFT to the commitments of the standard monomial basis polynomials (thanks to homomorphic property) is pretty mind blowing, for me anyway.   Similarly, applying it smoothly to the chunked setting is quite impressive and elegant imho.  I also like that this includes a fix to prevent potential misuse that could cause accidental invalid chunking of evaluations.

I've checked this over are carefully as I'm able given constraints and it looks good.  I proposed some suggestions and feedback.

There is a lot to absorb here and it is extremely important, so I think it definitely requires multiple reviewers.  If I find some more time I will examine it further.

In addition to the other reviewers, I think @querolita would be a very good perform to check it over carefully.

poly-commitment/src/commitment.rs
mimoo: oh yeah I think I meant

```
computed_lagrange_commitments[0].unshifted[0] == computed_lagrange_commitments[0].unshifted[1]
```

mimoo: mm, for some reason it's not always true for all index

Issue #374, opened 2022-03-12
max_poly_size and max_quot_size


mimoo:
Where are these values used? They are stored in both indexes as:

* `let max_quot_size = PERMUTS * cs.domain.d1.size as usize;`. Seems unused. I believe we used to use this to place an exact upperbound on `t`, but a re-read of the paper and talking with @arielgabizon seemed to indicate that we didn't need to enforce an upperbound on `t`. Also, we need to implement https://github.com/o1-labs/proof-systems/issues/147 anyway
* `let max_poly_size = srs.g.len();`. Looks useless to me since you can just call `srs.g.len()`

Sounds like a good idea to remove these values

duguorong009:
@mimoo @mrmr1993 
Please review the PR #877 I created for this issue.
In the PR, I only removed the `max_quot_size` field, not the `max_poly_size` field.

The reason I remain the `max_poly_size` is that it is somewhat useful when creating/initializing the `VerifierIndex`.
For example, there is a code part where the `srs` field could be empty & initialized by this `max_poly_size` field.
https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/verifier_index.rs#L288-L297
I studied the `VerifierIndex` codes & found that there could be the case that the `srs` field is empty - that is when the `VerifierIndex` is initialized from file.https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/verifier_index.rs#L314-L345
Here, if the `srs` param is `None` value, then the `srs` field of `VerifierIndex` is empty.

Please let me know your ideas on this study.

duguorong009:
@mimoo 
Maybe, can we close this? 
The PR is already merged. :smile_cat:

PR #877, opened 2022-11-26
Remove some unnecessary fields from indexes(both `prover` & `verifier`)

duguorong009:
**What's been done:**
  - Remove the `max_quot_size` field from both `prover_index` & `verifier_index`

Reference: #374


mimoo (approved):
awesome thanks a lot : )

mimoo:
making sure this works on the mina side: https://github.com/MinaProtocol/mina/pull/12324

PR #885, opened 2022-11-28
[commitment] rewrite MSM function over polycomm to make it clearer

mimoo:
clearer, but also it returns early when it can, and it only computes the `into_repr()` of the scalars once

### DIFF
--- a/poly-commitment/src/commitment.rs
+++ b/poly-commitment/src/commitment.rs
@@ -34,10 +34,8 @@ use super::evaluation_proof::*;
 /// A polynomial commitment.
 #[serde_as]
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct PolyComm<C>
-where
-    C: CanonicalDeserialize + CanonicalSerialize,
-{
+#[serde(bound = "C: CanonicalDeserialize + CanonicalSerialize")]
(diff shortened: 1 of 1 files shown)


poly-commitment/src/commitment.rs
```diff
@@ -213,44 +217,51 @@ impl<C: AffineCurve> PolyComm<C> {
...
+                .zip(&all_scalars)
+                // get rid of scalars that don't have an associated chunk
+                .filter_map(|(com, scalar)| com.unshifted.get(chunk).map(|c| (c, scalar)))
+                .unzip();
+
+            let chunk_msm = VariableBaseMSM::multi_scalar_mul::<C>(&points, &scalars);
+            unshifted.push(chunk_msm.into_affine());
+        }
+
+        let shifted_pairs: Vec<_> = com
```
mrmr1993, line 250: It's best not to allocate a `Vec` just to check whether the iterator is empty. You probably want `.peekable()` here and `.peek().is_some()` below.

mrmr1993 (approved):

poly-commitment/src/commitment.rs
mimoo: good point!

PR #889, opened 2022-11-29
Allow feature-flagged gates to be exposed to OCaml

mrmr1993:
This PR
* adds an `EnabledIf` constructor for `Expr` (and a `SkipIf` counterpart for the `PolishToken` variant)
  - the implementation for these is still to-do; they're not used by anything here, so everything still works
* tweaks the `linearization` functions to accept an `Option<FeatureFlags>`, which uses the `EnabledIf` form if no feature flags are given

This then generates code on the OCaml side that can be used to (optionally) enable the logic for these gates, making it easier to update pickles for new gate kinds.

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -447,7 +447,7 @@ pub enum FeatureFlag {
 }
 
 impl FeatureFlag {
-    pub fn is_enabled(&self) -> bool {
+    fn is_enabled(&self) -> bool {
         todo!("Handle features")
     }
 }
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -2257,7 +2257,7 @@ where
                 id.var_name()
             }
             EnabledIf(feature, e) => {
-                format!("enabled_if({:?}, (fun () -> {})", feature, e.ocaml(cache))
+                format!("enabled_if({:?}, (fun () -> {}))", feature, e.ocaml(cache))
             }
         }
     }
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -1837,11 +1837,13 @@ impl<F: Neg<Output = F> + Clone + One + Zero + PartialEq> Expr<F> {
                 mul_monomials(&x, &x)
             }
             EnabledIf(feature, x) => {
-                let e = x.monomials(ev);
-                HashMap::from_iter(
-                    e.into_iter()
-                        .map(|(m, c)| (m, Expr::EnabledIf(*feature, Box::new(c)))),
-                )
+                let mut monomials = x.monomials(ev);
+                for expr in monomials.values_mut() {
+                    let mut unflagged = Expr::zero();
+                    std::mem::swap(expr, &mut unflagged);
+                    *expr = Expr::EnabledIf(*feature, Box::new(unflagged))
+                }
+                monomials
             }
         }
     }
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -551,61 +551,61 @@ impl<F: FftField> PolishToken<F> {
         for t in toks.iter() {
             if skip_count > 0 {
                 skip_count -= 1;
-            } else {
-                use PolishToken::*;
-                match t {
-                    Alpha => stack.push(c.alpha),
-                    Beta => stack.push(c.beta),
-                    Gamma => stack.push(c.gamma),
-                    JointCombiner => {
-                        stack.push(c.joint_combiner.expect("no joint lookup was expected"))
+                continue;
+            }
+            use PolishToken::*;
+            match t {
+                Alpha => stack.push(c.alpha),
+                Beta => stack.push(c.beta),
+                Gamma => stack.push(c.gamma),
+                JointCombiner => {
+                    stack.push(c.joint_combiner.expect("no joint lookup was expected"))
+                }
+                EndoCoefficient => stack.push(c.endo_coefficient),
+                Mds { row, col } => stack.push(c.mds[*row][*col]),
+                ForeignFieldModulus(i) => {
+                    if let Some(modulus) = c.foreign_field_modulus.clone() {
+                        stack.push(ForeignElement::<F, 3>::from_biguint(modulus.clone())[*i])
                     }
-                    EndoCoefficient => stack.push(c.endo_coefficient),
-                    Mds { row, col } => stack.push(c.mds[*row][*col]),
-                    ForeignFieldModulus(i) => {
-                        if let Some(modulus) = c.foreign_field_modulus.clone() {
-                            stack.push(ForeignElement::<F, 3>::from_biguint(modulus.clone())[*i])
-                        }
-                    }
-                    VanishesOnLast4Rows => stack.push(eval_vanishes_on_last_4_rows(d, pt)),
-                    UnnormalizedLagrangeBasis(i) => {
-                        stack.push(unnormalized_lagrange_basis(&d, *i, &pt))
-                    }
-                    Literal(x) => stack.push(*x),
-                    Dup => stack.push(stack[stack.len() - 1]),
-                    Cell(v) => match v.evaluate(evals) {
-                        Ok(x) => stack.push(x),
-                        Err(e) => return Err(e),
-                    },
-                    Pow(n) => {
-                        let i = stack.len() - 1;
-                        stack[i] = stack[i].pow(&[*n as u64]);
-                    }
-                    Add => {
-                        let y = stack.pop().ok_or(ExprError::EmptyStack)?;
-                        let x = stack.pop().ok_or(ExprError::EmptyStack)?;
-                        stack.push(x + y);
-                    }
-                    Mul => {
-                        let y = stack.pop().ok_or(ExprError::EmptyStack)?;
-                        let x = stack.pop().ok_or(ExprError::EmptyStack)?;
-                        stack.push(x * y);
-                    }
-                    Sub => {
-                        let y = stack.pop().ok_or(ExprError::EmptyStack)?;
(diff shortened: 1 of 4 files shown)


kimchi/src/linearization.rs
```diff
@@ -151,10 +186,11 @@ pub fn linearization_columns<F: FftField + SquareRootField>(
 ///
 /// Will panic if the `linearization` process fails.
 pub fn expr_linearization<F: PrimeField + SquareRootField>(
-    feature_flags: &FeatureFlags<F>,
+    feature_flags: Option<&FeatureFlags<F>>,
```
mimoo, line 189: I'm not sure I understand why this is better. Now the code above is quite more complex. What's the rationale behind passing a `None` being better instead of a `FeatureFlags::default()`?

mimoo: ok I think I'm starting to understand, you're choosing to decide how the expression looks like at compile time (Some) or at runtime (None)

kimchi/src/circuits/expr.rs
```diff
@@ -2220,6 +2305,7 @@ where
                 cache.insert(*id, e.as_ref().clone());
                 id.latex_name()
             }
+            EnabledIf(feature, _) => format!("{:?}", feature),
```
mimoo, line 2308: you should recurse here no?

kimchi/src/circuits/expr.rs
```diff
@@ -1765,6 +1840,13 @@ impl<F: Neg<Output = F> + Clone + One + Zero + PartialEq> Expr<F> {
                 let x = x.monomials(ev);
                 mul_monomials(&x, &x)
             }
+            EnabledIf(feature, x) => {
+                let e = x.monomials(ev);
+                HashMap::from_iter(
+                    e.into_iter()
+                        .map(|(m, c)| (m, Expr::EnabledIf(*feature, Box::new(c)))),
```
mimoo, line 1847: can't you just do something like that:

```rust
let mut map = x.monomials(ev);
for expr in map.values_mut() {
  *expr = Expr::EnabledIf(*feature, Box::new(expr))
}
map
```

kimchi/src/linearization.rs
mimoo: maybe good to add some comment somewhere to say exactly this ^, this would help users of this function

kimchi/src/circuits/expr.rs
```diff
@@ -1257,6 +1295,17 @@ impl<F: FftField> Expr<ConstantExpr<F>> {
...
+            Expr::EnabledIf(feature, e) => {
+                let tok = PolishToken::SkipIf(*feature, 0);
+                res.push(tok);
+                let len_before = res.len();
+                /* Clone the cache, to make sure we don't try to access cached statements later
+                when the feature flag is off. */
+                let mut cache = cache.clone();
+                e.to_polish_(&mut cache, res);
+                let len_after = res.len();
+                res[len_before - 1] = PolishToken::SkipIf(*feature, len_after - len_before);
```
mimoo, line 1307: I would find it cleaner to create a temporary vec, and then extend `res` once we know the length. I always find code that goes back to modify a buffer a bit scary.

kimchi/src/circuits/expr.rs
```diff
@@ -1257,6 +1295,17 @@ impl<F: FftField> Expr<ConstantExpr<F>> {
                     }
                 }
             }
+            Expr::EnabledIf(feature, e) => {
+                let tok = PolishToken::SkipIf(*feature, 0);
```
mimoo, line 1299: sounds like this should be called `SkipIfNot` rather no?

kimchi/src/circuits/expr.rs
```diff
@@ -520,56 +545,68 @@ impl<F: FftField> PolishToken<F> {
...
-                JointCombiner => {
-                    stack.push(c.joint_combiner.expect("no joint lookup was expected"))
-                }
-                EndoCoefficient => stack.push(c.endo_coefficient),
-                Mds { row, col } => stack.push(c.mds[*row][*col]),
-                ForeignFieldModulus(i) => {
-                    if let Some(modulus) = c.foreign_field_modulus.clone() {
-                        stack.push(ForeignElement::<F, 3>::from_biguint(modulus.clone())[*i])
+            if skip_count > 0 {
+                skip_count -= 1;
```
mimoo, line 552: to avoid the indentation of the `else { ... }` you should just write `continue;` here

kimchi/src/circuits/expr.rs
```diff
@@ -468,7 +488,10 @@ pub enum PolishToken<F> {
     Gamma,
     JointCombiner,
     EndoCoefficient,
-    Mds { row: usize, col: usize },
+    Mds {
+        row: usize,
+        col: usize,
+    },
```
mimoo, line 494: I'm amazed that `cargo fmt` is not complaining :P

kimchi/src/circuits/expr.rs
```diff
@@ -433,6 +433,25 @@ impl Op2 {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(
+    feature = "ocaml_types",
+    derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::Enum)
+)]
+pub enum FeatureFlag {
+    ChaCha,
+    RangeCheck,
+    ForeignFieldAdd,
+    Xor,
```
mimoo, line 446: every tag should be documented here :o

kimchi/src/circuits/expr.rs
```diff
@@ -456,6 +475,7 @@ pub enum Expr<C> {
     UnnormalizedLagrangeBasis(i32),
     Pow(Box<Expr<C>>, u64),
     Cache(CacheId, Box<Expr<C>>),
+    EnabledIf(FeatureFlag, Box<Expr<C>>),
```
mimoo, line 478: should be documented

kimchi/src/circuits/expr.rs
```diff
@@ -520,56 +545,68 @@ impl<F: FftField> PolishToken<F> {
...
+                    }
+                    Store => {
+                        let x = stack[stack.len() - 1];
+                        cache.push(x);
+                    }
+                    Load(i) => stack.push(cache[*i]),
+                    SkipIf(feature, count) => {
+                        if !feature.is_enabled() {
+                            skip_count = *count;
+                            stack.push(F::zero());
```
mimoo, line 606: alternatively up there you can write:

```rust
while let Some(t) = toks.next() {
  // ...
    Skipif(feature, count) if !feature.is_enabled() => {
      toks.skip(count);
```

not sure I understand the `stack.push(F::zero())` yet though (still reading)

mimoo: interesting to see that this wouldn't work if you have a feature you're multiplying with something else (as pushing zero on the stack would cancel the feature). I guess we're not expecting this to happen but maybe good to keep in mind?

kimchi/src/circuits/expr.rs
```diff
@@ -612,6 +649,7 @@ impl<C> Expr<C> {
             }
             Pow(e, d) => d * e.degree(d1_size),
             Cache(_, e) => e.degree(d1_size),
+            EnabledIf(_, e) => e.degree(d1_size),
```
mimoo, line 652: not true no? The degree would be 0 if not enabled

kimchi/src/circuits/expr.rs
```diff
@@ -1327,6 +1377,13 @@ impl<F: FftField> Expr<ConstantExpr<F>> {
             UnnormalizedLagrangeBasis(i) => Ok(unnormalized_lagrange_basis(&d, *i, &pt)),
             Cell(v) => v.evaluate(evals),
             Cache(_, e) => e.evaluate_(d, pt, evals, c),
+            EnabledIf(feature, e) => {
+                if feature.is_enabled() {
+                    e.evaluate_(d, pt, evals, c)
+                } else {
+                    Ok(F::zero())
+                }
+            }
```
mimoo, line 1386:
```rust
EnabledIf(feature, e) if feature.is_enabled() => e.evaluate_(d, pt, evals, c),
EnabledIf(feature, e) => Ok(F::zero()),
```

shorter?

kimchi/src/circuits/expr.rs
```diff
@@ -1689,6 +1763,7 @@ impl<F: Neg<Output = F> + Clone + One + Zero + PartialEq> Expr<F> {
             VanishesOnLast4Rows => true,
             UnnormalizedLagrangeBasis(_) => true,
             Cache(_, x) => x.is_constant(evaluated),
+            EnabledIf(_, x) => x.is_constant(evaluated),
```
mimoo, line 1766: I would think that this should also test if it's enabled before

mimoo: oh wait I guess at this point we cannot IIUC

mimoo (approved):

kimchi/src/circuits/expr.rs
mrmr1993: It did this automatically, it wasn't me..

mimoo: maybe the latest rust update

kimchi/src/circuits/expr.rs
mrmr1993: Disagree, the variant names are descriptive enough.

kimchi/src/circuits/expr.rs
mrmr1993: The idea here is that we declare the terms opaquely here, then state the full constraints elsewhere, so that we're not inlining potentially disabled constraints.

kimchi/src/circuits/expr.rs
mrmr1993: I don't think that bloating the code to appease a CI rule is a good way to go. It's supposed to work for us, not the other way around.

PR #892, opened 2022-12-01
Don't create a shifted segment when the domain size is smaller than srs

mrmr1993:
This PR fixes an oversight in #870, which is triggered by proofs on the Mina side with very small circuit sizes.

### DIFF
--- a/poly-commitment/src/srs.rs
+++ b/poly-commitment/src/srs.rs
@@ -187,7 +187,7 @@ where
 
         // If the srs size does not exactly divide the domain size
         let shifted: Option<Vec<<G as AffineCurve>::Projective>> =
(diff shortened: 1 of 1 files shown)


poly-commitment/src/srs.rs
```diff
@@ -187,7 +187,7 @@ where
 
         // If the srs size does not exactly divide the domain size
         let shifted: Option<Vec<<G as AffineCurve>::Projective>> =
-            if num_unshifteds * self.g.len() == n {
+            if n < srs_size || num_unshifteds * srs_size == n {
```
jspada, line 190: May be good to add test (in proof systems code base) for this case?

jspada (approved):
Suggest adding a test for this case so it can be caught quicker.

Issue #693, opened 2022-07-23
Remove endo from SRS


mimoo:
the [SRS struct](https://github.com/o1-labs/proof-systems/blob/master/poly-commitment/src/srs.rs#L30) has two endo fields `endo_r` and `endo_q`:

```rust
pub struct SRS<G: CommitmentCurve> {
    /// The vector of group elements for committing to polynomials in coefficient form
    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
    pub g: Vec<G>,
    /// A group element used for blinding commitments
    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
    pub h: G,

    // TODO: the following field should be separated, as they are optimization values
    /// Commitments to Lagrange bases, per domain size
    #[serde(skip)]
    pub lagrange_bases: HashMap<usize, Vec<G>>,
    /// Coefficient for the curve endomorphism
    #[serde(skip)]
    pub endo_r: G::ScalarField,
    /// Coefficient for the curve endomorphism
    #[serde(skip)]
    pub endo_q: G::BaseField,
}
```

We should be able to remove these fields, as they are part of the `KimchiCurve` trait now: https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/curve.rs#L25

vivanraaj:
the "SRS.endo_r" and "SRS.endo_q" are used in https://github.com/o1-labs/proof-systems/blob/master/poly-commitment/src/evaluation_proof.rs#L244 and https://github.com/o1-labs/proof-systems/blob/master/poly-commitment/src/evaluation_proof.rs#L274. 

does the whole system design allows the kimchi trait to be imported into the files mentioned above to provide the "endo_r" and "endo_q" variables?

mimoo:
No because poly-commitment is a dependency of kimchi, and so you'd get a dependency cycle if you would do that. 

I guess there's two solutions:

1. Moving the `KimchiCurve` trait to a crate that can be imported by both would fix that (for example mina-curves).
2. Pass these endo values as argument in the functions that need them

I liked solution 2 initially, as I was thinking that a poly-commitment library should only be aware of one curve, and not two, but I guess our poly-commitment might be aware of that. So maybe solution 1 is better here. cc @mrmr1993 who might have a better opinion

duguorong009:
@mimoo 
I have one question about the issue.
[...]

PR #895, opened 2022-12-02
Remove `endo`-related fields from `SRS` struct

duguorong009:
**What's been done:**
Remove the `endo_r` & `endo_q` fields from `SRS` struct since they are irrelevant.

Reference: #693


mimoo:
This looks great! Can you also update the book to remove endos from the URS here: https://github.com/o1-labs/proof-systems/blob/master/book/src/specs/urs.md#endo-stuff ? I'm guessing that they should be documented next in https://github.com/o1-labs/proof-systems/blob/master/book/src/specs/poly-commitment.md but that's outside the scope of this PR :o

duguorong009:
Thanks! @mimoo 
Yes, I will try to update the document as recommended.

poly-commitment/src/evaluation_proof.rs
```diff
@@ -96,6 +96,8 @@ impl<G: CommitmentCurve> SRS<G> {
         RNG: RngCore + CryptoRng,
         G::BaseField: PrimeField,
     {
+        let (endo_q, endo_r) = endos::<G>();
```
mimoo, line 99: btw I think `G::endos()` would have been cleaner but I'm nit picking

PR #897, opened 2022-12-03
add a constant for the number of coefficient columns

mimoo:
we should probably wait for [that series of PR](https://github.com/o1-labs/proof-systems/pull/833) to land before landing this one


querolita (approved):
This helps in case the number of coeffs was different than that of columns.

Issue #886, opened 2022-11-29
create helper functions to move between domain size


mimoo:
In https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/domains.rs#L1 we define 4 domains of different sizes: d1, d2, d4, and d8. Each domain having twice as many elements are the previous one.

Evaluations in a specific domain size can quickly be casted in a _smaller_ domain by taking all the i-th elements, where i is the difference of size between the two domains. For example to go from d4 to d1, you'd pick every 4th element, and to go from d8 to d1 you'd pick every 8th element.

This results in a _lot_ of code that looks like this and is hard to parse:

```rust
let s = (plnm.domain().size / domain.size) as usize;
let v: Vec<_> = (0..(domain.size())).map(|i| plnm.evals[s * i]).collect();
```

sometimes it is even interlaced with some scaling:

```rust
x + evals.evals[(scale * i + (domain as usize) * shift) % evals.evals.len()]
```

it would be good to create helper functions to hide the low-level manipulation and make the code around it easier to parse. I can try to find more example of this pattern being used in the codebase, but grepping for these would be a good starting point!

duguorong009:
@mimoo 
I would like to take this issue!

duguorong009:
@mimoo
Atm, I found that the domain move is mostly done in `[commitment.rs](https://github.com/o1-labs/proof-systems/blob/master/poly-commitment/src/commitment.rs)` and `[expr.rs](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/expr.rs)`.
I think these appearances are mainly attributed to that the domain move/conversion is needed for polynomial commitment & aggregation(shrinking the polynomial degree) processes.
It would be best if you can confirm above findings & indicate me more appearances anywhere.

duguorong009:
@mimoo 
I tried to create a helper function for the issue.
TBH, I think this is a bit of `overkill`.
The reason is that the most of these patterns are used in kind of `shifting` elements.
```
let scale = (domain as usize) / (res_domain.0 as usize);
assert!(scale != 0);
let v: Vec<_> = (0..n)
  .into_par_iter()
  .map(|i| {
     x + evals.evals[(scale * i + (domain as usize) * shift) % evals.evals.len()]
  })
.collect();
```
Please let me know if I am wrong.

mimoo:
[...]

PR #899, opened 2022-12-05
Add the helper of moving between domain sizes in `poly-commitment`

duguorong009:
**What's been done:**
- Add the `to_domain` helper function to cast the domain size of commitment evaluations to smaller domain size, in `poly-commitment/src/commitment.rs` file
- Replace some codes with a new helper function

Reference: #886


kimchi/src/circuits/expr.rs
```diff
@@ -934,15 +935,16 @@ impl<'a, F: FftField> EvalResult<'a, F> {
...
+                let v: Vec<_> = evals_domain_size_cast(
+                    &evals.evals,
+                    domain as usize,
+                    res_domain.0 as usize,
+                    res_domain.1.size(),
+                    shift,
+                )
+                .into_par_iter()
+                .map(|v| x + v)
+                .collect();
```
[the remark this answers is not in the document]
duguorong009: Thanks for the valuable tips! @mimoo 
Please re-check the PR.
I followed your recommend in https://github.com/o1-labs/proof-systems/pull/899/commits/82a41b8cfb6dd7d10c66e0c7dd179c6f401dfd33

kimchi/src/circuits/expr.rs
```diff
@@ -987,10 +984,10 @@ impl<'a, F: FftField> EvalResult<'a, F> {
...
                 },
             ) => {
-                let scale = (d_sub as usize) / (d as usize);
-                assert!(scale != 0);
+                let es_sub_evals =
+                    to_domain(es_sub, d_sub as usize, d as usize, res_domain.1, s, None);
                 evals.evals.par_iter_mut().enumerate().for_each(|(i, e)| {
-                    *e += es_sub.evals[(scale * i + (d_sub as usize) * s) % es_sub.evals.len()];
+                    *e += es_sub_evals[i];
                 });
```
mimoo, line 991: I think we should think of a way to have that factored out as well, and as you said in https://github.com/o1-labs/proof-systems/issues/886#issuecomment-1337447961 not create this temporary unecessary vec

maybe a `to_domain_in_place` function?

poly-commitment/src/commitment.rs
```diff
@@ -879,6 +886,35 @@ pub fn inner_prod<F: Field>(xs: &[F], ys: &[F]) -> F {
...
+///
+/// ## Panics
+///
+/// Panics if `evals_domain_size` is smaller than `target_domain_size`.
+pub fn to_domain<F: FftField>(
+    evals: &Evaluations<F, D<F>>,
+    evals_domain_size: usize,
+    target_domain_size: usize,
+    target_domain: D<F>,
+    shift: usize,
```
mimoo, line 899: I think that should be an option

poly-commitment/src/commitment.rs
```diff
@@ -879,6 +886,35 @@ pub fn inner_prod<F: Field>(xs: &[F], ys: &[F]) -> F {
...
 
+/// Cast the evaluations in specfic domain size to the smaller domain size.
+///
+/// ## Panics
+///
+/// Panics if `evals_domain_size` is smaller than `target_domain_size`.
+pub fn to_domain<F: FftField>(
+    evals: &Evaluations<F, D<F>>,
+    evals_domain_size: usize,
+    target_domain_size: usize,
```
mimoo, line 897: can't you just do `target_domain.size()`?

mimoo: same for `evals_domain_size`

poly-commitment/src/commitment.rs
```diff
@@ -879,6 +886,35 @@ pub fn inner_prod<F: Field>(xs: &[F], ys: &[F]) -> F {
...
+        scale, 0,
+        "we can't move to a bigger domain without interpolating and reevaluating the polynomial"
+    );
+    let f = |i| {
+        if let Some(cst) = constant {
+            cst + evals.evals[(scale * i + evals_domain_size * shift) % evals.evals.len()]
+        } else {
+            evals.evals[(scale * i + evals_domain_size * shift) % evals.evals.len()]
+        }
+    };
```
mimoo, line 913: I think it's better to return a closure here, instead of having the branching be done in the closure unnecessarily

poly-commitment/src/commitment.rs
mimoo: yeah but since you use them to compute the ratio, they should give you a similar result

mimoo: btw this part of the mina book is relevant: https://o1-labs.github.io/proof-systems/plonk/domain.html

(and so is https://github.com/duguorong009/proof-systems/blob/create-move-domain-size-helper/kimchi/src/circuits/domains.rs)

duguorong009:
@mimoo 
In https://github.com/o1-labs/proof-systems/pull/899/commits/c6632fe7bdd34a4aff4fb90ca1751a42013731a3, I rolled back some `to_domain` uses. 
The reason is that those uses are just inappropriate(create extra Vecs, ...).
I think we can leave them & wrap up this PR. 
Pls let me know what you think.

mimoo:
sorry for the late response, I think there's still some unaddressed comments of me. Also, I think it's fine, we can think of taking care of the other stuff later.

important though, I think this code should live in the `utils/` crate, and extend something so that it's easier to use. (like I said in https://github.com/o1-labs/proof-systems/pull/899#discussion_r1044814591)

duguorong009:
@mimoo @mrmr1993 
Can you review the PR?
I moved the new util function to the `utils` crate.

duguorong009:
Closing this PR, in favor of #964 and #965.

PR #904, opened 2022-12-08
Use Two trait in the code

querolita:
This PR updates the codebase (specially tests) to use the Two trait for field elements.


mrmr1993 (commented):
This feels like a misstep to me; `from(2u64)` is already fairly clear, and other projects seem to manage without a `Two` trait quite capably.

I would prefer that we don't merge this.

querolita:
I guess two() is not as frequently used as two_pow(), which does save some space, but I agree we can leave it unmerged at least for now 👍🏻

mimoo:
my 2 cents is that the more helpers we introduce, the harder it is to learn the codebase

PR #912, opened 2022-12-10
Un-break utils

mrmr1993:
This PR
* adds a CI job to stop us from breaking things that don't have tests
* fixes the target that broke.

### DIFF
--- a/kimchi/src/circuits/polynomials/rot.rs
+++ b/kimchi/src/circuits/polynomials/rot.rs
@@ -56,7 +56,7 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
             CircuitGate {
                 typ: GateType::RangeCheck0,
                 wires: Wire::for_row(new_row + 1),
-                coeffs: vec![],
+                coeffs: vec![F::zero()],
             },
         ]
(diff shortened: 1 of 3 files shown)


kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -56,7 +56,7 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
             CircuitGate {
                 typ: GateType::RangeCheck0,
                 wires: Wire::for_row(new_row + 1),
-                coeffs: vec![],
+                coeffs: vec![F::zero()],
```
mimoo, line 59: `coeffs: vec![]` is the same here

mimoo (approved):

kimchi/src/circuits/polynomials/rot.rs
mrmr1993: It's not; the test fails without it.

mimoo: ah interesting, there must be some places that don't handle the coeffs gracefully : o

PR #920, opened 2022-12-13
Spec fixes (admonition to admonish)

jspada:


mimoo (approved):
thanks for the quick fix : )

PR #929, opened 2022-12-19
FFAdd sign update inside coefficients

querolita:
This PR addresses the final pending TODO inside the FFAdd gate that consists on moving the sign value to the coefficients of the gate. This is not only an aesthetic change, as it indeed makes sense that the operation being performed is known and part of the relation itself, but it also brings soundness to the final bound check of any FFAdd chain. In other words, we did not have enough space in copyable cells to both constrain that the sign of the final check is a "+" and that the final overflow flag is 1. By removing the sign from the layout, we can now constrain the overflow flag, and the sign can be publicly checked to be the right value. 

It is not clear to me though if we want to keep the `(sign - 1) * (sign + 1) = 0` constrain, as it is just a coefficients check and thus reducing further the number of constraints to 1, or if for some reason we may want to leave it as some sort of sanity check.

### DIFF
--- a/book/src/rfcs/foreign_field_add.md
+++ b/book/src/rfcs/foreign_field_add.md
@@ -201,17 +201,17 @@ with $r_{top} = r_2$ and $c = c_1$.
 
 ## Gadget
 
-The full foreign field addition/subtraction gadget will be composed by: 
+The full foreign field addition/subtraction gadget will be composed of: 
 - $1$ public input row containing the value $1$;
 - $n$ rows with `ForeignFieldAdd` gate type:
   - for the actual $n$ chained additions or subtractions;  
-- $1$ `ForeignFieldAdd` row for the final bound check;
-- $1$ `Zero` row for the final upper bound check.
+- $1$ `ForeignFieldAdd` row for the bound addition;
+- $1$ `Zero` row for the final bound addition.
 - $1$ `RangeCheck` gadget for the first left input of the chain $a_1 := a$;
 - Then, $n$ of the following set of rows:
   - $1$ `RangeCheck` gadget for the $i$-th right input $b_i$;
   - $1$ `RangeCheck` gadget for the $i$-th result which will correspond to the $(i+1)$-th left input of the chain $r_i = a_{i+1}$.
-- $1$ final `RangeCheck` gadget for the upper bound term $u$.
+- $1$ final `RangeCheck` gadget for the bound check $u$.
 
 A total of 20 rows with 15 columns in Kimchi for 1 addition. All ranges below are inclusive.
 
@@ -227,9 +227,30 @@ A total of 20 rows with 15 columns in Kimchi for 1 addition. All ranges below ar
 | 9n+7..9n+10      | `multi-range-check` | $u$     |
 
 
-This mechanism can chain foreign field additions together. Initially, there are $n$ foreign field addition gates, followed by a foreign field addition gate for the upper bound (whose current row corresponds to the next row of the last foreign field addition gate), and an auxiliary `Zero` row that holds the upper bound. At the end, an initial left input range check is performed, which is followed by a $n$ pairs of range check gates for the right input and intermediate result (which become the left input for the next iteration). After the chained inputs checks, a final range check on the upper bound takes place. 
+This mechanism can chain foreign field additions together. Initially, there are $n$ foreign field addition gates, followed by a foreign field addition gate for the bound addition (whose current row corresponds to the next row of the last foreign field addition gate), and an auxiliary `Zero` row that holds the upper bound. At the end, an initial left input range check is performed, which is followed by a $n$ pairs of range check gates for the right input and intermediate result (which become the left input for the next iteration). After the chained inputs checks, a final range check on the bound takes place. 
 
-Nonetheless, such an exhaustive set of checks are not necessary for completeness nor soundness. In particular, only the very final range check for the bound is required. Thus, a shorter gadget that is equally valid and takes $(8*n+4)$ less rows is possible, and it follows the next layout (with inclusive ranges):
+For example, chaining the following set of 3 instructions would result in a full gadget with 37 rows:
+
+$$add(add(add(a,b),c),d)$$
+
+| Row(s) | Gate type(s)        | Witness       |
+| ------ | ------------------- | ------------- |
+| 0      | `PublicInput`       | $1$           |
+| 1      | `ForeignFieldAdd`   | $a+b$         |
+| 2      | `ForeignFieldAdd`   | $(a+b)+c$     |
+| 3      | `ForeignFieldAdd`   | $((a+b)+c)+d$ |
+| 4      | `ForeignFieldAdd`   | bound         |
+| 5      | `Zero`              | bound         |
+| 6..9   | `multi-range-check` | $a$           |
+| 10..13 | `multi-range-check` | $b$           |
+| 14..17 | `multi-range-check` | $a+b$         |
+| 18..21 | `multi-range-check` | $c$           |
+| 22..25 | `multi-range-check` | $a+b+c$       |
+| 26..29 | `multi-range-check` | $d$           |
+| 30..33 | `multi-range-check` | $a+b+c+d$     |
+| 34..37 | `multi-range-check` | bound         |
+
+Nonetheless, such an exhaustive set of checks are not necessary for completeness nor soundness. In particular, only the very final range check for the bound is required. Thus, a shorter gadget that is equally valid and takes $(8*n+4)$ fewer rows could be possible if we can assume that the inputs of each addition are correct foreign field elements. It would follow the next layout (with inclusive ranges): 
 
 | Row(s)   | Gate type(s)                                          | Witness |
 | -------- | ----------------------------------------------------- | ------- |
@@ -239,6 +260,20 @@ Nonetheless, such an exhaustive set of checks are not necessary for completeness
 | n+2      | `Zero`                                                |         |
 | n+3..n+6 | `multi-range-check` for `bound`                       | $u$     |
 
+Otherwise, we would need range checks for each new input of the chain, but none for intermediate results; implying $4\cdot n$ fewer rows. 
+
+| Row(s)          | Gate type(s)                                          | Witness   |
+| --------------- | ----------------------------------------------------- | --------- |
+| 0               | public input row for soundness of bound overflow flag | $1$       |
+| 1..n            | `ForeignFieldAdd`                                     | $a_i+b_i$ |
+| n+1             | `ForeignFieldAdd`                                     |           |
+| n+2             | `Zero`                                                |           |
+| n+3..n+6        | `multi-range-check` for first left input              | $a_1$     |
+| n+7+4i..n+10+4i | `multi-range-check` for $i$-th right input            | $b_i$     |
+| 5n+7..5n+10     | `multi-range-check` for bound                         | $u$       |
+
+
+For more details see the Bound Addition section of the [Foreign Field Multiplication RFC](../rfcs/ffmul.md).
 
 ### Layout
 
@@ -254,8 +289,8 @@ For the full mode of tests of this gate, we need to perform 4 range checks to as
 | 4          | $b_1$ (copy)      |                   |
 | 5          | $b_2$ (copy)      |                   |
 | 6          | $q$               |
-| 7          | $c_0$             |
-| 8          | $c_1$             |
+| 7          | $c$               |
+| 8          |                   |
 | 9          |                   |
 | 10         |                   |
 | 11         |                   |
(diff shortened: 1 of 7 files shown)


jspada:
If the sign is part of the coefficients and is publicly visible/auditable and it's constrained to either 1 or -1, then I think it's fine to remove to duplicate `(sign - 1) * (sign + 1) = 0` constraint.

book/src/rfcs/foreign_field_add.md
```diff
@@ -248,7 +256,7 @@ For this gate, we need to perform 4 range checks to assert that the limbs of $a,
 | 6          | $q$               |
 | 7          | $c_0$             |
 | 8          | $c_1$             |
-| 9          | $s$               |
+| 9          |                   |
```
jspada, line 259: I think above needs update.  There's no `c_1` anymore.

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -138,7 +137,13 @@ where
     const CONSTRAINTS: u32 = 5;
 
     fn constraint_checks<T: ExprOps<F>>(env: &ArgumentEnv<F, T>) -> Vec<T> {
-        let foreign_modulus: [T; LIMB_COUNT] = array::from_fn(|i| env.coeff(i));
+        let foreign_modulus: [T; LIMB_COUNT] = [env.coeff(0), env.coeff(1), env.coeff(2)];
```
jspada, line 140: why switch from `array::from_fn`?

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -138,7 +137,13 @@ where
...
+
+        // stored as coefficient for better correspondance with the relation being proved
+        // this reduces the number of copy constraints needed to check the operation
+        // it also allows the final bound check to copy the overflow bit to be 1
+        // because otherwise it did not fit in the first 7 columns of the row
+        let sign = env.coeff(3);
```
jspada, line 146: Seems like a great soundness fix to me.  Well done!

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
```diff
@@ -149,23 +154,22 @@ where
...
         let result_lo = env.witness_next(0);
         let result_mi = env.witness_next(1);
         let result_hi = env.witness_next(2);
 
         // Field overflow and sign constraints
         let mut checks = vec![
-            // Field overflow bit is 0 or s.
-            field_overflow.clone() * (field_overflow.clone() - sign.clone()),
             // Sign flag is 1 or -1
+            // NOTE: not sure if we really need to check this, as it is part of the relation
```
jspada, line 169: I don't think so.

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
```diff
@@ -27,31 +30,40 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
     ///      [n+1]         -> 1 Zero row for bound result
     /// ]
     ///
-    pub fn create(
+    /// INTEGRATION:
+    /// - Wire the range check for result bound manually
+    /// - Connect to public input containing the 1 value for the overflow in the final bound check
+    pub fn create_chain_ffadd(
```
jspada, line 36: Just checking that we still have a way to create additions in a maximally flexible way?  In other words, we still have interfaces where we can perform the "external" checks (i.e. range checks and bound checks) however we like, so we can craft and optimise things around a particular use case...

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
```diff
@@ -27,31 +30,40 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
...
         }
+        let mut final_coeffs = foreign_field_modulus.to_vec();
+        final_coeffs.push(FFOps::Add.sign::<F>());
         // Then the final bound gate and the zero gate
         circuit_gates.append(&mut vec![
             CircuitGate {
                 typ: GateType::ForeignFieldAdd,
                 wires: Wire::for_row(next_row + num),
-                coeffs: foreign_field_modulus.to_vec(),
+                coeffs: final_coeffs,
```
jspada, line 66: I would just put `foreign_field_modulus.to_vec()` here and avoid temporary, single purpose variable.

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
```diff
@@ -62,22 +74,26 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
...
-    /// Create a single foreign field addition gate
+    /// Create a single foreign field addition gate. This is used for example in the final bound check.
     ///     Inputs
     ///         starting row
     ///     Outputs tuple (next_row, circuit_gates) where
     ///       next_row      - next row after this gate
     ///       circuit_gates - vector of circuit gates comprising this gate
     pub fn create_single_ffadd(
         start_row: usize,
+        operation: FFOps,
```
jspada, line 85: Ah, cool... so this is the one that we can use if we need max flexibility, etc.

document new parameter above.

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
```diff
@@ -27,6 +27,18 @@ pub enum FFOps {
...
 }
 
+/// Implementation of the FFOps enum
+impl FFOps {
+    /// Returns the sign of the operation as a field element, or zero if it is a multiplication
+    pub fn sign<F: PrimeField>(&self) -> F {
+        match self {
+            FFOps::Add => F::one(),
+            FFOps::Sub => -F::one(),
+            FFOps::Mul => F::zero(),
```
jspada, line 37: I don't think adding a Mul operation make sense, since both gates already have separate APIs and can't really be unified since they are structurally so different.  That is, for mul vs add, the operation type is implicit in the interface name you call.  However, I do think it makes sense to ffadd to have a `Sign` parameter or `FFOps` for add or subtract.

kimchi/src/tests/foreign_field_add.rs
```diff
@@ -29,9 +30,10 @@ use num_bigint::{BigUint, RandBigInt};
 use num_traits::FromPrimitive;
 use o1_utils::{
     foreign_field::{ForeignElement, HI, LO, MI, TWO_TO_LIMB},
-    FieldHelpers,
+    FieldHelpers, Two,
```
jspada, line 33: I think `Two` may have got dragged in by some other merge/branch that we don't want?

kimchi/src/tests/foreign_field_add.rs
```diff
@@ -148,60 +155,86 @@ static NULL_CARRY_BOTH: &[u8] = &[
...
-fn full_gadget<F: PrimeField + SquareRootField>(
-    num: usize,
+// Creates a circuit including the public input, chain of additions only
+// Inputs
+//  operations
+//  foreign field modulus
+// Outputs tuple (next_row, circuit_gates) where
+//  next_row      - next row after this gate
+//  circuit_gates - vector of circuit gates comprising this gate
+fn short_circuit<F: PrimeField + SquareRootField>(
```
jspada, line 165: Cool!  Is this a play on words?! :-)

jspada: I hope it doesn't "short circuit" or did you mean https://i.ytimg.com/vi/rROqDm_w8JM/movieposter.jpg

jspada (commented):
Amazing job -- great catch about the soundness failure!  Left some feedback, suggestions and comments.

book/src/rfcs/foreign_field_add.md
```diff
@@ -201,40 +201,48 @@ with $r_{top} = r_2$ and $c = c_1$.
...
-| 8i+4..8i+7  | `multi-range-check` for `right`  | $b$     |
-| 8i+8..8i+11 | `multi-range-check` for `result` | $r$     |
-|             |                                  |         |
-| 8n+4..8n+7  | `multi-range-check` for `bound`  | $u$     |
-|             |                                  |         |
-| 8n+i+8      | `ForeignFieldAdd`                |         |
-|             |                                  |         |
-| 9n+8        | `ForeignFieldAdd`                |         |
-| 9n+9        | `Zero`                           |         |
+The full foreign field addition/subtraction gadget will be composed by:
```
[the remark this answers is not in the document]
querolita: oh sure thanks!

book/src/rfcs/foreign_field_add.md
```diff
@@ -201,40 +201,48 @@ with $r_{top} = r_2$ and $c = c_1$.
...
+
+Nonetheless, such an exhaustive set of checks are not necessary for completeness nor soundness. In particular, only the very final range check for the bound is required. Thus, a shorter gadget that is equally valid and takes $(8*n+4)$ less rows is possible, and it follows the next layout (with inclusive ranges):
+
+| Row(s)   | Gate type(s)                                          | Witness |
+| -------- | ----------------------------------------------------- | ------- |
+| 0        | public input row for soundness of bound overflow flag | $1$     |
+| 1..n     | `ForeignFieldAdd`                                     |         |
+| n+1      | `ForeignFieldAdd`                                     |         |
+| n+2      | `Zero`                                                |         |
```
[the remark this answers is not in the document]
querolita: But okay, I guess if the very initial inputs were not real limbs, the field overflow could happen regardless.

book/src/rfcs/foreign_field_add.md
querolita: that is true, seems like I didn't modify the rfc in the last update to ffadd

kimchi/src/circuits/polynomials/foreign_field_add/circuitgates.rs
querolita: Oh because I was not sure it would be correctly parsed when the coefficients are not just 3 but 4. Meaning, I was not sure which 3 it would get to fit the LIMB_COUNT length. But I can revert that if it works fine.

kimchi/src/circuits/polynomials/foreign_field_add/gadget.rs
querolita: but I don't just need foreign_field_modulus there, but that plus the sign coefficient

kimchi/src/circuits/polynomials/foreign_field_add/witness.rs
querolita: I don't think it is worth having a Sign type, it is just internal use and used in just a couple of places. The sign kind of is inherent to the Add/Sub wording.

querolita: I will remove Mul then.

kimchi/src/tests/foreign_field_add.rs
querolita: Oh, the Two still exists, but not used everywhere as proposed in the other PR

kimchi/src/tests/foreign_field_add.rs
querolita: Hahaha it was not, none of them, but funny now that you say it XD

kimchi/src/tests/foreign_field_add.rs
querolita: It started being used partly in Rot a few weeks ago, but then the PR to use it everywhere else in the code was not accepted. I wonder if others may want to remove it completely, but it was really helpful for certain Keccak gates at least for me.

book/src/rfcs/foreign_field_add.md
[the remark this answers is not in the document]
jspada: My understanding was that the argument for allowing $\ge f$ inputs is restricted to the case when the input is the result of a previous addition.  That is, the inputs to addition are either the result of a previous addition or else something smaller than $f$.  Otherwise, without a range check on the actual values summed, I could start with an input equal to the max native modulus (i.e. $n - 1$) and trigger an overflow without hitting the max circuit size.  I think we should err on the side of caution and check with @mrmr1993 .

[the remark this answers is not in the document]
querolita: Well, this is kind of collateral to FFAdd, seems like we will have to address an efficient way when we look at the real use case indeed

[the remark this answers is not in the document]
jspada: Ah, so you think it's orthogonal?  Yeah, I guess we will build a custom layout with the flexible interface anyway and then the gadget creator can always constrain those if needed.

Still someone could use this gadget an end up with something possibly under-constrained, no?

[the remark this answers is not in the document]
querolita: I mean, right now the builder allows you to create a chain of additions, but any other check you may want to add needs to be included in a larger gadget. FFAdd specs warn about the needed checks you may want, so when we write a larger gadget we will be able to have plenty of flexibility on where to put each gate. If it is too monolithic, then it may not be useful for ecdsa as we may want to put things in between to chain with multiplications in the middle, etc.

jspada (approved):
Really great work!

PR #936, opened 2022-12-22
First version of our changelog file 😎 

querolita:
This PR introduces a first version of a changelog of Kimchi, starting on November 14th. Feel free to make any suggestions about the versioning numbers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,98 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.1.12] - 2022-12-19
+
+- **Changed**:
+  - `ForeignFieldMul` uses shifted plookups to check 3-bit length of the column 7 of the witness, reducing by 1 the number of constraints and using 1 fewer witness cell.
+  - The `Next` row of `ForeignFieldMul` uses 1 fewer witness cell and is moved to the last column of the `Curr` row.
+- **Removed**:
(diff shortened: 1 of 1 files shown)


mimoo (changes requested):
woop woop! So two important comments before we can get this in:

1. the version numbers should match the version numbers in the `Cargo.toml` file, which is (and has been) [0.1.0](https://github.com/o1-labs/proof-systems/blob/master/kimchi/Cargo.toml#L3) so far.
2. the first section should be something like "unreleased" so we can just list any new changes there until we perform a release. ([see this example](https://gist.github.com/juampynr/4c18214a8eb554084e21d6e288a18a2c))

CHANGELOG.md
```diff
@@ -76,23 +78,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 
-## [0.1.1] - 2022-11-16
+### [0.1.0] - 2022-11-16
 
 - **Changed**: 
   - The generic gate is implemented using the `Argument` trait.
 
 
-## [0.1.0] - 2022-11-14
+### [0.1.0] - 2022-11-14
```
mimoo, line 97: I don't think it makes sense to have many headers for 0.1.0, and actually I think it'll trick parsers who are probably not expecting this (unless I misunderstand how changelogs should be formatted)

also I think just having "first version of kimchi" under 0.1.0 is enough? There's no "change" since there was nothing before. Perhaps we can move all of this to the unreleased part if we want :o

mimoo: yeah, that's a good example: https://keepachangelog.com/en/1.0.0/

mimoo:
oh and I just thought about something else, we probably need a CHANGELOG.md per crate! (for example https://github.com/RustCrypto/elliptic-curves/)

mimoo:
sorry to hijack this one! https://github.com/o1-labs/proof-systems/pull/986

PR #937, opened 2023-01-05
[easy] Refactor Rot64 witness creation interface

querolita:
This is a refactor to make the witness creation and extension interface for the rotation gate simpler and easier to read. I realised the current version was not really comfortable when calling it from outside.

### DIFF
--- a/kimchi/src/circuits/polynomials/rot.rs
+++ b/kimchi/src/circuits/polynomials/rot.rs
@@ -282,42 +282,19 @@ fn init_rot64<F: PrimeField>(
 }
 
 /// Extends the rot rows to the full witness
-pub fn extend_rot_rows<F: PrimeField>(
+pub fn extend_rot<F: PrimeField>(
     witness: &mut [Vec<F>; COLUMNS],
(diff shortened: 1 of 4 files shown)


kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -333,12 +310,28 @@ fn create_witness_rot<F: PrimeField>(witness: &mut [Vec<F>; COLUMNS], word: u64,
     // Right input of the "FFAdd" for the bound equation
```
jspada, line 310: I'm not sure you should refer to this as `FFAdd` as it makes it sound like a`FFAdd` gate, instead maybe just say "addition"

jspada: Maybe I'm being paranoid, but above on line 308: `let rotated = shifted + excess;`
We are adding to `u64` here-- can you remind me what prevents an overflow?

jspada (approved):
Checked it.  Nice improvement that really neatens up the interface!

PR #938, opened 2023-01-05
[easy] Add a connect function for the range checks in FFAdd

querolita:
This PR adds a function `connect_ffadd_range_checks` inside the `Connect` trait to more easily configure the wiring of foreign field additions, given the row of the actual addition, and (optionally) any number of rows of the range checks of the left input, right input, and result. It can be reused by setting Left and Right to None for the bound check range check.


jspada (approved):
Nice simplification!  Looks good to me.

PR #942, opened 2023-01-09
[easy] extend function for Rotation in all tests

querolita:
This PR includes an actual extend function and uses it in all rot tests


querolita:
I am closing this PR because 941 already will include all of the gates.

PR #944, opened 2023-01-10
Remove dense polynomial forms

imeckler:
This PR removes dense polynomial forms of evaluated columns from the prover index, instead using evaluations forms directly to evaluate and open those polynomials. The result is (currently) one more size n FFT in the opening proof and reduced prover memory. It also paves the way for opening all the polynomials (in order to remove the linearization technique) without any increase in prover memory usage.

It does not currently allow the evaluations form polynomials to have size greater than n. This would require changing chunking to happen post-accumulation (rather than on a per-polynomial basis) which can be done in a follow up PR.

### DIFF
--- a/kimchi/src/lagrange_basis_evaluations.rs
+++ b/kimchi/src/lagrange_basis_evaluations.rs
@@ -120,24 +120,17 @@ mod tests {
         let x = Fp::rand(rng);
         let evaluator = LagrangeBasisEvaluations::new(domain, x);
 
-        let expected = (0..n)
-            .map(|i| {
-                let mut lagrange_i = vec![Fp::zero(); n];
-                lagrange_i[i] = Fp::one();
-                Evaluations::from_vec_and_domain(lagrange_i, domain)
-                    .interpolate()
-                    .evaluate(&x)
-            });
+        let expected = (0..n).map(|i| {
+            let mut lagrange_i = vec![Fp::zero(); n];
+            lagrange_i[i] = Fp::one();
+            Evaluations::from_vec_and_domain(lagrange_i, domain)
+                .interpolate()
+                .evaluate(&x)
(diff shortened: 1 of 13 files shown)


kimchi/src/prover.rs
```diff
@@ -295,7 +299,9 @@ where
             .for_each(|c| absorb_commitment(&mut fq_sponge, &c.commitment));
 
         //~ 1. Compute the witness polynomials by interpolating each `COLUMNS` of the witness.
-        //~    TODO: why not do this first, and then commit? Why commit from evaluation directly?
+        //~    As mentioned above, we commit using the evaluations form rather than the coefficients
```
mrmr1993, line 302: NB: You'll need to update the specs (`make -C book/specifications/kimchi`) if you update a spec comment.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -0,0 +1,196 @@
...
+        (&self.evals)
+            .into_par_iter()
+            .enumerate()
+            .map(|(i, e)| p_evals[stride * i] * e)
+            .sum()
+    }
+
+    /// Given the evaluations form of a polynomial, directly evaluate that polynomial at a point,
+    /// assuming that the given evaluations are either 0 or 1 at every point of the domain.
+    pub fn evaluate_zero_one<D: EvaluationDomain<F>>(&self, p: &Evaluations<F, D>) -> F {
```
mrmr1993, line 26: Nit: `evaluate_boolean` would be slightly clearer.

mrmr1993 (commented):
Fabulous!

mrmr1993 (approved):

mimoo (commented):
nice! left some comments :o

PR #951, opened 2023-01-14
Fix non-native field multiplication wraparound soundness regression

### LINKED ISSUE #953: Fix non-native field multiplication wraparound soundness regression

jspada:
Special thanks to @querolita for solving the wraparound instances for the tests!

Closes #953

### DIFF
--- a/kimchi/src/tests/foreign_field_mul.rs
+++ b/kimchi/src/tests/foreign_field_mul.rs
@@ -191,7 +191,6 @@ where
         }
     }
 
-    println!("witness[7][0] = {}", witness[7][0].to_biguint());
     if let Some(runner) = runner {
         // Perform full test that everything is ok before invalidation
         assert_eq!(runner.prove_and_verify::<EFqSponge, EFrSponge>(), Ok(()));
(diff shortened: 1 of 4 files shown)


kimchi/src/tests/foreign_field_mul.rs
```diff
@@ -927,6 +927,57 @@ fn test_invalid_carry1_hi_plookup() {
...
+    assert!(wraparound_0 - PallasField::one() >= PallasField::from(2u32).pow(&[12u64]));
+    assert!(two_to_9 * (wraparound_0 - PallasField::one()) >= PallasField::from(2u32).pow(&[12u64]));
+    // edge case: x + 8 is not a wraparound value
+    assert!(wraparound_0 + PallasField::from(8) >= PallasField::from(2u32).pow(&[12u64]));
+    assert!(two_to_9 * (wraparound_0 + PallasField::from(8)) >= PallasField::from(2u32).pow(&[12u64]));
+
+    // Invalid carry1_hi witness that causes wrap around to something less than 3-bits
+    let (result, witness) = run_test::<Vesta, VestaBaseSponge, VestaScalarSponge>(
+        true,
+        false, // Disable external checks so we can catch carry1_hi plookup failure
```
mrmr1993, line 955: No need now, but to make the code self-documenting, we really should either
* make this function call take a record
* create a dedicated enum `CheckConstraints::{WithGateConstraints, WithoutGateConstraints}`

mrmr1993 (approved):

kimchi/src/tests/foreign_field_mul.rs
jspada: It would also be good to
* make it only create CS once and then test many witnesses against it
* factor it out so could be used by other tests

PR #964, opened 2023-01-20
[WIP] helpers for domain conversions

mimoo:
CC @duguorong009 I'm creating a new PR since https://github.com/o1-labs/proof-systems/pull/899 is a forked PR and I can't figure a way to target it.

I added some code to experiment with a safer API to operate on evaluations while converting to smaller domains. I felt like the first version was a bit too dangerous. Now the API is quite complex which sucks... But I'm not sure how we can make it safe while making it simpler...

Another problem is, how do we access this information on the caller side? This would require some digging in the expression framework to see if this can be called with the correct arguments.


mimoo:
looks like this has moved to https://github.com/o1-labs/proof-systems/pull/965

mimoo:
reopening in case we don't go with the other one

PR #965, opened 2023-01-21
Eval subdomain casting helper compatible with expression framework

### LINKED ISSUE #886: create helper functions to move between domain size

jspada:
This PR builds on the awesome work of @duguorong009 and @mimoo to help get this improvement over the line.

Building on the PR #964 I found that in several places the expression framework does more than just add a constant, depending on the operation and situation.  It may
  * add or subtract a constant
  * multiply by a constant
  * merge two evals
  * add and subtract evaluation points
  * multiply evaluation points
  * square evaluation points
  
I also found that in many cases where domain casting is being done the circuit domain is not available.  Due to the above two issues, I extended the old API with a new function providing the required functionality.

Closes #886


kimchi/src/circuits/expr.rs
```diff
@@ -1063,12 +1053,11 @@ impl<'a, F: FftField> EvalResult<'a, F> {
                     evals: es_sub,
                 },
             ) => {
-                let scale = (d_sub as usize) / (d as usize);
-                assert!(scale != 0);
-                evals.evals.par_iter_mut().enumerate().for_each(|(i, e)| {
-                    *e += es_sub.evals[(scale * i + (d_sub as usize) * s) % es_sub.evals.len()];
-                });
-                Evals { evals, domain: d }
+                evals =
```
mrmr1993, line 1056: This (and other similar changes) change the memory allocation characteristics (`par_iter` vs `par_iter_mut`). Perhaps adding helpers that cache all the intermediate terms in a struct (`ReducedDomainEvals`?) and use them to index / iterate over the structure would help?

kimchi/src/circuits/expr.rs
```diff
@@ -1082,23 +1071,14 @@ impl<'a, F: FftField> EvalResult<'a, F> {
...
-                        es1.evals[(scale1 * i + (d1 as usize) * s1) % es1.evals.len()]
-                            + es2.evals[(scale2 * i + (d2 as usize) * s2) % es2.evals.len()]
-                    })
-                    .collect();
-
                 Evals {
                     domain: res_domain.0,
-                    evals: Evaluations::<F, D<F>>::from_vec_and_domain(v, res_domain.1),
+                    evals: es1.to_subdomain_unsafe(res_domain.1, (d1 as usize) * s1, |eval, i| {
+                        eval + es2.evals[(scale2 * i + (d2 as usize) * s2) % es2.evals.len()]
```
mrmr1993, line 1080: It's a pity that we still have to manually do this here.

jspada: Maybe there's a way to make a "merging" helper?

utils/src/evaluations.rs
```diff
@@ -167,19 +174,53 @@ impl<F: FftField> ExtendedEvaluations<F> for Evaluations<F, Radix2EvaluationDoma
...
-        };
-        let new_evals = (0..target_size).into_par_iter().map(f).collect();
+        self.to_subdomain_unsafe(target_domain, shift, |eval, _| eval + cst)
+    }
+
+    fn to_subdomain_unsafe<G: Sync + Send + Fn(F, usize) -> F>(
+        &self,
+        target_domain: Radix2EvaluationDomain<F>,
+        shift: usize,
+        map: G,
```
mimoo, line 185: imo passing a closure defeats the purpose of these helpers

utils/src/evaluations.rs
```diff
@@ -167,19 +174,53 @@ impl<F: FftField> ExtendedEvaluations<F> for Evaluations<F, Radix2EvaluationDoma
...
+                .into_par_iter()
+                .map(|i| map(self.evals[(scale * i + shift) % self.evals.len()], i))
+                .collect(),
+            target_domain,
+        )
     }
 }
 
+/// Identity evaluation mapping
+pub const fn identity_fn<F: FftField>(eval: F, _: usize) -> F {
```
mimoo, line 220: this should be inlined instead of being its own function

kimchi/src/circuits/expr.rs
mimoo: yeah it'd be nicer to have a function that clearly states what it's doing instead

PR #967, opened 2023-01-25
Split range check feature into two features (one for each gate)

### LINKED ISSUE #966: Add a distinct feature flag for RangeCheck1 in Kimchi

jspada:
Closes #966


kimchi/src/prover.rs
```diff
@@ -629,13 +626,12 @@ where
...
-                        .iter()
-                        .enumerate()
-                        .map(|(i, gate_type)| (*gate_type, &polys[i])),
-                );
+            if let Some(selector) = &index.column_evaluations.range_check0_selector8.as_ref() {
+                index_evals.insert(GateType::RangeCheck0, selector);
+            }
+
+            if let Some(selector) = &index.column_evaluations.range_check1_selector8.as_ref() {
+                index_evals.insert(GateType::RangeCheck1, selector);
```
querolita, line 634: Below, in the FFMul selector code, why are you enumerating across the gadget? Doesn't it contain only one term? I mean it to keep an uniform format with the rest, using insert instead.

querolita (approved):
I think there's no where else to change :)

kimchi/src/prover.rs
jspada: This is a relic of the beginnings of a uniform API would be called in a homogeneous way for single gates or gadgets of multiple gates.  But changes to ffmul code are not the subject of this PR, so I'd keep things focused and leave it how it is in this PR.

mrmr1993 (approved):

PR #971, opened 2023-01-27
Remove ChaCha gate as it's now unneeded

jspada:
As discussed with Matthew, with our new approach in zkoracles, this gate is no longer required and should be removed for simplicity and performance.  Removing this will also mean less work to do for verification of custom gates in wrap proofs.  There will be a couple separate PRs for the mina repo corresponding to this.


mrmr1993 (approved):

mimoo:
are we sure that zkapps will never have a need for this? It sounded like a nice thing to have :o

mimoo (changes requested):
cf my previous comment, but also please update the CHANGELOG.md in kimchi before getting this in! we have a CHANGELOG now :)

mrmr1993 (approved):

Issue #213, opened 2021-11-04
[kimchi] distinguish ZK_ROWS and PERM_FINAL_ACC


mimoo:
We currently use 2 zero-knowledge rows, but specify 3 in the code (under the constant `ZK_ROWS`). We use this to create enough space for three equations:

* one constraint for the final accumulator value of the permutation. In the original PLONK protocol this constraint does not exist as the constraint on the initial accumulator value takes care of that (because the accumulator wraps around). With [our way of doing zero-knowledge](https://o1-labs.github.io/mina-book/crypto/plonk/zkpm.html) we need to add that last constraint in a new row.
* two random values, for the zero-knowledge, as intended

It'd be nice to reflect that in the code.

Also, due to that we currently randomize the last three rows of the witness columns, which is not a big deal as they cannot be used by the circuit, but it would be good to document that :o)

random thought: intuitively, it sounds like we could use unused columns in zero gates (or other gates) to randomize the witnesses and save some rows. It doesn't save us that many rows though (2 is really negligible) and it wouldn't actually save us anything because we still need that space for the permutation. Nevermind.

mimoo:
Adding more color to this issue:

The relevant code is here: https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/prover.rs#L109

```rust
        //~ 1. Ensure we have room in the witness for the zero-knowledge rows.
        //~    We currently expect the witness not to be of the same length as the domain,
        //~    but instead be of the length of the (smaller) circuit.
        //~    If we cannot add `ZK_ROWS` rows to the columns of the witness before reaching
        //~    the size of the domain, abort.
        let length_witness = witness[0].len();
        let length_padding = d1_size
            .checked_sub(length_witness)
            .ok_or(ProofError::NoRoomForZkInWitness)?;
        if length_padding < ZK_ROWS as usize {
            return Err(ProofError::NoRoomForZkInWitness);
        }

        //~ 2. Pad the witness columns with Zero gates to make them the same length as the domain.
        //~    Then, randomize the last `ZK_ROWS` of each columns.
        for w in &mut witness {
            if w.len() != length_witness {
                return Err(ProofError::WitnessCsInconsistent);
            }

            // padding
            w.extend(std::iter::repeat(Fr::<G>::zero()).take(length_padding));

            // zk-rows
            for row in w.iter_mut().rev().take(ZK_ROWS as usize) {
                *row = Fr::<G>::rand(rng);
            }
        }
```

[...]

PR #974, opened 2023-01-30
Update some constants like `ZK_ROWS` for clear naming

duguorong009:
Reference: #213


kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -24,7 +24,7 @@ use super::runtime_tables;
 pub const CONSTRAINTS: u32 = 7;
 
 /// The number of random values to append to columns for zero-knowledge.
-pub const ZK_ROWS: usize = 3;
+pub const EVALS: usize = 2;
```
mimoo, line 27: So the weird thing is that in `polynomial/permutation.rs` we already have

```rust
pub const ZK_ROWS: u64 = 3;
```

IMO all of these should live either in `kimchi/src/lib.rs`, or in some constant file `kimchi/src/constants.rs`

I think we could also have two constants:

```rust
/// The number of points that are evaluated in the protocol (zeta and zeta * omega).
pub const EVALS: u64 = 2;

/// The number of rows that are added to a witness/circuit in order to provide zero-knowledge. We might actually only need `EVALS` but let's be double safe.
pub const ZK_ROWS: u64 = EVALS + 1;
```

(also `u64` makes more sense here than `usize`)

wdyt? It looks less awkward than having `EVALS + 1` everywhere

kimchi/src/circuits/constraints.rs
```diff
@@ -704,9 +704,10 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
         //~ 2. Create a domain for the circuit. That is,
         //~    compute the smallest subgroup of the field that
         //~    has order greater or equal to `n + ZK_ROWS` elements.
-        let domain = EvaluationDomains::<F>::create(gates.len() + ZK_ROWS as usize)?;
+        let domain =
+            EvaluationDomains::<F>::create(gates.len() + (ZK_ROWS + PERM_FINAL_ACC) as usize)?;
 
-        assert!(domain.d1.size > ZK_ROWS);
+        assert!(domain.d1.size > ZK_ROWS + PERM_FINAL_ACC);
```
mimoo, line 710: `PERM_FINAL_ACC` should be a permutation-related constant only (see https://www.youtube.com/watch?v=iY2ue8Kfsb0&t=1028s and https://o1-labs.github.io/proof-systems/plonk/zkpm.html)

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -65,11 +65,14 @@ use std::array;
 
 /// Number of constraints produced by the argument.
 pub const CONSTRAINTS: u32 = 3;
-pub const ZK_ROWS: u64 = 3;
+pub const ZK_ROWS: u64 = 2;
+pub const PERM_FINAL_ACC: u64 = 1;
```
mimoo, line 69: as I said in another comment I think ZK_ROWS should be moved (and keep the value 3, but by being computed as `EVALS + 1`)

The PERM_FINAL_ACC should definitely live here, but it needs some comment. Perhaps a link to https://o1-labs.github.io/proof-systems/plonk/zkpm.html and "The permutation constraint should not apply on the final row (not including [ZK_ROWS] otherwise it will wrap around. We only need to check the final value of the permutation accumulator on the final row")

mimoo: (btw I would recommend going through the permutation video I linked, I think this is a good opportunity to learn more about it o.o and ofc the associated explanation in the PLONK paper)

kimchi/src/circuits/lookup/constraints.rs
duguorong009: Yes, I agree with your idea.
Just one question, should we move the constants to `lib.rs` or `constants.rs`?
I think it is more meaningful to keep the constants as is.

mimoo: Let's say `constants.rs`. I don't think this is the current file makes sense : o

kimchi/src/circuits/constraints.rs
```diff
@@ -656,7 +656,7 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
         //~ 2. Create a domain for the circuit. That is,
         //~    compute the smallest subgroup of the field that
         //~    has order greater or equal to `n + ZK_ROWS` elements.
-        let domain = EvaluationDomains::<F>::create(gates.len() + ZK_ROWS as usize)?;
+        let domain = EvaluationDomains::<F>::create(gates.len() + (ZK_ROWS) as usize)?;
```
mimoo, line 659: ?

mimoo: number of such changes below as well

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -23,8 +23,12 @@ use super::runtime_tables;
...
 pub const CONSTRAINTS: u32 = 7;
 
-/// The number of random values to append to columns for zero-knowledge.
-pub const ZK_ROWS: usize = 3;
+/// The number of points that are evaluated in the protocol (zeta and zeta * omega).
+pub const EVALS: u64 = 2;
+
+/// The number of rows that are added to a witness/circuit in order to provide zero-knowledge.
+/// We might actually only need `EVALS` but let's be double safe.
+pub const ZK_ROWS: u64 = EVALS + 1;
```
mimoo, line 31: as said elsewhere, these should be moved to a constants.rs file

kimchi/src/constants.rs
```diff
@@ -0,0 +1,12 @@
+/// Number of constraints produced by the permutation argument.
+pub const PERM_CONSTRAINTS: u32 = 3;
```
mimoo, line 2: the PERM constants are actually specific to the permutation only, so I don't think they should be moved out of the permutation file : o

kimchi/src/constants.rs
```diff
@@ -0,0 +1,12 @@
...
+/// The number of points that are evaluated in the protocol (zeta and zeta * omega).
+pub const EVALS: u64 = 2;
+/// The permutation constraint should not apply on the final row  
+/// (not including [ZK_ROWS] otherwise it will wrap around.  
+/// We only need to check the final value of the permutation accumulator  
+/// on the final row.  
+/// Ref: https://o1-labs.github.io/proof-systems/plonk/zkpm.html
+pub const PERM_FINAL_ACC: u64 = 1;
+/// The number of rows that are added to a witness/circuit in order to provide zero-knowledge.
```
mimoo, line 11: nit: some spacing between all these constants would be nice : )

kimchi/src/constants.rs
```diff
@@ -0,0 +1,12 @@
+/// Number of constraints produced by the permutation argument.
```
mimoo, line 1: a module documentation would be nice (`//!`)

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -20,12 +21,6 @@ use CurrOrNext::{Curr, Next};
 
 use super::runtime_tables;
 
-/// Number of constraints produced by the argument.
-pub const CONSTRAINTS: u32 = 7;
```
mimoo, line 24: this shouldn't move as this is the number of constraints used by the lookup argument

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -63,9 +64,6 @@ use rand::{CryptoRng, RngCore};
 use rayon::prelude::*;
 use std::array;
 
-/// Number of constraints produced by the argument.
-pub const CONSTRAINTS: u32 = 3;
```
mimoo, line 67: same here, this shouldn't move

mimoo (commented):
left you more comments, also can you update the `CHANGELOG.md` to specify exactly what constants were renamed (if any) as these are public constants that might be used by external user of this library (this will also help us when we need to update the Mina side)

kimchi/src/circuits/polynomials/permutation.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/91350343175bf874d16ada51a0ee486a6135db73

kimchi/src/constants.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/91350343175bf874d16ada51a0ee486a6135db73

kimchi/src/constants.rs
duguorong009: Done

duguorong009:
@mimoo 
I found that the current kimchi book build is failing, and affecting the CI(both `build` & `clippy`).
Can you check the reason?
I spent some minutes, but cannot find why. :man_shrugging:

mimoo:
@duguorong009 this should be fixed in master now : o but if it is still happening @jspada would know better

duguorong009:
Then, is it better to invite & ask @jspada for more info?

duguorong009:
CI error has been fixed in latest PR! @mimoo

kimchi/src/constants.rs
```diff
@@ -0,0 +1,5 @@
+/// The number of points that are evaluated in the protocol (zeta and zeta * omega).
```
mimoo, line 1: this file needs a module doc comment (`//!`)

kimchi/src/prover.rs
```diff
@@ -692,8 +693,7 @@ where
             };
             // permutation
             let (mut t8, bnd) = {
-                let alphas =
-                    all_alphas.get_alphas(ArgumentType::Permutation, permutation::CONSTRAINTS);
+                let alphas = all_alphas.get_alphas(ArgumentType::Permutation, CONSTRAINTS);
```
mimoo, line 696: since a lot of modules have a `CONSTRAINTS` constant it's good to keep the path `permutation::CONSTRAINTS` here

kimchi/src/prover.rs
```diff
@@ -979,8 +979,7 @@ where
                 // the constraint system struct
 
                 // permutation (not part of linearization yet)
-                let alphas =
-                    all_alphas.get_alphas(ArgumentType::Permutation, permutation::CONSTRAINTS);
+                let alphas = all_alphas.get_alphas(ArgumentType::Permutation, CONSTRAINTS);
```
mimoo, line 982: same here

kimchi/src/circuits/polynomials/permutation.rs
mimoo: looks like PERM_FINAL_ACC isn't used anywhere?

duguorong009: Yes, it is not used anywhere.
The reason is that it is already included in `ZK_ROWS` constant(`constants.rs`).
I left this `PERM_FINAL_ACC` since it has comments which could be good guide for devs.

kimchi/src/circuits/lookup/constraints.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/91350343175bf874d16ada51a0ee486a6135db73

kimchi/src/circuits/constraints.rs
duguorong009: Yes, I just leave this variable in `permutations.rs` file.

kimchi/src/circuits/constraints.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/e8717474ce86614b4db72b3e06dc19e9de9cd469

kimchi/src/circuits/lookup/constraints.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/91350343175bf874d16ada51a0ee486a6135db73

kimchi/src/constants.rs
duguorong009: Added module doc in https://github.com/o1-labs/proof-systems/pull/974/commits/b2bb15aa2556a627152b1b800e7ebc5cb239a682

kimchi/src/constants.rs
duguorong009: Add module doc in https://github.com/o1-labs/proof-systems/pull/974/commits/b2bb15aa2556a627152b1b800e7ebc5cb239a682
Pls check if that is what you want, or similar one. @mimoo

kimchi/src/prover.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/b2bb15aa2556a627152b1b800e7ebc5cb239a682

kimchi/src/prover.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/974/commits/b2bb15aa2556a627152b1b800e7ebc5cb239a682

Issue #644, opened 2022-06-20
verify: move public input from proof to argument of the verify function


mimoo:
It doesn't make sense to have the public input be part of the proof

cc @rot256

imikushin:
Thinking of picking this up.

imikushin:
So, #976 implements this. What do folks think about adding `verifier_index` as an argument to `batch_verify()`? It makes the API a little cleaner, in my opinion. 

But, since I'm new here, I may be missing use cases when there is a verifier index per proof (which may be missing tests then).

PR #976, opened 2023-02-02
Move public input out of ProverProof type into an argument of verify

imikushin:
Implements #644

### DIFF
--- a/kimchi/src/tests/chacha.rs
+++ b/kimchi/src/tests/chacha.rs
@@ -79,6 +79,7 @@ fn chacha_prover() {
     let group_map = <Vesta as CommitmentCurve>::Map::setup();
 
     let start = Instant::now();
+    let public_input = witness[0][0..index.cs.public].to_vec();
     let proof =
         ProverProof::create::<BaseSponge, ScalarSponge>(&group_map, witness, &[], &index).unwrap();
     println!("{}{:?}", "Prover time: ".yellow(), start.elapsed());
@@ -88,7 +89,12 @@ fn chacha_prover() {
     println!("{}{:?}", "Verifier index time: ".yellow(), start.elapsed());
 
     let start = Instant::now();
-    match verify::<Vesta, BaseSponge, ScalarSponge>(&group_map, &verifier_index, &proof) {
+    match verify::<Vesta, BaseSponge, ScalarSponge>(
+        &group_map,
+        &verifier_index,
+        &proof,
+        &public_input,
+    ) {
         Err(error) => panic!("Failure verifying the prover's proofs in batch: {}", error),
         Ok(_) => {
             println!("{}{:?}", "Verifier time: ".yellow(), start.elapsed());
--- a/kimchi/src/verifier.rs
+++ b/kimchi/src/verifier.rs
@@ -18,6 +18,7 @@ use crate::{
     proof::{PointEvaluations, ProverProof, RecursionChallenge},
     verifier_index::VerifierIndex,
 };
+use ark_ec::AffineCurve;
 use ark_ff::{Field, One, PrimeField, Zero};
 use ark_poly::{EvaluationDomain, Polynomial};
 use commitment_dlog::commitment::{
@@ -104,6 +105,7 @@ where
         &self,
         index: &VerifierIndex<G>,
         public_comm: &PolyComm<G>,
+        public: &Vec<G::ScalarField>,
     ) -> Result<OraclesResult<G, EFqSponge>> {
         //~
         //~ #### Fiat-Shamir argument
@@ -271,12 +273,12 @@ where
         all_alphas.instantiate(alpha);
 
         // compute Lagrange base evaluation denominators
-        let w: Vec<_> = index.domain.elements().take(self.public.len()).collect();
+        let w: Vec<_> = index.domain.elements().take(public.len()).collect();
 
         let mut zeta_minus_x: Vec<_> = w.iter().map(|w| zeta - w).collect();
 
         w.iter()
-            .take(self.public.len())
+            .take(public.len())
             .for_each(|w| zeta_minus_x.push(zetaw - w));
 
         ark_ff::fields::batch_inversion::<G::ScalarField>(&mut zeta_minus_x);
@@ -284,13 +286,12 @@ where
         //~ 1. Evaluate the negated public polynomial (if present) at $\zeta$ and $\zeta\omega$.
         //~
         //~    NOTE: this works only in the case when the poly segment size is not smaller than that of the domain.
-        let public_evals = if self.public.is_empty() {
+        let public_evals = if public.is_empty() {
             [vec![G::ScalarField::zero()], vec![G::ScalarField::zero()]]
         } else {
             [
                 vec![
-                    (self
-                        .public
+                    (public
                         .iter()
                         .zip(zeta_minus_x.iter())
                         .zip(index.domain.elements())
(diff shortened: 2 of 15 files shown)


kimchi/src/verifier.rs
```diff
@@ -762,28 +763,30 @@ pub fn verify<G, EFqSponge, EFrSponge>(
...
+///     proofs: vector of tuples (Plonk proof, public input)
 ///     RETURN: verification status
 ///
 /// # Errors
 ///
 /// Will give error if `srs` of `proof` is invalid or `verify` process fails.
 pub fn batch_verify<G, EFqSponge, EFrSponge>(
     group_map: &G::Map,
-    proofs: &[(&VerifierIndex<G>, &ProverProof<G>)],
+    verifier_index: &VerifierIndex<G>,
```
imikushin, line 788: I assume (perhaps out of my zero knowledge of the subject), that the same verifier_index would be used when performing batch verification.

Please let me know if that assumption is wrong!

mimoo: no actually here we expect to be passed a list of tuples (verifier_index, proof) so each proof can be associated with a different index (as long as they share the same SRS length

kimchi/src/verifier.rs
```diff
@@ -803,26 +806,22 @@ where
         return Ok(());
     }
 
-    //~ 1. Ensure that all the proof's verifier index have a URS of the same length. (TODO: do they have to be the same URS though? should we check for that?)
```
mimoo, line 806: I don't think this code should be removed : o

also, when you remove this type of comments: `//~` you actually remove parts of the specification (that gets built in https://o1-labs.github.io/proof-systems/specs/kimchi.html#batch-verification-of-proofs (LaTeX there is broken currently))

kimchi/src/verifier.rs
mimoo: (so now we should expect the public input to be another element in the tuple)

kimchi/src/verifier.rs
```diff
@@ -104,6 +105,7 @@ where
         &self,
         index: &VerifierIndex<G>,
         public_comm: &PolyComm<G>,
+        public: &Vec<G::ScalarField>,
```
mimoo, line 108: can this be renamed `public_input`? I think it's clearer

kimchi/src/verifier.rs
imikushin: Fair point: if verifier index is passed in the tuple, this code should definitely stay in place. 

(A bit of annoyance with tuple is it's becoming too complex, and the linter doesn't like it, which means a helper datatype is needed.)

kimchi/src/verifier.rs
imikushin: Done

kimchi/src/verifier.rs
imikushin: Linter doesn't like the longer tuple, so introduced a struct. Seems to have made the code a little cleaner, actually.

kimchi/src/verifier.rs
imikushin: Rolled this one back.

kimchi/src/verifier.rs
```diff
@@ -104,6 +105,7 @@ where
         &self,
         index: &VerifierIndex<G>,
         public_comm: &PolyComm<G>,
+        public_input: &Vec<G::ScalarField>,
```
mimoo, line 108: I think clippy will yell at you here that it should be `&[G::ScalarField]`

mimoo: looks like it didn't, I'm confused o_O (usually it complains about that)

kimchi/src/verifier.rs
```diff
@@ -762,28 +763,39 @@ pub fn verify<G, EFqSponge, EFrSponge>(
     group_map: &G::Map,
     verifier_index: &VerifierIndex<G>,
     proof: &ProverProof<G>,
+    public_input: &Vec<G::ScalarField>,
```
mimoo, line 766: I think all these `&Vec` should be `&[]` as it's more flexible (takes ref to array or vec)

kimchi/src/verifier.rs
```diff
@@ -762,28 +763,39 @@ pub fn verify<G, EFqSponge, EFrSponge>(
...
+    }];
     batch_verify::<G, EFqSponge, EFrSponge>(group_map, &proofs)
 }
 
+/// Struct equivalent of tuple (verifier index, Plonk proof, public input)
+pub struct BatchedProof<'a, G: KimchiCurve> {
+    pub verifier_index: &'a VerifierIndex<G>,
+    pub proof: &'a ProverProof<G>,
+    pub public_input: &'a Vec<G::ScalarField>,
+}
```
mimoo, line 787: so a few comment here:

* can you move the data structure to the top of the file?
* can you document on each fields? since they are public. (e.g. "the [VerifierIndex] associated to this proof", "The proof to verify", "the public input used in the creation of this proof")
* I don't think `BatchedProof` is a good name, usually a batch is several proofs. I'm not sure what's a better name though... `ProofPackage`? `BundleToVerify`? `ReadyToVerify`? ...

kimchi/src/verifier.rs
```diff
@@ -762,28 +763,39 @@ pub fn verify<G, EFqSponge, EFrSponge>(
...
-///     index: `VerifierIndex`
 ///     RETURN: verification status
 ///
 /// # Errors
 ///
 /// Will give error if `srs` of `proof` is invalid or `verify` process fails.
 pub fn batch_verify<G, EFqSponge, EFrSponge>(
     group_map: &G::Map,
-    proofs: &[(&VerifierIndex<G>, &ProverProof<G>)],
+    proofs: &Vec<BatchedProof<G>>,
```
mimoo, line 798: same comment here

kimchi/src/verifier.rs
```diff
@@ -805,22 +817,36 @@ where
...
     // TODO: Account for the different SRS lengths
-    let srs = &proofs[0].0.srs();
-    for (index, _) in proofs.iter() {
-        if index.srs().g.len() != srs.g.len() {
+    let srs = &proofs[0].verifier_index.srs();
+    for BatchedProof {
+        verifier_index,
+        proof: _,
+        public_input: _,
+    } in proofs.iter()
```
mimoo, line 825: nit: you can simply write `for BatchedProof { verifier_index, .. } in &proofs` here

kimchi/src/tests/chacha.rs
```diff
@@ -79,6 +79,7 @@ fn chacha_prover() {
     let group_map = <Vesta as CommitmentCurve>::Map::setup();
 
     let start = Instant::now();
+    let public_input = witness[0][0..index.cs.public].to_vec();
```
mimoo, line 82: (note that you can also use the const `PUBLIC` instead of `index.cs.public`)

mimoo (approved):
This is great! Thank you so much for this change as well : ) Left you a few nits, but I think we can merge this now!

kimchi/src/verifier.rs
imikushin: Funny thing: there's already Context struct, that only lacks public input. I think I can refactor this a little more. Stay tuned for the follow-up PR.

imikushin: See #982

kimchi/src/verifier.rs
imikushin: Done in the follow-up PR

kimchi/src/verifier.rs
imikushin: Done

mrmr1993:
@mimoo please don't merge these without fixing the Mina side first.

PR #977, opened 2023-02-02
add 2pc related specs

xiangxiecrypto:
Add specs related to two-party computation, including:

-  garbled circuits and optimizations.
- oblivious transfer and its extension.
- 2pc protocol based on garbled circuit and oblivious transfer.


mimoo (approved):
beautiful :) thanks a lot!

PR #970, opened 2023-01-27
Bump Rust toolchain to 1.67.0 and run cargo update

imikushin:
To keep things fresh :)

PR #982, opened 2023-02-05
Use slice references in function arg types, verifier::Context

imikushin:
Follow-up: address nits from #976.

### DIFF
--- a/kimchi/benches/proof_criterion.rs
+++ b/kimchi/benches/proof_criterion.rs
@@ -22,7 +22,7 @@ pub fn bench_proof_creation(c: &mut Criterion) {
     group.sample_size(100).sampling_mode(SamplingMode::Auto);
     group.bench_function(
         format!("proof verification (SRS size 2^{})", ctx.srs_size()),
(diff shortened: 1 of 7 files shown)


kimchi/benches/proof_criterion.rs
```diff
@@ -22,7 +22,7 @@ pub fn bench_proof_creation(c: &mut Criterion) {
     group.sample_size(100).sampling_mode(SamplingMode::Auto);
     group.bench_function(
         format!("proof verification (SRS size 2^{})", ctx.srs_size()),
-        |b| b.iter(|| ctx.batch_verification(black_box(vec![proof_and_public.clone()]))),
+        |b| b.iter(|| ctx.batch_verification(black_box(&vec![proof_and_public.clone()]))),
```
mimoo, line 25: nit: you can just write `&[proof_and_public.clone()]` (I'm wondering why clippy doesn't say anything there again)

mimoo (approved):
excellent!

kimchi/benches/proof_criterion.rs
imikushin: It's much more aggressive in Rust v1.67.0 (see #970)

PR #983, opened 2023-02-06
Revert #976 and #982

mrmr1993:
This PR reverts the changes in #976 and #982, which break the Mina bindings.


querolita (approved):
Yeah, perhaps let's wait until we fork to add this kind of changes for the future... Specially if it breaks stuff in the mina side

mimoo:
The mina bindings are already broken. This is why I couldn't work on these changes

PR #984, opened 2023-02-06
Match sponge_debug output to that of OCaml implementation

jspada:


mrmr1993 (approved):

querolita (approved):
So basically it adds data to the sponge one by one instead of as a vector, right? I guess the outcome after all of the elements of the vector are absorbed should be the same, but this gives us more fine grained access to the per-step values, no?

jspada:
> So basically it adds data to the sponge one by one instead of as a vector, right? I guess the outcome after all of the elements of the vector are absorbed should be the same, but this gives us more fine grained access to the per-step values, no?

Yes - exactly

Issue #751, opened 2022-09-15
Enforce length of evaluations


mimoo:
All evaluations in `proof::EvaluationProof` and `proof:LookupEvaluationProof` should be of length 1. We should enforce that at the very beginning when receive a proof in `verifier.rs`.

duguorong009:
@mimoo 
I would like to take care of this issue. 😄 

One question. 
I think the fields `proof::EvaluationProof` and `proof:LookupEvaluationProof` are from `ProverProof`.
```
/// The proof that the prover creates from a [ProverIndex](super::prover_index::ProverIndex) and a `witness`.
#[serde_as]
#[derive(Clone, Serialize, Deserialize)]
#[serde(bound = "G: ark_serialize::CanonicalDeserialize + ark_serialize::CanonicalSerialize")]
pub struct ProverProof<G: AffineCurve> {
    /// All the polynomial commitments required in the proof
    pub commitments: ProverCommitments<G>,

    /// batched commitment opening proof
    pub proof: OpeningProof<G>,

    /// Two evaluations over a number of committed polynomials
    // TODO(mimoo): that really should be a type Evals { z: PE, zw: PE }
    pub evals: [ProofEvaluations<Vec<G::ScalarField>>; 2],

    /// Required evaluation for [Maller's optimization](https://o1-labs.github.io/mina-book/crypto/plonk/maller_15.html#the-evaluation-of-l)
    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
    pub ft_eval1: G::ScalarField,

    /// The public input
    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
    pub public: Vec<G::ScalarField>,

    /// The challenges underlying the optional polynomials folded into the proof
    pub prev_challenges: Vec<RecursionChallenge<G>>,
}
```
I would like to ask why it is needed to enforce the length of `Evaluations`, even though the `evals` field is already typed as `[ProofEvaluations; 2]`.

mimoo:
ah so, there are two evaluations points (zeta and zeta * omega) which is the 2 you're seeing here. But then for each `ProofEvaluations` we have vectors of field elements which represent evaluated chunks

note: evaluated chunks are basically taking a polynomial `f(x)` and seeing it as several polynomials of the same degree `f(x) = f_0(x) + 2^n f_1(x)` and then producing the evaluations `[f_0(z), f_1(z)]` for some evaluation point `z`.

[...]

PR #989, opened 2023-02-07
Enforce the length of evaluations of `Proof`

duguorong009:
Ref: #751


kimchi/src/verifier.rs
```diff
@@ -511,6 +547,8 @@ where
         ));
     }
 
+    check_proof_evals_len(proof);
```
mimoo, line 550: this needs a `//~ 1.` comment describing what's happening so that the check can be added to the specification here https://o1-labs.github.io/proof-systems/specs/kimchi.html

kimchi/src/verifier.rs
```diff
@@ -480,6 +480,42 @@ where
     }
 }
 
+fn check_proof_evals_len<G>(proof: &ProverProof<G>)
```
mimoo, line 483: ^ this needs a doc comment (`///`)

kimchi/src/verifier.rs
```diff
@@ -480,6 +480,42 @@ where
...
+        s,
+        coefficients,
+        lookup,
+        generic_selector,
+        poseidon_selector,
+    } = &proof.evals;
+
+    let check_eval_len = |eval: &PointEvaluations<Vec<_>>| {
+        assert!(eval.zeta.len() == 1);
+        assert!(eval.zeta_omega.len() == 1);
```
mimoo, line 500: these should return an error instead of panicking here : o

kimchi/src/verifier.rs
```diff
@@ -480,6 +480,42 @@ where
...
+    coefficients.iter().for_each(check_eval_len);
+    if let Some(lookup) = lookup {
+        lookup.sorted.iter().for_each(check_eval_len);
+        check_eval_len(&lookup.aggreg);
+        check_eval_len(&lookup.table);
+        if let Some(runtime) = &lookup.runtime {
+            check_eval_len(runtime);
+        }
+    }
+    check_eval_len(z);
```
mimoo, line 514: nit: it might be nicer to have these checks follow the order of the fields in the struct

kimchi/src/verifier.rs
```diff
@@ -480,6 +480,42 @@ where
...
+
+    let check_eval_len = |eval: &PointEvaluations<Vec<_>>| {
+        assert!(eval.zeta.len() == 1);
+        assert!(eval.zeta_omega.len() == 1);
+    };
+
+    w.iter().for_each(check_eval_len);
+    s.iter().for_each(check_eval_len);
+    coefficients.iter().for_each(check_eval_len);
+    if let Some(lookup) = lookup {
```
mimoo, line 506: it'd be nice to deconstruct `lookup` like you deconstruct `proof.evals` above : o so that the compiler yells at us if we forgot to check a field

mimoo (changes requested):
Nice! This is great besides the few comments I left. (Also we should keep in mind that we'll need to change this the day where we try to support larger circuits as we'll have more evaluations, but for now this is good!)

kimchi/src/verifier.rs
duguorong009: Pls check https://github.com/o1-labs/proof-systems/pull/989/commits/d05bdd1287a6d6d2f767eacbfde2163aa552a250

kimchi/src/verifier.rs
duguorong009: Pls check https://github.com/o1-labs/proof-systems/pull/989/commits/d05bdd1287a6d6d2f767eacbfde2163aa552a250

kimchi/src/verifier.rs
duguorong009: Pls check https://github.com/o1-labs/proof-systems/pull/989/commits/d05bdd1287a6d6d2f767eacbfde2163aa552a250

kimchi/src/verifier.rs
duguorong009: Same as above

kimchi/src/verifier.rs
duguorong009: Same as above

kimchi/src/verifier.rs
```diff
@@ -511,6 +571,9 @@ where
         ));
     }
 
+    //~ Check the lengh of evaluations inside the proof.
```
mimoo, line 574: last nit! but this needs to be `//~ 1.` so that it adds to [the list here](https://o1-labs.github.io/proof-systems/specs/kimchi.html#partial-verification).

kimchi/src/error.rs
```diff
@@ -44,6 +44,9 @@ pub enum VerifyError {
     #[error("the previous challenges have an unexpected length (expected {0}, got {1})")]
     IncorrectPrevChallengesLength(usize, usize),
 
+    #[error("the evaluation is of an unexpected size")]
```
mimoo, line 47: nit: a bit vague, perhaps a better error is "proof malformed: an evaluation was of the incorrect size (all evaluations are expected to be of length 1)"

mimoo:
left you a few comments, but also:

* it looks like CI is not passing
* even if you fix that, CI will still not be happy because the specifications haven't been updated with the new additional `//~` comment. You'll need to `cd book/specifications && make -C kimchi watch` (see [here](https://github.com/o1-labs/proof-systems/tree/master/book/specifications))

kimchi/src/verifier.rs
```diff
@@ -511,6 +571,9 @@ where
         ));
     }
 
+    //~ 1. Check the lengh of evaluations inside the proof.
```
mimoo, line 574: typo: length

mimoo (approved):
This is great thanks! I think there's more we can do here but I'll update the original issue

PR #1002, opened 2023-02-14
Run fast checks first in CI

mrmr1993:
This ensures that we don't get to the end of the slow CI run only to realise that one of the faster checks is failing right at the end.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,6 +33,34 @@ jobs:
           toolchain: "1.67.0"
           components: rustfmt, clippy
 
+      # https://github.com/Swatinem/rust-cache
+      - name: Cache Rust stuff
+        uses: Swatinem/rust-cache@v1
+
+      #
+      # Doc & Spec
+      #
+
+      - name: Install cargo-spec for specifications
+        uses: actions-rs/install@v0.1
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -37,46 +37,29 @@ jobs:
       - name: Cache Rust stuff
         uses: Swatinem/rust-cache@v1
 
-      - name: Setup OCaml (because of ocaml-gen)
-        run: |
-          sudo apt update
-          sudo apt install -y ocaml
```
mimoo, line 43: I don't think CI is going to work without OCaml (due to ocaml-rs that requires it)

mrmr1993: CI is happy, we don't have anything that uses OCaml directly. This was only here for the ocaml-gen tests, which are now in another repo.

o1-david: this is not tested in CI until it lands. I'll approve it but I'm pretty sure it's going to break CI :D as it was ocaml-rs that required it, not ocaml-gen (and we build with all features)

o1-david (approved):

jspada (approved):
Nice improvement.

PR #1005, opened 2023-02-14
[book] fix dependency versions

mimoo:
cc @querolita 

I tried installing locally and seeing what versions didn't work, then just wrote `cargo install mdbook-mermaid` and checked what version it managed to install.

I think overall this is probably not a great way to ensure that dependencies have the same versions across environments. I posted that issue here: https://github.com/rust-lang/mdBook/issues/2026 but there's probably a solution to this problem somewhere (perhaps it is nix :D)


querolita:
>  (perhaps it is nix :D)

I don't run proof-systems with nix, only the mina repo. But perhaps you mean that the CI for proof-systems uses nix inside?

querolita:
I got these warnings when `mdbook serve`-ing the specs generated in this branch:

````
Warning: The mdbook-mermaid preprocessor was built against version 0.4.27 of mdbook, but we're being called from version 0.4.25
Warning: The mdbook-toc preprocessor was built against version 0.4.21 of mdbook, but we're being called from version 0.4.25
````

Do you also get those or my local versions are outdated?

querolita (approved):
Approved! I introduced these changes in PR number 1008 in any case (but added a newer version of katex as well). So perhaps there's no need to merge it alone.

mimoo:
ah yeah I'll close this then

mimoo:
I meant maybe we need to introduce nix in this repo to fix these issues (as this is what nix is supposed to solve)

PR #1008, opened 2023-02-15
Fixing the specs

querolita:
This PR fixes the specs by addressing the `markdownlint` warnings (although this didn't work completely) and updating to the latest version of Katex (which did almost completely, solving the breaking in And section), and then substituting \$ for \wedge (which made all of LaTeX magically compile again).


jspada:
Basically this

https://katex.org/docs/error.html

PR #1011, opened 2023-02-16
Disable auto-spec for now until we switch to use cargo-spec as library

jspada:


mimoo (approved):
can you remove the `Cargo.toml` file in `book/` also?

PR #1012, opened 2023-02-16
Build kimchi spec using cargo-spec as library

jspada:


querolita (approved):
Oh really? Is it that simple?

mrmr1993 (approved):

kimchi/build.rs
```diff
@@ -1,15 +1,13 @@
...
-            .expect("failed to get status")
-            .success(),
-        "failed to generate specification markdown"
-    );
+    build::build(
+        "../book/specifications/kimchi/Specification.toml".into(),
+        Some("../book/src/specs/kimchi.md".into()),
+        OutputFormat::Markdown,
+    )
+    .expect("failed to generate specification");
```
mimoo, line 12: like we discussed this is not going to work, as it will not detect changes to rerun this command (unless you use the `cargo-rerun-something:...`

kimchi/build.rs
```diff
@@ -1,15 +1,13 @@
...
-    assert!(
-        Command::new("make")
-            .args(["-C", "../book/specifications/kimchi", "build"])
-            .status()
-            .expect("failed to get status")
-            .success(),
-        "failed to generate specification markdown"
-    );
+    build::build(
+        "../book/specifications/kimchi/Specification.toml".into(),
```
mimoo, line 8: is this going to work always? I think the `../` is not well specified and it's better to use an absolute path from a [cargo env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html), like so: 

```rust
let out_dir env::var("CARGO_MANIFEST_DIR").unwrap();
let spec_file = PathBuf::new(out_dir).join("../book/specifications/kimchi/Specification.toml");
```

at least, I think this should work despite not necessarily pointing to the [workspace](https://github.com/rust-lang/cargo/issues/3946#issuecomment-1402464796)

mimoo: (same for the output bellow)

kimchi/build.rs
mimoo: this didn't work?

```rust
let files = build::build(toml_spec, Some(output_file), OutputFormat::Markdown);
for file in files {
  println!(format!("cargo:rerun-if-changed={}", file));
}
```

jspada: I have not added that part yet because I found that it appears to be working without it and I want to read the documentation for `cargo:rerun-if-changed` carefully.

kimchi/build.rs
jspada: Again, this was not the final PR, just a start.  Thanks for tip 👍🏼

kimchi/build.rs
jspada: > Oh really? Is it that simple?

More or less, but we lose our second goal of linting markdown.

mrmr1993 (approved):

kimchi/build.rs
o1-david: ah interesting, I guess that makes sense. But yeah it won't work if you modify the template.md files : o so probably better to have it

jspada:
Spoke to David.  He approves

PR #1013, opened 2023-02-17
Optimisation: Don't compute combined_inner_product twice when verifying proofs

mrmr1993:


kimchi/src/verifier.rs
```diff
@@ -608,6 +626,7 @@ where
         polys,
         zeta1: zeta_to_domain_size,
         ft_eval0,
+        combined_inner_product,
```
querolita, line 629: Does this need an update to Mina?

poly-commitment/src/tests/batch_15_wires.rs
```diff
@@ -114,7 +114,34 @@ where
...
             open += start.elapsed();
 
-            (sponge.clone(), x, polymask, evalmask, comm, proof)
+            let combined_inner_product = {
+                let es: Vec<_> = comm
+                    .iter()
+                    .map(|(commitment, evaluations, degree_bound)| {
+                        let bound: Option<usize> = (|| {
+                            let b = (*degree_bound)?;
+                            let x = commitment.commitment.shifted?;
```
querolita, line 123: Is this double `commitment.commitment.shifted?` a typo? In `commitment.rs` this is done with just one level `commitment.shifted?`.

poly-commitment/src/tests/commitment.rs
```diff
@@ -81,13 +87,45 @@ impl AggregatedEvaluationProof {
             });
         }
 
+        let combined_inner_product = {
```
querolita, line 90: We should have a function that abstracts away this code because it is duplicated in some places across the PR.

querolita (approved):
I am approving this change, but please consider the comments about abstracting away `combined_inner_product`, and a question I have about how to instantiate it (double or single dereferencing to obtain `shifted`).

kimchi/src/verifier.rs
mrmr1993: Yes, I'll follow up with a PR shortly

poly-commitment/src/tests/batch_15_wires.rs
mrmr1993: Indeed: this is a blinded commitment, so we have to cast it to an unblinded one first

kimchi/src/verifier.rs
```diff
@@ -433,7 +433,25 @@ where
...
+                                li.runtime_tables_selector
+                                    .as_ref()
+                                    .map(|_| [Column::LookupRuntimeTable].into_iter())
+                                    .into_iter()
+                                    .flatten(),
+                            )
+                    })
+                    .into_iter()
+                    .flatten(),
+            ) {
```
mimoo, line 454: I'm puzzled as to how the lookup tests worked if we were not checking the lookup parts in the verification of the aggregated evaluation proof?

mrmr1993: We weren't using this `combined_inner_product` *except* on the pickles side. In practice, this means that everything appeared to work on the kimchi side, but pickles would be broken.

poly-commitment/src/tests/commitment.rs
mrmr1993: Lets do this as a follow-up, there's some urgency around getting this in.

kimchi/src/verifier.rs
mimoo: that's what I had guessed :D good job for spotting this : )

PR #1014, opened 2023-02-17
Remove unnecessary extra lookup tables in tests

mrmr1993:


kimchi/src/tests/range_check.rs
```diff
@@ -645,7 +644,6 @@ fn verify_range_check0_v1_test_lookups() {
             TestFramework::<Vesta>::default()
                 .gates(index.cs.gates.clone())
                 .witness(witness.clone())
-                .lookup_tables(vec![range_check::gadget::lookup_table()])
```
jspada, line 648: Oh, interesting. Are the lookup constraints still working without this?  Sounds like I've misunderstood something!

So it automatically knows to include the range-check lookup table without needing to specify it here?  And, by doing it here I was including an extra copy?

jspada (approved):

kimchi/src/tests/range_check.rs
mrmr1993: That's right, yea. The constraint system generator already associates gates with particular lookup tables

jspada: That's great.  Is it something new or has it always done that?

mrmr1993: It's always done that!

PR #1015, opened 2023-02-17
Account for lookup tables when computing the domain size

mrmr1993:
This PR builds upon #1014, using the size of the lookup tables to calculate the domain size required for a proof. The domain size is now at least `max(num_gates, num_lookups+1)+ZK_ROWS`.

This PR also removes the copy-pasted `1 << 13` padding from a large number of tests. Despite mostly using smaller lookup tables, it seems that this was blindly pasted everywhere, which contributed at least in part to the slow tests in CI.

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -653,10 +653,39 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
         // for some reason we need more than 1 gate for the circuit to work, see TODO below
         assert!(gates.len() > 1);
 
+        let lookup_features = LookupFeatures::from_gates(&gates, runtime_tables.is_some());
+
+        let num_lookups = {
+            let mut num_lookups: usize = lookup_tables
+                .iter()
+                .map(
+                    |LookupTable { data, id: _ }| {
+                        if data.len() > 0 {
+                            data[0].len()
+                        } else {
+                            0
+                        }
+                    },
+                )
+                .sum();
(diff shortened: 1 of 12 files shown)


kimchi/src/circuits/constraints.rs
```diff
@@ -653,10 +653,39 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
...
+            }
+            num_lookups
+        };
+
         //~ 2. Create a domain for the circuit. That is,
         //~    compute the smallest subgroup of the field that
         //~    has order greater or equal to `n + ZK_ROWS` elements.
-        let domain = EvaluationDomains::<F>::create(gates.len() + ZK_ROWS as usize)?;
+        let domain_size_lower_bound =
+            std::cmp::max(gates.len(), num_lookups + 1) + ZK_ROWS as usize;
```
jspada, line 687: Why is it + 1 on num_lookups?

jspada (approved):
This is a fantastic improvement.  I left only one question; otherwise looks great!

kimchi/src/circuits/constraints.rs
mrmr1993: On the last non-zk column, we have
```
LookupAggreg(Next) = random
```
because of the zk randomness, so we can't integrate a lookup value into the table from that row.
See the use of `VanishesOnLast4Rows` in lookup, which is used to do the same extra offsetting in the constraints.

PR #1016, opened 2023-02-17
Stop re-running a whole test setup for small tweaks to the witness

mrmr1993:
This PR restructures some tests to be less wasteful. Previously, they were taking about 3-4 mins each on CI because of poor caching.


jspada (approved):
Thank you -- this is a really awesome improvement to makes the  `TestFramework`, `ProverIndex` and `RuntimeTableCfg` clone-able so we can reuse test framework context / prover index for different witnesses.  Awesome!

PR #1021, opened 2023-02-19
Figure out what's broken in CI (and fix it)

imikushin:
CI is failing with
```
error: failed to run custom build command for `ocaml-sys v0.22.3`
```
even for things like adding a line to a markdown file.

This PR is to figure out why and hopefully fix it.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,11 @@ jobs:
         run: |
           npm install -g markdownlint-cli
 
+      - name: Setup OCaml (because of ocaml-gen)
+        run: |
+          sudo apt update
+          sudo apt install -y ocaml
+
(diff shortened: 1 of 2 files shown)


imikushin:
Fixed it. 
cc @mimoo

mimoo (approved):
awesome, thanks a lot for the fix :)

(cc @mrmr1993 )

.github/workflows/rust.yml
```diff
@@ -37,6 +37,11 @@ jobs:
         run: |
           npm install -g markdownlint-cli
 
+      - name: Setup OCaml (because of ocaml-gen)
```
mimoo, line 40: I guess we should change that comment at some point, now it's because of ocaml-rs, but no biggy ^^

PR #1025, opened 2023-02-23
Tweak tests to emit useful errors

mrmr1993:
This replaces the `assert!(foo.is_ok())` in tests with `foo.unwrap()`, emitting the underlying error message instead of a non-descriptive assertion.

### DIFF
--- a/kimchi/src/snarky/constraint_system.rs
+++ b/kimchi/src/snarky/constraint_system.rs
@@ -944,7 +944,7 @@ impl<Field: PrimeField> SnarkyConstraintSystem<Field> {
                             if !s1.is_zero() {
                                 self.union_find(x1);
                                 self.union_find(x2);
-                                assert!(self.union_finds.union(x1, x2).is_ok());
+                                self.union_finds.union(x1, x2).unwrap();
                             };
                         } else if
                         /* s1 x1 - s2 x2 = 0 */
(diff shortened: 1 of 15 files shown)


o1-david (approved):

querolita (approved):
Nice!

kimchi/src/snarky/constraint_system.rs
```diff
@@ -944,7 +944,7 @@ impl<Field: PrimeField> SnarkyConstraintSystem<Field> {
                             if !s1.is_zero() {
                                 self.union_find(x1);
                                 self.union_find(x2);
-                                assert!(self.union_finds.union(x1, x2).is_ok());
+                                self.union_finds.union(x1, x2).unwrap();
```
dkijania, line 947: Hello, please forgive my intrusiveness, but i wanted to know why we are not using expect in this case? Based on rust book [chapter 9](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html):

> Using expect instead of unwrap and providing good error messages can convey your intent and make tracking down the source of a panic easier. Because this error message starts with the text we specified... it will be easier to find where in the code this error message is coming from.

Probably it would be easier for a reader (like myself) to understand that the reason of test failure is an assertion fail instead just to read stack trace with inner exception description, which may also indicate general problem with test setup?

I think the same result we would receive if using:

`assert!(self.union_finds.union(x1, x2).is_ok(),"reason for failure");`

mimoo: yeah you're right, if you create a PR to change this I will approve it : o

mrmr1993: The `Error` already contains a descriptive error message, which this `assert!` is hiding. The whole point of this PR is not to obscure that error; adding an expect is only adding noise to the signal.

dkijania: Ok, got it thanks!

PR #998, opened 2023-02-10
Update arkworks to 0.4.0

imikushin:
Implements #980

PR #1027, opened 2023-02-24
expand on arkworks 0.4.0

mimoo:
to make the mina side (https://github.com/MinaProtocol/mina/pull/12717) work for https://github.com/o1-labs/proof-systems/pull/998 (as I can't work on top of a fork PR)

this should be good to review, I've reviewed all the commits of @imikushin already so someone should review the rest of the commits which do:

* reverse the bump to 1.67.1 for rust (this should go in its own PR)
* reverse the bump to proof system 0.2.0
* `into_bingint()` is `to_hex()` for the ocaml generation ONLY. The rest doesn't need to change. The reason is that what really changed is how Display is implemented for the type, it used to print as a hex string, but now it prints as a decimal string. So we use `to_hex()` instead.
* remove cargo-spec stuff from kimchi as it broke the wasm build

cc @jspada about the removal of the `build.rs`

cc @imikushin about this PR. I'm just using this for the Mina side. I rebased on top of your latest change (as of the last edit of this post).


mimoo:
this should be good to review, I've reviewed all the commits of @imikushin already so someone should review the rest of the commits which do:

* reverse the bump to 1.67.1 for rust (this should go in its own PR)
* reverse the bump to proof system 0.2.0
* `into_bingint()` is `to_hex()` for the ocaml generation ONLY. The rest doesn't need to change. The reason is that what really changed is how Display is implemented for the type, it used to print as a hex string, but now it prints as a decimal string. So we use `to_hex()` instead.
* remove cargo-spec stuff from kimchi as it broke the wasm build

imikushin (approved):
LGTM 😅

Issue #1004, opened 2023-02-14
delete circuit-construction crate


mimoo:
Since we're going to introduce a Rust implementation of snarky, we should delete the circuit-construction crate

PR #1028, opened 2023-02-24
deprecating the circuit construction lib

mimoo:
removing this crate since it's unused and we're not going to use it in the future

https://github.com/o1-labs/proof-systems/issues/1004


mimoo:
ping @fabrizio-m

PR #1030, opened 2023-02-27
Move build script to book

jspada:


kimchi/Cargo.toml
```diff
@@ -46,15 +45,13 @@ groupmap = { path = "../groupmap", version = "0.1.0" }
 mina-curves = { path = "../curves", version = "0.1.0" }
 o1-utils = { path = "../utils", version = "0.1.0" }
 mina-poseidon = { path = "../poseidon", version = "0.1.0" }
+mina-book = { path = "../book", version = "0.1.0" }
```
o1-david, line 48: this line defeats the purpose :o you're still going to rely on the dependencies there

book/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+description = "Documentation and specifications of the proof systems"
+repository = "https://github.com/o1-labs/proof-systems"
+homepage = "https://o1-labs.github.io/proof-systems/"
+documentation = "https://o1-labs.github.io/proof-systems/rustdoc/"
+readme = "README.md"
+edition = "2021"
+license = "Apache-2.0"
+
+[build-dependencies]
+cargo-spec = { version = "0.5.0" }
```
o1-david, line 13: you probably want to add this crate to the workspace `Cargo.toml`

kimchi/Cargo.toml
jspada: Well, our original goal was to make the spec as part of build, so without this it will not build, right?

jspada: Ah, I see what you mean.. we add it to top level workspace instead.

mimoo: yeah otherwise this still breaks the wasm build

PR #1032, opened 2023-02-28
Fix chunking for proofs

mrmr1993:
This PR fixes the logic for chunking proofs.

This was tested by overwriting [this line](https://github.com/o1-labs/proof-systems/blob/18722368f6db488139782caa811bb08b42fd8ae5/kimchi/src/prover_index.rs#L169) with
```
let mut srs = SRS::<G>::create(cs.domain.d1.size() / 2);
```

Chunking is still disabled in this PR; a follow-on PR will remove the checks that disable it and add appropriate tests.


kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -118,19 +159,44 @@ mod tests {
...
+        let evaluator_chunked =
+            LagrangeBasisEvaluations::new_with_chunked_segments(domain.size(), domain, x);
+        for (i, (evals, evals_chunked)) in evaluator
+            .evals
+            .iter()
+            .zip(evaluator_chunked.evals.iter())
+            .enumerate()
+        {
+            for (j, (evals, evals_chunked)) in evals.iter().zip(evals_chunked.iter()).enumerate() {
+                if evals != evals_chunked {
```
querolita, line 197: So basically this test tries to check that the `if` clause in `new` indeed calls the chunked counterpart, no? Would it make sense to check the chunked evaluations against single-chunk evaluations from a larger domain?

kimchi/src/proof.rs
```diff
@@ -61,6 +61,8 @@ pub struct LookupEvaluations<Evals> {
 #[serde_as]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ProofEvaluations<Evals> {
+    /// public input polynomials
+    pub public: Evals,
```
querolita, line 65: Interesting... So now the verifier won't compute evaluations on its own?

kimchi/src/prover.rs
```diff
@@ -810,25 +814,7 @@ where
...
-                let w = <G::ScalarField as UniformRand>::rand(rng);
-                t_comm
-                    .commitment
-                    .unshifted
-                    .push(index.srs.h.mul(w).into_affine());
-                t_comm.blinders.unshifted.push(w);
-            }
-            t_comm
-        };
+        let t_comm = { index.srs.commit(&quotient_poly, 7 * num_chunks, None, rng) };
```
querolita, line 817: What's 7 here?

kimchi/src/oracles.rs
```diff
@@ -20,8 +20,6 @@ where
     pub oracles: RandomOracles<G::ScalarField>,
     /// the computed powers of alpha
     pub all_alphas: Alphas<G::ScalarField>,
-    /// public polynomial evaluations
-    pub public_evals: [Vec<G::ScalarField>; 2],
```
querolita, line 24: So, since the public polynomials are now inside ProofEvaluations they no longer contribute to Fiat Shamir? Meaning, now the verifier does not compute the public polynomial evaluations? Why is this? Efficiency?

kimchi/src/prover.rs
```diff
@@ -994,12 +991,12 @@ where
...
 
                 // see https://o1-labs.github.io/mina-book/crypto/plonk/maller_15.html#the-prover-side
-                f.to_chunked_polynomial(index.max_poly_size)
+                f.to_chunked_polynomial(num_chunks, index.max_poly_size)
                     .linearize(zeta_to_srs_len)
             };
 
             let t_chunked = quotient_poly
-                .to_chunked_polynomial(index.max_poly_size)
+                .to_chunked_polynomial(7 * num_chunks, index.max_poly_size)
```
querolita, line 999: Here again that 7

kimchi/src/verifier.rs
```diff
@@ -212,9 +215,13 @@ where
         //~ 1. Derive $\alpha$ from $\alpha'$ using the endomorphism (TODO: details).
         let alpha = alpha_chal.to_field(endo_r);
 
-        //~ 1. Enforce that the length of the $t$ commitment is of size `PERMUTS`.
-        if self.commitments.t_comm.unshifted.len() != PERMUTS {
-            return Err(VerifyError::IncorrectCommitmentLength("t"));
+        //~ 1. Enforce that the length of the $t$ commitment is of size 7.
+        if self.commitments.t_comm.unshifted.len() != chunk_size * 7 {
```
querolita, line 219: Is this 7 meant to be PERMUTS? Why are we removing this constant? Seemed convenient...

kimchi/src/verifier.rs
```diff
@@ -327,8 +304,6 @@ where
         //~~ * poseidon selector
         //~~ * the 15 register/witness
         //~~ * 6 sigmas evaluations (the last one is not evaluated)
-        fr_sponge.absorb_multiple(&public_evals[0]);
-        fr_sponge.absorb_multiple(&public_evals[1]);
```
querolita, line 331: I am still a bit doubtful about not absorbing the public input...

poly-commitment/src/commitment.rs
```diff
@@ -522,10 +522,11 @@ impl<G: CommitmentCurve> SRS<G> {
     pub fn commit(
         &self,
         plnm: &DensePolynomial<G::ScalarField>,
+        num_chunks: usize,
```
querolita, line 525: Nice feature

poly-commitment/src/commitment.rs
```diff
@@ -587,6 +590,10 @@ impl<G: CommitmentCurve> SRS<G> {
             });
         }
 
+        for _ in unshifted.len()..num_chunks {
+            unshifted.push(G::zero());
```
querolita, line 594: Is this happening just when `is_zero`? Otherwise I was expecting the `else` clause to push `num_chunks` of them, doesn't it?

querolita (commented):
I left a few comments and questions. Still not accepting because I would like more eyes into this one.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -98,7 +108,38 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
...
+        let num_chunks = n / max_poly_size;
+        let mut evals = Vec::with_capacity(num_chunks);
+        for i in 0..num_chunks {
+            let mut x_pow = F::one();
+            let mut chunked_evals = vec![F::zero(); n];
+            for j in 0..max_poly_size {
+                chunked_evals[i * max_poly_size + j] = x_pow;
+                x_pow *= x;
+            }
+            domain.ifft_in_place(&mut chunked_evals);
```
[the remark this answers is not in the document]
mrmr1993: An inverse fast-Fourier transform. Identical reasoning to [here](https://github.com/o1-labs/proof-systems/blob/769362887694142e2eb46dbcd1ce305902ddffba/poly-commitment/src/srs.rs#L87), except that we're using powers of ζ instead of the commitments.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -98,7 +108,38 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
 
         // Denominators now contains the desired result.
         LagrangeBasisEvaluations {
-            evals: denominators,
+            evals: vec![denominators],
+        }
+    }
+
+    /// Compute all evaluations of the normalized lagrange basis polynomials of the
+    /// given domain at the given point. Runs in time O(domain size).
```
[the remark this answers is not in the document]
mrmr1993: Oops, yea. Copy-paste error

kimchi/src/lagrange_basis_evaluations.rs
mrmr1993: They will be different, unfortunately. The vector sum of the chunked evaluations should equal the unchunked ones, we could add that test if you think it's not adequately covered by https://github.com/o1-labs/proof-systems/pull/1033.

kimchi/src/prover.rs
mrmr1993: `d8.size() / d1.size() - 1`. It was falsely called `PERMUTS` before, abusing the coincidence that the constants are the same.

kimchi/src/oracles.rs
mrmr1993: They're already in the evaluations; we don't need to compute them again.

kimchi/src/prover.rs
mrmr1993: As above, so below

kimchi/src/verifier.rs
mrmr1993: See above; it was a misleading coincidence

kimchi/src/verifier.rs
mrmr1993: We do it when we absorb the other evals

poly-commitment/src/commitment.rs
mrmr1993: Seems to work in testing

kimchi/src/lagrange_basis_evaluations.rs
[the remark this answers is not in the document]
mrmr1993: Updated the comment.

kimchi/src/verifier.rs
querolita: Oh I see thanks

poly-commitment/src/evaluation_proof.rs
```diff
@@ -195,8 +198,12 @@ impl<G: CommitmentCurve> SRS<G> {
             let mut plnm = plnm.to_dense_polynomial();
             if !plnm_evals_part.is_empty() {
                 let n = plnm_evals_part.len();
+                let max_poly_size = self.g.len();
+                let num_chunks = n / max_poly_size;
                 plnm += &Evaluations::from_vec_and_domain(plnm_evals_part, D::new(n).unwrap())
-                    .interpolate();
+                    .interpolate()
```
querolita, line 204: Very clear interface

querolita (approved):
Excellent work! Left some small observations/comments, but overall great!

kimchi/src/prover.rs
```diff
@@ -1097,8 +1087,8 @@ where
         //~~ * poseidon selector
         //~~ * the 15 register/witness
         //~~ * 6 sigmas evaluations (the last one is not evaluated)
-        fr_sponge.absorb_multiple(&public_evals[0]);
-        fr_sponge.absorb_multiple(&public_evals[1]);
+        fr_sponge.absorb_multiple(&chunked_evals.public.as_ref().unwrap().zeta);
```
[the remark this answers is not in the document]
mrmr1993: `public` is an `Option` to retain support for the existing kimchi proofs deployed on Berkeley (which will not have a value for `public`, and instead which we compute as the verifier). Any proofs generated by this version of kimchi will have them (as you can see here: they're always generated above), but we then potentially drop this value to send the proof to the Berkeley network.

kimchi/src/prover.rs
```diff
@@ -1496,37 +1501,49 @@ pub mod caml {
...
         fn from(pp: (ProverProof<G>, Vec<G::ScalarField>)) -> Self {
-            Self {
-                commitments: pp.0.commitments.into(),
-                proof: pp.0.proof.into(),
-                evals: pp.0.evals.into(),
-                ft_eval1: pp.0.ft_eval1.into(),
-                public: pp.1.into_iter().map(Into::into).collect(),
-                prev_challenges: pp.0.prev_challenges.into_iter().map(Into::into).collect(),
+            let (public_evals, evals) = pp.0.evals.into();
+            CamlProofWithPublic {
```
jspada, line 1516: `Self`?

kimchi/src/verifier.rs
```diff
@@ -586,7 +620,15 @@ where
     }
 
     //~ 1. Check the length of evaluations inside the proof.
-    check_proof_evals_len(proof)?;
+    let chunk_size = {
+        let d1_size = verifier_index.domain.size();
```
jspada, line 624: Perhaps create a helper for this logic as it appears many places.

poly-commitment/src/commitment.rs
```diff
@@ -931,12 +942,14 @@ mod tests {
         let mut srs = SRS::<VestaG>::create(n / 2 + 1);
         srs.add_lagrange_basis(domain);
 
+        let num_chunks = (domain.size() + srs.g.len() - 1) / srs.g.len();
```
jspada, line 945: What's going on here?

poly-commitment/src/evaluation_proof.rs
```diff
@@ -151,9 +151,12 @@ impl<G: CommitmentCurve> SRS<G> {
...
                                 *x += scale * evals[i * stride];
                             });
-                        assert_eq!(omegas.unshifted.len(), 1);
-                        omega += &(omegas.unshifted[0] * scale);
-                        scale *= &polyscale;
+                        for j in 0..omegas.unshifted.len() {
+                            omega += &(omegas.unshifted[j] * scale);
+                            scale *= &polyscale;
+                        }
+                        // We assume here that we have no shifted segment.
```
jspada, line 158: Maybe add an assertion for this?

jspada (approved):
Looks good to me.  Left some comments, suggestions and questions.

Issue #1034, opened 2023-03-01
amortization benchmark


fabrizio-m:
Add a benchmark for the amortized batched verification.

PR #1040, opened 2023-03-06
amortization benchmark

fabrizio-m:
Add a benchmark for amortized verification, should cover most of #1034 but benchmarking memory usage is also of interest.


kimchi/benches/amortization.rs
```diff
@@ -0,0 +1,33 @@
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
```
mimoo, line 1: it'd be good to add some documentation here, I'm not sure what "amortization" stands for here. Also the `kimchi/README.md` has a benchmark section, it'd be good to add some words about this new benchmark there.

mimoo (approved):

kimchi/benches/amortization.rs
fabrizio-m: sorry, the message didn't appear until I merged, will add docs in next pr.

PR #1043, opened 2023-03-09
[tools] add a kimchi-asm tool to deserialize JSON circuits

mimoo:
This adds a binary to display JSON circuits as the made-up asm that exists in kimchi currently.


jspada (approved):
Really nice tool!  Will be useful.

PR #1033, opened 2023-02-28
Enable chunked proving

mrmr1993:
This PR builds on #1032, removing the checks that disabled chunked proving, and adding a test.

Only one of the added tests is enabled; the others are commented out to avoid increasing CI time too much.

Stats from my (admittedly slow) machine:
```
$ cargo test --release 2_to_16_unchunked_generic_gate_pub -- --nocapture

running 1 test
- time to create prover index: 39s
test tests::chunked::test_2_to_16_unchunked_generic_gate_pub has been running for over 60 seconds
- time to create proof: 18s
- time to verify: 337ms
test tests::chunked::test_2_to_16_unchunked_generic_gate_pub ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 191 filtered out; finished in 63.40s

$ cargo test --release 2_to_17_chunked_generic_gate_pub -- --nocapture

running 1 test
test tests::chunked::test_2_to_17_chunked_generic_gate_pub has been running for over 60 seconds
- time to create prover index: 154s
- time to create proof: 45s
- time to verify: 351ms
test tests::chunked::test_2_to_17_chunked_generic_gate_pub ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 191 filtered out; finished in 213.64s

$ cargo test --release 2_to_18_chunked_generic_gate_pub -- --nocapture

running 1 test
test tests::chunked::test_2_to_18_chunked_generic_gate_pub has been running for over 60 seconds
- time to create prover index: 623s
- time to create proof: 129s
- time to verify: 395ms
test tests::chunked::test_2_to_18_chunked_generic_gate_pub ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 191 filtered out; finished in 801.81s
```
Roughly, for each doubling of the number of constraints:
* The prover index appears to take ~4x as long to generate (including the SRS and domain)
* The proving time appears to take ~2.5x as long
* The verification time increased about 8%

PR #1045, opened 2023-03-10
Add zero-knowledge for chunked proofs

mrmr1993:
This PR builds upon #1033, adding zero-knowledge for chunked proofs.


kimchi/src/circuits/constraints.rs
```diff
@@ -682,12 +700,33 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
...
         //~ 2. Create a domain for the circuit. That is,
         //~    compute the smallest subgroup of the field that
-        //~    has order greater or equal to `n + ZK_ROWS` elements.
-        let domain_size_lower_bound =
-            std::cmp::max(gates.len(), num_lookups + 1) + ZK_ROWS as usize;
+        //~    has order greater or equal to `n + zk_rows` elements.
+        let (zk_rows, domain_size_lower_bound) = {
+            let mut zk_rows = 3;
+            let get_domain_size_lower_bound =
+                |zk_rows: u64| std::cmp::max(gates.len(), num_lookups + 1) + zk_rows as usize;
```
rbonichon, line 707: It looks like `gates` might change between calls to this function so I believe it would be incorrect to compute `max` outside of the function. Mark this comment as resolved if that's indeed the case.

kimchi/src/circuits/constraints.rs
```diff
@@ -682,12 +700,33 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
...
+                // chunks that we use, and also does not cause us to overflow the domain size.
+                // NB: We use iteration here rather than hard-coding an assumption about
+                // `compute_size_of_domain`s internals. In practice, this will never be executed
+                // more than once.
+                while {
+                    let domain_size = D::<F>::compute_size_of_domain(domain_size_lower_bound)
+                        .ok_or(SetupError::DomainCreation(
+                            "could not compute size of domain",
+                        ))?;
+                    zk_rows = ((16 * (domain_size / max_poly_size) + 4) / 7) as u64;
```
rbonichon, line 720: The `7` constant seems magic to my eyes. Could we give it a name? Or a comment ;-)

kimchi/src/circuits/lookup/constraints.rs
```diff
@@ -35,13 +32,14 @@ pub const ZK_ROWS: usize = 3;
...
+    zk_rows: usize,
     rng: &mut R,
 ) -> Evaluations<F, D<F>> {
     let n = d.size();
     let k = e.len();
-    assert!(k <= n - ZK_ROWS);
-    e.extend((0..((n - ZK_ROWS) - k)).map(|_| F::zero()));
-    e.extend((0..ZK_ROWS).map(|_| F::rand(rng)));
+    assert!(k <= n - zk_rows);
+    e.extend((0..((n - zk_rows) - k)).map(|_| F::zero()));
```
rbonichon, line 41: (very) minor: compute `n-zk_rows` just once

For later: is it worth thinking about not failing hard with assertions here (aka using an error monad) ?

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -63,60 +63,62 @@ use std::array;
...
 }
 
 /// Evaluates the polynomial
-/// (x - w^{n - 3}) * (x - w^{n - 2}) * (x - w^{n - 1})
-pub fn eval_zk_polynomial<F: FftField>(domain: D<F>, x: F) -> F {
-    let w3 = zk_w3(domain);
-    let w2 = domain.group_gen * w3;
-    let w1 = domain.group_gen * w2;
-    (x - w1) * (x - w2) * (x - w3)
+/// (x - w^{n - zk_rows}) * (x - w^{n - zk_rows + 1}) * (x - w^{n - 1})
```
rbonichon, line 105: 👍

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -63,60 +63,62 @@ use std::array;
...
+        term *= domain.group_gen;
+        acc *= x - term;
+    }
+    acc
 }
 
 /// The polynomial
-/// (x - w^{n - 4}) (x - w^{n - 3}) * (x - w^{n - 2}) * (x - w^{n - 1})
-pub fn vanishes_on_last_4_rows<F: FftField>(domain: D<F>) -> DensePolynomial<F> {
+/// (x - w^{n - i}) * (x - w^{n - i + 1}) * ... * (x - w^{n - 1})
```
rbonichon, line 83: 👍

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -335,7 +340,11 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>> ProverIndex<G> {
         //~
         //~ $\text{scalar} \cdot \sigma_6(x)$
         //~
-        let zkpm_zeta = self.cs.precomputations().zkpm.evaluate(&zeta);
+        let zkpm_zeta = self
+            .cs
+            .precomputations()
+            .permutation_vanishing_polynomial_m
```
rbonichon, line 346: I'm not sure I can easily get the difference between `permutation_vanishing_polynomial_m` and `permutation_vanishing_polynomial_l`. Do you think the `m` and `l` could be expanded in the identifier?

kimchi/src/prover.rs
```diff
@@ -180,19 +179,27 @@ where
...
-        //~    If we cannot add `ZK_ROWS` rows to the columns of the witness before reaching
+        //~    If we cannot add `zk_rows` rows to the columns of the witness before reaching
         //~    the size of the domain, abort.
         let length_witness = witness[0].len();
         let length_padding = d1_size
             .checked_sub(length_witness)
             .ok_or(ProverError::NoRoomForZkInWitness)?;
 
-        if length_padding < ZK_ROWS as usize {
+        let zero_knowledge_limit = (16 * num_chunks - 4) / 7 ;
```
rbonichon, line 189: Magic `7` again :-)

kimchi/src/verifier.rs
```diff
@@ -312,19 +314,24 @@ where
         //~ 1. Derive $v$ from $v'$ using the endomorphism (TODO: specify).
         let v = v_chal.to_field(endo_r);
 
+        println!("v: {}", v);
+
```
rbonichon, line 318: Should we consider printing only in debug mode?

kimchi/src/verifier.rs
```diff
@@ -312,19 +314,24 @@ where
...
         //~ 1. Sample $u'$ with the Fr-Sponge.
         let u_chal = fr_sponge.challenge();
 
         //~ 1. Derive $u$ from $u'$ using the endomorphism (TODO: specify).
         let u = u_chal.to_field(endo_r);
 
+        println!("u: {}", u);
```
rbonichon, line 325: Should we consider printing only in debug mode?

rbonichon (approved):
I only have minor comments to possibly address. This is nice.

kimchi/src/circuits/lookup/constraints.rs
mrmr1993: > is it worth thinking about not failing hard with assertions here

A failure here indicates an critical error on our side, so we really do want to fail hard instead of returning something to the user.

kimchi/src/circuits/polynomials/permutation.rs
mrmr1993: The `m` suffix used to mean 'monomial', and the `l` 'lagrange', corresponding respectively to the coefficients of the polynomial (`[a0, a1, ...]` for `f(x) = a0 x^0 + a1 x^1 + ...`) and the evaluations at each element of the domain (`[f(omega^0), f(omega^1), ...]`).

If you don't mind, I'm going to leave this as is for now, and fold them into the `EvaluatedColumnCoefficients` and `ColumnEvaluations` structures in the prover index respectively.

kimchi/src/verifier.rs
mrmr1993: Woops, caught me debugging :sweat_smile:

PR #1048, opened 2023-03-13
Use the same cache for all constraints, to avoid accidental aliasing

mrmr1993:
This is #1019, rebased onto master.

### DIFF
--- a/kimchi/src/circuits/polynomials/turshi.rs
+++ b/kimchi/src/circuits/polynomials/turshi.rs
@@ -208,7 +208,8 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
         alphas.register(ArgumentType::Gate(self.typ), Instruction::<F>::CONSTRAINTS);
 
         // Get constraints for this circuit gate
-        let constraints = circuit_gate_combined_constraints(self.typ, &alphas);
+        let constraints =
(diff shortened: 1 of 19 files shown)


kimchi/src/circuits/polynomials/turshi.rs
```diff
@@ -208,7 +208,8 @@ impl<F: PrimeField + SquareRootField> CircuitGate<F> {
         alphas.register(ArgumentType::Gate(self.typ), Instruction::<F>::CONSTRAINTS);
 
         // Get constraints for this circuit gate
-        let constraints = circuit_gate_combined_constraints(self.typ, &alphas);
+        let constraints =
+            circuit_gate_combined_constraints(self.typ, &alphas, &mut Cache::default());
```
jspada, line 212: I guess this isn't using the same one, but it doesn't matter because we're not using turshi

jspada: If so, then it could be source of future bugs, so maybe comment this as TODO?

jspada: Actually, I guess this old school verify function will be replaced anyway so nvm

jspada (approved):
Looks good to me.  One minor comment.

We currently don't cache anything in any of the new gates.  I wonder, could it be helpful in some places?

PR #1050, opened 2023-03-15
[bench] refactor precomputed SRS and fixes bench

mimoo:
I reckon tests should be much faster now that we'll directly use the precomputed SRS for (hopefully all) tests.

We should also update the metrics in README

Worthy notes: 

* we should add benches for compilation (constraint system + indexes creation)
* the precomputed SRS are only used for size 2^16, we could either increase that to 2^18 but I think 2^16 should cover most/all tests (and for benches we can simply compute them on the fly and eat the cost)

EDIT: looks like we don't get that many improvements, because the slow tests don't seem to be using the `TestFramework` or the `create_index_for_test` functions, and create indexes (or specifically the SRS) directly by themselves:

Screenshot 2023-03-15 at 3 07 09 PM

Screenshot 2023-03-15 at 3 08 21 PM

(cc @jspada @querolita, all the slow tests seem to be in the new tests :o)


kimchi/src/precomputed_srs.rs
```diff
@@ -0,0 +1,82 @@
...
+/// The path of the serialized SRS.
+fn get_srs_path<G: KimchiCurve>() -> PathBuf {
+    let base_path = std::env::var("CARGO_MANIFEST_DIR").expect("failed to get manifest path");
+    PathBuf::from(base_path)
+        .join("../srs")
+        .join(format!("{}.srs", G::NAME))
+}
+
+/// Obtains an SRS for a specific curve from disk.
+/// Panics if the SRS does not exists.
```
fabrizio-m, line 27: Given than it isn't an unrecoverable state, wouldn't it be better to return `Option` or `Result`?
Now if you don't want to panic the only way is to check the file beforehand, but I think it would be better if it didn't panic and if you want it to panic you can still use `unwrap`.

fabrizio-m (approved):

kimchi/src/precomputed_srs.rs
mimoo: I couldn't think of a case where you would not want to panic here, as it's probably a configuration issue and you'll realize that from the start. The situation is different now if you can request an SRS of a specific size or something :o

fabrizio-m: One case I can think of is "create if it doesn't exist".
But the main issue is that you go from result to panic, but not the other way around. If an error isn't unrecoverable, a library should ideally return a result and let the users decide if they should panic.
I also just noticed that it calls `get_srs_path` that can panic, those panics should also be the doc if we keep the panic approach.

PR #1063, opened 2023-04-13
Benchmarking RFC

fabrizio-m:


book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+This doesn't change that often, the main utility of these benchmarks is just to detect low utilization so that it can be improved.
+Hardware utilization could include many things, GPU acceleration for example if we were to start using it, but for now will only include CPU/core utilization.
+
+### Layers
+
+There are 3 main layers of the stack that we want to consider, kimchi, snarky and pickles. Given that they are implemented in different languages and exist in different repositories special considerations each layer will be required.
+
+## Where to run the benchmarks
+
+To start everyone with the resources should be able to run the benchmarks locally, additionally we may want some of the benchmarks to run in CI to detect regressions. Some benchmarks may be considerably expensive to run, a decision will have to be made about which benchmarks to run and how often we should run them.
```
rbonichon, line 48: Detecting regression in performance might also include thinking more about CI infrastructure than other benchmarks. Both for very short or very long tests.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Parameters
+
+We may be interested in parameterizing some benchmark, for example benchmarking verification time but doing it with different circuits sizes, that would show how performance scales with the size of the circuit.
```
rbonichon, line 52:
```suggestion
We may be interested in parameterizing some benchmarks, for example benchmarking verification time but doing it with different circuits sizes, that would show how performance scales with the size of the circuit.
```

rbonichon: I think this is a need, not a possible interest.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+| kimchi instructions      | proof-systems | time     | no               | yes              | yes             | no               |
+| kimchi memory            | proof-systems | memory   | yes              | yes              | yes             | no               |
+| kimchi memory flamegraph | proof-systems | memory   | no               | yes              | no              | no               |
+| snarky time              | mina/snarky*  | time     | yes              | yes              | yes             | yes              |
+| snarky memory            | mina/snarky*  | memory   | yes              | yes              | yes             | no               |
+| pickles time             | mina/pickles* | time     | yes              | yes              | yes             | yes              |
+| pickles memory           | mina/pickles* | memory   | yes              | yes              | yes             | no               |
+
+\* these have to be in Mina now because pickles is in Mina, and while snarky has it's repo, to use it requires a few things still in Mina, if it were possible change that it would make these benchmarks easier and probably cheaper.
```
rbonichon, line 71:
```suggestion
\* these have to be in Mina now because pickles is in Mina, and while snarky has its own repository, using it requires a few things still in Mina, so that we might want to consider some changes to make these benchmarks easier and probably cheaper.
```

rbonichon: Could you exemplify 1 or 2 changes that would be required?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Kimchi time
+
+One of the simplest measuring running time, can be an specific case or allow more flexibility by introducing parameters like for example different circuit sizes. We already have most of the infrastructure for this with criterion. Would ideally have a dedicated computer, considering that there aren't that many PRs in proof-systems the requirements shouldn't be that high.
```
rbonichon, line 75:
```suggestion
One of the simplest measuring running time, can be a specific case or allow more flexibility by introducing parameters like for example different circuit sizes. We already have most of the infrastructure for this with criterion. We would ideally have a dedicated computer. However, considering that there aren't that many PRs in proof-systems, the requirements shouldn't be that high.
```

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Kimchi time flamegraph
+
+This can be considered a variant of the previous, some more research is required but ideally can be generated from the already existing benchmarks with little to no changes.
```
rbonichon, line 79:
```suggestion
This can be considered a variant of the previous. Some more research is required but this information ideally can be generated from the already existing benchmarks with little to no changes.
```

rbonichon: Link to previous section

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+This may not be as interesting to have in CI, but can be useful to run locally for those working in optimizations.
```
rbonichon, line 80: I agree.

book/src/rfcs/benchmarking.md
rbonichon: Link to criterion?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Kimchi instructions
+
+This is basically counting the CPU instructions that takes to run some code, could be use as an alternative for CI given that it won't be affected by noise and can work in any computer. It would also be faster by running the code only once. The implementation will require a bit more research but one option is the crate iai, with some thinking it could be possible to have this and the time benchmarks share the same code.
```
rbonichon, line 84: Link to `iai` https://crates.io/crates/iai ?. Counting instructions is nice but counting CPU cycles might make more sense, especially on a CISCI-like architecture like `x86. However I see that iai can do that. Should we switch the instructions target to the cycles one?  Or be more generic wrt to the info that `iai` can generate.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+Is worth mentioning that the result are specific to the code and even the computer running the benchmark and thus not useful for comparisons with other projects.
+
+### Kimchi memory
+
+Similar to time, but with memory, the constraints are different in some ways, noise is not generally a problem and any computer cab be used, and also requires less sampling and should run faster. This is to some degree already implemented through a custom measurement for criterion, and in practice can be slower than measuring time, the instrumentation makes it run slower and the some limitations in criterion require to run the code more than necessary. With that in mind the best would be to run just a subset of these benchmark for regressions and let the others be run at discretion.
```
rbonichon, line 89:
```suggestion
Similar to time, but with memory, the constraints are different in some ways, noise is not generally a problem and any computer can be used, and also requires less sampling and should run faster. This is to some degree already implemented through a custom measurement for criterion, and in practice can be slower than measuring time, the instrumentation makes it run slower and the some limitations in criterion require to run the code more than necessary. With that in mind, the best would be to run just a subset of these benchmark for regressions and let the others be run at discretion.
```

rbonichon: Do you already have some ideas about which benchmarks would be marked as "regressions" or not?

rbonichon (commented):
The draft looks quite good.

I however suggested a number of minor typo fixes, rephrasing or extensions to the original text.

Could you mention the tools you'd  be going for for the tests in the OCaml world, as you did in Kimchi?

Another thing: would you be inclined to use 80 (or 100) characters line to ease the reading?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+We can differentiate to groups of people interested in benchmarks:
+
+- Developers working in the crypto stack, mostly our crypto team, where the main interest is to avoid regressions or make performance improvements.
+- People working outside of it, mostly users of our libraries, but could also include decision makers comparing other options or our marketing team, here the interest is more about more general metrics that can be compared with the competition or used to learn how to most efficiently use our technology.
+
+Must be mentioned that some overlap should be expected, some benchmarks may be useful for anyone while other just to some.
+
+### Metrics
+
+There are two main metrics, performance and memory, and a smaller one hardware utilization.
```
joseandro, line 20:
```suggestion
Three metrics are considered: performance, memory, and hardware utilization. The main metrics are performance and memory. The hardware utilization metric has a more limited scope.
```

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### Performance
+
+Basically execution time, this may be the hardest to measure, because the environment can have a considerable effect over the results, some high quality machine would be required, ideally a fully dedicated one to minimize noise, there is a way of measuring that overcomes noise problems that will be mentioned.
+It can be subdivided it 2 other metrics or alternative approaches:
+
+- Execution time: The most simple, run the program in question several times and take the average, it may become quite expensive if the program measured takes some time because that time will be multiplied by the number of samples, in those cases it can be made faster sacrificing precision. This benchmark is one of the most interesting for external users, and one of the main comparison point for evaluating the competition. It can also serve for internal developer as a regression test or to verify performance improvements.
```
joseandro, line 27:
```suggestion
: Execution time
The simplest approach. It runs the program several times and returns the overall average execution time. This benchmark might become slow in cases where the program takes too much time to run. In such scenarios, this benchmark can be made faster by reducing the number of samples (i.e., worst precision) it uses. This approach is one of the most interesting for external users and a key datapoint to compare how our proof-system fares against competitors. The execution time metric can also help our engineering teams as a metric to identify regressions or to verify performance improvements.
```

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
...
+
+- CPU instructions: This method instead of recording execution time counts the number of cpu instructions used, not very useful for external developer given that it doesn't translate directly to time. It is mostly useful to internal developer as an alternative to execution time, it is faster given that you only have to run it once to collect the information and can be run on cheap machines without significative noise given that execution time doesn't affect the number of instructions. Additionally the tools used to measure it also collect some other metrics that can be useful to aide optimization like cache misses and branch mispredictions.
```
joseandro, line 29:
```suggestion
: CPU instructions
Instead of recording execution time, this method counts the number of cpu instructions used in a program execution, not very useful for external developer given that it doesn't translate directly to time. It is mostly useful to our engineering teams as an alternative to execution time, it is faster given that you only have to run it once to collect the information, and can be run on cheap machines without significative noise given that execution time doesn't affect the number of instructions. Additionally, the tools used to measure it also collect some other metrics that can be useful to aide optimization like cache misses and branch mispredictions.
```

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,132 @@
...
+- The length of the hash chain can easily be a parameter allowing to measure different lengths. 
+- We want memory and time for compilation, proving and verification, it will be implemented as several benchmarks but they will share most code.
+- The number of constraints can be easily shown for each length.
+- This benchmark with some specific length can be used as regression tests in CI.
+- The full benchmark with several lengths can be shown at discretion of those interested running locally, or even added as optional to CI with some way to trigger it.
+
+### Pickles hash chain
+
+Same as kimchi hash chain but in pickles, we have less to cover here because we already have the kimchi ones, some time and memory benchmarks could be used for regressions and to observe the overhead, but having parameters like different circuits sizes may be redundant.
+A new benchmark possible would be to have an alternative circuit that uses recursion to make the hash chain instead of a single circuit, it could be interesting to have a comparison to the non recursive approach.
```
rbonichon, line 128:
```suggestion
A new  possible benchmark would be to have an alternative circuit that uses recursion to make the hash chain instead of a single circuit. It could be interesting to have a comparison to the non recursive approach.
```

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,132 @@
...
+
+### General kimchi benchmarks
+
+Some general benchmarks covering compilation, proving and verifying, both time and memory. Ideally all of them will parameterized with different circuit sizes and a subset can be used as regression tests. We could also add here benchmarks for the different operations of our polynomial commitments scheme.
```
rbonichon, line 132:
```suggestion
Some general benchmarks covering compilation, proving and verifying, both time and memory. Ideally all of them will be parameterized with different circuit sizes and a subset can be used as regression tests. We could also add here benchmarks for the different operations of our polynomial commitments scheme.
```

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,132 @@
...
+## Types of benchmarks
+
+There are different metrics that we can collect, and different ways of presenting them depending of who is interested in them.
+
+### Audience
+
+We can differentiate to groups of people interested in benchmarks:
+
+- Developers working in the crypto stack, mostly our crypto team, where the main interest is to avoid regressions or make performance improvements.
+- People working outside of it, mostly users of our libraries, but could also include decision makers comparing other options or our marketing team, here the interest is more about more general metrics that can be compared with the competition or used to learn how to most efficiently use our technology.
```
bkase, line 14: I don’t think this audience is entirely served by the current proposal. Decision makers for adopting this proof system stack at the application-level will want to understand the wall-clock time on commodity hardware of proving zkApps.

zkApp proofs (for now) are not proven in native execution even when they’re run server-side. Moreover, we probably need non-native benchmarks to help us internally decide how to prioritize building a zkapp flow that offloads some work to a native proving backend server-side.

Concretely, I think this translates to including in our benchmark matrix the following extra kinds of performance and memory tests:
{kimchi,pickles,snarky}-jsvm-{browser,server}

where “jsvm” is a mix of WASM and jsofocaml like SnarkyJS uses and browser means “run in the browser {V8 (Chrome), Safari, Firefox}” and server means running under a nodejs V8 environment. If we need to cut scope, just V8 for now in one of browser/server could be sufficient.

I am missing some context perhaps but I think for external audiences, these pickles web-based tests are either the most important benchmarks or second after native kimchi (for the purpose of being comparative against native proof systems).

bkase (changes requested):
I’m excited we’re spending energy on this benchmarking work. Thank you so much for this RFC!

Right now my main feedback is I think we’re missing one of the most important kinds of benchmarks for this project. I elaborate below. And, in light of this, more context on motivation which restates product requirements and rationale could help guide that discussion or give readers like me more context.

fabrizio-m (commented):
Some responses, more to come.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,109 @@
+# Benchmarking RFC
+
+This document proposes a high level approach to benchmarking the crypto stack, mostly kimchi, snarky and pickles. The main concern is to set up an infrastructure where it is possible to integrate different kinds of benchmarks with minimal effort. While the specific benchmarks are not an immediate concern, a few benchmarks of interest will be mentioned later.
+
+## Types of benchmarks
+
+There are different metrics that we can collect, and different ways of presenting them depending of who is interested in them.
```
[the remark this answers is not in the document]
fabrizio-m: done

book/src/rfcs/benchmarking.md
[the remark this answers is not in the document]
fabrizio-m: done

book/src/rfcs/benchmarking.md
fabrizio-m: done

book/src/rfcs/benchmarking.md
fabrizio-m: I added a bit at the end about leaving browser out of scope for now, but could be added at some point.

fabrizio-m:
I think we can go with `Core_bench` and `memtrace` for ocaml, at least for now should be enough, feel free to suggest alternatives.
I also noted that Github seems to not render definition list, I'll fix it with HTML or mimic it with some other feature when we are done with reviews.

rbonichon:
Don't bother with description lists if github does not support that for their markdown format.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Where to run the benchmarks
+
+To start everyone with the resources should be able to run the benchmarks
+locally, additionally we may want some of the benchmarks to run in CI to detect
+regressions. Some benchmarks may be considerably expensive to run, a decision
+will have to be made about which benchmarks to run and how often we should run
+them.
+
+## Parameters
```
bkase, line 108: Which explicitly will we parameterize and what will the parameters be?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+| pickles memory           | mina/pickles* | memory   | yes              | yes              | yes             | no               |
+
+\* these have to be in Mina now because pickles is in Mina, and while snarky has
+its own repository, using it requires a few things still in Mina, so that we
+might want to consider some changes to make these benchmarks easier and probably
+cheaper.
+
+### Kimchi time
+
+One of the simplest measuring running time, can be a specific case or allow more
```
bkase, line 141: Which will we choose? A specific case? What is the case? Something more flexible? Can you describe it in detail?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+flexibility by introducing parameters like for example different circuit sizes.
+We already have most of the infrastructure for this with
```
bkase, line 143: What is the remaining infrastructure beyond the criterion crate?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+[criterion](https://crates.io/crates/criterion). We would ideally have a
+dedicated computer. However, considering that there aren't that many PRs in
+proof-systems, the requirements shouldn't be that high.
+
+### Kimchi time flamegraph
+
+This can be considered a variant of the [previous](#kimchi-time). Some more
+research is required but this information ideally can be generated from the
```
bkase, line 151: I think the time for the research is in this document. If it’s too complex to include in the scope of this RFC, then we can remove it for now and address it in another rfc

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+already existing benchmarks with little to no changes.  This may not be as
+interesting to have in CI, but can be useful to run locally for those working in
+optimizations.
+
+### Kimchi instructions
+
+This is basically counting the CPU instructions that takes to run some code,
+could be use as an alternative for CI given that it won't be affected by noise
+and can work in any computer. It would also be faster by running the code only
+once. The implementation will require a bit more research but one option is the
```
bkase, line 161: Same with this one re research

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+running the benchmark and thus not useful for comparisons with other projects.
+
+### Kimchi memory
+
+Similar to time, but with memory, the constraints are different in some ways,
+noise is not generally a problem and any computer can be used, and also requires
+less sampling and should run faster. This is to some degree already implemented
+through a custom measurement for criterion, and in practice can be slower than
+measuring time, the instrumentation makes it run slower and the some limitations
+in criterion require to run the code more than necessary. With that in mind, the
```
bkase, line 174: Which are the subset? Can you be specific?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+best would be to run just a subset of these benchmark for regressions and let
+the others be run at discretion.
+
+### Kimchi memory flamegraph
+
+Similar to the time flamegraph, ideally can be made from the memory benchmark
+without much additional work.
```
bkase, line 181: What specifically needs to happen?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Snarky time
+
+This is similar to kimchi time, but the fact that it will be in the mina repo
+should be considered, CI will run much more often and costs may be higher, this
+would be a reason to prefer to have benchmarks in kimchi when possible.
```
bkase, line 187: Do you recommend removing this then? Or if it’s part of kimchi, how?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Snarky memory
+
+Here the main interest may be regressions, everything else can probably done at
+the kimchi level more precisely.
```
bkase, line 192: Agreed, what are the steps to achieve this?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Pickles time
+
+This is similar to snarky time, mostly differing in the overhead of the
+recursion. At this level beyond some regression we will probably have mostly
+application circuits being benchmarked.
```
bkase, line 198: Which applications and how will they be benchmarked specifically?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Pickles memory
+
+Similar to snarky, this will probably be mostly regression tests and maybe some
```
bkase, line 202: What will they be exactly?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+example application to give users an idea of what to expect.
+
+## Benchmarks
+
+Here is a non exhaustive list of benchmarks.
+
+### Kimchi hash chain
+
```
bkase, line 210: This section is a great start and has more detail, but there are still questions on precisely what we will do here? What will be used to determine the lengths that we set for the different runtimes? How will the tests be written? Where is the code, which libraries, etc

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+trigger it.
+
+### Pickles hash chain
+
+Same as kimchi hash chain but in pickles, we have less to cover here because we
+already have the kimchi ones, some time and memory benchmarks could be used for
+regressions and to observe the overhead, but having parameters like different
+circuits sizes may be redundant.  A new  possible benchmark would be to have an
+alternative circuit that uses recursion to make the hash chain instead of a
+single circuit. It could be interesting to have a comparison to the non
```
bkase, line 230: Can you add specifics here too?

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+recursive approach.
+
+### General kimchi benchmarks
```
bkase, line 233: Same here

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+# Prior art
+
+Right now there are just a few benchmarks covering specific cases that can be
+run for command line, mostly in the proof-systems repo. Almost all benchmarks
+are also limited to measuring only time.
+
+# Unresolved questions
+
+- While the tools for rust are mostly covered, specific tools for ocaml have yet
```
bkase, line 262: Do you intend on sorting out the ocaml parts before starting the project? If not, let’s remove them all from the RFC and make a new RFC later with the ocaml parts.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+example if we were to start using it, but for now will only include CPU/core
+utilization.
+
+### Layers
+
+There are 3 main layers of the stack that we want to consider, kimchi, snarky
+and pickles. Given that they are implemented in different languages and exist in
+different repositories special considerations each layer will be required.
+
+## Where to run the benchmarks
```
bkase, line 100: Can we run the benchmarks within the profiling frontend that Viable Systems created? https://perf.ci.openmina.com/builds . I bet the velocity team can help support this. I’d love to see explanations on how that would work in this RFC and I’d recommend looking into before deciding on benchmarking libraries because I know they’ve already instrumented some of the rust and ocaml code. I’ll invite someone from viable systems to comment here to describe how it works from their side.

bkase (commented):
Thank you for the explanations! I understand we’re descoping the browser runtime for now. Overall, I think everything is on the right path but I would love to see more specific examples/instructions throughout.

There may actually be a significant amount of work to sort out all the details and answer all the questions I posed. Alternatively, we can descope and remove some of the benchmarks for now so that we can start on the project earlier.

Note that some of the answers to my questions are present later in the doc so maybe you can lift up some of that content or mention in the section where relevant to point to specific parts later in the doc.

book/src/rfcs/benchmarking.md
```diff
@@ -0,0 +1,266 @@
...
+
+To start everyone with the resources should be able to run the benchmarks
+locally, additionally we may want some of the benchmarks to run in CI to detect
+regressions. Some benchmarks may be considerably expensive to run, a decision
```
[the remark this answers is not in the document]
fabrizio-m: I wasn't thinking of any in particular, it's more a consideration that we will have to do once in CI, relative to other benchmarks and our budget.
For each benchmark we will have to decide if it's worth having it in CI based on the running time we get.

book/src/rfcs/benchmarking.md
fabrizio-m: While I was thinking more about the general ability to have parameters, the only one I can think for now is circuit size.
It's basically running the same benchmark for different values of the parameter, for example running a prover time benchmark for circuits of size $2^n$ for $n$ in $[8..16]$.
I can make it more specific and list the parameters with only circuit size and then amend it if we want some other new parameters.

book/src/rfcs/benchmarking.md
fabrizio-m: With specific case I mean without parameters. I think it would be a per benchmark decision, we could choose for example circuit size $2^{16}$ that is similar to the mina circuits or have more flexibility through parameters and benchmark a range of sizes.

book/src/rfcs/benchmarking.md
fabrizio-m: In dependencies I don't we need anything we don't have already, there is the hardware, but I think that is not specific to this but more general to the whole RFC.
What remains is just some code, mostly thinking more in depth about the code structure now that there will be more than a single benchmark, creating utilities for that could help reduce code duplication, adding documentation and refactoring where necessary. Mostly small tasks that can be done in a day.

book/src/rfcs/benchmarking.md
fabrizio-m: Looked into it, at least for time flamegraphs it shouldn't be too hard, there is [cargo flamegraph](https://github.com/flamegraph-rs/flamegraph) that just wraps the binary and creates the flamegraph, it shouldn't require any change to the code.

book/src/rfcs/benchmarking.md
fabrizio-m: I think we can settle with `iai`, we already have it as dependency, it could require some more work that just running the time benchmark but should be a mere implementation detail.

book/src/rfcs/benchmarking.md
fabrizio-m: I don't think it can be specified at this point, it would depend on how much each benchmark costs, (both time and money), cost of all benchmarks, how much do we want to spend and what benchmarks seem more valuable relative to others.
We could start with running every single benchmark and removing the least valuables until we are under budget.

book/src/rfcs/benchmarking.md
fabrizio-m: I think is just a reason to prefer having benchmarks in `proof-systems` when possible, but some benchmarks will have to be at this level for their very nature.

book/src/rfcs/benchmarking.md
fabrizio-m: Not sure I understand the question, I decided to use `memtrace` since then for ocaml memory benchmarks, does that answer your question?

book/src/rfcs/benchmarking.md
fabrizio-m: I don't this section should that specific, for tools I decided to use `Core_bench` for ocaml time benchmarks, I can add that. About specific applications that would be in `## benchmarks`. This section is more about a common infrastructure/approach that can be reused for many benchmarks.

book/src/rfcs/benchmarking.md
fabrizio-m: The lengths chosen for regression tests are mostly arbitrary, the only constrains is CI cost, the process I can think of is choosing an arbitrary number and going up or down depending on how long it takes to run.
The code code should be Rust in the `proof-systems` repo, if that is precise enough, it shouldn't require additional dependencies over the already mentioned `criterion`.

fabrizio-m:
This RFC will be split into 2 parts to address the rust and ocaml codebase separately, #1083 addresses rust, and the one to address ocaml will be created later.

joseandro:
@bkase can you take another look? It's now only focused in Kimchi.

PR #1082, opened 2023-05-18
Add base58 encode/decode to secret key

MartinOndejka:
Add base58 encode/decode to SecKey in signer crate.


mrmr1993 (approved):
Very nice, thank you! There are some clippy linter errors though, would you mind fixing those up?

jspada (approved):
Great contribution.   Looks good to me!

PR #1083, opened 2023-05-25
Benchmarking RFC part 1: Rust

fabrizio-m:
Superseding #1063 to some degree, this focus only in the rust part and should only involve `proof-systems`.


dannywillems:
Should it not be in o1labs/rfcs?

dannywillems:
Closing this as staled. And we are also not working on RFC anymore.

PR #1084, opened 2023-06-01
Get rid of action rs and test more caml rust versions

dannywillems:
Discussed on Slack.


dannywillems:
Note that the CI takes a long time to run. OCaml seems to be compiled. However, the time is mostly spent in the runnig the tests. Caching should be added.

dannywillems:
We can use GH actions cache features: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows.

I would, in a job, check if Cargo.toml/Cargo.lock has changed (also OCaml deps), and if so, compute a hash of the files, do a fresh deps install, and cache in a key like `rust-${{ env.rust_toolchain_version }}-ocaml-${{ env.ocaml_version }}-hash-$HASH_CARGO_LOCKED`. The cache will be used by next jobs.
With this, we speed up the next jobs.

rbonichon (approved):
LGTM.

Same as for https://github.com/o1-labs/ocaml-gen/pull/17, would https://github.com/dtolnay/rust-toolchain be relevant here?

dannywillems:
> LGTM.
> 
> Same as for [o1-labs/ocaml-gen#17](https://github.com/o1-labs/ocaml-gen/pull/17), would https://github.com/dtolnay/rust-toolchain be relevant here?

See https://github.com/o1-labs/ocaml-gen/pull/17#issuecomment-1598853418

fabrizio-m (approved):

PR #1096, opened 2023-06-26
Add keccak sponge

mrmr1993:
This PR implements the `FrSponge` and `FqSponge` traits using a sponge backed by keccak256.

This sponge is explicitly designed to provide support for verification on the EVM, and so it builds a sponge out of the keccak256 hash function instead of using the keccak sponge directly.

This PR builds upon #1095, adding a test which exercises the BN128 pairing backend and the keccak sponge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1103,6 +1103,7 @@ dependencies = [
  "strum",
  "strum_macros",
  "thiserror",
+ "tiny-keccak",
  "turshi",
  "wasm-bindgen",
 ]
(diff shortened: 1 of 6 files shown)


querolita (approved):
Very nice! Nonetheless, I wonder if all parameters of this version of Keccak are the same ones as those we have been using for the gadgets. Would you mind checking this if it is not too time consuming? 

For example, on input the hexadecimal value `0x4920616d20746865206f776e6572206f6620746865204e465420776974682069642058206f6e2074686520457468657265756d20636861696e` does it return `0x63858e0487687c3eeb30796a3e9307680e1b81b860b01c88ff74545c2c314e36` ?

I am wondering whether it pads to a certain multiple for the input length, if it takes 64-bit or 32-bit words, if it is big or little endian, etc.

PR #1098, opened 2023-06-28
Fix potential soundness bug in rotation

querolita:
This is the Rust counterpart of this [Mina PR](https://github.com/MinaProtocol/mina/pull/13491)

It is not necessary right now, as we are not using rust gadgets but only Snarky's. Nonetheless, it is good to have the same code on both sides so it is easier to maintain.


rbonichon (approved):
LGTM

PR #1101, opened 2023-07-03
Benchmark cycles

fabrizio-m:
Adds support for benchmark based on performance counters including but not limited to cycles. It could be made significantly faster in the future by rewriting the criterion measurement to run the function a single time.

as an example:
`cargo criterion -p kimchi --bench perf`


dkijania:
For future reference on my local machine it fails to run criterion until i not implemented fix from:

`https://superuser.com/questions/980632/run-perf-without-root-rights`

PR #1106, opened 2023-07-06
Remove evaluated columns from linearization

mrmr1993:
This PR builds upon #1105, removing the newly-evaluated selectors from the linearization.

This PR is separated for the purposes of the corresponding Mina PR, where it results in the update of a large generated file, along with some other pickles changes.


rbonichon (approved):
LGTM

dannywillems (approved):

querolita (approved):

PR #1107, opened 2023-07-06
Use 'fixed hiding' for selector commitments

mrmr1993:
This PR is a follow-on from #1106, adjusting the blinding factor of selector commitments for the remaining optional gates, so that a 'zero' commitment does not occur for an unused gate.


rbonichon (approved):
LGTM

querolita (approved):

PR #1109, opened 2023-07-06
Evaluate selectors for optional custom gates

mrmr1993:
This PR is part of a chain, building upon #1107.

This PR adds evaluations for all of the optional custom gates, which will allow them to be removed from the linearization (in an upcoming PR).


kimchi/src/verifier.rs
```diff
@@ -437,6 +437,36 @@ where
...
+            .chain(if index.xor_comm.is_some() {
+                Some(Column::Index(GateType::Xor16))
+            } else {
+                None
+            })
+            .chain(if index.rot_comm.is_some() {
+                Some(Column::Index(GateType::Rot64))
+            } else {
+                None
+            })
```
rbonichon, line 469: Suggestion: write each optional commitment chain call using `Option.map`
```
            .chain(index.range_check0_comm.map(|_| Column::Index(GateType::RangeCheck0)))          
```

Bonus(?) point: it makes it look like the rest of the code

kimchi/src/verifier.rs
```diff
@@ -764,6 +821,37 @@ where
...
+    .chain(if verifier_index.xor_comm.is_some() {
+        Some(Column::Index(GateType::Xor16))
+    } else {
+        None
+    })
+    .chain(if verifier_index.rot_comm.is_some() {
+        Some(Column::Index(GateType::Rot64))
+    } else {
+        None
+    })
```
rbonichon, line 854: The `Option.map` refactoring would apply here as well.

rbonichon (approved):

querolita (approved):

dannywillems (approved):

kimchi/src/verifier.rs
rbonichon: Pushed the changes

PR #1111, opened 2023-07-07
Evaluate selectors for non-optional gates

mrmr1993:
This PR builds upon #1110, adding evaluations for all of the 'non-optional' gates, which will allow them to be removed from the linearization (in an upcoming PR).

This PR replaces #1105 as part of a rebase to make reduce the scope of changes needed for berkeley.

PR #1116, opened 2023-07-10
Stop linearizing optional gate selectors

mrmr1993:
This PR builds upon #1111, using the evaluations that it adds in place of the linearization.


rbonichon (approved):

kimchi/src/linearization.rs
```diff
@@ -310,6 +310,14 @@ pub fn linearization_columns<F: FftField + SquareRootField>(
     h.insert(Index(GateType::EndoMul));
     h.insert(Index(GateType::EndoMulScalar));
 
+    // optional columns
```
dannywillems, line 313: Should it not be conditioned based on the feature flags?

querolita (commented):
I am curious to solve the doubt that Danny brought up

PR #1117, opened 2023-07-10
Remove special-casing of lookup-related columns

mrmr1993:
This PR removes the specialized structure related to the lookup columns. This helps to avoid the erroneous assumptions that it lead to on the pickles side, simplifying the handling of columns in general.


rbonichon (approved):

querolita (commented):
I wonder if any of this is incompatible with having a case for `Lookup` gate in `plonk_constraint_sytem.ml`
https://github.com/MinaProtocol/mina/blob/d9c0d696c2dfeba7e0f2caa1cdd36811ecc28c42/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml#L1677

dannywillems (approved):
LGTM, except clippy is not happy :)

PR #1132, opened 2023-07-19
Normalize the runtime table configuration to "custom"

rbonichon:
Previously, we were considering two types of runtime table configurations, indexed and custom.
Indexed where supposed to represent "standard" array access, using indexed from 0 to N, which can be easily represented as a "custom" runtime table configurations whose first columns is from 0 to N.

It also simplifies the wasm binding as enum with structure values can not be represented using wasm-bindgen.


dannywillems (approved):

rbonichon:
!ci-build-me

PR #1181, opened 2023-08-09
Permutation argument in expression framework

mrmr1993:
This PR implements the permutation argument in the expression framework. This results in
* some tweaks to the expression framework to support the necessary constants and functions
* some changes to the verifier to use this expression version.

Note that this is not used by the prover, nor is it included in the total linearization. This is due to some quirks:
* computing as the prover, we will detect the target domain size as `> 8n`, due to an off-by-one error
* integrating this into the linearization will force the prover to use it in order to make it available to the verifier, because of how the verifier index is computed
* otherwise exposing it in the linearization will change how the pickles circuit is computed, making this an incompatible change.

This change would make it trivial to add an evaluation for the final permutation coefficients column. Again, this is not done here because it would break compatibility.


volhovm:
@mrmr1993 Still relevant? (seems so.) Should be rebased towards `master`?

PR #1183, opened 2023-08-17
Rust 1.71.1 upgrade

joseandro:
Rust 1.71.1 contains a patch for [CVE-2023-38497](https://www.cve.org/CVERecord?id=CVE-2023-38497)

In previous versions of Rust, Cargo did not respect the umask when extracting crate archives on UNIX-like systems. 
More information can be found [here](https://blog.rust-lang.org/2023/08/03/cve-2023-38497.html).


.github/workflows/rust.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   run_checks:
     strategy:
       matrix:
-        rust_toolchain_version: [1.67, 1.68, 1.69]
+        rust_toolchain_version: [1.71.1]
```
mrmr1993, line 21: I think we want to keep compatibility with the other versions for a while. Can you add this one instead of replacing the others?

PR #1185, opened 2023-08-19
[berkeley] Dedicated prover for a minimal circuit with given public input

mrmr1993:
This PR adds an optimised prover for kimchi, specialised to the circuit that asserts only the constraint `witness[0] = public_input` on every row.

This prover can be used to generate a 'stubbed' pickles proof, since the only requirements for such a proof are
* that the public input has the right format and contents, and
* that the proof itself is a valid kimchi proof in a berkeley-compatible configuration.

This can be used to replace the 'trivial zkApp prover' used extensively in the Mina protocol transaction tests, which should result in improved unit and integration test run times in the Mina CI.


volhovm:
@mrmr1993 Is it still relevant/ready for review?

I assume since berkeley is feature-frozen, I can at least re-direct it towards `master`. What do you think?

kimchi/src/public_input_only_prover.rs
```diff
@@ -0,0 +1,627 @@
...
+        // The polynomial `f(x) = 1`, without blinding.
+        // The generic gate is enabled on every row; combined with the coefficients, and with the
+        // mixin for public inputs, this encodes the equation `witness[0] = public_input`.
+        // Generic gate's equation is:
+        //    c_0 \cdot l + c_1 \cdot r + c_2 \cdot o + c_3 \cdot (l \times r) + c_4
+        // so we set c_0 = 1, and everything else to zero.
+        generic_comm: make_comm(srs.g[0]),
+        // The polynomials `f(x) = 0`, with blinding factor 1.
+        // This disables these gates.
+        psm_comm: make_comm(srs.h),
```
volhovm, line 104: I'm wondering if these commitments can be just `1` instead of `h^1`?

kimchi/src/public_input_only_prover.rs
```diff
@@ -0,0 +1,627 @@
...
+                // The polynomial `f(x) = 0`, with blinding factor 1.
+                // This blinding allows us to represent the commitment in affine coordinates.
+                srs.h
+            })
+        }),
+        // The polynomial `f(x) = 1`, without blinding.
+        // The generic gate is enabled on every row; combined with the coefficients, and with the
+        // mixin for public inputs, this encodes the equation `witness[0] = public_input`.
+        // Generic gate's equation is:
+        //    c_0 \cdot l + c_1 \cdot r + c_2 \cdot o + c_3 \cdot (l \times r) + c_4
```
volhovm, line 99: @mrmr1993 could you please double check this comment I added is what you had in mind?

kimchi/src/public_input_only_prover.rs
```diff
@@ -0,0 +1,627 @@
...
+                shifted: None,
+            };
+            let coefficients_form = DensePolynomialOrEvaluations::<_, D<_>>::DensePolynomial;
+
+            let mut polynomials_to_open = recursion_polynomials
+                .iter()
+                .map(|(p, d1_size)| (coefficients_form(p), None, non_hiding(*d1_size)))
+                .collect::<Vec<_>>();
+            // public polynomial
+            polynomials_to_open.push((coefficients_form(&public_poly), None, fixed_hiding(1)));
```
volhovm, line 493: Why not `fixed_hiding(0)` here?

volhovm (approved):
I understand about 80% of the changes (except perhaps bits related to ft computation), so given that it's a test prover anyway (I added a remark to the top of the file), I'm happy to merge.

PR #1201, opened 2023-08-31
make expressions generic over columns

fabrizio-m:
Adding a generic to `Expr` allowing to be generic over the columns, also adding an implementation for `Column`.

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -16,23 +16,26 @@ use ark_ff::{FftField, Field, One, PrimeField, Zero};
 use ark_poly::{
     univariate::DensePolynomial, EvaluationDomain, Evaluations, Radix2EvaluationDomain as D,
 };
+use constraints::ExprOps;
 use itertools::Itertools;
 use o1_utils::{foreign_field::ForeignFieldHelpers, FieldHelpers};
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
-use std::ops::{Add, AddAssign, Mul, Neg, Sub};
 use std::{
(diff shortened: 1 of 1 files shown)


fabrizio-m:
!ci-build-me

kimchi/src/circuits/expr.rs
```diff
@@ -468,20 +546,20 @@ impl FeatureFlag {
 /// the corresponding combination of the polynomials corresponding to
 /// the above variables should vanish on the PLONK domain.
 #[derive(Clone, Debug, PartialEq)]
-pub enum Expr<C> {
+pub enum Expr<C, Col: ColTrait = Column> {
```
querolita, line 549: I am a bit confused aboout this notation. 
1. What exactly was `C`?
2. Is `Column` the kind of "columns" we have in Berkeley Kimchi, and if we had a different configuration with a different system we would have to re-write a new `Expr` whose `ColTrait` was `ColumnNew` for example?

querolita (approved):
Nice looking so far :) 

I left one doubt regarding different column configurations, please reply. 

I made some changes to the witness interface to work generically over any number of columns [here](https://github.com/o1-labs/proof-systems/edit/zkvm/keccak/layout/kimchi/src/circuits/witness/mod.rs?pr=%2Fo1-labs%2Fproof-systems%2Fpull%2F1248) 

It is not necessary using that interface to write the witness generation code for a gate. But in any case, once both PRs (this and that one) are merged, I will be happy to help unify the notation so it matches this one's better, if you think it could make sense.

mrmr1993:
There's an equivalent version of this already merged to `master`. Closing.

PR #1230, opened 2023-09-06
abstract constants

fabrizio-m:


fabrizio-m:
!ci-build-me

querolita (approved):
Accepting, left one comment with a question 😃

querolita:
Just a remainder -> I believe we should be merging against master instead of develop

PR #1231, opened 2023-09-08
Domain computation: runtime table cfg is given as option, not list

dannywillems:
iter() was applied on the option, not the list.

Fixes https://github.com/MinaProtocol/mina/issues/14072

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -670,8 +670,7 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
                 )
                 .sum();
             // After that on the runtime tables
-            if runtime_tables.is_some() {
-                let runtime_tables = runtime_tables.as_ref().unwrap();
+            if let Some(runtime_tables) = runtime_tables.as_ref() {
                 for runtime_table in runtime_tables.iter() {
                     lookup_domain_size += runtime_table.len();
                 }
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -795,4 +795,32 @@ pub mod tests {
             Self::for_testing(gates)
         }
     }
+
+    #[test]
+    pub fn test_domains_computation_with_runtime_tables() {
+        let dummy_gate = CircuitGate {
+            typ: GateType::Generic,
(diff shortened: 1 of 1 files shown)


kimchi/src/circuits/constraints.rs
```diff
@@ -668,23 +669,28 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
                     },
                 )
                 .sum();
-            for runtime_table in runtime_tables.iter() {
-                num_lookups += runtime_table.len();
+            // After that on the runtime tables
+            if runtime_tables.is_some() {
+                let runtime_tables = runtime_tables.as_ref().unwrap();
```
rbonichon, line 674: Can we apply the `if let Some(x) = runtime_tables` construct here? (it would be consistent with the code later on L682

fabrizio-m, line 674: It may be better matching the `Option` or using `if let` in case you don't care about the `None` case.

rbonichon (approved):
LGTM

kimchi/src/circuits/constraints.rs
```diff
@@ -668,23 +669,28 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
...
+                for runtime_table in runtime_tables.iter() {
+                    lookup_domain_size += runtime_table.len();
+                }
             }
+            // And we add the built-in tables, depending on the features.
             let LookupFeatures { patterns, .. } = &lookup_features;
             for pattern in patterns.into_iter() {
```
fabrizio-m, line 681: You may be interested in `Iterator.filter_map` for this cases.

fabrizio-m (approved):

kimchi/src/circuits/constraints.rs
dannywillems: I do agree, maybe using instead a fold, but I would not add this in the scope of this PR.

kimchi/src/circuits/constraints.rs
dannywillems: See 58c9b33a7bde95f0cda4e06fc2a35b1dda752fc2

dannywillems: See 58c9b33a7bde95f0cda4e06fc2a35b1dda752fc2

PR #1232, opened 2023-09-08
Asserting the domain is big enough

dannywillems:
Even though we get a panic later, we do not get enough information to track down the information.

Not sure it would be useful now we have tests, still suggesting.


dannywillems:
Maybe useless, closing to clean the opened PR.

PR #1237, opened 2023-09-12
Lookup tables: introduce width method to get the width

dannywillems:
Using `self.data.len()` can be error-prone, and an index is easily introduced. It is also always better to abstract with an API using methods. In the future, I would recommend making the data field abstract.

Also, the length method is not fully correct. It supposes the user built the lookup table and know that all columns must have the same length. The first column, by construction, might have a different length.


rbonichon (approved):
LGTM

PR #1239, opened 2023-09-12
Lookup tables: get length using method instead of relying on struct f…

dannywillems:
…ield


rbonichon (approved):
Small improvements but improvements nonetheless. You have my go ahead!

PR #1240, opened 2023-09-12
Test runtime tables with ID zero

dannywillems:


kimchi/src/tests/lookup.rs
```diff
@@ -465,3 +465,139 @@ fn test_runtime_table_with_more_than_one_runtime_table_data_given_by_prover() {
...
+    // witness
+    let witness = {
+        let mut cols: [_; COLUMNS] = array::from_fn(|_col| vec![Fp::zero(); gates.len()]);
+
+        // only the first 7 registers are used in the lookup gate
+        let (lookup_cols, _rest) = cols.split_at_mut(7);
+
+        for row in 0..n_row {
+            // Lookup in table ID 0
+            lookup_cols[0][row] = Fp::zero();
```
jspada, line 509: Maybe ` = Fp::from(table_id)` would be clearer/neater.

kimchi/src/tests/lookup.rs
```diff
@@ -465,3 +465,139 @@ fn test_runtime_table_with_more_than_one_runtime_table_data_given_by_prover() {
...
+        .witness(witness)
+        .runtime_tables_setup(vec![cfg])
+        .setup()
+        .runtime_tables(vec![runtime_table])
+        .prove_and_verify::<BaseSponge, ScalarSponge>()
+        .unwrap();
+}
+
+#[test]
+fn test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_values() {
```
jspada, line 536: Seems like most of this code is the same as for `test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_fixed_values`

Maybe you can refactor them into a single function and just pass random values to one and fixed values to the other?

kimchi/src/tests/lookup.rs
```diff
@@ -465,3 +465,139 @@ fn test_runtime_table_with_more_than_one_runtime_table_data_given_by_prover() {
...
+    // run test
+    TestFramework::<Vesta>::default()
+        .gates(gates)
+        .witness(witness)
+        .runtime_tables_setup(vec![cfg])
+        .setup()
+        .runtime_tables(vec![runtime_table])
+        .prove_and_verify::<BaseSponge, ScalarSponge>()
+        .unwrap();
+}
```
jspada, line 603: Would be good to add some negative test cases to be sure things fail with invalid witness.

jspada (approved):
Nice stuff! Approved with some minor suggestions.

kimchi/src/tests/lookup.rs
dannywillems: I already added some negative tests. In particular, https://github.com/o1-labs/proof-systems/pull/1196.

dannywillems: They are prefixed with `test_negative`. It might not be the best naming convention, though…

kimchi/src/tests/lookup.rs
dannywillems: On it.

kimchi/src/tests/lookup.rs
dannywillems: Will be removed in the refacto suggested in https://github.com/o1-labs/proof-systems/pull/1240/files#r1322922879

kimchi/src/tests/lookup.rs
dannywillems: See 8193de769e55bf8f1e6da7be885a13b5f795b759

PR #1243, opened 2023-09-13
Sparse lookup tables for Keccak

querolita:
It creates a 2-column lookup table containing the 2^16 elements and their expansion to 64 bits.

~~Still pending to figure out if this is enough (together with some logic changes in the lookup argument to look at single columns) or if we need 3 tables instead (to have access to each column separately), or if we prefer to have more columns in the gate.~~

It additionally creates two more tables containing the above columns separately.

Closes https://github.com/o1-labs/proof-systems/issues/1223


kimchi/src/circuits/lookup/tables/bits16.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use super::BITS16_TABLE_ID;
+
+//~ The lookup table for 16-bits
+
+/// Returns the lookup table for all 16-bit values
+///
+/// # Panics
+///
+/// Will panic if `data` is invalid.
```
mrmr1993, line 12: Can this panic? I think you can remove this line, `data` is entirely internal

kimchi/src/circuits/lookup/tables/reset.rs
```diff
@@ -0,0 +1,34 @@
...
+
+//~ The lookup table for 16-bit expansion for Keccak words encoding.
+//~ This is a 2-column table containing the reset sparse representation of the 16-bit values.
+//~ The first column contains the 16-bit values, and the second column contains their expansion to 64-bit values.
+
+/// Returns the sparse lookup table
+///
+/// # Panics
+///
+/// Will panic if `data` is invalid.
```
mrmr1993, line 14: Likewise, delete

kimchi/src/circuits/lookup/tables/sparse.rs
```diff
@@ -0,0 +1,28 @@
...
+use ark_ff::Field;
+
+//~ The lookup table for 16-bit expansion for Keccak words encoding.
+//~ This is a 1-column table containing the sparse representation of all 16-bit preimages.
+
+/// Returns the sparse lookup table
+///
+/// # Panics
+///
+/// Will panic if `data` is invalid.
```
mrmr1993, line 11: Here also

mrmr1993 (changes requested):
LGTM, but see comments

kimchi/src/circuits/lookup/tables/bits16.rs
querolita: Oh thanks. I believe this comment comes from a copy-paste of another file though.

PR #1245, opened 2023-09-13
Add creation function for lookup structures used in Wasm

rbonichon:
This is necessary to be able to build Rust values within o1js-bindings

### DIFF
--- a/kimchi/src/circuits/lookup/lookups.rs
+++ b/kimchi/src/circuits/lookup/lookups.rs
@@ -550,3 +550,58 @@ fn lookup_pattern_constants_correct() {
         assert_eq!((pat, pat.max_joint_size()), (pat, max_joint_size as u32));
     }
 }
+
+#[cfg(feature = "wasm_types")]
+pub mod wasm {
+    use super::*;
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupPatterns {
+        #[wasm_bindgen::prelude::wasm_bindgen(constructor)]
+        pub fn new(
+            xor: bool,
+            lookup: bool,
+            range_check: bool,
+            foreign_field_mul: bool,
+        ) -> LookupPatterns {
+            LookupPatterns {
+                xor,
+                lookup,
+                range_check,
+                foreign_field_mul,
+            }
+        }
+    }
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupFeatures {
+        #[wasm_bindgen::prelude::wasm_bindgen(constructor)]
+        pub fn new(
+            patterns: LookupPatterns,
+            joint_lookup_used: bool,
+            uses_runtime_tables: bool,
+        ) -> LookupFeatures {
+            LookupFeatures {
+                patterns,
+                joint_lookup_used,
+                uses_runtime_tables,
+            }
+        }
+    }
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupInfo {
+        #[wasm_bindgen::prelude::wasm_bindgen(constructor)]
+        pub fn new(
+            max_per_row: usize,
+            max_joint_size: u32,
+            features: LookupFeatures,
+        ) -> LookupInfo {
+            LookupInfo {
+                max_per_row,
+                max_joint_size,
+                features,
+            }
+        }
+    }
+}


kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -550,3 +550,50 @@ fn lookup_pattern_constants_correct() {
...
 }
+
+#[cfg(feature = "wasm_types")]
+pub mod wasm {
+    use super::*;
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupPatterns {
+        #[wasm_bindgen::prelude::wasm_bindgen]
+        pub fn new(xor: bool, lookup: bool, range_check: bool, foreign_field_mul: bool) -> Self {
```
mitschabaude, line 561: nit: can we actually call this `create()`? Here and elsewhere? This is consistent with `Wire` and also, having a static `new` method without it being a constructor is a bit weird

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -550,3 +550,50 @@ fn lookup_pattern_constants_correct() {
...
+                patterns,
+                joint_lookup_used,
+                uses_runtime_tables,
+            }
+        }
+    }
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupInfo {
+        // Not quite sure about the actual types for max_per_row and max_joint_size
```
mitschabaude, line 589: they're JS `number`s. your code should be fine in that it preserves any reasonably large value

mitschabaude (approved):
just one nit!

kimchi/src/circuits/lookup/lookups.rs
rbonichon: One of the structure (`LookupInfo`) already has a  `create` function which conflicts here. Hence this sub-optimal solution.

There are 2 workarounds:
1. Have the same name (not `create` then) in all the involved types for creation functions
2. Change the name of the existing `create` function to something else (like `create_from_features`)

rbonichon: I just implemented 2. But this clearly needs the approval of `proof-systems` gatekeepers.

kimchi/src/circuits/lookup/lookups.rs
rbonichon: Hum so they're `f64` then?

mitschabaude: Hmm, I think internally in JS, `number` is either `f64` or `i32`, depending on the circumstance, but when passed to Wasm will be converted to the type matching the input type of the generated Wasm function. In this case definitely to a 32-bit integer, `i32`. (Wasm types don't distinguish between `u32` and `i32`, instead the individual instructions treat them as one or the other)

rbonichon: JS numbers are clearly double

mitschabaude: that article also mentions that these doubles are converted to int32 for some operations:

> JavaScript has some lower-level functions that deal with the binary encoding of integer numbers, most notably [bitwise operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#bitwise_shift_operators) and [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) objects. Bitwise operators always convert the operands to 32-bit integers.

what the article doesn't mention is that there are known ways of writing code that the compiler will take as hint that it should use the int32 representation for several operations in a row without converting back to double

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -550,3 +550,50 @@ fn lookup_pattern_constants_correct() {
...
     }
 }
+
+#[cfg(feature = "wasm_types")]
+pub mod wasm {
+    use super::*;
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupPatterns {
+        #[wasm_bindgen::prelude::wasm_bindgen]
```
dannywillems, line 560: Why don't you use constructor?

dannywillems: See https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-rust-exports/constructor.html

dannywillems: Is it related to this PR: https://github.com/o1-labs/o1js-bindings/pull/146/files?

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -550,3 +550,50 @@ fn lookup_pattern_constants_correct() {
...
+                uses_runtime_tables,
+            }
+        }
+    }
+
+    #[wasm_bindgen::prelude::wasm_bindgen]
+    impl LookupInfo {
+        // Not quite sure about the actual types for max_per_row and max_joint_size
+        #[wasm_bindgen::prelude::wasm_bindgen]
+        pub fn create(max_per_row: i32, max_joint_size: i32, features: LookupFeatures) -> Self {
```
dannywillems, line 591: why don't you use u32 directly?

dannywillems (changes requested):
See comments above.

kimchi/src/circuits/lookup/lookups.rs
rbonichon: Done

dannywillems (approved):

PR #1252, opened 2023-09-21
quadricization

fabrizio-m:
Adds a method to `Expr` that will make it degree $<=2$ by adding columns and extracting expressions into them.
This still has to be incorporated into the prover/verifier and thus shouldn't be merged alone.

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -284,7 +284,7 @@ impl<C: ColTrait> Variable<C> {
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 /// An arithmetic expression over
 ///
 /// - the operations *, +, -, ^
@@ -479,7 +479,7 @@ impl Cache {
 }
 
 /// A binary operation
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Op2 {
     Add,
     Mul,
@@ -498,7 +498,7 @@ impl Op2 {
 }
 
 /// The feature flags that can be used to enable or disable parts of constraints.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord, Hash)]
 #[cfg_attr(
     feature = "ocaml_types",
     derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::Enum)
@@ -605,7 +605,7 @@ impl ColTrait for Column {
 /// This represents a PLONK "custom constraint", which enforces that
 /// the corresponding combination of the polynomials corresponding to
 /// the above variables should vanish on the PLONK domain.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Expr<C, Col: ColTrait = Column> {
     Constant(C),
     Cell(Variable<Col>),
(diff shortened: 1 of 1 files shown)


kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+            (exp, new_constraints)
+        }
+        fn pow_to_mul(self, p: u64) -> Self {
+            let e = Box::new(self);
+            let mul = |a, b| Expr::BinOp(Op2::Mul, a, b);
+            let e_2 = Box::new(Expr::Square(e.clone()));
+            let new_exp = match p {
+                2 => *e_2,
+                3 => mul(e, e_2),
+                4..=8 => {
```
querolita, line 3281: This looks cool

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+        ) -> Expr<C, ColumnExtended<Col>> {
+            let trivial_map = |c| ColumnExtended::Inner(c);
+            let degree = self.degree(4) / 4;
+            if degree <= 2 {
+                return self.map_column(&trivial_map);
+            }
+
+            type E<C, Col> = Expr<C, ColumnExtended<Col>>;
+            let e = match self {
+                e @ Expr::VanishesOnLast4Rows
```
querolita, line 3308: Be careful, in master this has a different name

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+                | e @ Expr::UnnormalizedLagrangeBasis(_)
+                | e @ Expr::Constant(_)
+                | e @ Expr::Cell(_) => e.map_column(&trivial_map),
+
+                Expr::Double(e) => E::<C, Col>::Double(Box::new(e.reduce_degree_to_2_rec(rec))),
+                Expr::Square(e) => E::<C, Col>::Square(Box::new(e.reduce_degree_to_1(rec))),
```
querolita, line 3314: Why are you reducing degree to 1? How does this even look like?

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+                        let e2 = Box::new(e2.reduce_degree_to_2_rec(rec));
+                        E::<C, Col>::BinOp(op, e1, e2)
+                    }
+                    Op2::Mul => {
+                        let e1 = Box::new(e1.reduce_degree_to_1(rec));
+                        let e2 = Box::new(e2.reduce_degree_to_1(rec));
+                        E::<C, Col>::BinOp(Op2::Mul, e1, e2)
+                    }
+                },
+                //only supporting 1..=8 for simplicity, but arbitrary exponent could be
```
querolita, line 3327: You mean because in Kimchi we didn't support constraints which are higher than degree 8, right?

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+            };
+            e
+        }
+
+        fn reduce_degree_to_1(
+            self,
+            rec: &mut ExprRecorder<C, Col>,
+        ) -> Expr<C, ColumnExtended<Col>> {
+            let trivial_map = |c| ColumnExtended::Inner(c);
+            let degree = self.degree(4) / 4;
```
querolita, line 3355: What's the meaning of this line?

querolita (commented):
Left a few comments to understand the scope better

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+        hash::Hash,
+        ops::Neg,
+    };
+
+    ///extension of columns with necessary extra columns
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+    pub enum ColumnExtended<S: ColTrait + Eq + Hash> {
+        Inner(S),
+        ///extra column used to reduce degree of expressions
+        Extended(usize),
```
volhovm, line 3169: Would it make sense to document that `usize` means scaling coefficient (if I understand correctly?).

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+        ) -> Option<crate::proof::PointEvaluations<F>> {
+            match self {
+                ColumnExtended::Inner(s) => s.evaluate(evals),
+                ///TODO: a change in evals will be required so that it can also provide access to the extended columns
+                ColumnExtended::Extended(e) => todo!(),
+            }
+        }
+
+        fn make_witness(i: usize) -> Self {
+            Self::Inner(S::make_witness(i))
```
volhovm, line 3205: So we never allocate things in this new column?

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+    }
+
+    impl<C: Hash + Eq, Col: ColTrait + Hash + Eq> ExprRecorder<C, Col> {
+        fn new() -> Self {
+            Self {
+                recorded_exprs: Default::default(),
+                next: 0,
+            }
+        }
+        fn get_id(&mut self, e: Expr<C, ColumnExtended<Col>>) -> usize {
```
volhovm, line 3229: If I understand the semantics of `ExprRecorder`, it's basically accumulating expressions, to later sort them according to the insertion order. So it's like an ordered hashmap.

In this case, `get_id` is a bit misleading, I'd call it `insert` or `register` instead?

kimchi/src/circuits/expr.rs
volhovm: I guess `reduce_degree_to_1` will try to reduce the argument to `2` eventually, then put it into this extended row, and after that it's treated as a degree 1 expression. Similar to how `Op2::Mul` works?

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+                    Op2::Mul => {
+                        let e1 = Box::new(e1.reduce_degree_to_1(rec));
+                        let e2 = Box::new(e2.reduce_degree_to_1(rec));
+                        E::<C, Col>::BinOp(Op2::Mul, e1, e2)
+                    }
+                },
+                //only supporting 1..=8 for simplicity, but arbitrary exponent could be
+                //implemented, ideally an efficient fixed exponent algorithm
+                Expr::Pow(_, 0) => {
+                    panic!("no");
```
volhovm, line 3330: Why not return const 1? Is `Pow(_, 0)` generally not an allowed expression?

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+        fn reduce_degree_to_1(
+            self,
+            rec: &mut ExprRecorder<C, Col>,
+        ) -> Expr<C, ColumnExtended<Col>> {
+            let trivial_map = |c| ColumnExtended::Inner(c);
+            let degree = self.degree(4) / 4;
+            if degree <= 1 {
+                return self.map_column(&trivial_map);
+            }
+            let extended = |id| {
```
volhovm, line 3359: Is there a reason to define this as a lambda at this step? It doesn't seem to capture any external variables, so why not just inline it into the last row of `reduce_degree_to_1`?

kimchi/src/circuits/expr.rs
```diff
@@ -3149,3 +3149,246 @@ pub mod test {
...
+        ) -> (
+            Expr<C, ColumnExtended<Col>>,
+            Vec<Expr<C, ColumnExtended<Col>>>,
+        ) {
+            let mut rec = ExprRecorder::new();
+            let exp = self.reduce_degree_to_2_rec(&mut rec);
+            let new_constraints = rec.into_constraints();
+            (exp, new_constraints)
+        }
+        fn pow_to_mul(self, p: u64) -> Self {
```
volhovm, line 3274: Maybe generically would look better, like do a bit decomposition and then in the loop multiply the accumulator by the power of `e` if necessary. I think it also doesn't create more expressions than necessary.

dannywillems:
Closing as it seems to be in `kimchi/src/folding/quadraticization.rs`

PR #1257, opened 2023-09-27
[develop] Add rust to 1.71 and 1.72 to CI

dannywillems:


kimchi/src/tests/framework.rs
```diff
@@ -123,7 +123,7 @@ where
         let start = Instant::now();
 
         let lookup_tables = std::mem::take(&mut self.lookup_tables);
-        let runtime_tables_setup = mem::replace(&mut self.runtime_tables_setup, None);
+        let runtime_tables_setup = self.runtime_tables_setup.take();
```
fabrizio-m, line 126: was this requested by clippy?

fabrizio-m (approved):

kimchi/src/tests/framework.rs
dannywillems: Yes

PR #1261, opened 2023-10-02
Tests for zkvm Keccak witness code

querolita:
This PR includes a few tests whose goal is to check that the witness generation code works as expected.

It still does not contain tests for the gates. That will come when the proof system is compatible with this new configuration.


kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+    state_f.append(&mut tail);
+
+    println!("State F:");
+    print_matrix(&state_f);
+    println!("State G:");
+    print_matrix(&next[0..100]);
+}
+
+// Sets up test for a given message and desired input bytelength
+fn test_keccak<G: KimchiCurve>(message: BigUint) -> BigUint
```
dannywillems, line 114: It does not "test" anything. I would rename it in `setup`

kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+    BigUint::from_bytes_be(&hash)
+}
+
+#[test]
+fn test_bitwise_sparse_representation() {
+    assert_eq!(expand(0xFFFF), 0x1111111111111111);
+
+    let word_a: u64 = 0x70d324ac9215fd8e;
+    let dense_a = decompose(word_a);
+    let real_dense_a = [0xfd8e, 0x9215, 0x24ac, 0x70d3];
```
dannywillems, line 152: Nit: I tend to use `expected` instead of `real`.

kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+        hash.push(FieldHelpers::to_bytes(&witness[200 + b][hash_row])[0]);
+        print!("{:02x}", hash[b]);
+    }
+    println!();
+    println!();
+
+    BigUint::from_bytes_be(&hash)
+}
+
+#[test]
```
dannywillems, line 146: I think there are also some property based testing we could apply.
We could generate random inputs and verify the properties are respected.

kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+                + shifts_sum_ab[4 + i] * 2
+                + shifts_sum_ab[8 + i] * 4
+                + shifts_sum_ab[12 + i] * 8
+        )
+    }
+}
+
+#[test]
+// Tests a random block of 1080 bits
+fn test_random_block() {
```
dannywillems, line 212: This seems to be a test vector. Do you have a source we can check with?

kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+        BigUint::from_hex("832588523900cca2ea9b8c0395d295aa39f9a9285a982b71cc8475067a8175f38f235a2234abc982a2dfaaddff2895a28598021895206a733a22bccd21f124df1413858a8f9a1134df285a888b099a8c2235eecdf2345f3afd32f3ae323526689172850672938104892357aad32523523f423423a214325d13523aadb21414124aaadf32523126"),
+    );
+    let hash_random =
+        BigUint::from_hex("845e9dd4e22b4917a80c5419a0ddb3eebf5f4f7cc6035d827314a18b718f751f");
+    assert_eq!(claim_random, hash_random);
+}
+
+#[test]
+// Test hash of message zero with 1 byte
+fn test_dummy() {
```
dannywillems, line 223: Is it really "dummy"? Poor message `0` :(.

kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+    let claim1 = test_keccak::<Pallas>(BigUint::from_bytes_be(&[0x00]));
+    let hash1 =
+        BigUint::from_hex("bc36789e7a1e281436464229828f817d6612f7b477d66591ff96a9e064bcc98a");
+    assert_eq!(claim1, hash1);
+}
+
+#[test]
+// Test hash of message zero with 1 byte
```
dannywillems, line 231: Not sure the comment is correct.

dannywillems (approved):
We can merge this and add a follow-up for my comments below. We can therefore merge step by step each PR.

kimchi/src/tests/keccak.rs
```diff
@@ -1,54 +1,237 @@
...
+    let padded_len = {
+        let mut sized = message.to_bytes_be();
+        sized.resize(bytelength - sized.len(), 0);
+        pad(&sized).len()
+    };
+    let _index = create_test_constraint_system::<G>(padded_len);
+    let witness = create_keccak_witness::<G>(message);
+
+    for r in 1..=24 {
+        print_witness::<G::ScalarField>(&witness, r);
```
dannywillems, line 128: For a test, I would avoid writing to the stdout.

kimchi/src/tests/keccak.rs
querolita: It is a random block of 1080 that I generated offline so I could hardcode the expected hash. You can check the value with this online tool for example: https://leventozturk.com/engineering/sha3/

If instead we want to be able to check the expected hash dynamically, we should make use of a library exposing an implementation of keccak.

PR #1263, opened 2023-10-02
Lookups: Improve assertions and tests regarding empty rows and table duplicates

volhovm:
- Introduces test for lookup table creation function here:
https://github.com/o1-labs/proof-systems/pull/1263/files#diff-c712332ec88dcf26c81cb945fa135596e8816276fbec24228af3922546cc5443R384
- Fix the bug in checking the existence of the zero row (the check was not actually done properly)
- Move a check on inconsistent table lengths into the table implementation itself.
- Add a check on duplicate table IDs and fix several tests that were mistakenly double-adding the tables.

This resolves MinaProtocol/mina#14070 and MinaProtocol/mina#14097.


volhovm:
@mimoo Question: `test_zero_table_zero_row` will return `true` if _at least one_ element of the row is zero. While the description of the function implies it should return `true` when _all the elements of the row_ are zero. Is this a bug or a documentation issue?

https://github.com/o1-labs/proof-systems/blob/develop/kimchi/src/circuits/lookup/tables/mod.rs#L35

(You were in `git blame` for this line, feel free to redirect me to someone else if you're not the author)

kimchi/src/circuits/lookup/index.rs
```diff
@@ -217,41 +215,63 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
...
-                let fixed_lookup_tables_ids: Vec<i32> = lookup_tables.iter().map(|lt| lt.id()).collect();
-                check_id_duplicates(fixed_lookup_tables_ids.iter(),"fixed lookup table duplicates")?;
+                // If there is a gate using a lookup table, this table must not be added
+                // explicitly to the constraint system.
+                let fixed_gate_joint_ids: Vec<i32> = fixed_lookup_tables
+                    .iter()
+                    .map(|lt| lt.id())
+                    .chain(gate_lookup_tables.iter().map(|lt| lt.id()))
+                    .collect();
+                check_id_duplicates(
```
dannywillems, line 245: Let's have a test checking that the error is raised.

kimchi/src/circuits/lookup/index.rs
```diff
@@ -217,41 +215,63 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
...
+
+                //~ 3. Concatenate explicit runtime lookup tables with the ones (implicitly) used by gates.
+                let mut lookup_tables: Vec<LookupTable<_>> = fixed_lookup_tables
+                    .into_iter()
+                    .chain(gate_lookup_tables)
+                    .collect();
+
+                let fixed_lookup_tables_ids: Vec<i32> =
+                    lookup_tables.iter().map(|lt| lt.id()).collect();
+                check_id_duplicates(
```
dannywillems, line 258: Same here

kimchi/src/prover_index.rs
```diff
@@ -211,7 +211,9 @@ pub mod testing {
             override_srs_size,
             |d1: D<G::ScalarField>, size: usize| {
                 let log2_size = size.ilog2();
+                // Run test_srs_serialization test to generate test SRS & enable this
                 let mut srs = if log2_size <= precomputed_srs::SERIALIZED_SRS_SIZE {
+                    // @volhovm: on my machine for d=16 (2.2Mb SRS file) this takes 110s. Why???
```
dannywillems, line 216: As discussed, it is because of the missing `--release` flag.

kimchi/src/circuits/lookup/index.rs
```diff
@@ -411,6 +433,8 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
                     lookup_table8.push(eval);
                 }
 
+                // @volhovm: Do we need to enforce that there is at least one table
```
mrmr1993, line 436: We should actually ensure that *either* we have a table with ID 0 *or* that there was space for us to insert a 0 value.

kimchi/src/circuits/lookup/lookups.rs
```diff
@@ -204,7 +204,7 @@ impl LookupInfo {
         &self,
         domain: &EvaluationDomains<F>,
         gates: &[CircuitGate<F>],
-    ) -> (LookupSelectors<Evaluations<F>>, Vec<LookupTable<F>>) {
+    ) -> (LookupSelectors<Evaluations<F>>, HashSet<LookupTable<F>>) {
```
mrmr1993, line 207: `HashSet` ordering is non-deterministic (and we probably shouldn't be hashing a whole lookup table). Revert.

kimchi/src/circuits/lookup/tables/mod.rs
```diff
@@ -1,45 +1,64 @@
...
-/// Enumerates the different 'fixed' lookup tables used by individual gates
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub enum GateLookupTable {
-    Xor,
-    RangeCheck,
-}
-
 /// A table of values that can be used for a lookup, along with the ID for the table.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
mrmr1993, line 10: Don't hash/check equality on a `Vec<Vec<F>>`. It could be large!

kimchi/src/prover_index.rs
mrmr1993: We should document this somewhere.

kimchi/src/tests/lookup.rs
```diff
@@ -200,7 +200,7 @@ fn test_runtime_table() {
     let len = first_column.len();
 
     let mut runtime_tables_setup = vec![];
-    for table_id in 0..num {
+    for table_id in 1..num + 1 {
```
mrmr1993, line 203: If we don't have a `table_id` of 0, this should fail :thinking:

kimchi/src/circuits/lookup/lookups.rs
volhovm: Hashset order being nondeterministic is true, but the problem is deeper then: we build tables right now on the result of this hashset converted to a vector: we return `gate_tables.iter().convert()`, where `gate_tables` is a hashset. What I did instead is to return it directly without converting to vec, because it might be more convenient to check for duplicates.

The question is: how do you want tables sorted between each other, and is there any difference at all?

kimchi/src/circuits/lookup/tables/mod.rs
volhovm: It is absolutely true and I realise it, but otherwise we can't really check if there are any truly duplicate tables in our setup. In that case we would only be able to check table id collisions. I guess it's fine.

kimchi/src/prover_index.rs
volhovm: @mrmr1993 @dannywillems Adding the following into the top-lvl `Cargo.toml` fixes the issue:
```
[profile.test]
opt-level = 1 # Otherwise SRS deserialization in tests takes forever
```
and I don't think it's terribly unreasonable, unless we really want our non-release test build to be fully unoptimised?

kimchi/src/circuits/lookup/index.rs
volhovm: "space to insert zero value" is checking that the total number of lookup table rows is one less than `d1` domain minus zk rows?

kimchi/src/prover_index.rs
dannywillems: > We should document this somewhere.

https://github.com/o1-labs/proof-systems/blob/develop/CONTRIBUTING.md#development

kimchi/src/circuits/lookup/tables/mod.rs
```diff
@@ -1,45 +1,64 @@
-use ark_ff::{FftField, One, Zero};
+use ark_ff::{Field, One, Zero};
 use poly_commitment::PolyComm;
 use serde::{Deserialize, Serialize};
+use thiserror::Error;
 
 pub mod range_check;
 pub mod xor;
 
-//~ spec:startcode
```
dannywillems, line 8: Why is it moved down? If no reason, it is adding noise to the PR.

kimchi/src/circuits/lookup/tables/mod.rs
```diff
@@ -54,19 +73,47 @@ where
...
+/// Enumerates the different 'fixed' lookup tables used by individual gates
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum GateLookupTable {
+    Xor,
+    RangeCheck,
 }
 
 /// Returns the lookup table associated to a [`GateLookupTable`].
-pub fn get_table<F: FftField>(table_name: GateLookupTable) -> LookupTable<F> {
+pub fn get_table<F: Field>(table_name: GateLookupTable) -> LookupTable<F> {
```
dannywillems, line 116: Why?

kimchi/src/circuits/constraints.rs
```diff
@@ -688,15 +688,13 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
...
-                        } else {
-                            data[0].len()
-                        }
-                    },
-                )
+                .map(|lt| {
+                    if lt.data().is_empty() {
+                        0
+                    } else {
+                        lt.data()[0].len()
```
dannywillems, line 695: This is the width. There is a method for this.

kimchi/src/circuits/lookup/index.rs
```diff
@@ -217,21 +215,64 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
                 // product is 1, we cannot use those rows to store any values.
                 let max_num_entries = d1_size - zk_rows - 1;
 
-                //~ 2. Get the lookup selectors and lookup tables (TODO: how?)
+                //~ 2. Get the lookup selectors and lookup tables that are specified implicitly
+                // by the lookup gates (TODO: how?)
```
dannywillems, line 219: Remove TODO?

kimchi/src/circuits/lookup/index.rs
```diff
@@ -217,21 +215,64 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
...
                 let (lookup_selectors, gate_lookup_tables) =
                     lookup_info.selector_polynomials_and_tables(domain, gates);
 
-                //~ 3. Concatenate runtime lookup tables with the ones used by gates
-                let mut lookup_tables: Vec<_> = gate_lookup_tables
+                // Checks whether an iterator contains any duplicates, and if yes, raises
+                // a corresponding LookupTableIdCollision error.
+                fn check_id_duplicates<'a, I: Iterator<Item = &'a i32>>(
```
dannywillems, line 225: Nit: as it is strictly relatd to tables, you can move this in `tables/mod.rs`

kimchi/src/circuits/lookup/tables/mod.rs
volhovm: It makes more sense in terms of file structure:
1. First we define `LookupTable`
2. Then there's code on fixed tables (xor & range). Note that in the original this code is split in two, `gate_table` and `impl GateLookupTable` were somewhat astray from everything else.
3. Then there's generic code on merging tables.

Ideally fixed lookup tables should be in a separate module, but I thought it might create too many conflicts with the customizable kimchi effort, where it'll be probably done later anyway.

kimchi/src/prover_index.rs
volhovm: I put a comment with a note and the link to the `get_srs` function.

kimchi/src/circuits/constraints.rs
```diff
@@ -688,15 +688,13 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
...
-                    |LookupTable { data, id: _ }| {
-                        if data.is_empty() {
-                            0
-                        } else {
-                            data[0].len()
-                        }
-                    },
-                )
+                .map(|lt| {
+                    if lt.data().is_empty() {
```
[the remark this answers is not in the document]
volhovm: Addressed.

kimchi/src/circuits/constraints.rs
volhovm: Addressed.

kimchi/src/circuits/lookup/index.rs
volhovm: Not mine, but done :)

kimchi/src/circuits/lookup/index.rs
volhovm: Since we don't use this function anywhere except `create` and it's somewhat tailored for it (works with `LookupError` that is defined in `index.rs`), I'd say it's an OK place for it? Otherwise it'd complicate dependencies and would be not obvious where it's used/why it's defined publicly.

dannywillems:
Changes have to be done in https://github.com/o1-labs/o1js-bindings/tree/main/kimchi/wasm.
```
error[E0451]: field `id` of struct `kimchi::circuits::lookup::tables::LookupTable` is private
  --> src/pasta_fp_plonk_index.rs:47:13
   |
47 |             id: wasm_lt.id.into(),
   |             ^^^^^^^^^^^^^^^^^^^^^ private field

error[E0451]: field `data` of struct `kimchi::circuits::lookup::tables::LookupTable` is private
  --> src/pasta_fp_plonk_index.rs:48:13
   |
48 |             data: wasm_lt.data.0,
   |             ^^^^^^^^^^^^^^^^^^^^ private field

error[E0451]: field `id` of struct `kimchi::circuits::lookup::tables::LookupTable` is private
  --> src/pasta_fq_plonk_index.rs:44:13
   |
44 |             id: wasm_lt.id.into(),
   |             ^^^^^^^^^^^^^^^^^^^^^ private field

error[E0451]: field `data` of struct `kimchi::circuits::lookup::tables::LookupTable` is private
  --> src/pasta_fq_plonk_index.rs:45:13
   |
45 |             data: wasm_lt.data.0,
   |             ^^^^^^^^^^^^^^^^^^^^ private field

```

dannywillems:
Please git merge develop and check again against mina.

dannywillems (approved):

kimchi/src/circuits/lookup/index.rs
```diff
@@ -217,21 +215,64 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
...
                 let (runtime_table_offset, runtime_selector) =
                     if let Some(runtime_tables) = &runtime_tables {
+                        let runtime_tables_ids: Vec<i32> =
+                            runtime_tables.iter().map(|rt| rt.id).collect();
+                        check_id_duplicates(runtime_tables_ids.iter(), "runtime table duplicates")?;
+                        check_id_duplicates(
+                            runtime_tables_ids
+                                .iter()
+                                .chain(fixed_lookup_tables_ids.iter()),
+                            "duplicates between runtime and fixed tables",
```
mrmr1993, line 273: We explicitly don't want this. This should never have landed without backing this out.

PR #1270, opened 2023-10-04
Add feature flag for rust WASM nightly build

mrmr1993:


dannywillems:
I think it is fixed by the update to a newer nightly version. I encountered it when rebasing https://github.com/MinaProtocol/mina/pull/14225 on develop after merging some changes to chunking. It disappeared after bumping up the version of rust nightly to nightly-2023-09-01

dannywillems:
I would suggest to close this one @mrmr1993

PR #1273, opened 2023-10-09
Remove noisy `debug_sponge`

mrmr1993:
This PR removes the `debug_sponge` feature. This currently makes the test output extremely noisy, and the OCaml counterpart that would have made this useful was never implemented.


dannywillems:
I remember there is an id in the sponge implementation in Pickles, which was used for debugging only also. You might want to remove it there too?

dannywillems (approved):

PR #1276, opened 2023-10-10
Improve errors around domain creation

mrmr1993:
This PR tweaks the errors returned around domain creation so that they have informative content.

This comes from the debugging for chunking, where the current messages are effectively useless.


rbonichon (approved):
Nice changes that improve readability!

PR #1277, opened 2023-10-10
Fixup edge-case around chunking in constraint system

mrmr1993:
This PR fixes an error where we calculate `0` chunks (and then get an underflow, with predictably disastrous results).


dannywillems (approved):
It does also suppose that the values `domain_size` and `max_poly_size` are powers of two.

rbonichon (approved):
I know this one :-D

PR #1284, opened 2023-10-17
Merge back to master

mrmr1993:


querolita:
I am observing that this does not bring the changes about runtime tables to master (and zk rows, etc). Is this meant to be like this so that master remains being compatible with Mina?

dannywillems (approved):

PR #1285, opened 2023-10-18
Folding

fabrizio-m:
Adds support for folding.

### DIFF
--- a/kimchi/src/folding/error.rs
+++ b/kimchi/src/folding/error.rs
@@ -68,9 +68,9 @@ pub(crate) fn eval_exp_error<'a, F: Field, E: FoldingEnv<F>>(
     }
 }
 
-pub(crate) fn compute_error<'a, F: Field, E: FoldingEnv<F>>(
+pub(crate) fn compute_error<F: Field, E: FoldingEnv<F>>(
     exp: &IntegratedFoldingExpr<F>,
-    env: &'a E,
+    env: &E,
     u: (F, F),
 ) -> Vec<F> {
     let add = |a, b| EvalLeaf::bin_op(a, b, |a, b| *a + b, |a, b| *a += b);
--- a/kimchi/src/folding/error.rs
+++ b/kimchi/src/folding/error.rs
@@ -1,0 +1,126 @@
+use super::expressions::{FoldingExp, IntegratedFoldingExpr};
+use super::FoldingEnv;
+use crate::folding::expressions::{Degree, FoldingColumn};
+use crate::folding::EvalLeaf;
+use ark_ff::Field;
+
+pub(crate) fn eval_exp_error<'a, F: Field, E: FoldingEnv<F>>(
+    exp: &FoldingExp<F>,
+    env: &'a E,
+    side: bool,
+) -> EvalLeaf<'a, F> {
+    let degree = exp.degree();
+    use EvalLeaf::*;
+    use FoldingColumn::*;
+    use FoldingExp::*;
+
+    match exp {
+        Constant(c) => EvalLeaf::Const(*c),
+        Cell(col) => match col {
+            Witness(i) => Col(env.witness(*i, side)),
+            WitnessExtended(i) => Col(env.witness_ext(*i, side)),
+            Index(i) => Col(env.index(i)),
+            Coefficient(i) => Col(env.coefficient(*i)),
+            //TODO: maybe specialize
+            Shift => Col(env.shift()),
+        },
+        Double(e) => {
+            let col = eval_exp_error(e, env, side);
+            col.map(Field::double, |f| {
+                Field::double_in_place(f);
+            })
+        }
+        FoldingExp::Square(e) => {
+            let col = eval_exp_error(e, env, side);
+            col.map(Field::square, |f| {
+                Field::square_in_place(f);
+            })
+        }
+        FoldingExp::Add(e1, e2) => {
+            let a = eval_exp_error(e1, env, side);
+            let b = eval_exp_error(e2, env, side);
+            EvalLeaf::bin_op(a, b, |a, b| *a + b, |a, b| *a += b)
+        }
+        FoldingExp::Sub(e1, e2) => {
+            let a = eval_exp_error(e1, env, side);
+            let b = eval_exp_error(e2, env, side);
+            EvalLeaf::bin_op(a, b, |a, b| *a - b, |a, b| *a -= b)
+        }
+        FoldingExp::Mul(e1, e2) => match (degree, e1.degree()) {
+            (Degree::Two, Degree::One) => {
+                let a = eval_exp_error(e1, env, side);
+                let b = eval_exp_error(e2, env, !side);
+                let first = EvalLeaf::bin_op(a, b, |a, b| *a * b, |a, b| *a *= b);
+                let a = eval_exp_error(e1, env, !side);
+                let b = eval_exp_error(e2, env, side);
+                let second = EvalLeaf::bin_op(a, b, |a, b| *a * b, |a, b| *a *= b);
+                EvalLeaf::bin_op(first, second, |a, b| *a + b, |a, b| *a += b)
+            }
+            _ => {
+                let a = eval_exp_error(e1, env, side);
+                let b = eval_exp_error(e2, env, side);
+                EvalLeaf::bin_op(a, b, |a, b| *a * b, |a, b| *a *= b)
+            }
+        },
+        FoldingExp::UnnormalizedLagrangeBasis(i) => EvalLeaf::Result(env.lagrange_basis(i)),
+        //TODO: use cache
+        FoldingExp::Cache(_id, exp) => eval_exp_error(exp, env, side),
+    }
+}
+
+pub(crate) fn compute_error<'a, F: Field, E: FoldingEnv<F>>(
+    exp: &IntegratedFoldingExpr<F>,
+    env: &'a E,
+    u: (F, F),
+) -> Vec<F> {
+    let add = |a, b| EvalLeaf::bin_op(a, b, |a, b| *a + b, |a, b| *a += b);
+    let sub = |a, b| EvalLeaf::bin_op(a, b, |a, b| *a - b, |a, b| *a -= b);
+    let scale = |t, s| EvalLeaf::bin_op(t, EvalLeaf::Const(s), |a, b| *a * b, |a, b| *a *= b);
+
+    let t_0 = EvalLeaf::Result(env.zero_vec());
+    let t_0 = exp.degree_0.iter().fold(t_0, |t_0, (exp, sign)| {
+        //could be true or false, doesn't matter for constant terms
+        let e = eval_exp_error(exp, env, true);
+        if *sign {
+            add(t_0, e)
+        } else {
+            sub(t_0, e)
+        }
+    });
+    let t_0 = scale(t_0, (u.0 * u.1).double());
+
+    let t_1l = EvalLeaf::Result(env.zero_vec());
+    let t_1r = EvalLeaf::Result(env.zero_vec());
+    let (t_1l, t_1r) = exp
+        .degree_0
+        .iter()
+        .fold((t_1l, t_1r), |(t_1l, t_1r), (exp, sign)| {
+            let el = eval_exp_error(exp, env, true);
+            let er = eval_exp_error(exp, env, false);
+            if *sign {
+                (add(t_1l, el), add(t_1r, er))
+            } else {
+                (sub(t_1l, el), sub(t_1r, er))
+            }
+        });
+    let t_1l = scale(t_1l, u.1);
+    let t_1r = scale(t_1r, u.0);
+    let t_1 = EvalLeaf::bin_op(t_1l, t_1r, |a, b| *a + b, |a, b| *a += b);
+
+    let t_2 = EvalLeaf::Result(env.zero_vec());
+    let t_2 = exp.degree_0.iter().fold(t_2, |t_2, (exp, sign)| {
+        //true or false matter in some way, but not at the top level call
+        let e = eval_exp_error(exp, env, true);
+        if *sign {
+            add(t_2, e)
+        } else {
+            sub(t_2, e)
+        }
+    });
+    let t = add(add(t_0, t_1), t_2);
+
+    match t {
+        EvalLeaf::Result(res) => res,
+        _ => unreachable!(),
+    }
+}
--- a/kimchi/src/folding/error.rs
+++ b/kimchi/src/folding/error.rs
@@ -4,10 +4,25 @@ use crate::folding::expressions::{Degree, FoldingColumn};
 use crate::folding::EvalLeaf;
 use ark_ff::Field;
 
+#[derive(Clone, Copy)]
+pub enum Side {
+    Left,
+    Right,
+}
+impl std::ops::Not for Side {
+    type Output = Self;
+
+    fn not(self) -> Self::Output {
+        match self {
+            Side::Left => Side::Right,
+            Side::Right => Side::Left,
+        }
+    }
+}
 pub(crate) fn eval_exp_error<'a, F: Field, E: FoldingEnv<F>>(
     exp: &FoldingExp<F>,
     env: &'a E,
-    side: bool,
+    side: Side,
 ) -> EvalLeaf<'a, F> {
     let degree = exp.degree();
     use EvalLeaf::*;
@@ -80,7 +95,7 @@ pub(crate) fn compute_error<F: Field, E: FoldingEnv<F>>(
     let t_0 = EvalLeaf::Result(env.zero_vec());
     let t_0 = exp.degree_0.iter().fold(t_0, |t_0, (exp, sign)| {
         //could be true or false, doesn't matter for constant terms
-        let e = eval_exp_error(exp, env, true);
+        let e = eval_exp_error(exp, env, Side::Left);
         if *sign {
             add(t_0, e)
         } else {
@@ -95,8 +110,8 @@ pub(crate) fn compute_error<F: Field, E: FoldingEnv<F>>(
         .degree_0
         .iter()
         .fold((t_1l, t_1r), |(t_1l, t_1r), (exp, sign)| {
-            let el = eval_exp_error(exp, env, true);
-            let er = eval_exp_error(exp, env, false);
+            let el = eval_exp_error(exp, env, Side::Left);
+            let er = eval_exp_error(exp, env, Side::Right);
             if *sign {
                 (add(t_1l, el), add(t_1r, er))
             } else {
@@ -110,7 +125,7 @@ pub(crate) fn compute_error<F: Field, E: FoldingEnv<F>>(
     let t_2 = EvalLeaf::Result(env.zero_vec());
     let t_2 = exp.degree_0.iter().fold(t_2, |t_2, (exp, sign)| {
         //true or false matter in some way, but not at the top level call
-        let e = eval_exp_error(exp, env, true);
+        let e = eval_exp_error(exp, env, Side::Left);
         if *sign {
             add(t_2, e)
         } else {
--- a/kimchi/src/folding/expressions.rs
+++ b/kimchi/src/folding/expressions.rs
@@ -123,7 +123,6 @@ impl<C> Default for IntegratedFoldingExpr<C> {
         }
     }
 }
-// fn extract_terms<C: Clone, F>(exp: FoldingExp<C>) -> Vec<Term<C>>
 pub fn extract_terms<C: Clone, F>(exp: FoldingExp<C>, terms: &mut Vec<Term<C>>, f: &F)
 where
     F: Fn(Term<C>) -> Term<C>,
--- a/kimchi/src/folding/expressions.rs
+++ b/kimchi/src/folding/expressions.rs
@@ -1,0 +1,301 @@
+use crate::circuits::{
+    expr::{Column, Expr, FeatureFlag, Op2, Variable},
+    gate::CurrOrNext,
+};
+
+use super::{CacheId, GateType};
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum FoldingColumn {
+    Witness(usize),
+    ///for the extra columns added by quadricization
+    #[allow(dead_code)]
+    WitnessExtended(usize),
+    Index(GateType),
+    Coefficient(usize),
+    ///basically X, to allow accesing the next row
+    Shift,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+///A simplified expression to be use in folding
+pub enum FoldingExp<C> {
+    Constant(C),
+    Cell(FoldingColumn),
+    Double(Box<FoldingExp<C>>),
+    Square(Box<FoldingExp<C>>),
+    Add(Box<FoldingExp<C>>, Box<FoldingExp<C>>),
+    Sub(Box<FoldingExp<C>>, Box<FoldingExp<C>>),
+    Mul(Box<FoldingExp<C>>, Box<FoldingExp<C>>),
+    UnnormalizedLagrangeBasis(i32),
+    Cache(CacheId, Box<FoldingExp<C>>),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Degree {
+    Zero,
+    One,
+    Two,
+}
+impl<C> FoldingExp<C> {
+    pub(super) fn degree(&self) -> Degree {
+        use Degree::*;
+        match self {
+            FoldingExp::Constant(_) => Zero,
+            FoldingExp::Cell(col) => match col {
+                FoldingColumn::Witness(_) | FoldingColumn::WitnessExtended(_) => One,
+                _ => Zero,
+            },
+            FoldingExp::Double(e) => e.degree(),
+            FoldingExp::Square(e) => &e.degree() * &e.degree(),
+            FoldingExp::Add(e1, e2) | FoldingExp::Sub(e1, e2) | FoldingExp::Mul(e1, e2) => {
+                e1.degree() + e2.degree()
+            }
+            FoldingExp::UnnormalizedLagrangeBasis(_) => Zero,
+            FoldingExp::Cache(_, e) => e.degree(),
+        }
+    }
+}
+impl std::ops::Add for Degree {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        use Degree::*;
+        match (self, rhs) {
+            (Zero, Zero) => Zero,
+            (Zero, One) | (One, Zero) | (One, One) => One,
+            (_, Two) | (Two, _) => Two,
+        }
+    }
+}
+impl std::ops::Mul for &Degree {
+    type Output = Degree;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        use Degree::*;
+        match (self, rhs) {
+            (Zero, Zero) => Zero,
+            (Zero, One) | (One, Zero) => One,
+            (Zero, Two) | (One, One) | (Two, Zero) => Two,
+            (One, Two) | (Two, One) | (Two, Two) => panic!("degree over 2"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Term<C> {
+    pub exp: FoldingExp<C>,
+    pub sign: bool,
+    pub degree: Degree,
+}
+
+impl<C> Term<C> {
+    fn new(exp: FoldingExp<C>, degree: Degree) -> Self {
+        Self {
+            exp,
+            sign: true,
+            degree,
+        }
+    }
+}
+impl<C> std::ops::Neg for Term<C> {
+    type Output = Self;
+
+    fn neg(self) -> Self::Output {
+        Term {
+            sign: !self.sign,
+            ..self
+        }
+    }
+}
+///A simplified expression with all terms separated by degree
+pub struct IntegratedFoldingExpr<C> {
+    pub(super) degree_0: Vec<(FoldingExp<C>, bool)>,
+    pub(super) degree_1: Vec<(FoldingExp<C>, bool)>,
+    pub(super) degree_2: Vec<(FoldingExp<C>, bool)>,
+}
+
+impl<C> Default for IntegratedFoldingExpr<C> {
+    fn default() -> Self {
+        Self {
+            degree_0: vec![],
+            degree_1: vec![],
+            degree_2: vec![],
+        }
+    }
+}
+// fn extract_terms<C: Clone, F>(exp: FoldingExp<C>) -> Vec<Term<C>>
+pub fn extract_terms<C: Clone, F>(exp: FoldingExp<C>, terms: &mut Vec<Term<C>>, f: &F)
+where
+    F: Fn(Term<C>) -> Term<C>,
+{
+    use Degree::*;
+    match exp {
+        e @ FoldingExp::Constant(_) => {
+            let term = f(Term::new(e, Zero));
+            terms.push(term)
+        }
+        FoldingExp::Cell(col) => {
+            let degree = match col {
+                FoldingColumn::Witness(_) | FoldingColumn::WitnessExtended(_) => One,
+                _ => Zero,
+            };
+            let exp = FoldingExp::Cell(col);
+            let term = f(Term::new(exp, degree));
+            terms.push(term)
+        }
+        FoldingExp::Double(e) => {
+            let f = |t| {
+                let Term { exp, sign, degree } = t;
+                let exp = FoldingExp::Double(Box::new(exp));
+                f(Term { exp, sign, degree })
+            };
+            let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+            extract_terms(*e, terms, &f);
+        }
+        FoldingExp::Square(e) => {
+            let mut temp = Vec::with_capacity(4);
+            extract_terms(*e.clone(), &mut temp, f);
+            for term in temp {
+                let Term { exp, sign, degree } = term;
+                let exp = Box::new(exp);
+                let f = |t| {
+                    let Term {
+                        exp: exp2,
+                        sign: sign2,
+                        degree: degree2,
+                    } = t;
+                    let degree = &degree * &degree2;
+                    let sign = sign != sign2;
+                    //TODO: specialize square case
+                    let exp = FoldingExp::Mul(exp.clone(), Box::new(exp2));
+                    Term { exp, sign, degree }
+                };
+                let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+                extract_terms(*e.clone(), terms, &f);
+            }
+        }
+        FoldingExp::Add(e1, e2) => {
+            extract_terms(*e1, terms, f);
+            extract_terms(*e2, terms, f);
+        }
+        FoldingExp::Sub(e1, e2) => {
+            extract_terms(*e1, terms, f);
+            let f = |t: Term<C>| f(-t);
+            let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+            extract_terms(*e2, terms, &f);
+        }
+        FoldingExp::Mul(e1, e2) => {
+            let mut temp = Vec::with_capacity(4);
+            extract_terms(*e1, &mut temp, f);
+            for term in temp {
+                let Term { exp, sign, degree } = term;
+                let exp = Box::new(exp);
+                let f = |t| {
+                    let Term {
+                        exp: exp2,
+                        sign: sign2,
+                        degree: degree2,
+                    } = t;
+                    let degree = &degree * &degree2;
+                    let sign = sign == sign2;
+                    let exp = FoldingExp::Mul(exp.clone(), Box::new(exp2));
+                    Term { exp, sign, degree }
+                };
+                let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+                extract_terms(*e2.clone(), terms, &f);
+            }
+        }
+        e @ FoldingExp::UnnormalizedLagrangeBasis(_) => {
+            let term = f(Term::new(e, Zero));
+            terms.push(term)
+        }
+        FoldingExp::Cache(id, e) => {
+            let f = |t| {
+                let Term { exp, sign, degree } = t;
+                let exp = FoldingExp::Cache(id, Box::new(exp));
+                Term { exp, sign, degree }
+            };
+            let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+            extract_terms(*e, terms, &f);
+        }
+    }
+}
+
+///simplyfies the expression, and separates it into terms of different degree y
+pub fn folding_expression<C: Copy, F: Fn(&FeatureFlag) -> bool>(
+    exp: &Expr<C>,
+    flag_resolver: &F,
+) -> IntegratedFoldingExpr<C> {
+    let simplied = simplify_expression(exp, flag_resolver);
+    let mut terms = vec![];
+    extract_terms(simplied, &mut terms, &|t| t);
+    let mut integrated = IntegratedFoldingExpr::default();
+    for term in terms.into_iter() {
+        let Term { exp, sign, degree } = term;
+        let t = (exp, sign);
+        match degree {
+            Degree::Zero => integrated.degree_0.push(t),
+            Degree::One => integrated.degree_1.push(t),
+            Degree::Two => integrated.degree_2.push(t),
+        }
+    }
+    integrated
+}
+fn simplify_expression<C: Copy, F: Fn(&FeatureFlag) -> bool>(
+    exp: &Expr<C>,
+    flag_resolver: &F,
+) -> FoldingExp<C> {
+    match exp {
+        Expr::Constant(c) => FoldingExp::Constant(*c),
+        Expr::Cell(v) => {
+            let Variable { col, row } = v;
+            let col = match col {
+                Column::Witness(i) => FoldingColumn::Witness(*i),
+                Column::Index(i) => FoldingColumn::Index(*i),
+                Column::Coefficient(i) => FoldingColumn::Coefficient(*i),
+                _ => unreachable!(),
+            };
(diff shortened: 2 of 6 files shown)


kimchi/src/folding/test.rs
```diff
@@ -0,0 +1,30 @@
+use crate::folding::expressions::{extract_terms, FoldingColumn, FoldingExp};
+
+#[test]
+fn test_term_separation() {
```
dannywillems, line 4: What does the test do exactly?

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+        vec![zero; self.len]
+    }
+}
+pub trait Sponge<G: CommitmentCurve> {
+    fn challenge(absorbe: &PolyComm<G>) -> G::ScalarField;
+}
+
+pub struct FoldingScheme<G: CommitmentCurve, S: Sponge<G>, C: SRS<G>> {
+    _sponge: PhantomData<S>,
+    // original_exp: Expr<ConstantExpr<G::ScalarField>>,
```
dannywillems, line 157: Can be removed?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+impl<C> Default for IntegratedFoldingExpr<C> {
+    fn default() -> Self {
+        Self {
+            degree_0: vec![],
+            degree_1: vec![],
+            degree_2: vec![],
+        }
+    }
+}
+// fn extract_terms<C: Clone, F>(exp: FoldingExp<C>) -> Vec<Term<C>>
```
dannywillems, line 126: Can be removed?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+        let t = (exp, sign);
+        match degree {
+            Degree::Zero => integrated.degree_0.push(t),
+            Degree::One => integrated.degree_1.push(t),
+            Degree::Two => integrated.degree_2.push(t),
+        }
+    }
+    integrated
+}
+fn simplify_expression<C: Copy, F: Fn(&FeatureFlag) -> bool>(
```
dannywillems, line 244: Could you add tests please?

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,126 @@
+use super::expressions::{FoldingExp, IntegratedFoldingExpr};
+use super::FoldingEnv;
+use crate::folding::expressions::{Degree, FoldingColumn};
+use crate::folding::EvalLeaf;
+use ark_ff::Field;
+
+pub(crate) fn eval_exp_error<'a, F: Field, E: FoldingEnv<F>>(
+    exp: &FoldingExp<F>,
+    env: &'a E,
+    side: bool,
```
dannywillems, line 10: What is side used for?

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn map<M: Fn(&F) -> F, I: Fn(&mut F)>(self, map: M, in_place: I) -> Self {
+        use EvalLeaf::*;
+        match self {
+            Const(c) => Const(map(&c)),
+            Col(col) => {
+                let res = col.iter().map(map).collect();
+                Result(res)
+            }
+            Result(mut col) => {
+                for cell in col.iter_mut() {
```
dannywillems, line 48: If you want to apply functional programming, why not using `for_each` instead of the evil `for`? :)

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+    fn default() -> Self {
+        Self {
+            degree_0: vec![],
+            degree_1: vec![],
+            degree_2: vec![],
+        }
+    }
+}
+// fn extract_terms<C: Clone, F>(exp: FoldingExp<C>) -> Vec<Term<C>>
+pub fn extract_terms<C: Clone, F>(exp: FoldingExp<C>, terms: &mut Vec<Term<C>>, f: &F)
```
volhovm, line 127: Regarding the function style I'm tempted to say that it should not operate on a mutable append-only vec (note that append-only property is not enforced on the type level), but instead return `Vec<Term>` which will be concatenated, but maybe that's not as efficient. 

Another point is that I don't see why you have to pass this `map` argument `f` into the next call instead of mapping elements in the current iteration and calling `extract_terms` recursively on the already-transformed data. I guess also efficiency?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+    flag_resolver: &F,
+) -> IntegratedFoldingExpr<C> {
+    let simplied = simplify_expression(exp, flag_resolver);
+    let mut terms = vec![];
+    extract_terms(simplied, &mut terms, &|t| t);
+    let mut integrated = IntegratedFoldingExpr::default();
+    for term in terms.into_iter() {
+        let Term { exp, sign, degree } = term;
+        let t = (exp, sign);
+        match degree {
```
volhovm, line 236: Stupid question, but: isn't it a problem that `integrated` is now sorted differently than `terms`. Does the order in which the terms are extracted matter at all?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+                CurrOrNext::Curr => *cell,
+                CurrOrNext::Next => FoldingExp::Mul(cell, shift),
+            }
+        }
+        //may be better to replace by mul by 2
+        Expr::Double(e) => {
+            let e = simplify_expression(e, flag_resolver);
+            FoldingExp::Double(Box::new(e))
+        }
+        //may be better to replace by mul by itself
```
volhovm, line 270: I intuitively agree with both square and double -- I don't see how pushing them deeper into processing hierarchy is useful in terms of performance (and semantics wise it's better to transform them into `a+a` and `a*a` right here).

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+        Expr::IfFeature(flag, e1, e2) => {
+            if flag_resolver(flag) {
+                simplify_expression(e1, flag_resolver)
+            } else {
+                simplify_expression(e2, flag_resolver)
+            }
+        }
+        Expr::Pow(_, _) => {
+            unreachable!()
```
volhovm, line 300: Why is this unreachable? :thinking:

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+        //TODO: check that this is fine
+        Expr::VanishesOnZeroKnowledgeAndPreviousRows => todo!(),
```
volhovm, line 303: What's a semantics of this even?

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    len: usize,
+}
+impl<'a, G, I, W> FoldingEnv<G::ScalarField> for Env<'a, G, I, W>
+where
+    G: CommitmentCurve,
+    I: InstanceTrait<G>,
+    W: WitnessTrait<G>,
+{
+    fn witness(&self, i: usize, side: bool) -> &Vec<G::ScalarField> {
+        match side {
```
volhovm, line 113: Yeah, agreeing with Danny's question, `side` could be a two-valued enum (Left/Right). This is unclear.

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    fn coefficient(&self, i: usize) -> &Vec<G::ScalarField> {
+        &self.coefficients[i]
+    }
+
+    fn shift(&self) -> &Vec<G::ScalarField> {
+        self.shift
+    }
+
+    fn lagrange_basis(&self, i: &i32) -> Vec<G::ScalarField> {
```
volhovm, line 138: Why not give it just `i32`, it's cheap to copy anyway?

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    pub fn fold_instance_witness_pair<I, W, A, B>(
+        &self,
+        a: A,
+        b: B,
+    ) -> (RelaxedInstance<G, I>, RelaxedWitness<G, W>, PolyComm<G>)
+    where
+        I: InstanceTrait<G>,
+        W: WitnessTrait<G>,
+        A: RelaxablePair<G, I, W>,
+        B: RelaxablePair<G, I, W>,
```
volhovm, line 205: Is there a value in assuming `A != B` here?

volhovm (commented):
I don't feel confident enough yet to see some technical details, but maybe that's for the next pass. Generally it looks good!

I would ask to:
- `rustic`-format everything
- Add comments to at least the fields of the most important structs. This PR contains a lot of non-trivial knowledge and you're the only person having it rn, so it's of a very high value to leave some comments.

kimchi/src/folding/test.rs
fabrizio-m: actually, it doesn't test anything, I just used it to see the output, I'll make it non default.

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+        A: RelaxableInstance<G, I>,
+        B: RelaxableInstance<G, I>,
+    {
+        let a: RelaxedInstance<G, I> = a.relax(self.zero_commitment.clone());
+        let b: RelaxedInstance<G, I> = b.relax(self.zero_commitment.clone());
+        let challenge = S::challenge(&error_commitment);
+        RelaxedInstance::combine(a, b, challenge).add_error(&error_commitment, challenge)
+    }
+
+    //this can be reused for quadricization
```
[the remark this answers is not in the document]
fabrizio-m: it is not strictly part of folding, it is basically the missing part for the quadricization optimization, I let it there to deleted after I integrate quadricization and implement the optimization, but I can remove it now if you prefer.

kimchi/src/folding/mod.rs
fabrizio-m: yes, I'll remove it.

kimchi/src/folding/expressions.rs
fabrizio-m: What kind of test do you have in mind? it isn't very far from a `From` implementation, I suppose I can write the expected results for some inputs and assert they are equal to the actual output, something like that?

kimchi/src/folding/error.rs
fabrizio-m: to choose between one instance/witness and the other

kimchi/src/folding/mod.rs
fabrizio-m: I didn't want to apply anything in particular, generally I prefer `for in` for doing side effects or mutations with each element, and I find it clearer than `for_each` unless I already have a closure available.

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,170 @@
...
+            error_vec: zero_poly.to_vec(),
+        }
+    }
+}
+pub struct RelaxedInstance<G: CommitmentCurve, I: InstanceTrait<G>> {
+    instance: I,
+    pub u: G::ScalarField,
+    error_commitment: PolyComm<G>,
+}
+trait Relaxable<G: CommitmentCurve>: InstanceTrait<G> + Sized {
```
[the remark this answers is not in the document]
fabrizio-m: Indeed, remnant of a previous approach that ended up not using, rust won't detect unused traits and there are a few others I also forget to remove.

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    fn relax(self, zero_poly: &[G::ScalarField]) -> RelaxedWitness<G, Self> {
+        let witness = self;
+        RelaxedWitness {
+            witness,
+            error_vec: zero_poly.to_vec(),
+        }
+    }
+}
+pub struct RelaxedInstance<G: CommitmentCurve, I: InstanceTrait<G>> {
```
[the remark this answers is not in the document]
fabrizio-m: there is actually inter dependency, instances can be relaxed, but relaxed instances are generic over an inner instance, do you still think it is better to move them?

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,170 @@
...
+    ) -> (RelaxedInstance<G, I>, RelaxedWitness<G, W>) {
+        let (instance, witness) = self;
+        (
+            RelaxableInstance::relax(instance, zero_commitment),
+            RelaxableWitness::relax(witness, zero_poly),
+        )
+    }
+}
+
+trait Pair2<G: CommitmentCurve> {
```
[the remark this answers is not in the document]
fabrizio-m: will remove

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,170 @@
...
+        } = self;
+        let error_commitment = &error - &error_commitment.scale(challenge);
+        RelaxedInstance {
+            instance,
+            u,
+            error_commitment,
+        }
+    }
+
+    pub(super) fn combine(a: Self, b: Self, challenge: <G>::ScalarField) -> Self {
```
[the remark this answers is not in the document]
fabrizio-m: maybe using the same name was not the best idea, the struct level methods are extensions based on the method of the underlying type implementing the trait.
I just need `RelaxedInstance` and the like to be opaque types and use the methods internally.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+            FoldingExp::Cache(_, e) => e.degree(),
+        }
+    }
+}
+impl std::ops::Add for Degree {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        use Degree::*;
+        match (self, rhs) {
```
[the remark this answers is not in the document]
fabrizio-m: The idea is to support only degree $<=2$ at folding level, and to support arbitrary degree by converting expressions above it into degree 2 by using quadricization.
But maybe it would be better to do the quadricization work internally inside folding, still this type in particular would only handle degree $<=2$ or maybe $<=3$ if I implement and optimization I thought for the challenges.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+{
+    use Degree::*;
+    match exp {
+        e @ FoldingExp::Constant(_) => {
+            let term = f(Term::new(e, Zero));
+            terms.push(term)
+        }
+        FoldingExp::Cell(col) => {
+            let degree = match col {
+                FoldingColumn::Witness(_) | FoldingColumn::WitnessExtended(_) => One,
```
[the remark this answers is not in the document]
fabrizio-m: Answering also the above comment, the idea is that an expression can have an arbitrary number of terms, each of a different degree, while the `Term` should have a single term.
If you think of expressions as multivariate polynomials, for folding they can't have mixed terms of different degrees, I need to separate them to be able to handle them in different ways later.
Here `extract_terms` should take a potentially mixed expression and segregate the terms, the `degree` field of `Term` should be just what you would get from.
Maybe I misunderstood your question, we can probably remove the `degree` field and instead call `exp.degree()` without much change to the code, if that si a better answer.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+            let f = |t| {
+                let Term { exp, sign, degree } = t;
+                let exp = FoldingExp::Double(Box::new(exp));
+                f(Term { exp, sign, degree })
+            };
+            let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+            extract_terms(*e, terms, &f);
+        }
+        FoldingExp::Square(e) => {
+            let mut temp = Vec::with_capacity(4);
```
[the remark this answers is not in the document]
fabrizio-m: not exactly 4, just at most 4 most of the time, it just felt optimal to me at that time, but `Vec::new()` would also work.

kimchi/src/folding/expressions.rs
fabrizio-m: Yeah, concatenation is the most aesthetic for me, I just didn't want to allocate so many `Vec`s, but maybe the cost is not significant and it's better to do refactor it that way.
I had some issues with the compiler and `f` which led to the unnatural use of recursion, don't look very deep into it, it's just the first approach I could get to compile.
In the latest versions of Rust you get a slightly more understandable error instead of a crash, I may be able to refactor it later into a more straightforward way.

kimchi/src/folding/expressions.rs
fabrizio-m: I think it isn't an issue as long as `folding_expression` is pure, and even if the order where to be different each time, the expression should evaluate to the same.

kimchi/src/folding/expressions.rs
fabrizio-m: I should comment that, quadricization already flattens any `Pow` into a tree of `Mul`, thus there shouldn't be any `Pow` at this point.

kimchi/src/folding/expressions.rs
fabrizio-m: I'm not completely sure right now, it was added after I started this branch, I believe that it is 0 in the las 4 rows used for ZK, but it replaced another variant that already did that making me think that it does something more.
It is low priority now given that we won't be using ZK with folding for now at least.

kimchi/src/folding/mod.rs
fabrizio-m: will change

kimchi/src/folding/mod.rs
fabrizio-m: it just so happens that in most cases I already have `&i32`, it was just less work than having to dereference at each call.

kimchi/src/folding/mod.rs
fabrizio-m: It allows you to fold for example `(RelaxedInstance<G, I>, RelaxedWitness<G, W>)` with `(I,W)`.
Basically, you can provide an instance, witness or pair that could or not be relaxed and fold it with another one that could also be relaxed or not, and it will be relaxed internally if needed.
For IVC you usually have at each step a relaxed pair of accumulated computation and a non-relaxed pair for the new step.

kimchi/src/folding/expressions.rs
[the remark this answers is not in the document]
volhovm: Can you elaborate a bit more? I'm not sure how quadricization works, maybe there's somewhere I can read about it?

kimchi/src/folding/expressions.rs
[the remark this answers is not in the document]
volhovm: But... why do you extract 4 elements most of the time? (1) I'm thinking if the maximum expected degree is two, the average "square" term will contain a constant, so I would expect to only extract one term? (2) my alternative understanding is that this `e` can contain anything, like a highly nested expression, and `4` is kind of just random here. Which understanding is closer?...

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+    env: &E,
+    u: (F, F),
+) -> Vec<F> {
+    let add = |a, b| EvalLeaf::bin_op(a, b, |a, b| *a + b, |a, b| *a += b);
+    let sub = |a, b| EvalLeaf::bin_op(a, b, |a, b| *a - b, |a, b| *a -= b);
+    let scale = |t, s| EvalLeaf::bin_op(t, EvalLeaf::Const(s), |a, b| *a * b, |a, b| *a *= b);
+
+    let t_0 = EvalLeaf::Result(env.zero_vec());
+    let t_0 = exp.degree_0.iter().fold(t_0, |t_0, (exp, sign)| {
+        //could be true or false, doesn't matter for constant terms
```
querolita, line 97: I guess here you mean could be Left or Right

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+                (sub(t_1l, el), sub(t_1r, er))
+            }
+        });
+    let t_1l = scale(t_1l, u.1);
+    let t_1r = scale(t_1r, u.0);
+    let t_1 = EvalLeaf::bin_op(t_1l, t_1r, |a, b| *a + b, |a, b| *a += b);
+
+    let t_2 = EvalLeaf::Result(env.zero_vec());
+    let t_2 = exp.degree_0.iter().fold(t_2, |t_2, (exp, sign)| {
+        //true or false matter in some way, but not at the top level call
```
querolita, line 127: Here again, Left Right

querolita: In any case, this comment is a bit hand-wavy and does not contain much information an outsider can understand.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum FoldingColumn {
+    Witness(usize),
+    ///for the extra columns added by quadricization
+    #[allow(dead_code)]
+    WitnessExtended(usize),
+    Index(GateType),
+    Coefficient(usize),
+    ///basically X, to allow accesing the next row
+    Shift,
```
querolita, line 16: Oh so this is like Curr and Next you mean?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+        match self {
+            FoldingExp::Constant(_) => Zero,
+            FoldingExp::Cell(col) => match col {
+                FoldingColumn::Witness(_) | FoldingColumn::WitnessExtended(_) => One,
+                _ => Zero,
+            },
+            FoldingExp::Double(e) => e.degree(),
+            FoldingExp::Square(e) => &e.degree() * &e.degree(),
+            FoldingExp::Add(e1, e2) | FoldingExp::Sub(e1, e2) | FoldingExp::Mul(e1, e2) => {
+                e1.degree() + e2.degree()
```
querolita, line 51: Why is the degree of the addition and subtraction the sum of both sides' instead of the maximum between them? I understand for multiplication, but for the others?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+impl std::ops::Add for Degree {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        use Degree::*;
+        match (self, rhs) {
+            (Zero, Zero) => Zero,
+            (Zero, One) | (One, Zero) | (One, One) => One,
```
querolita, line 65: like, here inside `add` if you have (One, One) the result is One and not 2

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+impl std::ops::Mul for &Degree {
+    type Output = Degree;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        use Degree::*;
+        match (self, rhs) {
+            (Zero, Zero) => Zero,
+            (Zero, One) | (One, Zero) => One,
+            (Zero, Two) | (One, One) | (Two, Zero) => Two,
```
querolita, line 78: Okay, in multiplication it is the addition of both degrees. Seems fine.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,305 @@
...
+            (One, Two) | (Two, One) | (Two, Two) => panic!("degree over 2"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Term<C> {
+    pub exp: FoldingExp<C>,
+    pub sign: bool,
+    pub degree: Degree,
```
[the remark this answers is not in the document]
querolita: Mmmmm I also have the feeling that I am not getting the semantics quite right. See my comments above regarding the Add and Sub.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+            extract_terms(*e2, terms, f);
+        }
+        FoldingExp::Sub(e1, e2) => {
+            extract_terms(*e1, terms, f);
+            let f = |t: Term<C>| f(-t);
+            let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+            extract_terms(*e2, terms, &f);
+        }
+        FoldingExp::Mul(e1, e2) => {
+            let mut temp = Vec::with_capacity(4);
```
querolita, line 187: Where does the 4 come from?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+            for term in temp {
+                let Term { exp, sign, degree } = term;
+                let exp = Box::new(exp);
+                let f = |t| {
+                    let Term {
+                        exp: exp2,
+                        sign: sign2,
+                        degree: degree2,
+                    } = t;
+                    let degree = &degree * &degree2;
```
querolita, line 198: why is the degree multiplied instead of added here?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+                let exp = FoldingExp::Cache(id, Box::new(exp));
+                Term { exp, sign, degree }
+            };
+            let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+            extract_terms(*e, terms, &f);
+        }
+    }
+}
+
+///simplyfies the expression, and separates it into terms of different degree y
```
querolita, line 223: this means a vector of as many terms as different degrees in the expression?

kimchi/src/folding/expressions.rs
querolita: Yeah, something similar at least to be able to see at a first glance how they should look like

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
...
+            error_vec: zero_poly.to_vec(),
+        }
+    }
+}
+pub struct RelaxedInstance<G: CommitmentCurve, I: InstanceTrait<G>> {
+    instance: I,
+    pub u: G::ScalarField,
+    error_commitment: PolyComm<G>,
+}
+pub struct RelaxedWitness<G: CommitmentCurve, W: WitnessTrait<G>> {
```
querolita, line 36: What is a relaxed witness and instance?

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+
+mod error;
+mod expressions;
+mod instance_witness;
+#[cfg(test)]
+mod test;
+
+pub(crate) trait FoldingEnv<F> {
+    fn witness(&self, i: usize, side: Side) -> &Vec<F>;
+    fn witness_ext(&self, i: usize, side: Side) -> &Vec<F>;
```
querolita, line 23: maybe use verbose for this one so it is easier to parse/find, given that the enum uses the verbose version of this (meaning, WitnessExtended)

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn index(&self, i: &GateType) -> &Vec<F>;
+    fn coefficient(&self, i: usize) -> &Vec<F>;
+    fn shift(&self) -> &Vec<F>;
+    fn lagrange_basis(&self, i: &i32) -> Vec<F>;
+    ///a vec of just zeros of the same length as other columns
+    fn zero_vec(&self) -> Vec<F>;
```
querolita, line 29: maybe zeros is equally meaningful

querolita (commented):
Left a few comments here and there to help me digest the idea behind. Using more comments would definitely be useful for knowledge sharing, given that the corresponding RFC is not too explicit either.

kimchi/src/folding/instance_witness.rs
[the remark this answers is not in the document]
volhovm: Perhaps not then, gotcha.

kimchi/src/folding/instance_witness.rs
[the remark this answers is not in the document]
volhovm: I think one thing I do not find intuitive in this hierarchy of structs/traits is that `InstanceTrait` and `WitnessTrait` have the `relax` function default implementation. Not all the instances/witnesses are relaxable, and you have a specific trait for those which are: `RelaxedInstance`/`RelaxedWitness`. Maybe a better way would be to use the `relax` from them.

It's a nit perhaps but this hierarchy is quite confusing.

kimchi/src/folding/mod.rs
volhovm: Getting back to the previous comment on struct/trait hierarchy: I think it should be `I: RelaxableInstance, W: RelaxableWitness`, since not all the instances are relaxable. It makes sense, as you say, that you can merge relaxed pair with a standard (but relaxable) pair, but the naming is a bit confusing. 

I'll resolve this particular discussion.

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+        Term {
+            sign: !self.sign,
+            ..self
+        }
+    }
+}
+///A simplified expression with all terms separated by degree
+pub struct IntegratedFoldingExpr<C> {
+    pub(super) degree_0: Vec<(FoldingExp<C>, bool)>,
+    pub(super) degree_1: Vec<(FoldingExp<C>, bool)>,
```
volhovm, line 113: `degree_1` and `degree_2` are not used anywhere atm. I'd expect to see them in `compute_error` for cross term T computation, but there you only use `degree_0` -- I assume everything written passes tests, so I'm missing smth?

volhovm (commented):
Left some comments... half is what we discussed yesterday, another half is new.

kimchi/src/folding/instance_witness.rs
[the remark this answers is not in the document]
fabrizio-m: actually, all instances/pairs are relaxable, the ones already relaxed have a trivial implementation that is just the identity function.
And there is no specific traits for `RelaxedInstance`/`RelaxedWitness`, just type-level methods.
I see it as similar to taking `Into<T>` instead of `T` as the argument in some function for example.

kimchi/src/folding/expressions.rs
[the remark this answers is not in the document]
fabrizio-m: the assumption would be for higher degree expressions to not appear, with the detail that the degree is computed only counting the witness columns, fixed columns are "free".
But I think I'll integrate quadricization into folding, that should make everything clearer.

kimchi/src/folding/expressions.rs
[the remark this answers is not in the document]
fabrizio-m: right, I was thinking of the (a+b)(c+b) case, but it isn't necessarily the most common.

kimchi/src/folding/expressions.rs
fabrizio-m: in some way it is the actual polynomial, the order shouldn't matter, at some point it will have to be turned into `Expr` again to provide it to the proof system, and I'll change it later so that it is handled by folding itself and you don't have to know `IntegratedFoldingExpr` even exists.

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    match exp {
+        Constant(c) => EvalLeaf::Const(*c),
+        Cell(col) => match col {
+            Witness(i) => Col(env.witness(*i, side)),
+            WitnessExtended(i) => Col(env.witness_ext(*i, side)),
+            Index(i) => Col(env.index(i)),
+            Coefficient(i) => Col(env.coefficient(*i)),
+            //TODO: maybe specialize
+            Shift => Col(env.shift()),
```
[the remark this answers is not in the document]
jspada: same question

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+            let a = eval_exp_error(e1, env, side);
+            let b = eval_exp_error(e2, env, side);
+            EvalLeaf::bin_op(a, b, |a, b| *a + b, |a, b| *a += b)
+        }
+        FoldingExp::Sub(e1, e2) => {
+            let a = eval_exp_error(e1, env, side);
+            let b = eval_exp_error(e2, env, side);
+            EvalLeaf::bin_op(a, b, |a, b| *a - b, |a, b| *a -= b)
+        }
+        FoldingExp::Mul(e1, e2) => match (degree, e1.degree()) {
```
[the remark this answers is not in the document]
jspada: agree

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+    type Output = Self;
+
+    fn not(self) -> Self::Output {
+        match self {
+            Side::Left => Side::Right,
+            Side::Right => Side::Left,
+        }
+    }
+}
+pub(crate) fn eval_exp_error<'a, F: Field, E: FoldingEnv<F>>(
```
jspada, line 22: would be good to document this function.

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+        FoldingExp::Sub(e1, e2) => {
+            let a = eval_exp_error(e1, env, side);
+            let b = eval_exp_error(e2, env, side);
+            EvalLeaf::bin_op(a, b, |a, b| *a - b, |a, b| *a -= b)
+        }
+        FoldingExp::Mul(e1, e2) => match (degree, e1.degree()) {
+            (Degree::Two, Degree::One) => {
+                let a = eval_exp_error(e1, env, side);
+                let b = eval_exp_error(e2, env, !side);
```
jspada, line 67: maybe use an enum for the sides?

kimchi/src/folding/error.rs
```diff
@@ -0,0 +1,141 @@
...
+        FoldingExp::UnnormalizedLagrangeBasis(i) => EvalLeaf::Result(env.lagrange_basis(i)),
+        //TODO: use cache
+        FoldingExp::Cache(_id, exp) => eval_exp_error(exp, env, side),
+    }
+}
+
+pub(crate) fn compute_error<F: Field, E: FoldingEnv<F>>(
+    exp: &IntegratedFoldingExpr<F>,
+    env: &E,
+    u: (F, F),
```
jspada, line 89: maybe give these names, i.e., `(u1, u2) : (F: F)`

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+    WitnessExtended(usize),
+    Index(GateType),
+    Coefficient(usize),
+    ///basically X, to allow accesing the next row
+    Shift,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+///A simplified expression to be use in folding
+pub enum FoldingExp<C> {
```
jspada, line 21: Would be nice to comment these

kimchi/src/folding/expressions.rs
jspada: good question.  maybe it should be `max(e1.degree(), e2.degree())` ?

kimchi/src/folding/expressions.rs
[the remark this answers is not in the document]
jspada: do you assert somewhere degree of input is $\le 2$

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+impl<C> Default for IntegratedFoldingExpr<C> {
+    fn default() -> Self {
+        Self {
+            degree_0: vec![],
+            degree_1: vec![],
+            degree_2: vec![],
+        }
+    }
+}
+pub fn extract_terms<C: Clone, F>(exp: FoldingExp<C>, terms: &mut Vec<Term<C>>, f: &F)
```
jspada, line 126: Please comment/document this function

kimchi/src/folding/expressions.rs
jspada: ditto

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+                    let sign = sign == sign2;
+                    let exp = FoldingExp::Mul(exp.clone(), Box::new(exp2));
+                    Term { exp, sign, degree }
+                };
+                let f: Box<dyn Fn(Term<C>) -> Term<C>> = Box::new(f);
+                extract_terms(*e2.clone(), terms, &f);
+            }
+        }
+        e @ FoldingExp::UnnormalizedLagrangeBasis(_) => {
+            let term = f(Term::new(e, Zero));
```
jspada, line 208: What's happening here?

kimchi/src/folding/expressions.rs
jspada: what do you mean by "pure"?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+) -> FoldingExp<C> {
+    match exp {
+        Expr::Constant(c) => FoldingExp::Constant(*c),
+        Expr::Cell(v) => {
+            let Variable { col, row } = v;
+            let col = match col {
+                Column::Witness(i) => FoldingColumn::Witness(*i),
+                Column::Index(i) => FoldingColumn::Index(*i),
+                Column::Coefficient(i) => FoldingColumn::Coefficient(*i),
+                _ => unreachable!(),
```
jspada, line 255: Is this necessary? should it be `panic!`?

kimchi/src/folding/expressions.rs
```diff
@@ -0,0 +1,304 @@
...
+            };
+            let cell = Box::new(FoldingExp::Cell(col));
+            let shift = Box::new(FoldingExp::Cell(FoldingColumn::Shift));
+            match row {
+                CurrOrNext::Curr => *cell,
+                CurrOrNext::Next => FoldingExp::Mul(cell, shift),
+            }
+        }
+        //may be better to replace by mul by 2
```
jspada, line 264: in terms of simplicity, performance or both?

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
+use super::{CommitmentCurve, One, PolyComm};
+
+pub trait InstanceTrait<G: CommitmentCurve>: Sized {
```
jspada, line 3: please comment all the interface

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn relax(self, zero_commit: PolyComm<G>) -> RelaxedInstance<G, Self> {
+        let instance = self;
+        RelaxedInstance {
+            instance,
+            u: G::ScalarField::one(),
+            error_commitment: zero_commit,
+        }
+    }
+}
+pub trait WitnessTrait<G: CommitmentCurve>: Sized {
```
jspada, line 15: same

kimchi/src/folding/instance_witness.rs
jspada: Aren't they just the ones with the error terms allowed?

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
...
+        self.relax(zero_poly)
+    }
+}
+impl<G: CommitmentCurve, W: WitnessTrait<G>> RelaxableWitness<G, W> for RelaxedWitness<G, W> {
+    fn relax(self, _zero_poly: &[G::ScalarField]) -> RelaxedWitness<G, W> {
+        self
+    }
+}
+
+// pub trait RelaxableInstance<G: KimchiCurve, >: seal::Seal<G, I, 0> {
```
jspada, line 67: please add rustdoc comments for these interfaces

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+impl<G: CommitmentCurve, I: InstanceTrait<G>> RelaxedInstance<G, I> {
+    pub fn add_error(self, error_commitment: &PolyComm<G>, challenge: G::ScalarField) -> Self {
+        let RelaxedInstance {
+            instance,
+            u,
+            error_commitment: error,
+        } = self;
+        let error_commitment = &error - &error_commitment.scale(challenge);
```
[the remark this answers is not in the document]
jspada: nice idea.  would be clearer imo

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
...
+            error_commitment: e1,
+        } = a;
+        let RelaxedInstance {
+            instance: ins2,
+            u: u2,
+            error_commitment: e2,
+        } = b;
+        let instance = I::combine(ins1, ins2, challenge);
+        let u = u1 + u2 * challenge;
+        let error_commitment = &e1 + &e2.scale(challenge_square);
```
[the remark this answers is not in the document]
jspada: nice idea

kimchi/src/folding/instance_witness.rs
```diff
@@ -0,0 +1,157 @@
...
+            witness: a,
+            error_vec: mut e1,
+        } = a;
+        let RelaxedWitness {
+            witness: b,
+            error_vec: e2,
+        } = b;
+        let challenge = challenge * challenge;
+        let witness = W::combine(a, b, challenge);
+        for (a, b) in e1.iter_mut().zip(e2.into_iter()) {
```
jspada, line 151: nice

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn bin_op<M: Fn(&F, &F) -> F, I: Fn(&mut F, &F)>(
+        a: Self,
+        b: Self,
+        map: M,
+        in_place: I,
+    ) -> Self {
+        use EvalLeaf::*;
+        match (a, b) {
+            (Const(a), Const(b)) => Const(map(&a, &b)),
+            (Const(a), Col(b)) | (Col(b), Const(a)) => {
```
jspada, line 64: what about `(Col(a), Const(b))`, for example?

kimchi/src/folding/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+            (Result(mut a), Result(b)) => {
+                for (a, b) in a.iter_mut().zip(b.iter()) {
+                    in_place(a, b);
+                }
+                Result(a)
+            }
+        }
+    }
+}
+struct Env<'a, G, I, W>
```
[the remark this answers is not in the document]
jspada: It's used in the rust code some places to describe the "environment" in which something is being done.

jspada (commented):
Amazing work @fabrizio-m - left some questions and comments to help out.

PR #1288, opened 2023-10-19
Bring `witness/index-cell` to Keccak branches

querolita:
Needed for the Keccak layout


querolita:
Merged because witness/index-cell branch had already been merged to master, so the diff was +0-0

PR #1291, opened 2023-10-20
Kimchi + KZG prover

gabrielbosio:


gabrielbosio:
Closing because target branch doesn't have KZG implemented.

PR #1299, opened 2023-10-25
Adds trivial support for rotation by 0 and 64 bits

querolita:
This is to match the interface being created in o1js [in this PR](https://github.com/o1-labs/o1js/pull/1182).

This does not modify the constraints. This only affects the witness generation code for the Kimchi gadgets, which are only used for internal testing.

Now rotation by 0 and 64 bits are allowed. Nonetheless, it is not recommended since it adds 3 unneeded rows to the circuit, as it is trivial that the rotated word is just the same as the input.

### DIFF
--- a/kimchi/src/circuits/polynomials/rot.rs
+++ b/kimchi/src/circuits/polynomials/rot.rs
@@ -328,8 +328,6 @@ pub fn extend_rot<F: PrimeField>(
     rot: u32,
     side: RotMode,
 ) {
-    assert!(rot < 64, "Rotation value must be less than 64");
-    assert_ne!(rot, 0, "Rotation value must be non-zero");
     let rot = if side == RotMode::Right {
         64 - rot
     } else {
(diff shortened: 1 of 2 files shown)


kimchi/src/circuits/polynomials/rot.rs
```diff
@@ -328,8 +328,6 @@ pub fn extend_rot<F: PrimeField>(
     rot: u32,
     side: RotMode,
 ) {
-    assert!(rot < 64, "Rotation value must be less than 64");
```
jspada, line 331: Don't you want `assert!(rot <= 64,` instead?

jspada (approved):
Approved with one minor suggestion.

kimchi/src/circuits/polynomials/rot.rs
querolita: Oh true, thanks for the nit, I missed that line!

PR #1304, opened 2023-10-31
Revamp mina book structure

volhovm:
- Turn old RFCs into sections of the mina book
- Editorial: try to improve weird structure, inconsistent notation, broken latex, vague language... where possible. No /content/ changes are included in this PR.


book/src/SUMMARY.md
```diff
@@ -27,56 +27,46 @@
...
+
 
 # Pickles & Inductive Proof Systems
 
 - [Overview](./fundamentals/zkbook_ips.md)
 - [Accumulation](./pickles/accumulation.md)
 - [Deferred Computation](./pickles/deferred.md)
-- [Passthough & Me-Only](./pickles/passthrough.md)
-
-# RFCs
```
querolita, line 68: Not sure if we want to integrate this section inside Kimchi directly or if we prefer to keep track of the "changes" we made to the above.

querolita (approved):
Overall looks good to me, but I wonder if the rest will like that the RFCs section is now integrated somewhere else.

dannywillems:
Note: this had to target master.

volhovm:
@dannywillems why do you suggest this should target `master`? I explicitly made it target `develop`.

PR #1307, opened 2023-10-31
Pretty print info according to CLI frequency specification

rbonichon:


optimism/src/mips/witness.rs
```diff
@@ -225,9 +264,94 @@ impl<Fp: Field> Env<Fp> {
...
+    fn page_address(&self) -> (u32, usize) {
+        let address = self.instruction_pointer;
+        let page = (address >> PAGE_ADDRESS_SIZE) as u32;
+        let page_address = (address & (PAGE_ADDRESS_MASK as u32)) as usize;
+        (page, page_address)
+    }
+
+    fn get_opcode(&mut self) -> Option<u32> {
+        let (page_id, page_address) = self.page_address();
+        for (page_index, memory) in self.memory.iter_mut() {
```
mrmr1993, line 307: Nit: we're not mutating, so this can just be `iter`

mrmr1993 (approved):

optimism/src/mips/witness.rs
rbonichon: Done!

dannywillems (approved):

PR #1308, opened 2023-11-01
Adding sparse lookup tables (`Sparse`, `Reset`,  `Bits16`)

querolita:
Needed for Keccak, making this PR independent of the chain of PRs so it can be merged in advanced.

Revival of https://github.com/o1-labs/proof-systems/pull/1243 which got merged non-intendedly.

It creates a 2-column lookup table containing the 2^16 elements and their expansion to 64 bits.

~Still pending to figure out if this is enough (together with some logic changes in the lookup argument to look at single columns) or if we need 3 tables instead (to have access to each column separately), or if we prefer to have more columns in the gate.~

It additionally creates two more tables containing the above columns separately.

Closes https://github.com/o1-labs/proof-systems/issues/1223

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -344,8 +344,15 @@ pub const XOR_TABLE_ID: i32 = 0;
 /// The range check table ID.
 pub const RANGE_CHECK_TABLE_ID: i32 = 1;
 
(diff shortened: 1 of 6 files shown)


mrmr1993 (approved):

querolita:
Closing because this is only relevant in mips

PR #1309, opened 2023-11-01
Bring `kimchi/sparse-lookups` into Keccak branches

querolita:


dannywillems:
I suppose we ignore this PR as we move to RAMlookup

dannywillems (changes requested):
I suppose we ignore this PR as we move to RAMlookup

PR #1319, opened 2023-11-03
Support for `meta.json` file

rbonichon:
Add symbol to debugging/info output

### DIFF
--- a/optimism/src/cannon.rs
+++ b/optimism/src/cannon.rs
@@ -197,6 +197,10 @@ mod tests {
       "size": 0
     }]}"#;
 
+    fn deserialize_meta_sample() -> Meta {
+        serde_json::from_str::<Meta>(META_SAMPLE).unwrap()
+    }
+
     #[test]
     fn test_meta_deserialize_from_file() {
         let path = "meta_test.json";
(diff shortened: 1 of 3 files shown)


dannywillems (changes requested):
LGTM, but would be nice to have a meta.json file example and see if it is parsed correctly with a test

optimism/src/cannon.rs
```diff
@@ -152,3 +137,165 @@ impl Start {
...
+    fn test_find_address_symbol() {
+        let meta = deserialize_meta_sample();
+
+        assert_eq!(
+            meta.find_address_symbol(69633),
+            Some("internal/cpu.processOptions".to_string())
+        );
+        assert_eq!(
+            meta.find_address_symbol(69632),
+            Some("internal/cpu.processOptions".to_string())
```
dannywillems, line 297:
```suggestion
            Some("runtime.text".to_string())

```
?

dannywillems, line 297: Duplicated?

optimism/src/cannon.rs
```diff
@@ -152,3 +137,165 @@ impl Start {
...
+                    }
+                }
+            },
+        );
+
+        match res {
+            Ok(idx) => Some(self.symbols[idx].name.to_string()),
+            Err(_) => None,
+        }
+        // for e in self.symbols.iter() {
```
dannywillems, line 180: Old?

dannywillems (approved):
LGTM modulo last comments, but Clippy is grumpy.

optimism/src/cannon.rs
rbonichon: No. Not testing the same value.

optimism/src/cannon.rs
rbonichon: Yep. removed

PR #1321, opened 2023-11-07
Fix serialization of preimage key

rbonichon:
This was improperly deserialized as `String`, while we know it's a 256 bits key. Hence a `[u8; 32]` is more appropriate.


dannywillems (approved):
Lgtm, history looks weird. I thought GH would only give the commits different from master.

rbonichon:
I'll clean up the history. I saw some @mrmr1993's commits in there as well

PR #1334, opened 2023-11-16
Deduplicate lookup tables

mrmr1993:
This PR fixes a bug in the domain size computed for lookup tables. This allows us to prove with a smaller domain size when ffmul and some other range check lookup gate is used, and also to revert the workaround [here](https://github.com/MinaProtocol/mina/pull/14405/commits/c2d9b3b844630ddb3d604ec8b9ea76edd5360ab5).


dannywillems (approved):
Approving, but would have been nice to have a test, even it looks fine regarding the code. Opening a follow-up issue for a 10 min task.

PR #1338, opened 2023-11-22
Basic support for preimage oracle process within witness environment

rbonichon:


mrmr1993 (commented):
LGTM, thanks for this. It seems like CI is failing though

PR #1350, opened 2023-11-29
Fix old statements about implemented lookup tables

dannywillems:


book/src/kimchi/lookup.md
```diff
@@ -125,7 +126,10 @@ $$
 
 Since we would like to allow multiple table lookups per row, we define multiple **queries**, where each query is associated with a **lookup selector**.
 
-At the moment of this writing, the `ChaCha` gates all perform $4$ queries in a row. Thus, $4$ is trivially the largest number of queries that happen in a row.
+Previously, ChaCha20 was implemented in Kimchi but has been removed as unneeded.
```
rbonichon, line 129: suggestion: s/as unneeded/as it has become unneeded.

rbonichon (approved):
LGTM. I only have a single minor suggestion.

book/src/kimchi/lookup.md
dannywillems, line 129:
```suggestion
Previously, ChaCha20 was implemented in Kimchi but has been removed as it has become unneeded.
```

dannywillems: Fixed.

PR #1351, opened 2023-11-29
Update criterion benchmarks for kimchi

kantp:
Trying to run the criterion benchmarks for kimchi, I ran into a couple of minor issues (documentation, a macOS dependent thing, and duplicated names for different benchmarks which prevented criterion from running).

With these changes, I am able to run the criterion benchmarks for kimchi.

There's also a minor refactoring to remove some code duplication.

### DIFF
--- a/kimchi/benches/proof_criterion.rs
+++ b/kimchi/benches/proof_criterion.rs
@@ -9,7 +9,7 @@ pub fn bench_proof_creation(c: &mut Criterion) {
         let ctx = BenchmarkCtx::new(size);
 
         group.bench_function(
-            format!("proof creation ({} gates)", ctx.num_gates),
+            format!("proof creation (SRS size 2^{{{}}}, {} gates)", ctx.srs_size(), ctx.num_gates),
             |b| b.iter(|| black_box(ctx.create_proof())),
         );
     }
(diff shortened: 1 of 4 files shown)


kimchi/benches/proof_criterion.rs
```diff
@@ -5,26 +5,30 @@ pub fn bench_proof_creation(c: &mut Criterion) {
...
 
-    let ctx = BenchmarkCtx::new(14);
-    group.bench_function(
-        format!("proof creation (SRS size 2^{})", ctx.srs_size()),
-        |b| b.iter(|| black_box(ctx.create_proof())),
-    );
-
-    let proof_and_public = ctx.create_proof();
+        group.bench_function(
+            format!("proof creation ({} gates)", ctx.num_gates),
```
dannywillems, line 12: Can you add the SRS size, in log2 please? We often think in terms of log2 domain size for the benchmarks. It is good to have it on the stdout.

kimchi/benches/proof_criterion.rs
```diff
@@ -5,26 +5,30 @@ pub fn bench_proof_creation(c: &mut Criterion) {
...
-        format!("proof verification (SRS size 2^{})", ctx.srs_size()),
-        |b| b.iter(|| ctx.batch_verification(black_box(&vec![proof_and_public.clone()]))),
-    );
+
+    for size in [10, 14] {
+        let ctx = BenchmarkCtx::new(size);
+        let proof_and_public = ctx.create_proof();
+
+        group.bench_function(
+            format!("proof verification ({} gates)", ctx.num_gates),
```
dannywillems, line 27: Same comment than above.

dannywillems (changes requested):
Small changes. Otherwise, LGTM.

kimchi/benches/proof_criterion.rs
kantp: Sure!

(The reason I had added the number of gates was that the size in log2 alone was not leading to unique names, but having both solves the issue and keeps the log2 size).

kantp:
Thanks for the quick review! I made the changes you requested.

dannywillems (approved):

dannywillems:
Accepted. Running the CI and will merge it.

PR #1366, opened 2023-12-01
lui instruction

dannywillems:


optimism/src/mips/interpreter.rs
```diff
@@ -270,7 +326,16 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
         ITypeInstruction::AndImmediate => (),
         ITypeInstruction::OrImmediate => (),
         ITypeInstruction::XorImmediate => (),
-        ITypeInstruction::LoadUpperImmediate => (),
+        ITypeInstruction::LoadUpperImmediate => {
+            // lui $reg, [most significant 16 bits of immediate]
+            let rt = env.get_instruction_part(InstructionPart::RT);
+            let immediate_value = env.get_immediate();
```
querolita, line 332: According to [this](https://www.dsi.unive.it/~gasparetto/materials/MIPS_Instruction_Set.pdf),

> 6 Load constant into upper 16 bits.
> Lower 16 bits are set to zero.

 shouldn't this multiply the immediate value by 2^16 and then overwrite the register?

optimism/src/mips/interpreter.rs
```diff
@@ -387,4 +452,25 @@ mod tests {
...
+            rs: 0b00000,
+            rt: 0b00001,
+            rd: 0b00000,
+            shamt: 0b00000,
+            funct: 0b001010,
+        };
+        interpret_itype(&mut dummy_env, ITypeInstruction::LoadUpperImmediate);
+        assert_eq!(
+            dummy_env.registers[REGISTER_AT as usize],
+            0xa,
```
querolita, line 473: like, shouldnt this be 0x0a0000 instead?

querolita (commented):
Not sure if lui should behave exactly like this or not

optimism/src/mips/interpreter.rs
```diff
@@ -270,7 +326,16 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
         ITypeInstruction::AndImmediate => (),
         ITypeInstruction::OrImmediate => (),
         ITypeInstruction::XorImmediate => (),
-        ITypeInstruction::LoadUpperImmediate => (),
+        ITypeInstruction::LoadUpperImmediate => {
+            // lui $reg, [most significant 16 bits of immediate]
+            let rt = env.get_instruction_part(InstructionPart::RT);
+            let immediate_value = env.get_immediate() << 16;
```
querolita, line 332: looks good to me now

querolita (approved):

PR #1369, opened 2023-12-01
Implement jr

dannywillems:

PR #1371, opened 2023-12-02
Interpreter: implement sw, jr, lw and fix previously implemented addiu + add docs

dannywillems:
Contains (now closed): https://github.com/o1-labs/proof-systems/pull/1370 and https://github.com/o1-labs/proof-systems/pull/1369

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -363,17 +363,22 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
             return;
         }
         ITypeInstruction::AddImmediateUnsigned => {
+            // addiu: R[rt] <- R[rs] + signed_extended_imm
+            //                 -----------
+            //                     res
+            // Despite the name, it is adding a signed extended value!
+            // The difference with addi: doesn't cause overflow.
             let rs = env.get_instruction_part(InstructionPart::RS);
             let rt = env.get_instruction_part(InstructionPart::RT);
-            let immediate = env.get_immediate();
+            let signed_extended_imm = env.get_immediate();
             debug!(
                 "Instr: addiu {}, {}, {}",
                 Env::debug_register(&rt),
                 Env::debug_register(&rs),
-                immediate
+                Env::debug_signed_16bits_variable(&signed_extended_imm)
             );
-            let register_rs = env.fetch_register_checked(&rs);
-            let res = register_rs + immediate;
+            let r_rs = env.fetch_register_checked(&rs);
+            let res = Env::add_16bits_signed_offset(&r_rs, &signed_extended_imm);
             env.overwrite_register_checked(&rt, &res);
             env.set_instruction_pointer(env.get_instruction_pointer() + Env::constant(4u32));
             // TODO: update next_instruction_pointer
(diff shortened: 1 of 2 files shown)


optimism/src/mips/interpreter.rs
```diff
@@ -527,21 +634,71 @@ mod tests {
...
+        assert_eq!(dummy_env.instruction_pointer, ip + 4);
+    }
+
+    #[test]
+    fn test_unit_addiu_instruction_with_signed_value() {
+        // Setup
+        let mut rng = rand::thread_rng();
+        let mut dummy_env = dummy_env();
+        let ip = dummy_env.instruction_pointer;
+        // Add a value bigger than 12 to avoid underflow
```
querolita, line 663: would this be a problem otherwise? meaning, if in this instruction the content of the register was something smaller than 12, it would still produce a value, no? Even if "too large"

optimism/src/mips/witness.rs
```diff
@@ -100,6 +100,100 @@ fn memory_size(total: usize) -> String {
 impl<Fp: Field> InterpreterEnv for Env<Fp> {
     type Variable = u32;
 
+    // Debug
+    fn debug_signed_16bits_variable(v: &Self::Variable) -> String {
+        // FIXME: It does suppose u32 is between 0 and 2 ^ 16
```
querolita, line 105: but then why is it called signed?

optimism/src/mips/witness.rs
```diff
@@ -100,6 +100,100 @@ fn memory_size(total: usize) -> String {
...
+        let v = *v;
+        // handle 0b1000000000000000
+        if v == 1 << 15 {
+            return String::from("0");
```
querolita, line 109: Oh, so what's `0b0000000000000000`? Is it also zero, like are there two representations of zero?

optimism/src/mips/witness.rs
```diff
@@ -100,6 +100,100 @@ fn memory_size(total: usize) -> String {
...
+        }
+        let bit_sign = v >> 15;
+        if bit_sign == 1 {
+            let v = !(v - 1) & 0x0000ffff;
```
querolita, line 113: so it is not one leading bit to indicate sign, and the rest is the binary "absolute value"? Instead, you need to subtract 1 to the value if it is negative, then negate it all, and then get the 16 rightmost bits and that's the absolute value?

optimism/src/mips/witness.rs
```diff
@@ -100,6 +100,100 @@ fn memory_size(total: usize) -> String {
...
+
+        // handle 0b1000000000000000
+        if v == 1 << 15 {
+            return x;
+        }
+
+        let bit_sign = v >> 15;
+        if bit_sign == 1 {
+            let v = !(v - 1) & 0x0000ffff;
+            if x == 0 {
```
querolita, line 171: cannot you add 0 and a negative value?

optimism/src/mips/witness.rs
```diff
@@ -100,6 +100,100 @@ fn memory_size(total: usize) -> String {
...
+                panic!("Underflow expected: x = {}, v = -{}", x, v)
+            }
+            x - v
+        } else {
+            x + v
```
querolita, line 176: but instead if x was negative instead of v, could you do it? that seems confusing

optimism/src/mips/witness.rs
```diff
@@ -100,6 +100,100 @@ fn memory_size(total: usize) -> String {
...
+        Self::Variable,
+        Self::Variable,
+        Self::Variable,
+        Self::Variable,
+    ) {
+        let v1 = (*value >> 24) & ((1 << (32 - 24)) - 1);
+        let v2 = (*value >> 16) & ((1 << (24 - 16)) - 1);
+        let v3 = (*value >> 8) & ((1 << (16 - 8)) - 1);
+        let v4 = *value & ((1 << 8) - 1);
+        (v1, v2, v3, v4)
```
querolita, line 194: Mmmmm iirc, I believe in other parts of the code you refer to v4 as the most significant chunk but here it looks like it is the least significant one. Maybe try to unify the standard?

dannywillems:
Closing.

PR #1381, opened 2023-12-04
Add support for scratch_state to MIPS VM

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/1380, adding the `alloc_scratch` helper to the environment trait. This will be the basis of all intermediate storage for the gate constraints.

In order to do this, this PR adds
* a `Column` type (currently missing: `InstructionCounter`, since it's unused here)
* a `Position` type to the trait
* `alloc_scratch` to allocate a new position in the scratch state
* a `write_column` helper in the witness environment, to allow for writing to the scratch state (currently unused).

### DIFF
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -4,6 +4,7 @@ use crate::{
         PAGE_SIZE,
     },
     mips::{
+        column::Column,
         interpreter::{
             self, ITypeInstruction, Instruction, InstructionPart, InstructionParts, InterpreterEnv,
             JTypeInstruction, RTypeInstruction,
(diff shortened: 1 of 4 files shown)


optimism/src/mips/witness.rs
```diff
@@ -211,6 +220,12 @@ impl<Fp: Field> Env<Fp> {
         }
     }
 
+    pub fn write_column(&mut self, column: Column, value: u64) {
+        match column {
+            Column::ScratchState(idx) => self.scratch_state[idx] = value.into(),
+        }
```
querolita, line 226: I guess as soon as we add more variants into Column then we will have to add the branch `_ => panic!("Unexpected column"),` in the match, as in the demo?

querolita (approved):

PR #1382, opened 2023-12-05
Use scratch allocations when fetching memory

mrmr1993:
This PR uses the scratch state implemented in https://github.com/o1-labs/proof-systems/pull/1381 to pull values into the local witness from memory.

The access function is marked `unsafe`, following the rust guidelines: users of the function should be careful to add the constraints manually, because it will not do it for them. We are not currently being this careful.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -209,7 +209,10 @@ pub trait InterpreterEnv {
 
     /// Fetch the memory value at address `addr` and store it in local position `output`.
     ///
-    /// This is unsafe: no lookups or other constraints are added as part of this operation.
+    /// # Safety
+    ///
+    /// No lookups or other constraints are added as part of this operation. The caller must
+    /// manually add the lookups for this memory operation.
     unsafe fn fetch_memory(
         &mut self,
         addr: &Self::Variable,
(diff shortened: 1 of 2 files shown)


optimism/src/mips/interpreter.rs
```diff
@@ -367,10 +377,39 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
                 addr.clone()
             );
             // We load 4 bytes, i.e. one word.
-            let v0 = env.fetch_memory(&addr_with_offset);
-            let v1 = env.fetch_memory(&(addr_with_offset.clone() + Env::constant(1)));
-            let v2 = env.fetch_memory(&(addr_with_offset.clone() + Env::constant(2)));
-            let v3 = env.fetch_memory(&(addr_with_offset.clone() + Env::constant(3)));
+            let v0 = unsafe {
+                // FIXME: A safe wrapper should be exposed in the trait, and it must add the
```
querolita, line 381: Thanks for the comment

optimism/src/mips/interpreter.rs
```diff
@@ -207,7 +207,17 @@ pub trait InterpreterEnv {
 
     fn fetch_register_checked(&self, register_idx: &Self::Variable) -> Self::Variable;
 
-    fn fetch_memory(&mut self, addr: &Self::Variable) -> Self::Variable;
+    /// Fetch the memory value at address `addr` and store it in local position `output`.
```
querolita, line 210: Oh so now the storing also happens inside the fetch, thanks for the doc comment

querolita (approved):

PR #1383, opened 2023-12-05
Add other accessors necessary for zkVM memory

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/1382, adding counterpart operations for
* writing to memory
* reading the 'access index' for a memory address
* writing the 'access index' for a memory address.

These primitives are essential for the correct operation of RAMLookup, ensuring that reads are 'temporal' (i.e. you can't read a value in the past that you wrote in the future).

### DIFF
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -47,7 +47,7 @@ pub struct Env<Fp> {
     pub instruction_parts: InstructionParts<u32>,
     pub instruction_counter: usize,
     pub memory: Vec<(u32, Vec<u8>)>,
-    pub memory_write_index: Vec<(u32, Vec<usize>)>,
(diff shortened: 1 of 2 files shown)


optimism/src/mips/witness.rs
```diff
@@ -47,7 +47,7 @@ pub struct Env<Fp> {
     pub instruction_parts: InstructionParts<u32>,
     pub instruction_counter: usize,
     pub memory: Vec<(u32, Vec<u8>)>,
```
querolita, line 49: Wasn't the content of the memory and registers of type `Variable`? Or does this casting only happen inside the interpreter?

querolita (approved):

dannywillems (approved):

optimism/src/mips/witness.rs
mrmr1993: The `Variable` type is only in the interpreter, yea.

PR #1384, opened 2023-12-05
Expose `InterpreterEnv::instruction_counter`

mrmr1993:
This PR builds upon https://github.com/o1-labs/proof-systems/pull/1383, making the instruction counter available for use in the VM interpreter.

Currently, the counter is hard-coded to `u32`. This limit will not be big enough, but should be fine until we reach full e2e testing.

### DIFF
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -45,7 +45,7 @@ impl SyscallEnv {
 
 pub struct Env<Fp> {
     pub instruction_parts: InstructionParts<u32>,
-    pub instruction_counter: usize,
+    pub instruction_counter: u32, // TODO: u32 will not be big enough..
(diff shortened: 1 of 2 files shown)


optimism/src/mips/witness.rs
```diff
@@ -435,6 +439,8 @@ impl<Fp: Field> Env<Fp> {
         }
 
         interpreter::interpret_instruction(self, opcode);
+
+        self.instruction_counter += 1;
```
querolita, line 443: So all steps are instructions and all instructions increase the instruction counter.

Actually I was thinking about modeling the Keccak hash function as a `KType` instruction that was triggered by the `RTypeInstruction::SyscallWritePreimage`, but then KType's should not call step. Just a thought for myself.

querolita (approved):

dannywillems (approved):

PR #1386, opened 2023-12-05
Add safe helpers for reading and writing memory

mrmr1993:
This PR adds helpers for safely reading and writing memory. This PR builds upon https://github.com/o1-labs/proof-systems/pull/1385 and its ancestors, emitting lookups for the RAMLookup read/write operations.

This PR also adds a stub `range_check64` helper; this is easy to fill in later, and orthogonal to the main body of the PR.

PR #1387, opened 2023-12-05
Add safe helpers for reading and writing registers in the zkVM

mrmr1993:
This PR mirrors https://github.com/o1-labs/proof-systems/pull/1386, but for registers instead of RAM.

This current implementation comes with some caveats:
* a register may be accessed multiple times in the course of an operation, and so we need to handle the accesses more deliberately than with memory. This is not included in this PR.
* the zero register is handled incorrectly. Writing to it will change the value of the register.
  - This requires the `numerator` in the register to be a variable, which is out of scope for this PR.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -196,6 +196,7 @@ pub enum ITypeInstruction {
 #[derive(Copy, Clone, Debug)]
 pub enum LookupTable {
     MemoryLookup,
+    RegisterLookup,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


optimism/src/mips/interpreter.rs
```diff
@@ -451,10 +555,10 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
             let rs = env.get_instruction_part(InstructionPart::RS);
             let rt = env.get_instruction_part(InstructionPart::RT);
             debug!("Fetching register: {:?}", rs);
-            let register_rs = env.fetch_register_checked(&rs);
+            let register_rs = env.read_register(&rs);
             let immediate = env.get_immediate();
             let res = register_rs + immediate;
-            env.overwrite_register_checked(&rt, &res);
+            env.write_register(&rt, res);
```
querolita, line 561: I like how this reads better

querolita (approved):

dannywillems (approved):

PR #1392, opened 2023-12-05
Respect the 'branch delay slot' in the zkVM

mrmr1993:
The MIPS architecture has a branch delay slot: every jump is executed *2 instructions later*, rather than 1. This PR builds upon https://github.com/o1-labs/proof-systems/pull/1390, updating the instruction pointer logic to handle this correctly.


dannywillems (approved):
I'll approve (modulo broken test :)). I'm curious about this. I thought initially we were keeping the next IP because of pseudo instructions and the branch instructions that require some computations with the IP and some parts of the instruction. I thought we could do without it. Let's see!

PR #1394, opened 2023-12-05
Basic implementation of sll

mrmr1993:
This PR implements the `sll` instruction. This only implements the basic witness logic, and does not add any constraints.

This PR builds upon https://github.com/o1-labs/proof-systems/pull/1393, which adds parsing for rtype instructions.

PR #1395, opened 2023-12-05
Add implementation of `jr` instruction

mrmr1993:
This PR implements the `jr` instruction. This builds on top of https://github.com/o1-labs/proof-systems/pull/1394.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -466,7 +466,7 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
         let pos = env.alloc_scratch();
         unsafe { env.bitmask(&instruction, 32, 26, pos) }
     };
-    let _rs = {
+    let rs = {
         // FIXME: Requires a range check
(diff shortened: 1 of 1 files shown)


dannywillems (approved):
I am a bit annoying, but while running programs in the VM, I liked seeing instructions decoded, like
```
debug!("Instr: jr {}", Env::debug_register(&rs));
```
Is it something you'd agree to add for each instruction?
In addition to some documentation for each?
Can be smth like:
```
            // lw: R[rt] <- M[R[rs] + sign_extended_imm]
            //                -------------------------
            //                         address
            //              ----------------------------
            //                        value
            ```
            
The helpers are in the last PRs I opened :). Might be useful for you to debug too.

dannywillems:
Also, do you have any documentation regarding this next IP?

mrmr1993:
> Do you have any documentation regarding this next IP?

https://www.cs.cmu.edu/afs/cs/academic/class/15740-f97/public/doc/mips-isa.pdf

More importantly, it's what cannon does, and so we should match it if we want a proven version of cannon's execution.

PR #1396, opened 2023-12-05
Add sign extension, use it for existing I-type instructions

mrmr1993:
This PR adds a `sign_extend` helper function to the VM interpreter, and uses it to correct the behaviour of some existing instructions where the high bit of the immediate value is set.

As part of these fixes, this also corrects the behaviour of `Load32`, which was mistakenly attempting to read from the register index as an address.

This PR builds upon https://github.com/o1-labs/proof-systems/pull/1395.


optimism/src/mips/interpreter.rs
```diff
@@ -646,7 +656,8 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
         ITypeInstruction::AddImmediateUnsigned => {
```
dannywillems, line 656: :face_exhaling:

optimism/src/mips/interpreter.rs
```diff
@@ -438,6 +438,15 @@ pub trait InterpreterEnv {
     ) -> Self::Variable;
 
     fn set_halted(&mut self, flag: Self::Variable);
+
+    fn sign_extend(&mut self, x: &Self::Variable, bitlength: u32) -> Self::Variable {
```
dannywillems, line 442: I added in some tests in https://github.com/o1-labs/proof-systems/pull/1371 if you want :)

dannywillems:
Note also the value 0 has two representations. Would be good to import the tests.

dannywillems (approved):
Accepting, but opening follow-ups I would like we address later.

PR #1407, opened 2023-12-05
Add a sign to lookup numerators

mrmr1993:
This PR builds on https://github.com/o1-labs/proof-systems/pull/1405. This adds a `Signed` type (and a `Sign` enum to support it), so that the numerator of a register write can be invalidated by passing a zeroed variable to it.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -157,6 +157,18 @@ pub enum ITypeInstruction {
 }
 
 #[derive(Copy, Clone, Debug)]
+pub enum Sign {
+    Pos,
+    Neg,
+}
+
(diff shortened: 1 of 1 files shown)


dannywillems (approved):
Approving, but curious to see a usecase of:
> register write can be invalidated by passing a zeroed variable to it.

mrmr1993:
> Approving, but curious to see a usecase of:
> 
> > register write can be invalidated by passing a zeroed variable to it.

The zero register is the best example.

That said, it's also useful for folding, where the final instance may have some rows left unused after the program exits. We have to put *something* there, so being able to 'turn off' a row by disabling the lookups (and thus any impact on the execution trace) is appealing.

PR #1413, opened 2023-12-05
Implement `xori`

mrmr1993:


dannywillems (approved):

optimism/src/mips/witness.rs
```diff
@@ -266,6 +266,17 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         res
     }
 
+    unsafe fn xor_witness(
+        &mut self,
+        x: &Self::Variable,
+        y: &Self::Variable,
+        position: Self::Position,
+    ) -> Self::Variable {
+        let res = x ^ y;
```
dannywillems, line 275: I don't like auto dereferencing in Rust... I don't know if that can cause some interesting and subtle bugs.

dannywillems: Oh thanks. That was more a comment than a request, though.

PR #1414, opened 2023-12-05
Implement `lb`

mrmr1993:


optimism/src/mips/interpreter.rs
```diff
@@ -926,7 +926,19 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
...
             return;
         }
-        ITypeInstruction::Load8 => (),
+        ITypeInstruction::Load8 => {
+            let base = env.read_register(&rs);
+            let dest = rt;
+            let offset = env.sign_extend(&immediate, 16);
+            let addr = base + offset;
+            let v0 = env.read_memory(&addr);
+            let value = env.sign_extend(&v0, 8);
```
dannywillems, line 935: Are we sure it does set all first 24 bits to 0?

dannywillems (commented):
I think sign_extend sets to first bits to 1, no? We want to load in `dest`:
`0...0xxxxxxxx` where `xxxxxxxx` are the 8 bits of `rs + offset`

mrmr1993:
`sign_extend` takes the highest bit in the input and repeats it in the higher bits, preserving both the value and sign. E.g.
```
5i8 = 00000101b
sign_extend((5i8 as u8) as u32, 8) = 00000000000000000000000000000101b = 5i32
-5i8 = 0b11111011
sign_extend((-5i8 as u8) as u32, 8) = 0b11111111111111111111111111111011 = -5i32
```

PR #1415, opened 2023-12-05
Implement `srl`

mrmr1993:


dannywillems (approved):
Interesting that overflows are not documented. Do you know?

mrmr1993:
> Interesting that overflows are not documented. Do you know?

I don't know what this means.

PR #1417, opened 2023-12-05
Implement `sltiu`

mrmr1993:


optimism/src/mips/interpreter.rs
```diff
@@ -949,7 +949,19 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
             return;
         }
         ITypeInstruction::SetLessThanImmediate => (),
-        ITypeInstruction::SetLessThanImmediateUnsigned => (),
+        ITypeInstruction::SetLessThanImmediateUnsigned => {
```
dannywillems, line 952: Another unsigned in the name which works with signed value!! 
image

dannywillems (approved):

optimism/src/mips/interpreter.rs
mrmr1993: ?

dannywillems: Agree for the offset. It might be only me who thought that having `Unsigned` in the name would make clear that all computations, even the offset, would be unsigned :').

PR #1420, opened 2023-12-05
Add missing parsing for `bltz` and `bgez`

mrmr1993:
These opcodes fall under the special `regimm` category, and were missed when transferring over the opcode parsing.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -134,6 +134,8 @@ pub enum ITypeInstruction {
     BranchNeq,                    // bne
     BranchLeqZero,                // blez
     BranchGtZero,                 // bgtz
+    BranchLtZero,                 // bltz
+    BranchGeqZero,                // bgez
     AddImmediate,                 // addi
     AddImmediateUnsigned,         // addiu
     SetLessThanImmediate,         // slti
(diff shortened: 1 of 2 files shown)


optimism/src/mips/interpreter.rs
```diff
@@ -134,6 +134,8 @@ pub enum ITypeInstruction {
     BranchNeq,                    // bne
     BranchLeqZero,                // blez
     BranchGtZero,                 // bgtz
+    BranchLtZero,                 // bltz
```
dannywillems, line 137: image
there are more apparently

optimism/src/mips/witness.rs
```diff
@@ -461,6 +461,14 @@ impl<Fp: Field> Env<Fp> {
                         panic!("Unhandled instruction {:#X}", instruction)
                     }
                 },
+                0x01 => {
+                    // RegImm instructions
+                    match (instruction >> 16) & 0x1F {
```
dannywillems, line 466: Ok... took me some seconds to get it....

dannywillems (commented):
LGTM, but more to come.

optimism/src/mips/interpreter.rs
mrmr1993: They are also unused / unimplemented by cannon.

dannywillems (approved):

PR #1427, opened 2023-12-06
Implement `divu`

mrmr1993:


dannywillems (approved):
LGTM. Crazy no exception is raised...

mrmr1993:
> Crazy no exception is raised...

Cannon doesn't guard these, so doing a division by 0 will crash cannon. As such, I was just going to make it unprovable :)

dannywillems:
> > Crazy no exception is raised...
> 
> Cannon doesn't guard these, so doing a division by 0 will crash cannon. As such, I was just going to make it unprovable :)

Opened follow-up issue: https://github.com/o1-labs/proof-systems/issues/1447

PR #1429, opened 2023-12-06
Implement `mul`

mrmr1993:


optimism/src/mips/interpreter.rs
```diff
@@ -899,7 +912,19 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
...
-        RTypeInstruction::MultiplyToRegister => (),
+        RTypeInstruction::MultiplyToRegister => {
+            let rs = env.read_register(&rs);
+            let rt = env.read_register(&rt);
+            let res = {
+                // FIXME: Constrain
+                let pos = env.alloc_scratch();
+                unsafe { env.mul_signed_witness(&rs, &rt, pos) }
+            };
+            env.write_register(&rd, res);
```
dannywillems, line 923: Do you have the spec for MUL?
I would have expected to write in HI/LO.

dannywillems (commented):
See comment

optimism/src/mips/interpreter.rs
mrmr1993: You're thinking of `mult`, it's a different instruction.

mrmr1993: Oh no, *I'm* thinking of the wrong instruction. Lemme fix.

mrmr1993: Actually, wrong again. This `MultiplyToRegister` is indeed correct :man_facepalming:

dannywillems (approved):

PR #1433, opened 2023-12-06
Implement `SyscallReadOther`

mrmr1993:


optimism/src/mips/interpreter.rs
```diff
@@ -784,8 +785,33 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
...
+            let mut check_equal = |expected_fd_id: u32| {
+                // FIXME: Requires constraints
+                let pos = env.alloc_scratch();
+                unsafe { env.test_zero(&(fd_id.clone() - Env::constant(expected_fd_id)), pos) }
+            };
+            let is_stdin = check_equal(FD_STDIN);
+            let is_preimage_read = check_equal(FD_PREIMAGE_READ);
+            let is_hint_read = check_equal(FD_HINT_READ);
+
+            // FIXME: Should assert that `is_preimage_read` and `is_hint_read` cannot be true here.
```
querolita, line 801: do you mean using `env.test_zero` or just a boolean check with normal asserts?

optimism/src/mips/interpreter.rs
```diff
@@ -784,8 +785,33 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
...
+            let other_fd = Env::constant(1) - is_stdin - is_preimage_read - is_hint_read;
+
+            // We're either reading stdin, in which case we get `(0, 0)` as desired, or we've hit a
+            // bad FD that we reject with EBADF.
+            let v0 = other_fd.clone() * Env::constant(0xFFFFFFFF);
+            let v1 = other_fd * Env::constant(0x9); // EBADF
```
querolita, line 807: What's does this 0x9 stand for?

querolita (commented):
Looks good, left some comments. But I couldn't find any resource explaining what `SyscallReadOther` is meant to do, so I do not feel confident enough to approve yet.

optimism/src/mips/interpreter.rs
mrmr1993: I mean by adding a constraint, which is not exposed in the API yet.

optimism/src/mips/interpreter.rs
```diff
@@ -784,8 +785,33 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
         RTypeInstruction::JumpAndLinkRegister => (),
         RTypeInstruction::SyscallMmap => (),
         RTypeInstruction::SyscallExitGroup => (),
+        RTypeInstruction::SyscallReadHint => (),
         RTypeInstruction::SyscallReadPreimage => (),
-        RTypeInstruction::SyscallReadOther => (),
+        RTypeInstruction::SyscallReadOther => {
+            let fd_id = env.read_register(&Env::constant(4));
```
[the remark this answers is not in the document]
mrmr1993: My personal preference is that we don't define a constant that's local and used once in one place. This constant 4 is 4 because the read syscall reads register 4 to find out the FD that it's reading from.
```rust
const REGISTER_FD_ID: u32 = 4;
let fd_id = env.read_register(&Env::constant(REGISTER_FD_ID));
```
is just forcing us to write the same thing 3 times instead of 1.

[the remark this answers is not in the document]
querolita: Okay if it only gets used once. But if it becomes something we write over and over again it could start making sense.

optimism/src/mips/interpreter.rs
querolita: like, it stands for `error bad file`? I just felt like it was quite arbitrary `0x9` and didn't know where it came from.

querolita (approved):

PR #1443, opened 2023-12-06
Add `lookups` vector to `Env` struct

querolita:
This PR includes a vector of lookups inside the `Env` struct. It is used by the `add_lookup` function inside the `InterpreterEnv` trait. 

The type, however, is not clear to me. Shouldn't it be `Lookup<Fp>` instead of `Lookup<Self::Variable>`? Specially when `type Variable = u32`.


querolita:
Closing because the MIPS line of work is dealing with it

PR #1444, opened 2023-12-06
Keccak columns for MIPS zkVM

querolita:
This PR adds a preliminary layout of the Keccak-related columns by adding the variant `KeccakState(KeccakColumn)` inside `Column` and then `KeccakColumn` holds all variants of Keccak columns.

This adds no constraints whatsoever. It just integrates some needed notation in the new format.

Optimizations after constraints are ready: reuse column indices for Round and Sponge, so it does not add more columns unnecessarily (these are mutually exclusive modes of operation depending on the flags values).


optimism/src/mips/keccak/mod.rs
```diff
@@ -0,0 +1,67 @@
+pub mod column;
+
+pub const ZKVM_KECCAK_COLS: usize = 1965 + 6;
+pub const RATE: usize = 1088;
+pub const RATE_IN_BYTES: usize = RATE / 8;
+pub const DIM: usize = 5;
+pub const QUARTERS: usize = 4;
+
+fn grid_20(x: usize, q: usize) -> usize {
+    q + QUARTERS * x
```
mrmr1993, line 10: Please either inline the constant here or change the name. Personally I think it would be clearer to have
```rust
fn matrix_2d_4(x: usize, y: usize) -> usize {
    y + 4 * x
}
```
and similarly
```rust
fn matrix_3d_4_5(x: usize, y: usize, z: usize) -> usize {
    z + 4 * (y + 5 * x)
}
```
for both of the below.

optimism/src/mips/keccak/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+fn grid_80(i: usize, x: usize, q: usize) -> usize {
+    q + QUARTERS * (x + DIM * i)
+}
+
+fn grid_100(y: usize, x: usize, q: usize) -> usize {
```
mrmr1993, line 17: This function is the same as above, please don't name according to caller semantics.

mrmr1993 (commented):
LGTM, but see comments

optimism/src/mips/keccak/mod.rs
```diff
@@ -0,0 +1,67 @@
+pub mod column;
```
mrmr1993, line 1: On further reflection, shouldn't this be a separate module in `optimism/src` instead of living in MIPS?

querolita: Sounds good, will move it outside

optimism/src/mips/keccak/mod.rs
querolita: Yeah, it is the same but I wanted to have different indices names to facilitate readability of cells afterwards in the constraints. Meaning, `i` ranges `[0..4)` and `y` ranges `[0..5)`, so using the same parameter name could be confusing.

PR #1445, opened 2023-12-06
A check only dummy value looked up, without XOR table

dannywillems:


kimchi/src/tests/lookup.rs
```diff
@@ -582,3 +582,50 @@ fn test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_v
...
+    let seed: [u8; 32] = thread_rng().gen();
+    eprintln!("Seed: {:?}", seed);
+    let mut rng = StdRng::from_seed(seed);
+    let max_len: u32 = 100u32;
+    let max_table_id: i32 = 100;
+
+    // No zero-length table
+    let len = 1u32 + rng.gen_range(0u32..max_len);
+    // No table of ID 0
+    let table_id: i32 = 1 + rng.gen_range(0i32..max_table_id);
```
volhovm, line 597: Remind me: the `RANGE_CHECK` table will not be included because there is no lookup gate into it? This is why `table_id == 1` will work here and won't collide with `RANGE_CHECK_TABLE_ID`?

kimchi/src/tests/lookup.rs
```diff
@@ -582,3 +582,50 @@ fn test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_v
 
     setup_successfull_runtime_table_test(vec![cfg], vec![runtime_table], lookups);
 }
+
+#[test]
```
volhovm, line 586:
```suggestion
#[test]
/// Performs a bunch of lookups into a table with (1) non-zero id, (2) non-zero keys and values. Assumes that 
/// the xor lookup is disabled. Looks up the all-zeroes XOR dummy value.
```

kimchi/src/tests/lookup.rs
```diff
@@ -582,3 +582,50 @@ fn test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_v
...
+    let max_len: u32 = 100u32;
+    let max_table_id: i32 = 100;
+
+    // No zero-length table
+    let len = 1u32 + rng.gen_range(0u32..max_len);
+    // No table of ID 0
+    let table_id: i32 = 1 + rng.gen_range(0i32..max_table_id);
+    // No index 0 in the table.
+    let indices: Vec<Fp> = (0..len)
+        .map(|_| 1 + rng.gen_range(0u32..max_len))
```
volhovm, line 600:
```suggestion
        .map(|_| rng.gen_range(1u32..max_len))
```
For uniformity with how `values` are computed.

volhovm (approved):

kimchi/src/tests/lookup.rs
dannywillems: Correct.

kimchi/src/tests/lookup.rs
dannywillems: I made the test name explicit.

PR #1450, opened 2023-12-06
Remove register names

mrmr1993:
This PR removes the list of register names from the VM.

The names clutter the interpreter file, and the semantic meanings are meant for users of MIPS (i.e. they provide a standard for how variables should be used so that functions etc. compose cleanly), but they only serve as a distraction/confusion for the VM implementation itself.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -18,66 +18,6 @@ pub const SYSCALL_READ: u32 = 4003;
 pub const SYSCALL_WRITE: u32 = 4004;
 pub const SYSCALL_FCNTL: u32 = 4055;
 
-// Source: https://www.doc.ic.ac.uk/lab/secondyear/spim/node10.html
-// Reserved for assembler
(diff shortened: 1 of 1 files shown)


optimism/src/mips/interpreter.rs
```diff
@@ -18,66 +18,6 @@ pub const SYSCALL_READ: u32 = 4003;
 pub const SYSCALL_WRITE: u32 = 4004;
 pub const SYSCALL_FCNTL: u32 = 4055;
 
-// Source: https://www.doc.ic.ac.uk/lab/secondyear/spim/node10.html
-// Reserved for assembler
-pub const REGISTER_AT: u32 = 1;
```
querolita, line 23: But wouldn't this complicate readability whenever the interpreter has to use them? I find it clearer if these arbitrary numbers have a constant name that I can relate to, rather than just the plain values.

mrmr1993: The names don't help with readability, they're just actively confusing. These names are how the user should think about them; we're not the user.

PR #1454, opened 2023-12-06
[develop] Lookup: a table with ID 0 *must* contain a row with 0 only

dannywillems:
Regression test for https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/tables/mod.rs#L85

### DIFF
--- a/kimchi/src/tests/lookup.rs
+++ b/kimchi/src/tests/lookup.rs
@@ -583,6 +583,55 @@ fn test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_v
     setup_successfull_runtime_table_test(vec![cfg], vec![runtime_table], lookups);
 }
 
+// This test verifies that if there is a table with ID 0, it contains a row with only zeroes.
+// This is to enforce the constraint we have on the so-called "dummy value".
+// FIXME: see https://github.com/o1-labs/proof-systems/issues/1460
+// We should test the error message, "expected" argument of the macro won't be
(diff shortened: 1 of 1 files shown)


kimchi/src/tests/lookup.rs
```diff
@@ -582,3 +582,98 @@ fn test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_v
...
+    // lookups with (0, 0, 0).
+    let witness = array::from_fn(|_col| vec![Fp::zero(); num_lookups]);
+
+    let err = TestFramework::<Vesta>::default()
+        .gates(gates)
+        .witness(witness)
+        .lookup_tables(lookup_tables)
+        .setup()
+        .prove::<BaseSponge, ScalarSponge>()
+        .unwrap_err();
```
mrmr1993, line 630: It's not the proving that fails, it's the setup. You'll need to update your test, and perhaps the framework to have a `setup_err` helper

dannywillems: Follow-up issue: https://github.com/o1-labs/proof-systems/issues/1460.
I will add the macro `should_panic` atm, and link with this issue.

volhovm (approved):
LGTM, the test is the new variant of the one here (that was reverted):
https://github.com/o1-labs/proof-systems/pull/1263/files#diff-8e3bb71e9bda46ef09ab12111bd3c07d8d20b4185fc96f195a2d8a25d7d865c7R321

PR #1455, opened 2023-12-06
Lookup: add more documentation regarding the dummy value

dannywillems:


kimchi/src/circuits/lookup/index.rs
```diff
@@ -394,6 +394,15 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
                 }
 
                 //~ 6. Pad the end of the concatened table with the dummy value.
+                //     By padding with 0, we constraint the table with ID 0 to
```
volhovm, line 397: The comment/semantics makes a lot of sense to me, but I'm not 100% sure. I think an alternative (but subjectively more complicated) semantics would be to just throw an error, and let user add the fake value explicitly? 

Maybe a quick "yes" from @mrmr1993 would be really helpful.

volhovm (approved):
LGTM but would ideally want a look from Matthew.

kimchi/src/circuits/lookup/index.rs
```diff
@@ -394,6 +394,15 @@ impl<F: PrimeField + SquareRootField> LookupConstraintSystem<F> {
...
+                //     have a zero entry.
+                //     This is for the rows which do not have a lookup selector,
+                //     see ../../../../book/src/kimchi/lookup.md.
+                //     The zero entry row is contained in the built-in XOR table.
+                //     An error is raised when creating the CS if a user-defined
+                //     table is defined with ID 0 without a row contain zeroes.
```
dannywillems, line 403:
```suggestion
                //     table is defined with ID 0 without a row containing zeroes.
```

PR #1457, opened 2023-12-06
Fix 1362: add the dummy entry in the domain size count

### LINKED ISSUE #1362: Flaky test: tests::lookup::test_runtime_table_only_one_table_with_id_zero_with_non_zero_entries_random_values
https://github.com/o1-labs/proof-systems/actions/runs/7047117196/job/19180286206

dannywillems:
Fix #1362

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -736,20 +736,25 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
 
         let lookup_domain_size = {
             // First we sum over the lookup table size
+            let mut has_table_with_id_0 = false;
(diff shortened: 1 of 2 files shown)


kimchi/src/circuits/constraints.rs
```diff
@@ -736,20 +736,25 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
 
         let lookup_domain_size = {
             // First we sum over the lookup table size
+            let mut has_table_with_id_0 = false;
```
dannywillems, line 739: Taste: I don't like using mutable values, but it avoids iterating a second time in the vector. We could use a `any`.

querolita (approved):

PR #1468, opened 2023-12-07
Include basic operations for constraints as used in mips-demo

querolita:
This PR adds the type `E<F>` and ports the functions `boolean`, `combine`, and `curr_cell`.

I thought we might not be using `next_cell` after all, but it including it again is straightforward.


querolita:
Closing for now as the code I am using for Keccak won't use these eventually

PR #1481, opened 2023-12-07
Implement `sh`

mrmr1993:


dannywillems:
Arg! Fixing the test!

dannywillems:
It is a flaky test, re-running should be ok. Fix in https://github.com/o1-labs/proof-systems/pull/1484

PR #1489, opened 2023-12-08
Resurrect lookup PR1263 [do not merge]

volhovm:
This resurrects https://github.com/o1-labs/proof-systems/pull/1263 which was reverted in https://github.com/o1-labs/proof-systems/pull/1378 due to the bug that is to be understood within this PR.

Important! Note that the first commit /still/ has the bug MinaProtocol/mina#14657. It will be fixed in the following commits explicitly, so that the bug can be hunted down.

Plan before this can be re-merged:
- [X] Resurrect the PR as it is, squashing it all into a single commit, resolve conflicts
- [X] Check that the `zero_row_zero_id` still makes sense
- [X] Explicitly revert the "colliding runtime+lookup ids" behaviour.
- [ ] Understand why the panic-within-panic bug happened in `mina`
- [ ] Make desired changes, confirm the bug does not reproduce anymore, cover it with a test if possible

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -1648,8 +1648,8 @@ If lookup is used, the following values are added to the common index:
 To create the index, follow these steps:
 
 1. If no lookup is used in the circuit, do not create a lookup index
-2. Get the lookup selectors and lookup tables (TODO: how?)
-3. Concatenate runtime lookup tables with the ones used by gates
+2. Get the lookup selectors and lookup tables that are specified implicitly
+3. Concatenate explicit runtime lookup tables with the ones (implicitly) used by gates.
 4. Get the highest number of columns `max_table_width`
    that a lookup table can have.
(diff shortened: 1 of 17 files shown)


volhovm:
@dannywillems I confirm that the `test_zero_table_zero_row` test that is resurrected within this PR captures the "nonzero row with zero entry" scenario we discussed earlier. In this scenario, the error must be raised, but it is not since `has_zero_entry` in current `master` returns false positives (current resurrected PR fixes this).

For the context, the scenario could be alternatively captured by adding such a row into https://github.com/o1-labs/proof-systems/pull/1454/files#diff-30d4661dcab0b5e55e4bb6f27ea2c313bded3d2fc088df9d16f6d0fdde981cc1R603 (by either adding zero into indices, or into values, but not into both).

volhovm:
Most chunks of these were merged independently. What was not resurrected is creating a private `LookupTable` constructor, but that affects too much code and is not very useful, so low priority.

PR #1490, opened 2023-12-08
Handle SIGINT from parent to terminate child process

### LINKED ISSUE #1487: Close children processes gracefully upon SIGINT (and other termination signals)
I don't know if it is expected, but we get:
```
t=2023-12-07T19:34:31+0100 lvl=eror msg="failed to close pre-image server" err="signal: interrupt"
```
when stopping the zkVM with `CTRL-C`.

Not urgent IMHO, and definitely not in the scope of this PR, but would be nice to open a follow-up.

_Originally posted by @dannywillems in https://github.com/o1-labs/proof-systems/issues/1469#issuecomment-1845941574_

rbonichon:
Should fix #1487

### DIFF
--- a/optimism/src/main.rs
+++ b/optimism/src/main.rs
@@ -21,6 +21,10 @@ use mina_poseidon::{
     sponge::{DefaultFqSponge, DefaultFrSponge},
 };
 
+use log::debug;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
 type Fp = ark_bn254::Fr;
 type SpongeParams = PlonkSpongeConstantsKimchi;
 type BaseSponge = DefaultFqSponge<ark_bn254::g1::Parameters, SpongeParams>;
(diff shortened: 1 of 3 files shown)


dannywillems:
Is it still required? @rbonichon

rbonichon:
> Is it still required? @rbonichon

Guess so for it's still not handled.

optimism/src/main.rs
```diff
@@ -77,6 +85,10 @@ pub fn main() -> ExitCode {
 
     let mut env = mips_witness::Env::<ark_bn254::Fr>::create(cannon::PAGE_SIZE as usize, state, po);
 
+    // Install signal catcher
+    let term = Arc::new(AtomicBool::new(false));
+    signal_hook::flag::register(signal_hook::consts::SIGTERM, Arc::clone(&term)).unwrap();
```
volhovm, line 90: This seems exactly equivalent to the code on lines 68-70, I assume that's not on purpose?

optimism/src/main.rs
```diff
@@ -124,7 +136,7 @@ pub fn main() -> ExitCode {
             next: std::array::from_fn(|_| Vec::with_capacity(domain_size)),
         };
 
-    while !env.halt {
+    while !term.load(Ordering::Relaxed) && !env.halt {
```
volhovm, line 139: I'm not exactly sure I understand what's happening... I assume after registering the hook, SIGINT will not work as usual, but will instead just modify the mutable variable? 
- In this case: Is it intended that the Keccak proof generation (lines 256-270) do not depend on `term`? Will that part be interrupted by SIGINT?

volhovm (commented):
Could you please clarify the logic? I'm unsure whether if all the parts of `main` are properly interrupted by SIGINT.

dannywillems:
Closing as old and stale.

PR #1504, opened 2023-12-10
Implement preimage write syscall

mrmr1993:


optimism/src/mips/interpreter.rs
```diff
@@ -1060,7 +1061,157 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
...
+                let pos = env.alloc_scratch();
+                unsafe { env.bitmask(&write_length, 2, 0, pos) }
+            };
+            let register_idx = {
+                let registers_left_to_write_after_this = {
+                    // FIXME: Requires a range check
+                    let pos = env.alloc_scratch();
+                    // The virtual register is 32 bits wide, so we can just read 6 bytes. If the
+                    // register has an incorrect value, it will be unprovable and we'll fault.
+                    unsafe { env.bitmask(&write_length, 6, 2, pos) }
```
dannywillems, line 1083: Trying to understand this.

mrmr1993: The write syscall will request 32 bytes, but we want to write aligned-word-wise.

dannywillems (approved):

optimism/src/mips/interpreter.rs
querolita: I wonder if what's being written is the claimed hash output, whereas it might be SyscallReadPreimage what reads any arbitrary length.

Issue #1359, opened 2023-11-30
Add mdbook-linkcheck


dannywillems:
Could you please change it to `[Extended Lookup Tables](../kimchi/extended-lookup-tables.md)`?

Also do we have some mechanism to check broken references? Like https://docs.rs/mdbook-linkcheck/latest/mdbook_linkcheck/

_Originally posted by @volhovm in https://github.com/o1-labs/proof-systems/pull/1346#discussion_r1410394956_

volhovm:
This PR introduces the desired feature: https://github.com/Michael-F-Bryan/mdbook-linkcheck/pull/81.

But even if it's merged we'd need to wait for mdbook-linkcheck to get released if I understand correctly? (or can we use a dev version?)

dannywillems:
> This PR introduces the desired feature: [Michael-F-Bryan/mdbook-linkcheck#81](https://github.com/Michael-F-Bryan/mdbook-linkcheck/pull/81).
> 
> But even if it's merged we'd need to wait for mdbook-linkcheck to get released if I understand correctly? (or can we use a dev version?)

Yes, let's wait for a release.
Is the project still maintained?

volhovm:
It seems maintained. I'll ping `Micheal-F-Bryan` again in a week. It's all done so I'm a bit annoyed it can't be merged only because I'm not getting any replies...

PR #1505, opened 2023-12-11
Book: repair links with mdbook-linkcheck (manually)

volhovm:
PR 1/2 of https://github.com/o1-labs/proof-systems/issues/1359 that for now does a manual pass over the book and fixes link inconsistencies. 

(The future PR 2/2 will actually integrate the checker after https://github.com/Michael-F-Bryan/mdbook-linkcheck/pull/81/is merged)

### DIFF
--- a/book/Makefile
+++ b/book/Makefile
@@ -9,23 +9,20 @@ MDBOOK_LINKCHECK_VERSION = 0.7.7
 MDBOOK_MERMAID_VERSION = 0.12.6
 MDBOOK_TOC_VERSION = 0.14.1
 
-
-all: deps check build serve
-
 #
-# Installs the dependencies required to serve or build the book
+# use `make deps` to install the dependencies required to serve or build the book
 #
 
 deps:
 	cargo install "mdbook@$(MDBOOK_VERSION)"
 	cargo install "mdbook-admonish@$(MDBOOK_ADMONISH_VERSION)"
 	cargo install "mdbook-katex@$(MDBOOK_KATEX_VERSION)"
-	cargo install "mdbook-linkcheck@$(MDBOOK_LINKCHECK_VERSION)"
+	# cargo install "mdbook-linkcheck@$(MDBOOK_LINKCHECK_VERSION)"
 	cargo install "mdbook-mermaid@$(MDBOOK_MERMAID_VERSION)"
 	cargo install "mdbook-toc@$(MDBOOK_TOC_VERSION)"
 
 #
-# Checks if your installed dependencies match what we've listed above
+# use `make check` to check if your installed dependencies match what we've listed above
 #
 
 define check_version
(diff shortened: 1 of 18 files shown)


book/Makefile
```diff
@@ -9,20 +9,23 @@ MDBOOK_LINKCHECK_VERSION = 0.7.7
...
 	cargo install "mdbook-admonish@$(MDBOOK_ADMONISH_VERSION)"
 	cargo install "mdbook-katex@$(MDBOOK_KATEX_VERSION)"
-	# cargo install "mdbook-linkcheck@$(MDBOOK_LINKCHECK_VERSION)"
+	cargo install "mdbook-linkcheck@$(MDBOOK_LINKCHECK_VERSION)"
 	cargo install "mdbook-mermaid@$(MDBOOK_MERMAID_VERSION)"
 	cargo install "mdbook-toc@$(MDBOOK_TOC_VERSION)"
 
 #
-# use `make check` to check if your installed dependencies match what we've listed above
+# Checks if your installed dependencies match what we've listed above
```
querolita, line 28: what command does this now instead? If it is not done automatically, I would keep the comment just in case someone wants to check this manually. Same with the other comments that have been changed in the Makefile.

querolita (approved):
Looks good to me, but take a look at the comment, as I wouldn't like to provide less guidance for users by removing concrete explanations.

book/Makefile
volhovm: The same command below, `make check`. It's just that everything in makefile was duplicated, there was a target "`mytarget`" and a comment "use `make mytarget` to ...".

If it clarified but not confused things I can return it back?

dannywillems (changes requested):
Please do a PR with only link fixes.
`make serve` does not work locally because of the newly added configuration for linkcheck, which, I suppose, is fixed by your PR upstream.
It breaks local compilation of the documentation.

book/book.toml
```diff
@@ -33,3 +33,10 @@ command = "mdbook-mermaid"
 [preprocessor.toc]
 command = "mdbook-toc"
 renderer = ["html"]
+
+
+# See docs: https://github.com/Michael-F-Bryan/mdbook-linkcheck
```
dannywillems, line 38: Revert. Waiting for PR to be approved and new released in dep.

volhovm:
@dannywillems reverted the `Makefile` related changes.

book/src/specs/kimchi.md
```diff
@@ -2225,9 +2226,9 @@ The prover then follows the following steps to create the proof:
 1. Evaluate the same polynomials without chunking them
    (so that each polynomial should correspond to a single value this time).
 1. Compute the ft polynomial.
-   This is to implement [Maller's optimization](https://o1-labs.github.io/mina-book/crypto/plonk/maller_15.html).
+   This is to implement [Maller's optimization](https://o1-labs.github.io/mina-book/kimchi/maller_15.html).
```
dannywillems, line 2229: I would use the local version.

book/src/specs/kimchi.md
```diff
@@ -2225,9 +2226,9 @@ The prover then follows the following steps to create the proof:
...
 1. construct the blinding part of the ft polynomial commitment
-   [see this section](https://o1-labs.github.io/mina-book/crypto/plonk/maller_15.html#evaluation-proof-and-blinding-factors)
+   [see this section](https://o1-labs.github.io/mina-book/kimchi/maller_15.html#evaluation-proof-and-blinding-factors)
```
dannywillems, line 2231: Same here

dannywillems (approved):
LGTM. Merging. You can address, if you agree with, the comments in a follow-up PR.

PR #1508, opened 2023-12-11
Boolean trait for the Keccak environment

querolita:
This PR adds a few basic boolean operations for the Keccak environment (that will be used inside constraints). It updates the `KeccakEnvironment` trait to use them in flags.


mrmr1993 (approved):
Very clean, nicely done!

PR #1301, opened 2023-10-26
Single-row Optimizations on top of KeccakRound gate

querolita:
These optimizations still keep the `KeccakRound` gate in one single row. Other optimizations will build on top of this one to split across rows to create narrower chips. Here, by "Old" I mean the design shown in the [RFC](https://github.com/o1-labs/rfcs/pull/27/). 

- Remove intermediate states and embed them in constraints
- Simplify rotation inside Theta removing bound_c and lookups for quotient_c (which just needs a boolean check)
- Remove bound_e following the discussion about optimizing rotation with the product team

Old column count: 2344
New column count: 1965

Old constraint count: 754
New constraint count: 389

Old lookup count: 1760
New lookup count: 1620

Old prover time: 23s x 1 hash | 1871s x 1000 hash
New prover time: 19s x 1 hash | 1691s x 1000 hash

Old setup time: 57s x 1 hash | 517s x 1000 hash
New setup time: 57s x 1 hash | 479s x 1000 hash

Old verifier time: 6516ms x 1 hash | 6797ms x 1000 hash
New verifier time: 5600ms x 1 hash | 6021ms x 1000 hash

PR #1511, opened 2023-12-11
Constraints for Keccak gate (without lookups)

querolita:
This PR brings constraints for the keccak gate for the mips demo repo, as proposed in the 1-row optimization of 1965 columns ([here](https://github.com/o1-labs/proof-systems/pull/1301)). 

It also includes constraints for the values of the selectors (flags) to decide whether it is a sponge or a round type of gate, which were not included in the above PR as this was encoded as gate coefficients before.


querolita:
Closing because I will be splitting this PR across smaller chunks

PR #1518, opened 2023-12-11
Extension to frequency specification

rbonichon:
Allow to define a range in a `rust`-like manner `lo..[hi]`, that is from `lo` until `hi` (excluded) if specified, or until the end if not.


mrmr1993 (approved):
:heart:

PR #1520, opened 2023-12-12
zkVM: implement Page encoder

dannywillems:


dannywillems:
This doesn't work:
```
+
+        let value: &str = r#"{"index": 0,"data":"eJzswAENAAAAwiD7p7bHBwMAAADyHgAA//8QAAAB"}"#;
+        let decoded_page: Page = serde_json::from_str(value).unwrap();
+        let res = serde_json::to_string(&decoded_page).unwrap();
+        assert_eq!(res, value);
```
Investigating.

dannywillems:
ok, found. It is the compression level. It is default in Go, not in Rust.

dannywillems:
I'm giving up at the moment the fix. It should be fine as it is decoded to the same value.

mrmr1993 (approved):

PR #1523, opened 2023-12-12
zkVM: add snapshot-state-at option

dannywillems:
Good to go. The test below doesn't affect the behavior of decoding.
The file can be loaded into Cannon also.

My tests are: run until a certain step with the default state.json, take a snapshot regularly, look at step N what the instruction and the debug line, stop, reload with a snapshot at before (say `N - M1`), check the log is the same, redo with a bunch of `M_(i + 1) > M_(i)`

-----

Still in progress. Ready for partial review. I am not sure it comes from this PR. Trying to fix this diff:
```shell
➜  optimism git:(dw/dump-state-cli) ✗ jq '.memory[0]' dump-state-0.json
{
  "index": 0,
  "data": "eJztwAENAAAAwiD7p7bHBwMAAADyDhAAAAE="
}
➜  optimism git:(dw/dump-state-cli) ✗ jq '.memory[0]' state.json       
{
  "index": 0,
  "data": "eJzswAENAAAAwiD7p7bHBwMAAADyHgAA//8QAAAB"
}
```


dannywillems:
I'm gonna rename in `snapshot-state-at`.

optimism/src/mips/witness.rs
```diff
@@ -807,6 +811,49 @@ impl<Fp: Field> Env<Fp> {
         None
     }
 
+    fn snapshot_state_at(&mut self, at: &StepFrequency) {
+        if self.should_trigger_at(at) {
+            let filename = format!("snapshot-state-{}.json", self.instruction_counter);
```
rbonichon, line 816: Minor quip: shouldn't the format follow the `snapshot-fmt` CLI argument?

rbonichon (approved):
LGTM. There's a possible improvement that I would like to see but I'm happy if this is a follow-up issue as well.

optimism/src/mips/witness.rs
dannywillems: Good idea, but I couldn't see how it was used. It seems it isn't used, right?

PR #1525, opened 2023-12-12
Dw/serializer page additional test

dannywillems:
The encoder parameters should be different. It seems that the go library doesn't use the same default than the rust. However, the decoding is correct, which is what we expect.


volhovm:
@dannywillems this PR is 2 months old, is it still WIP?

dannywillems:
Let's close it. I couldn't find the right parameters in the libflate library. Not critical to have.

PR #1528, opened 2023-12-13
Print the current instruction and step on halt

mrmr1993:
This PR aims to help us debug when the zkVM halts unexpectedly, by printing the offending instruction and step number so that we can reproduce it.


dannywillems:
Can you target master please? @mrmr1993

dannywillems (approved):

PR #1531, opened 2023-12-13
Use an environment variable to snapshot the state into

dannywillems:
Using the VM configuration might be better. Let me know what you think.

### DIFF
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -813,7 +813,12 @@ impl<Fp: Field> Env<Fp> {
 
     fn snapshot_state_at(&mut self, at: &StepFrequency) {
         if self.should_trigger_at(at) {
-            let filename = format!("snapshot-state-{}.json", self.instruction_counter);
+            let snapshot_directory =
+                std::env::var("ZKVM_SNAPSHOT_DIRECTORY").unwrap_or(".".to_string());
+            let filename = format!(
+                "{}/snapshot-state-{}.json",
+                snapshot_directory, self.instruction_counter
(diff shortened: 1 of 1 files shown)


optimism/src/mips/witness.rs
```diff
@@ -813,7 +813,12 @@ impl<Fp: Field> Env<Fp> {
 
     fn snapshot_state_at(&mut self, at: &StepFrequency) {
         if self.should_trigger_at(at) {
-            let filename = format!("snapshot-state-{}.json", self.instruction_counter);
+            let snapshot_directory =
+                std::env::var("ZKVM_SNAPSHOT_DIRECTORY").unwrap_or(".".to_string());
```
rbonichon, line 817: There is a command-line argument for that as well (see https://github.com/o1-labs/proof-systems/pull/1523#discussion_r1425375789)

rbonichon: Well not really for the directory parameter.

rbonichon (commented):
Why not but shouldn't this be handled through the `snapshot-fmt` which would recursively create directory on need?

optimism/src/mips/witness.rs
dannywillems: I totally agree, it is far better with the parameter. I don't see how to use it tbh. Why was it added?

rbonichon: It's a Cannon Cli argument, so we need to be able to parse it 😆

dannywillems:
Closing the PR at the moment. Not a priority.

PR #1534, opened 2023-12-13
Split op-program and cannon

dannywillems:
We should be able to run the binaries individually

### DIFF
--- a/optimism/run-cannon.sh
+++ b/optimism/run-cannon.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+make -C ./ethereum-optimism/op-program op-program
 make -C ./ethereum-optimism/cannon cannon
 
 ./ethereum-optimism/cannon/bin/cannon load-elf --path=./ethereum-optimism/op-program/bin/op-program-client.elf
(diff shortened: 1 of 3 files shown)


optimism/run-cannon.sh
```diff
@@ -0,0 +1,26 @@
...
+./ethereum-optimism/cannon/bin/cannon load-elf --path=./ethereum-optimism/op-program/bin/op-program-client.elf
+
+./ethereum-optimism/cannon/bin/cannon run \
+  --pprof.cpu \
+  --info-at "${INFO_AT:-%10000000}" \
+  --proof-at never \
+  --stop-at "${STOP_AT:-never}" \
+  --input "{CANNON_STATE_FILENAME:-./state.json}" \
+  -- \
+  ./ethereum-optimism/op-program/bin/op-program \
```
rbonichon, line 15: I was wondering whether the scripts should assume that `op-program` is in your path - same for `cannon`, etc. I for sure never had it in `./ethereum-optimism` directory

dannywillems: `ethereum-optimism` is initialized as a submodule with a compatible version. We do assume the user will run `git submodule init` and `git submodule update`.
We should run `make -C ethereum-optimism op-program` in the script though. Fixing it.

rbonichon: The `submodule` part is not in the README 👿

PR #1537, opened 2023-12-13
Use write instead of write_all and remove flush

dannywillems:
Flush: does not seem to change anything
Write returns the length that has been written. write_all does not.

### DIFF
--- a/optimism/src/preimage_oracle.rs
+++ b/optimism/src/preimage_oracle.rs
@@ -110,10 +110,10 @@ impl PreImageOracle {
     pub fn get_preimage(&mut self, key: [u8; 32]) -> Preimage {
         let RW(ReadWrite { reader, writer }) = &mut self.oracle_client;
 
-        let r = writer.write_all(&key);
+        let r = writer.write(&key);
         assert!(r.is_ok());
-        let r = writer.flush();
-        assert!(r.is_ok());
+        // We check that 32 bytes have been written
+        assert_eq!(r.unwrap(), 32);
 
         debug!("Reading response");
         let mut buf = [0_u8; 8];


mrmr1993 (commented):
I'm concerned that this is fragile to the buffer size and go's behaviour. Does this fix anything?

Also, I think we still want to flush, to ensure that we don't move on before the go process has read all of the data

rbonichon (commented):
I agree that we could use `write` here. But `flush` is more useful with `write` than `write_all` since `write` may not write everything right now (it can be buffered). 

`write_all` does not return the size since it "writes all". To me, `write` + the `assert_eq!` is basically `write_all`. Is that not so?

dannywillems:
> write_all does not return the size since it "writes all". To me, write + the assert_eq! is basically write_all. Is that not so?

The doc doesn't say it. The doc says it calls write, one by one, until the end of the buffer. It might also be less efficient to call `write_all` as, it seems, to call N times the `write` system call. `write` with a buffer of size N will call only once the syscall `write`. In any case, we *should*, for consistency, check that we actually wrote 32 bytes.

> I agree that we could use write here. But flush is more useful with write than write_all since write may not write everything right now (it can be buffered).

We don't need to flush if we know we exactly wrote the 32 bytes (or N bytes for other calls). The next stream we send to the file will be different, and we will check also the right amount of bytes have been written correctly. Also, rust-analyzer gives me a flush implementation returning simply `Ok(())`.

dannywillems:
> I'm concerned that this is fragile to the buffer size and go's behaviour. Does this fix anything?

The buffer has a fixed-length of 32 bytes. Go will wait for 32 bytes, see [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-preimage/oracle.go#L53). Therefore, it is the exact behavior expected by Go. Also, it will writes the 32 bytes in one-shot, not one by one.

dannywillems:
> Also, I think we still want to flush, to ensure that we don't move on before the go process has read all of the data

When we wrote exactly 32 bytes, the function will continue [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-preimage/oracle.go#L59). Calling flush is unnecessary if we do check we sent 32 bytes. Also, as said [here](https://github.com/o1-labs/proof-systems/pull/1537#issuecomment-1855861761), `flush` doesn't seem to do anything.

mrmr1993 (changes requested):
`libc::write` may choose to write fewer than 32 bytes. I would prefer to close this PR

dannywillems:
> libc::write may choose to write fewer than 32 bytes.

Right, I did not know! Good point. However, we do check if we wrote the 32 bytes.

PR #1541, opened 2023-12-14
[easy] Create Keccak traits

querolita:
This PR collects existing Keccak-related operations located in the `mod.rs` file into meaningful traits for easier reusability inside the MIPS repo (future PRs coming).


optimism/src/keccak/witness.rs
```diff
@@ -0,0 +1,42 @@
+use ark_ff::Field;
```
dannywillems, line 1: I don't know if this is supposed to be here. It is not used and there are some missing primitives in the KeccakEnv like `update_step`.

dannywillems (commented):
See comment about witness.rs. I would suggest to remove it from this PR as it seems to be used later.

optimism/src/keccak/witness.rs
querolita: You're absolutely right, I don't know how this one got in here!

dannywillems (approved):

PR #1542, opened 2023-12-14
Single row Keccak optimizations (without proof-systems modifications)

querolita:
This contains the changes from https://github.com/o1-labs/proof-systems/pull/1301, ignoring the changes to the proof system. 

It gets rid of redundancy in the Keccak layout to reduce the number of columns by ~400. 

- Remove intermediate states and embed them in constraints
- Simplify rotation inside Theta removing bound_c and lookups for quotient_c (which just needs a boolean check)
- Remove bound_e following the discussion about optimizing rotation with the product team
- 64 Less constraints in squeeze, only for the 256-bits of the output instead of the whole state.

The main reason for getting these changes in, is to be able to avoid code duplication between the kimchi crate and the mips crate (mainly regarding witness creation).

### DIFF
--- a/kimchi/src/circuits/polynomials/keccak/circuitgates.rs
+++ b/kimchi/src/circuits/polynomials/keccak/circuitgates.rs
@@ -1,19 +1,13 @@
 //! Keccak gadget
-//! -------------
-//! The Keccak gadget is a circuit that implements the Keccak hash function
-//! for 64-bit words, output length of 256 bits and bit rate of 1088 bits.
-//!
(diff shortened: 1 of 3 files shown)


dannywillems:
- `state_c` is removed. It initially corresponded to `C[x]` in the reference, the first intermediate array which is created to compute the sum of `A[x][i]`. It is removed as it can be inlined in the next computation for `D[x]`. The same logic happens for `state_e` and `state_d`.

dannywillems (approved):

PR #1548, opened 2023-12-14
[easy] Implement `hash` function inside `KeccakInterpreter`

querolita:
This is a preliminary trigger for the Keccak interpreter, on input the preimage data. More PRs are coming. In particular, take into account that this still does not implement the communication channel, so the preimage is passed as input instead of being read from memory.


dannywillems (approved):
LGTM, except require a git merge/git rebase because we removed witness.rs in the first PR.

PR #1550, opened 2023-12-14
[easy] Implement step of Keccak interpreter

querolita:
This function is meant to be the step that repeats over and over again in the execution of the Keccak circuit.

### DIFF
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -36,12 +36,21 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
         // Run all steps of hash
         while self.curr_step.is_some() {
             self.step();
-            self.update_step();
         }
     }
 
+    // FIXME: read preimage from memory and pad and expand
     fn step(&mut self) {
-        todo!()
+        // Reset columns to zeros to avoid conflicts between steps
+        self.null_state();
+
+        // FIXME sparse notation
+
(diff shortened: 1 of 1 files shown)


optimism/src/keccak/witness.rs
```diff
@@ -36,12 +36,21 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
...
+        // Reset columns to zeros to avoid conflicts between steps
+        self.null_state();
+
+        // FIXME sparse notation
+
+        match self.curr_step.unwrap() {
+            KeccakStep::Sponge(typ) => self.run_sponge(typ),
+            KeccakStep::Round(i) => self.run_round(i),
+        }
+        self.update_step();
```
dannywillems, line 53: It seems that [here](https://github.com/o1-labs/proof-systems/pull/1549/files#diff-4681765c9d01206fa0a308b5bb0532934fb259d49efa85a4ec2dd88c41bf6554R39), `update_step` is run after step. Is it expected?

dannywillems: Ignore. Fixed above.

dannywillems (approved):

optimism/src/keccak/witness.rs
querolita: Exactly, I preferred to move the `update_step` inside `step` itself so it was less confusing from the higher level function.

PR #1552, opened 2023-12-14
Implement sponge execution inside Keccak interpreter

querolita:
Includes `run_sponge`, `run_squeeze` and `run_absorb`.

### DIFF
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -1,11 +1,11 @@
 use ark_ff::Field;
-use kimchi::circuits::polynomials::keccak::{Keccak, ROUNDS};
+use kimchi::circuits::polynomials::keccak::{Keccak, CAPACITY_IN_BYTES, RATE_IN_BYTES, ROUNDS};
 
 use super::{
     column::KeccakColumn,
     environment::KeccakEnv,
     interpreter::{Absorb, KeccakInterpreter, KeccakStep, Sponge},
-    DIM, QUARTERS,
+    DIM, HASH_BYTELENGTH, QUARTERS, WORDS_IN_HASH,
 };
 
 impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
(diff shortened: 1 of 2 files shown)


optimism/src/keccak/witness.rs
```diff
@@ -83,15 +83,76 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
...
+        // Rest is zero thanks to null_state
+
+        // TODO: more updates to the env?
+    }
+
+    fn run_absorb(&mut self, absorb: Absorb) {
+        self.set_flag_absorb(absorb);
+
+        // Compute witness values
+        let ini_idx = self.block_idx * RATE_IN_BYTES;
```
dannywillems, line 122: The drawing I did during our call. Only if you want to add it.
```
        //    Permutation + Mode of Operation (Sponge)
        //    state -> permutation(state) -> state'
        //              ----> either [0] or state'
        //             |            new state = Exp(block)
        //             |         ------------------------
        //    Absorb: state  + [  block      +     0...0 ]
        //                       1088 bits          512
        //            ----------------------------------
        //                         XOR STATE
```

dannywillems: Nit (not blocking, and only a personal taste): I would simply reorganize the variable definition in a different order
```
        let ini_idx = self.block_idx * RATE_IN_BYTES;
        let mut block = self.padded[ini_idx..ini_idx + RATE_IN_BYTES].to_vec();
        // Pad with zeros
        block.append(&mut vec![0; CAPACITY_IN_BYTES]);

        let old_state = self.prev_block.clone();
        let new_state = Keccak::expand_state(&block);
        let xor_state = old_state
            .iter()
            .zip(new_state.clone())
            .map(|(x, y)| x + y)
            .collect::<Vec<u64>>();
```

optimism/src/keccak/witness.rs
```diff
@@ -83,15 +83,76 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
...
+        let xor_state = old_state
+            .iter()
+            .zip(new_state.clone())
+            .map(|(x, y)| x + y)
+            .collect::<Vec<u64>>();
+
+        // Write absorb-related columns
+        for i in 0..QUARTERS * DIM * DIM {
+            self.write_column(KeccakColumn::SpongeOldState(i), old_state[i]);
+            self.write_column(KeccakColumn::SpongeNewState(i), new_state[i]);
```
dannywillems, line 139: As discussed, maybe it can be simplified later as the new state is only used temporally. Opening a follow-up.

dannywillems (approved):

optimism/src/keccak/witness.rs
querolita: I included them in https://github.com/o1-labs/proof-systems/pull/1559/

PR #1558, opened 2023-12-14
Witness creation for Iota permutation in the interpreter

querolita:
This reuses the witness code in Kimchi. It adds some methods to obtain the fields of the Iota struct.

### DIFF
--- a/kimchi/src/circuits/polynomials/keccak/witness.rs
+++ b/kimchi/src/circuits/polynomials/keccak/witness.rs
@@ -355,16 +355,25 @@ impl Chi {
 /// Values involved in Iota permutation step
 pub struct Iota {
     state_g: Vec<u64>,
+    rc: Vec<u64>,
 }
 
 impl Iota {
-    fn create(state_f: Vec<u64>, round: usize) -> Self {
+    pub fn create(state_f: &[u64], round: usize) -> Self {
         let rc = Keccak::sparse(RC[round]);
-        let mut state_g = state_f.clone();
+        let mut state_g = state_f.to_vec();
         for (i, c) in rc.iter().enumerate() {
             state_g[i] = state_f[i] + *c;
         }
-        Self { state_g }
+        Self { state_g, rc }
+    }
+
+    pub fn state_g(&self, i: usize) -> u64 {
+        self.state_g[i]
+    }
+
+    pub fn rc(&self, i: usize) -> u64 {
+        self.rc[i]
     }
 }
 
@@ -428,7 +437,7 @@ pub fn extend_keccak_witness<F: PrimeField>(witness: &mut [Vec<F>; KECCAK_COLS],
             let chi = Chi::create(&pirho.state_b);
 
             // Iota
-            let iota = Iota::create(chi.state_f, round);
+            let iota = Iota::create(&chi.state_f, round);
 
             // Initialize the round row
             witness::init(
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -1,8 +1,8 @@
 use ark_ff::Field;
 use kimchi::{
     circuits::polynomials::keccak::{
-        witness::{Chi, PiRho, Theta},
-        Keccak, CAPACITY_IN_BYTES, RATE_IN_BYTES, RC, ROUNDS,
+        witness::{Chi, Iota, PiRho, Theta},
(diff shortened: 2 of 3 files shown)


kimchi/src/circuits/polynomials/keccak/witness.rs
```diff
@@ -355,16 +355,25 @@ impl Chi {
...
     state_g: Vec<u64>,
+    rc: Vec<u64>,
 }
 
 impl Iota {
-    fn create(state_f: Vec<u64>, round: usize) -> Self {
+    pub fn create(state_f: &[u64], round: usize) -> Self {
         let rc = Keccak::sparse(RC[round]);
-        let mut state_g = state_f.clone();
+        let mut state_g = state_f.to_vec();
```
dannywillems, line 364: Apparently there is no map in Rust with the index. We could have simplify the code and remove one list iteration.

kimchi/src/circuits/polynomials/keccak/witness.rs
```diff
@@ -355,16 +355,25 @@ impl Chi {
 /// Values involved in Iota permutation step
 pub struct Iota {
     state_g: Vec<u64>,
+    rc: Vec<u64>,
```
dannywillems, line 358: Optim: as it is fixed, we could have it static somewhere. It will be in the binary directly, and not in memory.

dannywillems: Open https://github.com/o1-labs/proof-systems/issues/1605

querolita: https://github.com/o1-labs/proof-systems/issues/1606

optimism/src/keccak/witness.rs
```diff
@@ -257,7 +256,15 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
...
-        todo!()
+    fn run_iota(&mut self, state_f: &[u64], round: usize) {
+        let iota = Iota::create(state_f, round);
+
+        // Update columns
+        for i in 0..QUARTERS * DIM * DIM {
+            self.write_column(KeccakColumn::NextState(i), iota.state_g(i));
+        }
+        for i in 0..QUARTERS {
+            self.write_column(KeccakColumn::RoundConstants(i), iota.rc(i));
```
dannywillems, line 267: I think we do not need columns for the round constants. It is well-defined by the Keccak instance, and static.

kimchi/src/circuits/polynomials/keccak/witness.rs
querolita: https://github.com/o1-labs/proof-systems/issues/1607

kimchi/src/circuits/polynomials/keccak/witness.rs
```diff
@@ -355,16 +355,25 @@ impl Chi {
...
 }
 
 impl Iota {
-    fn create(state_f: Vec<u64>, round: usize) -> Self {
+    pub fn create(state_f: &[u64], round: usize) -> Self {
         let rc = Keccak::sparse(RC[round]);
```
dannywillems, line 363: And following [this](https://github.com/o1-labs/proof-systems/pull/1558/files#r1430165407), we could get rid of this line, and have it static somewhere.

optimism/src/keccak/witness.rs
dannywillems: And [this](https://github.com/o1-labs/proof-systems/pull/1558/files#r1430167003) is related.

dannywillems (approved):
Modulo comments. Let's merge and get back to this later as it might require changes out of scope of this PR.

optimism/src/keccak/witness.rs
querolita: The goal for having these is to be able to check in a lookup table that the right round constant is being used in the current round. 

But if it makes more sense to model it differently I will be glad to adapt it.

dannywillems: >The goal for having these is to be able to check in a lookup table that the right round constant is being used in the current round.

Oh, I see. Never thought about this. I do not know if it is important.

dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/1608 to keep track of it.

PR #1560, opened 2023-12-14
Constraints for flags configuration

querolita:
These constraints were not in the Kimchi gate because flags were publicly accessible as gate coefficients encoding the mode of operation.

### DIFF
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -1,4 +1,8 @@
-use crate::keccak::{column::KeccakColumn, environment::KeccakEnv};
+use crate::keccak::{
+    column::KeccakColumn,
+    environment::{KeccakEnv, KeccakEnvironment},
+    BoolOps,
+};
 use ark_ff::Field;
 
 use super::E;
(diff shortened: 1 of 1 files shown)


optimism/src/keccak/constraints.rs
```diff
@@ -26,8 +30,38 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
...
     fn constraints(&mut self) {
-        todo!();
         // CORRECTNESS OF FLAGS
+        {
+            // TODO: remove redundancy if any
+
+            // Booleanity of sponge flags
+            {
+                // Absorb is either true or false
+                self.constrain(Self::boolean(self.absorb()));
```
dannywillems, line 40: Follow-up: I think the operations defined in BoolOps overlap with some from the interpreter env.

dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/1609

optimism/src/keccak/constraints.rs
```diff
@@ -26,8 +30,38 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
...
+                // Squeeze is either true or false
+                self.constrain(Self::boolean(self.squeeze()));
```
dannywillems, line 42: Nit: I would have called the method `is_boolean`. Not part of this PR.

optimism/src/keccak/constraints.rs
```diff
@@ -26,8 +30,38 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
...
+                // Root is either true or false
+                self.constrain(Self::boolean(self.root()));
+                // Pad is either true or false
+                self.constrain(Self::boolean(self.pad()));
+            }
+            // Mutually exclusiveness of flags
+            {
+                // Squeeze and Root are not both true
+                self.constrain(Self::either_false(self.squeeze(), self.root()));
```
dannywillems, line 51: What's the expected output of `either_false`?
If at least one of the parameters is `false` (i.e. `0`), return `true` (i.e. `1`)?

querolita: if at least one of them is false, then the constrain is satisfied -> the constrain is not satisfied if both are true -> you cannot have an absorb root in a squeeze sponge

optimism/src/keccak/constraints.rs
querolita: oh, yeah I can modify this. I just inherited a function we had in kimchi that was called just `boolean`

optimism/src/keccak/constraints.rs
querolita: Probably, they were created in parallel

dannywillems (approved):

PR #1561, opened 2023-12-14
Constraints for the Keccak sponge

querolita:
The difference between these and the ones found in the Kimchi gate are the additional degree resulting from multiplying by the selector flag in each constraint, and removing the constraints related to the padding because this will be handled differently in the MIPS project.

### DIFF
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -4,8 +4,9 @@ use crate::keccak::{
     BoolOps,
 };
 use ark_ff::Field;
+use kimchi::circuits::polynomials::keccak::{DIM, QUARTERS};
 
-use super::E;
+use super::{E, WORDS_IN_HASH};
 
 pub trait Constraints {
     type Column;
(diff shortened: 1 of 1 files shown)


optimism/src/keccak/constraints.rs
```diff
@@ -64,6 +65,47 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
         }
 
         // SPONGE CONSTRAINTS
+        {
+            for z in self.sponge_zeros() {
```
dannywillems, line 69: Follow-up/reminder: check this twice when the final structure is merged.

dannywillems (approved):

optimism/src/keccak/constraints.rs
querolita: actually, I realized of an optimization that reduces by 32 constraints in the sponge, I am creating a follow-up branch called `keccak/32-less-constraints`

querolita: Okay, false alarm. Then I need to check that the other 32 entries are unchanged (xor by zero), that doesnt change things then).

PR #1563, opened 2023-12-14
Constraints of the PiRho permutation step

querolita:
The difference between these and the ones found in the Kimchi gate is an extra degree due to multiplying by the `is_round()` selector.

### DIFF
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -4,7 +4,7 @@ use crate::keccak::{
     {ArithOps, BoolOps, E, WORDS_IN_HASH},
 };
 use ark_ff::Field;
-use kimchi::circuits::polynomials::keccak::{DIM, QUARTERS};
+use kimchi::circuits::polynomials::keccak::{DIM, OFF, QUARTERS};
 
 pub trait Constraints {
     type Column;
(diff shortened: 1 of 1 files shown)


optimism/src/keccak/constraints.rs
```diff
@@ -159,6 +163,76 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
                     }
                 }
             } // END theta
+
+            // STEP pirho: 5 * 5 * (2 + 4 * 1) = 150 constraints
+            for (y, col) in OFF.iter().enumerate() {
+                for (x, off) in col.iter().enumerate() {
+                    let word_e = Self::from_quarters(&self.keccak_state.pi_rho_dense_e, Some(y), x);
+                    let quo_e =
```
dannywillems, line 171: Don't we have to check it is a boolean?

dannywillems:
Sorry, I merged two others PR into this one.

dannywillems (commented):
See comment

optimism/src/keccak/constraints.rs
querolita: Not here. Here instead, we need to check that quotient_e is <2^64.

We check it is a boolean for state_c because the rotation in step theta is just by 1 bit. That means that the quotient can only be 1 or 0.

dannywillems: Ah, yes, right!

dannywillems (approved):

PR #1568, opened 2023-12-14
Support for padding (witness and constraints)

querolita:
This PR adds witness generation code and constraints for the Keccak padding. The strategy is twofold: on the one hand, we must check that the padding occurs at the end of the state and not anywhere else; on the other hand, we must check that the padding is correct, according to the predefined suffix for each length.

The idea is to have additional columns that will act as boolean byte selectors to indicate the constraints what bytes are involved in the padding. Then, in order to check that these are correctly located at the end of the data, we constrain that the combination of those flags (that will have to be set to 1 consecutively) results in the "all-one-word", meaning 2^pad_len - 1, for which we add a column as well. Since we already had a column for the padding length, we will check that TwoToPad is correct using a lookup table (that will come in a subsequent PR). 

In order to check that the pad being used is correct, we add the PadSuffix columns which consist of 5 field elements (the first at most 12 bytes, and the rest of at most 31 bytes, altogether at most 1088 bits) that correspond to the correct padding that should be used for each pad length. For this, we will use a lookup in a subsequent PR. For now, we just make sure that combining each chunk of bytes that are involved in the padding result in the desired pad suffix given by the newly added columns.


optimism/src/keccak/witness.rs
```diff
@@ -15,6 +15,28 @@ use super::{
     DIM, HASH_BYTELENGTH, QUARTERS, WORDS_IN_HASH,
 };
 
+pub(crate) fn pad_blocks<Fp: Field>(pad_bytelength: usize) -> Vec<Fp> {
+    // Blocks to store padding. The first one uses at most 12 bytes, and the rest use at most 31 bytes.
+    let mut blocks = vec![Fp::zero(); 5];
+    let mut pad = [Fp::zero(); RATE_IN_BYTES];
+    pad[RATE_IN_BYTES - pad_bytelength] = Fp::one();
+    pad[RATE_IN_BYTES - 1] += Fp::from(0x80u8);
```
dannywillems, line 23: Nit: just an assignment would be a bit faster, as it won't have to compute the addition. Here it does one addition + one new allocation for the result of the addition.

optimism/src/keccak/witness.rs
```diff
@@ -15,6 +15,28 @@ use super::{
...
+    blocks[0] = pad
+        .iter()
+        .take(12)
+        .fold(Fp::zero(), |acc, x| acc * Fp::from(256u32) + *x);
```
dannywillems, line 27: If bitshift is implemented on Fp, it would be faster than the multiplication.

optimism/src/keccak/witness.rs
```diff
@@ -15,6 +15,28 @@ use super::{
     DIM, HASH_BYTELENGTH, QUARTERS, WORDS_IN_HASH,
 };
 
+pub(crate) fn pad_blocks<Fp: Field>(pad_bytelength: usize) -> Vec<Fp> {
```
dannywillems, line 18: Follow-up: I would prefer to have a fixed-length array instead of a vector as we know the size. Better for performances as it is known at compile time.

dannywillems (approved):
Nit improvements, for a follow-up.

optimism/src/keccak/witness.rs
querolita: This is because `0x81` is a possibility (adding 0x80 to 0x01). Are you suggesting that a set of if-else would be slightly faster?

PR #1571, opened 2023-12-15
Feature/generalize expr variable

mrmr1993:


dannywillems:
Seems to be in master. Closing.

PR #1574, opened 2023-12-15
Add columns and environment helpers for padding

querolita:
This is the first of a series of PRs meant to add support for padding in the MIPS Keccak module. 

This PR adds a few columns needed to select bytes involved in the padding, access the value $2^{len}$, and the claimed correct padding for a given length. 

It also provides some helpers in the environment to access these columns, and combinations of them that will be used inside the constraints.

### DIFF
--- a/optimism/src/keccak/environment.rs
+++ b/optimism/src/keccak/environment.rs
@@ -38,6 +38,7 @@ impl<Fp: Field> KeccakEnv<Fp> {
     pub fn write_column(&mut self, column: KeccakColumn, value: u64) {
         self.keccak_state[column] = Self::constant(value.into());
     }
+
     pub fn null_state(&mut self) {
         self.keccak_state = KeccakColumns::default();
     }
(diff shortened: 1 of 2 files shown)


optimism/src/keccak/environment.rs
```diff
@@ -148,6 +149,18 @@ pub(crate) trait KeccakEnvironment {
 
     fn length(&self) -> Self::Variable;
 
+    fn two_to_pad(&self) -> Self::Variable;
```
dannywillems, line 152: For a follow-up: would be nice to document each function, even if it is one-line

dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/1616

dannywillems (approved):
I'd like to see at the end of the PR train if we cannot simplify the code. Let's see. Acceptig in the meantime.

PR #1575, opened 2023-12-15
Add variants of lookup tables that will be used in Keccak

querolita:
This PR adds a few table IDs that will be needed for the lookups inside Keccak.


optimism/src/keccak/environment.rs
```diff
@@ -17,7 +17,7 @@ pub struct KeccakEnv<Fp> {
     /// Constraints that are added to the circuit
     pub(crate) constraints: Vec<E<Fp>>,
     /// Values that are looked up in the circuit
-    pub(crate) _lookup_terms: [Vec<Lookup<E<Fp>>>; 2], // at most 2 values are looked up at a time
+    pub(crate) _lookup_terms: Vec<Lookup<E<Fp>>>, // at most 5 values are looked up at a time
```
dannywillems, line 20: I would prefer to keep a fixed-length vector, and simply increase the size.

dannywillems (approved):

optimism/src/keccak/environment.rs
querolita: The old type of an array was a typo

PR #1582, opened 2023-12-15
Lookups for padding

querolita:
This PR adds the lookup required to check the padding. This is being done in the same table of 137 rows (from 0 to 136) including the padding length, the term 2^len, and the suffixes corresponding to the up to 1088 bits of padding.


optimism/src/mips/interpreter.rs
```diff
@@ -138,10 +138,8 @@ pub enum LookupTable {
     ResetLookup,
     // 24-row table with all possible values for round and their round constant in expanded form
     RoundConstantsLookup,
-    // All [0..136] values of possible padding lengths and the 5 corresponding pad suffixes with the 10*1 rule
+    // All [0..136] values of possible padding lengths, the value 2^len, and the 5 corresponding pad suffixes with the 10*1 rule
     PadLookup,
-    // Powers of two, from 2^0 to 2^136
-    PowersOfTwoLookup,
```
dannywillems, line 144: Why removed?

querolita: Oh, because I integrated it inside the same table for padding. Now the table contains one column for the length, another one for the 2^len, and then 5 for the suffixes.

PR #1585, opened 2023-12-15
Use `Literal` to implement `pow`

mrmr1993:


dannywillems:
Would be nice to have regression tests for the print results. Not urgent, but good for also learning the expression framework.

dannywillems:
> Would be nice to have regression tests for the print results. Not urgent, but good for also learning the expression framework.

Opened https://github.com/o1-labs/proof-systems/issues/1622. I remember we talked with @volhovm to play with the expression framework.

dannywillems (approved):

PR #1586, opened 2023-12-15
Lookups for sponge

querolita:
This PR adds the lookups for the sponge of Keccak. It involves lookups to the Sparse table to check that Shift1, Shift2, Shift3 are correct; lookups to the Bytes table to check that the bytes of the sponge are in range; and lookups to the Reset table using the composition of every 2 bytes as the dense column and the Shifts0, to make sure they are correctly decomposed.


optimism/src/mips/interpreter.rs
```diff
@@ -140,6 +140,9 @@ pub enum LookupTable {
     RoundConstantsLookup,
     // All [0..136] values of possible padding lengths, the value 2^len, and the 5 corresponding pad suffixes with the 10*1 rule
     PadLookup,
+    // All values that can be stored in a byte
+    // TODO: model as RangeCheck16 for x and a scaled x' = x * 2^8
+    ByteLookup,
```
mrmr1993, line 145: TODO-me: it's probably more efficient to have an 8-bit table, since it's amortised across the prood

mrmr1993 (approved):

optimism/src/mips/interpreter.rs
querolita: Ah nice, then I will remove that comment

PR #1590, opened 2023-12-15
Add `Cache` to `Operations`

mrmr1993:


kimchi/src/circuits/expr.rs
```diff
@@ -2631,78 +2659,153 @@ impl<F: PrimeField> FormattedOutput for ConstantExprInner<F> {
...
     }
-    fn ocaml(&self) -> String {
+    fn ocaml(&self, cache: &mut HashMap<CacheId, Self>) -> String {
         use ConstantExprInner::*;
         match self {
-            Challenge(x) => x.ocaml(),
-            Constant(x) => x.ocaml(),
+            Challenge(x) => {
+                let mut inner_cache = HashMap::new();
+                let res = x.ocaml(&mut inner_cache);
```
dannywillems, line 2667: I suppose the cache is given for the future; it is useless at the moment. Challenges could be a list of values, for instance when we use the IPA.

dannywillems (approved):

PR #1600, opened 2023-12-18
[easy] Numerator helpers to write or read only one

querolita:
This PR adds two helpers inside `Signed` when the numerator indicates that it reads or writes only one entry. It is meant to reduce the number of lines of code required to define a `Lookup`.

Convention used: minus one is to read from table, plus one is to write to table.


mrmr1993 (approved):
Love this :heart:

PR #1613, opened 2023-12-18
Constants for offset of commencement of Keccak witnesses to be used in MIPS columns

querolita:
This PR introduces a series of constants to substitute the "arbitrary-looking" offsets in the constraints, explaining where they come from. They are a bit verbose, feel free to recommend any other convention for them. What's important here is to be able to export them to the MIPS line of work so that `KeccakColumns` can be done more efficiently, reducing the columns by ~800 (by reusing vector entries depending on the mode of operation).


dannywillems (approved):
LGTM if it makes your life easier. I don't have strong opinions.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -99,25 +149,70 @@ where
...
         // THETA
-        let state_a = grid!(100, env.witness_curr_chunk(0, 100));
-        let shifts_c = grid!(80, env.witness_curr_chunk(100, 180));
-        let dense_c = grid!(20, env.witness_curr_chunk(180, 200));
-        let quotient_c = grid!(5, env.witness_curr_chunk(200, 205));
-        let remainder_c = grid!(20, env.witness_curr_chunk(205, 225));
-        let dense_rot_c = grid!(20, env.witness_curr_chunk(225, 245));
-        let expand_rot_c = grid!(20, env.witness_curr_chunk(245, 265));
+        let state_a = grid!(
+            100,
```
dannywillems, line 153: I suppose here it is always the length.

dannywillems: I don't know if you want to replace it.

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
```diff
@@ -220,12 +315,13 @@ where
         let mut constraints = vec![];
 
         // LOAD WITNESS
-        let old_state = env.witness_curr_chunk(0, 100);
-        let new_block = env.witness_curr_chunk(100, 200);
-        let zeros = env.witness_curr_chunk(168, 200);
-        let xor_state = env.witness_next_chunk(0, 100);
-        let bytes = env.witness_curr_chunk(200, 400);
-        let shifts = env.witness_curr_chunk(400, 800);
+        let old_state = env.witness_curr_chunk(SPONGE_OLD_STATE_OFF, SPONGE_OLD_STATE_LEN);
```
dannywillems, line 318: Do you want the length of the new state offset?

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: Yeah, it is always the length. But because of how the macro is defined, it does not recognize the pattern matching if I use a constant instead of the hard-coded number 😭

kimchi/src/circuits/polynomials/keccak/circuitgates.rs
querolita: well in this case it is equivalent (because the offset of the old state is 0), but let's do new state offset so they all have the same shape

PR #1624, opened 2023-12-20
Use a seedable PRNG for randomized tests

dannywillems:
Part of https://github.com/MinaProtocol/mina/issues/14664

### DIFF
--- a/kimchi/src/tests/lookup.rs
+++ b/kimchi/src/tests/lookup.rs
@@ -23,9 +23,13 @@ type BaseSponge = DefaultFqSponge<VestaParameters, SpongeParams>;
 type ScalarSponge = DefaultFrSponge<Fp, SpongeParams>;
 
 fn setup_lookup_proof(use_values_from_table: bool, num_lookups: usize, table_sizes: Vec<usize>) {
+    let seed: [u8; 32] = thread_rng().gen();
+    eprintln!("Seed: {:?}", seed);
+    let mut rng = StdRng::from_seed(seed);
+
     let lookup_table_values: Vec<Vec<_>> = table_sizes
         .iter()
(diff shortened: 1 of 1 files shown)


kimchi/src/tests/lookup.rs
```diff
@@ -23,9 +23,13 @@ type BaseSponge = DefaultFqSponge<VestaParameters, SpongeParams>;
 type ScalarSponge = DefaultFrSponge<Fp, SpongeParams>;
 
 fn setup_lookup_proof(use_values_from_table: bool, num_lookups: usize, table_sizes: Vec<usize>) {
+    let seed: [u8; 32] = thread_rng().gen();
+    eprintln!("Seed: {:?}", seed);
```
volhovm, line 27: It's a really good change, but I'm also wondering whether there's a way to not always print the seed into stderr, but instead only when the test fails. Maybe some lightweight test framework for rust could do it for us? I'm worried that it's too custom and all the tests will print it differently => less uniform => harder to debug. Or we don't really need to care?

volhovm (approved):

kimchi/src/tests/lookup.rs
dannywillems: I would like to have something better than this repeated 3 lines. There is a lack of equivalent to qcheck in Rust. Having a library for this would solve what you're mentioning.

dannywillems: I hope this "hack" is only temporary.

dannywillems: Open https://github.com/o1-labs/proof-systems/issues/1625

rbonichon (approved):

querolita (approved):

PR #1626, opened 2023-12-20
Prover

mrmr1993:


optimism/src/main.rs
```diff
@@ -40,12 +40,113 @@ pub fn main() -> ExitCode {
 
     env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
 
-    let mut env = witness::Env::<ark_bn254::Fq>::create(cannon::PAGE_SIZE as usize, state, po);
+    let domain_size = 1 << 15;
+
+    let domain =
+        kimchi::circuits::domains::EvaluationDomains::<ark_bn254::Fr>::create(domain_size).unwrap();
+
+    let srs = {
```
dannywillems, line 48: Follow-up: use a public SRS

dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/1648 and linked in the code, see https://github.com/o1-labs/proof-systems/pull/1626/commits/0e44be9d8da27a6effd569833feffbb3f5299dd0

querolita: > use a public SRS

do you mean available SRS's which are known to having been created in a ceremony instead of generated in the code?

optimism/src/main.rs
```diff
@@ -40,12 +40,113 @@ pub fn main() -> ExitCode {
...
+        srs
+    };
+
+    let mut env = witness::Env::<ark_bn254::Fr>::create(cannon::PAGE_SIZE as usize, state, po);
+
+    let mut accumulator = proof::ProofInputs::<
+        ark_ec::short_weierstrass_jacobian::GroupAffine<ark_bn254::g1::Parameters>,
+    >::default();
+
+    let new_chunk = || proof::WitnessColumns {
```
querolita, line 65: what is exactly a chunk in this context? chunks as in kimchi in case the circuit was larger than the domain size?

optimism/src/main.rs
```diff
@@ -40,12 +40,113 @@ pub fn main() -> ExitCode {
...
+        }
+        current_chunk
+            .instruction_counter
+            .push(ark_bn254::Fr::from(env.instruction_counter));
+        // TODO
+        use ark_ff::UniformRand;
+        current_chunk
+            .error
+            .push(ark_bn254::Fr::rand(&mut rand::rngs::OsRng));
+        if current_chunk.instruction_counter.len() == 1 << 15 {
```
querolita, line 99: Oh, or does chunk refer to each of the instances that will be folded?

querolita: Like, in case we wanted to split the full circuit into instances of equal length that is a divisor of the circuit size?

optimism/src/main.rs
```diff
@@ -40,12 +40,113 @@ pub fn main() -> ExitCode {
...
+            poly_commitment::pairing_proof::PairingProof<ark_ec::bn::Bn<ark_bn254::Parameters>>,
+            BaseSponge,
+            ScalarSponge,
+        >(domain, &srs, &mut accumulator, current_chunk);
+    }
+
+    {
+        let proof = proof::prove::<
+            _,
+            poly_commitment::pairing_proof::PairingProof<ark_ec::bn::Bn<ark_bn254::Parameters>>,
```
querolita, line 132: I would personally try to import these up-front so the code does not look so verbose. Otherwise it is a bit hard to parse what is actually being called.

optimism/src/mips/proof.rs
```diff
@@ -0,0 +1,444 @@
...
+        }
+    };
+    let commitments = {
+        let WitnessColumns {
+            scratch,
+            instruction_counter,
+            error,
+        } = &polys;
+        let comm = |evals: &Evaluations<G::ScalarField, D<G::ScalarField>>| {
+            srs.commit_evaluations_non_hiding(domain.d1, evals)
```
querolita, line 93: non hiding because Optimism is not asking for zk?

optimism/src/mips/proof.rs
```diff
@@ -0,0 +1,444 @@
...
+>(
+    domain: EvaluationDomains<G::ScalarField>,
+    srs: &OpeningProof::SRS,
+    inputs: ProofInputs<G>,
+) -> Proof<G, OpeningProof>
+where
+    OpeningProof::SRS: Sync,
+{
+    let ProofInputs { evaluations } = inputs;
+    let rng = &mut rand::rngs::OsRng;
```
querolita, line 153: shouldn't we obtain the randomness source from the sponge instead so it matches the verifier's?

querolita (commented):
Looks good to me, but I left a few comments that I would like to chat about before approving. I will review the following ones now, maybe some of these questions become clearer by then.

optimism/src/main.rs
mrmr1993: It's a pre-folding witness. The domain size we're using is 2^15, so a 'chunk' is 2^15 constraints/rows.

optimism/src/mips/proof.rs
mrmr1993: Non-hiding because it simplifies things for now.

optimism/src/mips/proof.rs
mrmr1993: It's only used for the opening proof, where it doesn't matter anyway. I'll inline it there.

optimism/src/main.rs
```diff
@@ -40,12 +42,106 @@ pub fn main() -> ExitCode {
...
+
+    use mina_poseidon::{
+        constants::PlonkSpongeConstantsKimchi,
+        sponge::{DefaultFqSponge, DefaultFrSponge},
+    };
+    type Fp = ark_bn254::Fr;
+    type SpongeParams = PlonkSpongeConstantsKimchi;
+    type BaseSponge = DefaultFqSponge<ark_bn254::g1::Parameters, SpongeParams>;
+    type ScalarSponge = DefaultFrSponge<Fp, SpongeParams>;
+    type OpeningProof = PairingProof<Bn<ark_bn254::Parameters>>;
```
querolita, line 83: Awesome, looks much clearer now

querolita (approved):

PR #1632, opened 2023-12-20
Speed up proving with better parallelisation

mrmr1993:


optimism/src/mips/proof.rs
```diff
@@ -24,6 +25,70 @@ pub struct WitnessColumns<G> {
...
+}
+
+impl<G: Send + std::fmt::Debug> FromParallelIterator<G> for WitnessColumns<G> {
+    fn from_par_iter<I>(par_iter: I) -> Self
+    where
+        I: IntoParallelIterator<Item = G>,
+    {
+        let mut iter_contents = par_iter.into_par_iter().collect::<Vec<_>>();
+        let error = iter_contents.pop().unwrap();
+        let instruction_counter = iter_contents.pop().unwrap();
```
querolita, line 51: Silly question whose answer should be "yes", but asking just in case: does push and pop for vectors behave in "the expected way"? Meaning, that the elements in the resulting vector being popped occur in the same order as the ones being pushed 🤔

optimism/src/mips/proof.rs
```diff
@@ -64,40 +129,17 @@ pub fn fold<
...
+        inputs
+            .par_iter()
+            .map(|evals: &Vec<G::ScalarField>| {
+                let evals = Evaluations::<G::ScalarField, D<G::ScalarField>>::from_vec_and_domain(
+                    evals.clone(),
+                    domain.d1,
+                );
+                srs.commit_evaluations_non_hiding(domain.d1, &evals)
+            })
+            .collect::<WitnessColumns<_>>()
```
querolita, line 142: Very nice looking

querolita (approved):
Lovely

optimism/src/mips/proof.rs
mrmr1993: Yes :)

PR #1633, opened 2023-12-20
Performance improvement: reuse vector memory allocations in the zkVM

mrmr1993:


optimism/src/main.rs
```diff
@@ -62,14 +62,24 @@ pub fn main() -> ExitCode {
...
 
-    let new_chunk = || proof::WitnessColumns {
+    let reset_chunk = |witness_columns: &mut proof::WitnessColumns<Vec<_>>| {
+        let proof::WitnessColumns {
+            scratch,
+            instruction_counter,
+            error,
+        } = witness_columns;
+        // Resize without deallocating
+        scratch.iter_mut().for_each(Vec::clear);
```
querolita, line 72: Oh, so then it is assumed that since each chunk will have the same size, it is fine if the same memory allocations get "nullified" instead of deallocated. And then they can be reassigned again to the new values? Nice

querolita (approved):
Very nice

PR #1637, opened 2023-12-20
[easy] Split `NextState` column across 2 variants for readability

querolita:
Even if this PR seems to increase the number of total columns by 100, a coming PR will eventually reduce it by 800. This is just a preliminary step so the diff is smaller.

It also fixes a notation deviation between `new_block` (which referred in Kimchi to just 1088 bits) and `new_state` (which refers to all 1600 bits).

### DIFF
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -83,7 +83,7 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
                 // Absorbs the new block by performing XOR with the old state
                 self.constrain(
                     self.is_absorb()
-                        * (self.next_state(i) - (self.old_state(i) + self.new_block(i))),
+                        * (self.xor_state(i) - (self.old_state(i) + self.new_block(i))),
                 );
                 // In absorb, Check shifts correspond to the decomposition of the new state
                 self.constrain(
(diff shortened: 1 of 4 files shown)


optimism/src/keccak/constraints.rs
```diff
@@ -246,7 +246,7 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
             // STEP iota: 4 constraints
             for (q, c) in self.round_constants().iter().enumerate() {
                 self.constrain(
-                    self.is_round() * (self.next_state(q) - (state_f[0][0][q].clone() + c.clone())),
+                    self.is_round() * (self.state_g(q) - (state_f[0][0][q].clone() + c.clone())),
```
volhovm, line 249: Is it OK that you only constrain `state_g` here once`? My very vague intuition suggests that that it should appear in two places, similarly to `new_state(i)`...?

volhovm (approved):

optimism/src/keccak/constraints.rs
querolita: For transparency, in our meeting we discussed how the step in the interpreter should take the value of `state_g` to instantiate the `prev_step` of a round. But now that I am checking the code, I can see that `prev_step` is indeed used to initialize sponges and for the first theta, but it is not done in-between rounds. Will be creating upcoming PR to address this issue.

PR #1639, opened 2023-12-20
[easy] Add `curr` and `next` row vectors inside `KeccakColumns`

querolita:
This PR adds two more elements inside the `KeccakColumns` struct in order to keep track of the columns located in the `Curr` and `Next` rows of the Kimchi implementation. The goal of this addition is to be able to remove 800 columns eventually by providing an unified "storage", but allowing for a fine-grained access to each `KeccakColumn` variant.

This PR is meant to make the diff look smaller. Coming PR will actually remove those redundant columns.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -3,7 +3,7 @@ use std::ops::{Index, IndexMut};
 use ark_ff::{One, Zero};
 use serde::{Deserialize, Serialize};
 
-use super::grid_index;
+use super::{grid_index, ZKVM_KECCAK_COLS_CURR, ZKVM_KECCAK_COLS_NEXT};
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


mrmr1993 (commented):
I don't think this approach works for the last column of the domain, nor for folding, without special consideration. Have you thought about / described how to handle those? It seems like it would be better not to merge this.

mrmr1993 (approved):
We chatted on a call. This is just a naming convention, there's no relationship to the 'next row' in the proof's rows themselves.

querolita:
> I don't think this approach works for the last column of the domain, nor for folding, without special consideration. Have you thought about / described how to handle those? It seems like it would be better not to merge this.

Solved issue -> the creation of these vectors do not mean that the next row is actually being used, it is just a naming convention to match Kimchi's layout. Nonetheless, I am creating a follow-up PR to constrain inputs and outputs as a memory model. Otherwise, only the witness is doing it, but there are no constraints for the correctness of it.

PR #1640, opened 2023-12-20
[easy] Create functions to access columns as slices

querolita:
The reason for this PR is that when columns get reduced by ~800, there won't be any more vectors inside `KeccakColumns` for individual states, and we need some helpers to access them.

### DIFF
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -5,7 +5,7 @@ use crate::keccak::{
 };
 use ark_ff::Field;
 use kimchi::circuits::polynomials::keccak::{
-    constants::{DIM, QUARTERS, RATE_IN_BYTES},
+    constants::{DIM, QUARTERS, RATE_IN_BYTES, SPONGE_ZEROS_LEN},
     OFF,
 };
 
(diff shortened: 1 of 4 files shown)


mrmr1993 (approved):

mrmr1993 (changes requested):
After looking over later PRs, it doesn't seem like we actually need to clone these, and it has a negative impact on the performance. Could you return slice references instead?

optimism/src/keccak/constraints.rs
```diff
@@ -244,9 +245,10 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
             } // END chi
 
             // STEP iota: 4 constraints
-            for (q, c) in self.round_constants().iter().enumerate() {
+            for (q, c) in self.round_constants().to_vec().iter().enumerate() {
```
mrmr1993, line 248: If you're going to clone anyway (via `to_vec`), you should probably just use `into_iter` and change `c.clone()` to `c` below.

optimism/src/keccak/constraints.rs
```diff
@@ -73,22 +73,23 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
 
         // SPONGE CONSTRAINTS
         {
-            for z in self.sponge_zeros() {
+            for i in 0..SPONGE_ZEROS_LEN {
                 // Absorb phase pads with zeros the new state
-                self.constrain(self.is_absorb() * z.clone());
+                self.constrain(self.is_absorb() * self.sponge_zeros()[i].clone());
```
mrmr1993, line 78: If this is going to be immediately cloned anyway, this makes sense to be a vector.

optimism/src/keccak/constraints.rs
```diff
@@ -73,22 +73,23 @@ impl<Fp: Field> Constraints for KeccakEnv<Fp> {
...
             }
             for i in 0..QUARTERS * DIM * DIM {
                 // In first absorb, root state is all zeros
-                self.constrain(self.is_root() * self.old_state(i));
+                self.constrain(self.is_root() * self.old_state(i).clone());
```
mrmr1993, line 82: Likewise for these.

mrmr1993 (commented):
The conversion from `Vec` to `[_]` might have been a bit overzealous. It seems like it doesn't make sense for those values that are now cloned immediately.

mrmr1993 (commented):
Approving to unblock, getting the slice/vec balance right is just a nit.

mrmr1993 (approved):

optimism/src/keccak/constraints.rs
querolita: Thanks for finding the nit, I will trace this issue here and address it asap.

https://github.com/o1-labs/proof-systems/issues/1678

PR #1644, opened 2023-12-20
[easy] Limit number of public fields inside `KeccakColumns`

querolita:
To prevent undesired side-effects.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -44,25 +44,26 @@ pub enum KeccakColumn {
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct KeccakColumns<T> {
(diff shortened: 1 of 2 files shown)


optimism/src/keccak/column.rs
```diff
@@ -74,9 +75,18 @@ impl<T: Clone> KeccakColumns<T> {
     ) -> &mut T {
         &mut self.curr[offset + grid_index(length, i, y, x, q)]
     }
+
     pub fn chunk(&self, offset: usize, length: usize) -> Vec<T> {
         self.curr[offset..offset + length].to_vec().clone()
     }
+
+    pub(crate) fn rc(&self) -> Vec<T> {
```
mrmr1993, line 83: Nit: Characters are cheap; `round_constants` would be more readable here.

mrmr1993 (approved):

PR #1653, opened 2024-01-03
[easy] Read/Write modes for Signed numerators in Lookups

querolita:
This PR adds an enum `LookupMode` whose variants are `Read` and `Write`, which act as wrappers for the positive and negative signs of the numerator of RAMLookups. Together with a flag selector argument which acts as the magnitude of elements added to the tables, these changes will facilitate the creation of lookups inside the Keccak workflow. In particular:

- thanks to the mode, witness and constraints will be able to reuse the same `lookups()` function to either "create" or "consume" the values in the table; and,
- thanks to the flag, lookups will be easily "switched on/off" depending on whether the Keccak step is a sponge or a round, without conflicts due to the reused vector positions to save 800 columns.

### DIFF
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -149,6 +149,14 @@ impl<T: One> Signed<T> {
             },
         }
     }
+
+    pub fn read_one() -> Self {
+        Self::new(LookupMode::Read, None)
+    }
+
+    pub fn write_one() -> Self {
+        Self::new(LookupMode::Write, None)
+    }
 }
 
 #[derive(Copy, Clone, Debug)]
--- a/optimism/src/mips/interpreter.rs
+++ b/optimism/src/mips/interpreter.rs
@@ -122,41 +122,9 @@ pub enum Sign {
 }
 
 #[derive(Copy, Clone, Debug)]
-pub struct Signed<T> {
-    pub sign: Sign,
-    pub magnitude: T,
-}
-
-#[derive(Copy, Clone)]
 pub enum LookupMode {
     Read,
     Write,
-}
-
-impl<T: One> Signed<T> {
-    /// Creates a new Signed element, either a Read or a Write, and can be null if the flag is zero
-    // TODO: if the flag trick works, then RC does not need to be length 25 anymore nor nonzero default column values
-    // FIXME: check what is the value of the flags at the witness stage
-    pub fn new(rw: LookupMode, flag: Option<T>) -> Self {
-        match rw {
-            LookupMode::Read => Self {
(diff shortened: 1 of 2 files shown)


optimism/src/mips/interpreter.rs
```diff
@@ -127,18 +127,35 @@ pub struct Signed<T> {
...
+    Write,
+}
+
 impl<T: One> Signed<T> {
-    pub fn read_one() -> Self {
-        Self {
-            sign: Sign::Neg,
-            magnitude: T::one(),
+    /// Creates a new Signed element, either a Read or a Write, and can be null if the flag is zero
+    // TODO: if the flag trick works, then RC does not need to be length 25 anymore nor nonzero default column values
```
mrmr1993, line 138: Can you explain the 'flag trick' here? The magnitude should only be used when we use the final tables, to account for the number of times we lookup a value.
For example, if we lookup values
```
[ 0, 0, 6, 6, 0, 6, 0 ]
```
in table
```
[ 0, 1, 2, 3, 4, 5, 6, 7 ]
```
then we would expect the table magnitudes to be
```
[ 4, 0, 0, 0, 0, 0, 3, 0 ]
```
so that all of the `1`s from the values cancel out with the table's magnitudes.

optimism/src/mips/interpreter.rs
```diff
@@ -127,18 +127,35 @@ pub struct Signed<T> {
...
+    // FIXME: check what is the value of the flags at the witness stage
+    pub fn new(rw: LookupMode, flag: Option<T>) -> Self {
```
mrmr1993, line 140: Please remove `flag` here

mrmr1993 (changes requested):

optimism/src/mips/interpreter.rs
mrmr1993: Oh, never mind, I see that it's either `0` or `1`

querolita: I was expecting the magnitude to represent something like _the number of times this value is added to the table_ . In particular, when I use them inside Keccak, the magnitude that is inferred from flags is either 1 or 0. Meaning, it either adds the lookup (as either positive or negative), or it does not because the sign is multiplied by zero. I had understood that was how magnitude was being used after reading the code of the MIPs interpreter where magnitude is sometimes `if_is_true`.

querolita: Meaning, the lookups are there, it doesn't mean that different rows have a different set of lookups. It's just when variables are actual witness values that those terms in the lists would get nullified.

mrmr1993 (approved):
Much clearer

PR #1654, opened 2024-01-03
[easy] Helper for `RangeCheck16` lookup table

querolita:
This PR is the first one trying to simplify the `Lookup` trait for `KeccakEnv` by adding a helper to easily create lookups to the `RangeCheck16` lookup table. Not only does it clean up the interface, but it also allows for switch on/off functionality thanks to the flag argument of `Signed`.

### DIFF
--- a/optimism/src/keccak/lookups.rs
+++ b/optimism/src/keccak/lookups.rs
@@ -20,7 +20,7 @@ pub(crate) trait Lookups {
     fn add_lookup(&mut self, lookup: Lookup<Self::Variable>);
 
     /// Adds all lookups of Self
-    fn lookups(&mut self);
+    fn lookups(&mut self, rw: LookupMode);
 
     /// Adds a lookup to the RangeCheck16 table
     fn lookup_rc16(&mut self, rw: LookupMode, flag: Self::Variable, value: Self::Variable);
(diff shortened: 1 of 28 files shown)


optimism/src/keccak/lookups.rs
```diff
@@ -182,4 +171,12 @@ impl<Fp: Field> Lookups for KeccakEnv<Fp> {
             }
         }
     }
+
+    fn lookup_rc16(&mut self, rw: LookupMode, flag: Self::Variable, value: Self::Variable) {
+        self.add_lookup(Lookup {
+            numerator: Signed::new(rw, Some(flag)),
```
mrmr1993, line 177: I personally find this less clear than the alternative. Perhaps it makes more sense to expand the `Lookup` structure to become:
```rust
Lookup<F> {
    mode: LookupMode,
    /// The number of times that this lookup value should be added to / subtracted from the lookup accumulator.
    magnitude_contribution: F,
    table_id: LookupTable,
    value: Vec<F>,
}
```

querolita: Ha! Good point 🤔

querolita: Addressed here! https://github.com/o1-labs/proof-systems/pull/1653

mrmr1993 (approved):

PR #1655, opened 2024-01-03
[easy] Reuse code for lookups inside Keccak constraints and witness

querolita:
This PR leverages the `LookupMode` enum to easily reuse the lookups code in the Keccak workflow for both witness and constraints stages (until this PR, the code was only reading lookups which were never written, so this PR fixes that pending issue).

### DIFF
--- a/optimism/src/keccak/lookups.rs
+++ b/optimism/src/keccak/lookups.rs
@@ -17,11 +17,11 @@ pub(crate) trait Lookups {
         + Clone;
     type Fp: std::ops::Neg<Output = Self::Fp>;
 
-    /// Adds a given lookup to the environment
+    /// Adds a given Lookup to the environment
     fn add_lookup(&mut self, lookup: Lookup<Self::Variable>);
 
     /// Adds all lookups of Self
-    fn lookups(&mut self);
+    fn lookups(&mut self, rw: LookupMode);
 }
 
 impl<Fp: Field> Lookups for KeccakEnv<Fp> {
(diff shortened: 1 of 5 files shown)


optimism/src/keccak/lookups.rs
```diff
@@ -33,7 +33,7 @@ impl<Fp: Field> Lookups for KeccakEnv<Fp> {
         self.lookups.push(lookup);
     }
 
-    fn lookups(&mut self) {
+    fn lookups(&mut self, rw: LookupMode) {
```
mrmr1993, line 36: This will only ever be used with read, surely? Revert?

optimism/src/keccak/lookups.rs
```diff
@@ -42,7 +42,8 @@ impl<Fp: Field> Lookups for KeccakEnv<Fp> {
             // Power of two corresponds to 2^pad_length
             // Pad suffixes correspond to 10*1 rule
             // Note: When FlagLength=0, TwoToPad=1, and all PadSuffix=0
-            self.add_lookup(Lookup::read_one(
+            self.add_lookup(Lookup::new(
```
mrmr1993, line 45: `Lookup::new` is barely less verbose than the record. It feels like it would be clearer to have it expanded and delete the helper, like
```rust
self.add_lookup(Lookup {
    mode: rw,
    magnitude: Self::Variable::one(),
    table_id: LookupTable::PadLookup,
    value: vec![...],
});
```

mrmr1993 (changes requested):

optimism/src/keccak/lookups.rs
querolita: absolutely, the sole reason that I changed `read_one` for `new(LookupMode::Read`, was that I thought `self.lookups(rw)` would be called from both `Write` and `Read` modes. Given that now I know modes should be fixed ahead of time and these ones will always be `Read`, it makes sense to revert this change. I will check everything I changed in this PR, because maybe I can just close it, or make it much smaller in scope.

querolita:
Closing PR because reusability between constraints and witness lookups were not necessary after all.

PR #1656, opened 2024-01-03
[easy] Helper for `Reset` lookup table

querolita:
This PR is the second one trying to simplify the `Lookup` trait for `KeccakEnv` by adding a helper to easily create lookups to the `Reset` lookup table. Not only does it clean up the interface, but it also allows for switch on/off functionality thanks to the flag argument of `Signed`.

### DIFF
--- a/optimism/src/keccak/lookups.rs
+++ b/optimism/src/keccak/lookups.rs
@@ -24,6 +24,15 @@ pub(crate) trait Lookups {
 
     /// Adds a lookup to the RangeCheck16 table
     fn lookup_rc16(&mut self, rw: LookupMode, flag: Self::Variable, value: Self::Variable);
+
+    /// Adds a lookup to the Reset table
+    fn lookup_reset(
+        &mut self,
+        rw: LookupMode,
+        flag: Self::Variable,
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):

optimism/src/keccak/lookups.rs
```diff
@@ -24,6 +24,15 @@ pub(crate) trait Lookups {
 
     /// Adds a lookup to the RangeCheck16 table
     fn lookup_rc16(&mut self, rw: LookupMode, flag: Self::Variable, value: Self::Variable);
+
+    /// Adds a lookup to the Reset table
+    fn lookup_reset(
+        &mut self,
+        rw: LookupMode,
```
mrmr1993, line 31: In this environment, we probably only ever want to read. Suggest removing this argument.

mrmr1993 (changes requested):

optimism/src/keccak/lookups.rs
querolita: yes, I am propagating the changes from base branches

PR #1658, opened 2024-01-03
[easy] Use read and write helpers inside MIPS interpreter

querolita:
This cleans up the interpreter code a bit.


querolita:
Addressed here:

https://github.com/o1-labs/proof-systems/pull/1653

PR #1665, opened 2024-01-03
Fix: `prev_block` was not being updated with value of `state_g` in-between rounds

querolita:
This PR includes a missing update that was not being performed inside the interpreter when running rounds. In particular, the output `state_g` was not being stored inside `self.prev_block`, so subsequent rounds could be loading old states instead.

For the rest (sponges), I have checked that the `prev_block` was being updated accordingly, this seems to be the only place where this was missing.

I realized about this issue when going through Misha's related comment in [this PR](https://github.com/o1-labs/proof-systems/pull/1637), asking about constraints between consecutive steps.

### DIFF
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -308,8 +308,8 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
         let state_g = iota.state_g();
 
         // Update columns
-        for i in 0..STATE_LEN {
-            self.write_column(KeccakColumn::IotaStateG(i), state_g[i]);
+        for (i, g) in state_g.iter().enumerate() {
+            self.write_column(KeccakColumn::IotaStateG(i), *g);
         }
         for i in 0..QUARTERS {
             self.write_column(KeccakColumn::RoundConstants(i), iota.rc(i));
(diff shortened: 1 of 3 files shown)


optimism/src/keccak/witness.rs
```diff
@@ -302,15 +303,18 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
...
 
-    fn run_iota(&mut self, state_f: &[u64], round: usize) {
+    fn run_iota(&mut self, state_f: &[u64], round: usize) -> Vec<u64> {
         let iota = Iota::create(state_f, round);
+        let state_g = iota.state_g();
 
         // Update columns
-        for i in 0..STATE_LEN {
-            self.write_column(KeccakColumn::IotaStateG(i), iota.state_g(i));
+        for (i, g) in state_g.iter().enumerate() {
```
volhovm, line 311: I suggest to add an additional assert that this value is equal to `STATE_LEN`.

volhovm (approved):

optimism/src/keccak/witness.rs
querolita: https://github.com/o1-labs/proof-systems/issues/1666

PR #1667, opened 2024-01-03
Reverting 1596

querolita:
~The recently merged PR https://github.com/o1-labs/proof-systems/pull/1596, added one dummy entry to the round constants so that when Keccak was running a sponge step with FlagRound=0, the lookup for the round constant would not fail. This made sense when lookups were "static", meaning they were always "added". Nonethless, thanks to the new functionality to select with flags, we can execute that lookup only when `is_round()=1`, and thus the entry is in the table. At any other time, when `is_sponge()=1` and `round()=0`, the lookup won't be triggered.~

~This simplifies the code as it sets the round constants to their "natural" shape with just 24 entries.~

_NOTE: if rows without Keccak constraints also need to instantiate Keccak columns to some default value, then having an alternative definition of `is_round()` will make sense, so it is not the negation of `is_sponge`. Or even, have another flag of `is_keccak`._ 

This PR was initially reverting https://github.com/o1-labs/proof-systems/pull/1596 because I thought that using the `is_round()` flag to select the round constants lookups would be enough to make all constraints satisfiable. Nonetheless, I have realised that we might have to keep those changes nonetheless in order to create an alternative definition of `is_round()` which is not the negation of `is_sponge()`, for the integration with the MIPS instructions. 

For this reason, the commits after [2e3ff20](https://github.com/o1-labs/proof-systems/pull/1667/commits/2e3ff204bf144add09bb47c496535820819eea8b), simply add the alternative definition. Maybe in the future we might want another flag for `is_mips()` instead, but this might increase the degree of all Keccak constraints. We can probably optimize it differently.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1322,6 +1322,7 @@ name = "kimchi_optimism"
 version = "0.1.0"
 dependencies = [
  "ark-bn254",
+ "ark-ec",
  "ark-ff",
  "ark-poly",
  "base64",
@@ -1340,6 +1341,7 @@ dependencies = [
  "os_pipe",
  "poly-commitment",
  "rand",
+ "rayon",
  "regex",
  "rmp-serde",
  "serde",
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -248,6 +248,19 @@ pub trait Literal: Sized {
     fn to_literal_ref(&self) -> Option<&Self::F>;
 }
 
+impl<F: Field> Literal for F {
+    type F = F;
+    fn literal(x: Self::F) -> Self {
+        x
+    }
+    fn to_literal(self) -> Result<Self::F, Self> {
+        Ok(self)
+    }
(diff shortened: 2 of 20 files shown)


querolita:
We might have to revert this one again if as soon as we integrate Keccak with the rest of MIPS, we realize that we need "dummy" witness values for the Keccak columns that satisfy all constraints. In particular, we would need to be able to allow that a row can be not a sponge but also not a round (meaning, a MIPS instruction row). In the Keccak work alone, round and sponges are mutually exclusive (it's either one or the other, both cannot be zero). And in that case, we would need again an alternative definition of `is_round` that is not the negation of `is_sponge()`, by adding again the inverse of the round number and claiming that round is not zero (which would be the default for dummy Keccak rows, when any other MIPS instruction is being executed).

querolita:
Closing for now, because I created smaller https://github.com/o1-labs/proof-systems/pull/1680 and https://github.com/o1-labs/proof-systems/pull/1681 for it.

Issue #1676, opened 2024-01-04
Check next states in-between Keccak steps


querolita:
This issue is meant to track what's missing from the big issue #1341

In particular, this issue will be considered done as soon as:

- [ ] Input/Output between Keccak steps (needs double check if step and columns are sufficient) -> After meeting with Matthew we observed that the witness code was not enough. We need to add "copy constraints" between Keccak steps, and with folding the natural way towards this is the memory model.

PR #1679, opened 2024-01-08
Constrain input and output between consecutive Keccak steps

querolita:
This PR uses memory-styled lookups to constrain information in-between consecutive Keccak steps. Meaning, that the output of the i-th step is the input of the (i+1)-th step. Although steps consecution was already handled within a `KeccakEnv` instance and the input/output witness was code was done, it was not constrained anywhere else. This PR fixes that.

In the code there's 2 TODO's accounting for the final output of the hash and the inputs for it. This should be done in a similar fashion to encode the communication channel between the syscalls and keccak.

Closes https://github.com/o1-labs/proof-systems/pull/1676

### DIFF
--- a/optimism/src/keccak/environment.rs
+++ b/optimism/src/keccak/environment.rs
@@ -30,7 +30,9 @@ pub struct KeccakEnv<Fp> {
     /// How many blocks are left to absrob (including current absorb)
     pub(crate) blocks_left_to_absorb: u64,
     /// What step of the hash is being executed (or None, if just ended)
-    pub(crate) curr_step: (u64, Option<KeccakStep>),
+    pub(crate) keccak_step: Option<KeccakStep>,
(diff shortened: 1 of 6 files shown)


optimism/src/keccak/environment.rs
```diff
@@ -30,7 +30,7 @@ pub struct KeccakEnv<Fp> {
     /// How many blocks are left to absrob (including current absorb)
     pub(crate) blocks_left_to_absorb: u64,
     /// What step of the hash is being executed (or None, if just ended)
-    pub(crate) curr_step: Option<KeccakStep>,
+    pub(crate) curr_step: (u64, Option<KeccakStep>),
```
mrmr1993, line 33: Nit: would be nice to separate these with names, I don't think they benefit from coupling

mrmr1993 (approved):

PR #1681, opened 2024-01-09
Alternative `is_round()` definition

querolita:
Uses `is_nonzero()` helper.

This way, there are no conflicting definitions of `is_round()`, because it no longer is the negation of `is_sponge()` (to allow for both to be false at the same time, in case it is a MIPS instruction row).


mrmr1993 (commented):
The rows won't intermingle, so this should be unnecessary. Suggest closing.

PR #1685, opened 2024-01-10
Trigger hash of `KeccakInterpreter` from Syscall

querolita:
This PR executes the hash of the `KeccakInterpreter` inside `request_preimage_write()`, which is called from the `interpret_rtype()` function, when the instruction is a `SyscallReadPreimage`. 

The idea here is to add a counter of the number of bytes of the preimage read so far inside the MIPS `Env`. When this value reaches the total length of the preimage, it's time to call the hash inside the interpreter of Keccak. This creates a new `KeccakEnv`, which triggers the witness creation code.

Some things remain to be done:
- Lookups between MIPS and Keccak (hash [here](https://github.com/o1-labs/proof-systems/pull/1687) and preimage [here](https://github.com/o1-labs/proof-systems/pull/1688))
> How and where to trigger Keccak constraints?
> - TBD when constraints of MIPS are also integrated with folding.

> NOTE: Right now I am doubting if defining the witness columns of Keccak as `keccak_state: KeccakColumns<E<Fp>>` is sufficient or not. In the sense that this keeps track of _one row_, but it is not a vector for all the steps. I wonder how this is handled in the MIPS side, and whether the same could be applied in Keccak. Specially considering what Matthew pointed out [here](https://github.com/o1-labs/proof-systems/pull/1681#pullrequestreview-1811231174) about rows not intermingling. 
> - After talking to Matthew, I will adapt the `KeccakInterpreter` trait so that each `request_preimage_write` calls single `step()` instead of the full `hash()`. The loop inside `hash()` checking if there are any other steps to run should be moved to a sub-loop inside the `main.rs` halt loop instead.

### DIFF
--- a/optimism/src/keccak/environment.rs
+++ b/optimism/src/keccak/environment.rs
@@ -33,6 +33,22 @@ pub struct KeccakEnv<Fp> {
     pub(crate) curr_step: Option<KeccakStep>,
 }
 
+impl<Fp: Field> Default for KeccakEnv<Fp> {
+    fn default() -> Self {
+        Self {
+            constraints: vec![],
+            lookups: vec![],
+            prev_block: vec![],
+            padded: vec![],
+            block_idx: 0,
+            keccak_state: KeccakColumns::default(),
+            pad_len: 0,
+            blocks_left_to_absorb: 0,
+            curr_step: None,
+        }
+    }
+}
+
 impl<Fp: Field> KeccakEnv<Fp> {
     pub fn write_column(&mut self, column: KeccakColumn, value: u64) {
         self.keccak_state[column] = Self::constant(value);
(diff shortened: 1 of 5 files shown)


optimism/src/mips/witness.rs
```diff
@@ -608,6 +609,16 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
             }
         }
         self.write_column(pos, actual_read_len);
+
+        // Update the total number of preimage bytes read so far
+        self.preimage_bytes_read = Some(self.preimage_bytes_read.unwrap() + actual_read_len);
```
dannywillems, line 614: Note: we should reset to 0 the number of bytes read.

optimism/src/keccak/environment.rs
```diff
@@ -35,6 +35,23 @@ pub struct KeccakEnv<Fp> {
     pub(crate) step_counter: u64,
 }
 
+impl<Fp: Field> Default for KeccakEnv<Fp> {
+    fn default() -> Self {
+        Self {
+            constraints: vec![],
```
dannywillems, line 41: Counter for the index of the Keccak we are computing.

dannywillems:
Note: we have to populate the Keccak polynomials (i.e. like for the interpreter)

dannywillems (approved):

PR #1686, opened 2024-01-11
Merge o1js-main into develop

mitschabaude:
merge PR parallel to o1js: https://github.com/o1-labs/o1js/pull/1365


poly-commitment/src/srs.rs
```diff
@@ -289,3 +290,37 @@ where
...
+            })
+            .collect();
+
+        const MISC: usize = 1;
+        let [h]: [G; MISC] = array::from_fn(|i| {
+            let mut h = Blake2b512::new();
+            h.update("srs_misc".as_bytes());
+            h.update((i as u32).to_be_bytes());
+            point_of_random_bytes(&m, &h.finalize())
+        });
```
rbonichon, line 318: Is the call to `array::from_fn` really warranted here? Or is it to be somewhat "future-proof" wrt to the `MISC` constant?

rbonichon (approved):
Only a nitpick. Otherwise LGTM

poly-commitment/src/srs.rs
mitschabaude: tbh I don't know - when I made this change I copied the code from `SRS::create` (further above) and added `.into_par_iter()` to make it parallel 😅

PR #1687, opened 2024-01-11
Communication channel of preimage key

querolita:
This PR adds lookups on the Keccak side and the MIPS side to communicate that the preimage key corresponds to the claimed hash. 

Here, because only 31 bytes are needed (the MSB of the hash is dropped to store the identifier 0x02 in most cases), these 248 bits can be stored as a single field element in the table. A leading value encoding the hash index is included.

A similar approach shall be followed for the preimage data, except that here they will be stored as bytes (https://github.com/o1-labs/proof-systems/pull/1688).

### DIFF
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -1,5 +1,8 @@
 use crate::cannon::{Page, State};
 use crate::keccak::interpreter::KeccakInterpreter;
+use crate::keccak::lookups::Lookups;
+use crate::keccak::ArithOps;
+use crate::mips::interpreter::{Lookup, LookupTable};
 use crate::{
     cannon::{
         Hint, Meta, Start, StepFrequency, VmConfiguration, PAGE_ADDRESS_MASK, PAGE_ADDRESS_SIZE,
@@ -18,6 +21,7 @@ use crate::{
 };
 use ark_ff::Field;
 use core::panic;
+use kimchi::o1_utils::Two;
 use log::{debug, info};
 use std::array;
 use std::fs::File;
@@ -58,6 +62,7 @@ pub struct Env<Fp> {
     pub preimage_oracle: PreImageOracle,
     pub preimage: Option<Vec<u8>>,
     pub preimage_bytes_read: Option<u64>,
+    pub preimage_key: Option<[u8; 32]>,
     pub keccak_env: Option<KeccakEnv<Fp>>,
 }
 
@@ -573,6 +578,7 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
             }
             let preimage = self.preimage_oracle.get_preimage(preimage_key).get();
             self.preimage = Some(preimage);
+            self.preimage_key = Some(preimage_key);
         }
 
         const LENGTH_SIZE: usize = 8;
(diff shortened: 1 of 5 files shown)


optimism/src/mips/witness.rs
```diff
@@ -616,6 +622,18 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         if self.preimage_bytes_read.unwrap() == preimage_len as u64 {
             let mut keccak_env = KeccakEnv::<Fp>::default();
             keccak_env.hash(self.preimage.as_ref().unwrap());
+            match self.preimage_key {
+                Some(preimage_key) => {
+                    let bytes31 = (1..32).fold(Fp::zero(), |acc, i| {
+                        acc * Fp::two_pow(8) + Fp::from(preimage_key[i])
+                    });
+                    keccak_env.add_lookup(Lookup::read_one(
```
dannywillems, line 630: Typo: one has to be considered as a read, and the other one as a write. Here we write on both channels.

optimism/src/mips/witness.rs
```diff
@@ -729,6 +747,7 @@ impl<Fp: Field> Env<Fp> {
             preimage_oracle,
             preimage: state.preimage,
             preimage_bytes_read: Some(0),
+            preimage_key: None,
```
dannywillems, line 750: Can be removed, see above.

optimism/src/mips/witness.rs
```diff
@@ -616,6 +622,18 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
             let mut keccak_env = KeccakEnv::<Fp>::default();
             keccak_env.hash(self.preimage.as_ref().unwrap());
+            match self.preimage_key {
+                Some(preimage_key) => {
+                    let bytes31 = (1..32).fold(Fp::zero(), |acc, i| {
+                        acc * Fp::two_pow(8) + Fp::from(preimage_key[i])
+                    });
+                    keccak_env.add_lookup(Lookup::read_one(
+                        LookupTable::SyscallLookup,
+                        vec![<KeccakEnv<Fp> as ArithOps>::constant_field(bytes31)],
```
dannywillems, line 632: Q: check the endianness.

optimism/src/mips/witness.rs
```diff
@@ -616,6 +622,18 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         if self.preimage_bytes_read.unwrap() == preimage_len as u64 {
             let mut keccak_env = KeccakEnv::<Fp>::default();
             keccak_env.hash(self.preimage.as_ref().unwrap());
+            match self.preimage_key {
+                Some(preimage_key) => {
+                    let bytes31 = (1..32).fold(Fp::zero(), |acc, i| {
```
dannywillems, line 627: Q: should we also check the hash type? (i.e. is `0x02`, `0x01`). It is indicated by the hint syscall IIRC. Maybe it has to be checked in the MIPS world/tape.

optimism/src/mips/witness.rs
dannywillems, line 630: Note: we need to enforce the order the preimage key, i.e. which keccak hash is referring it to.

```
i + r * keccak preimage key
```

querolita: We were confused, apparently there's indeed one read and one write. Maybe we were reading at a different part of the file.

querolita: the hash index, correct

optimism/src/mips/witness.rs
```diff
@@ -58,6 +62,7 @@ pub struct Env<Fp> {
     pub preimage_oracle: PreImageOracle,
     pub preimage: Option<Vec<u8>>,
     pub preimage_bytes_read: Option<u64>,
+    pub preimage_key: Option<[u8; 32]>,
```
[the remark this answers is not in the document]
querolita: But then we need to compute it again, is it worth it?

optimism/src/mips/witness.rs
dannywillems: Opening https://github.com/o1-labs/proof-systems/issues/1698. I don't remember our conclusion yesterday @querolita. Not important atm as we only have 0x02.

optimism/src/mips/witness.rs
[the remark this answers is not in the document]
dannywillems: Opening a follow-up issue.

PR #1694, opened 2024-01-16
Improve code & book documentation

volhovm:
This is the part with some small refactorings I was doing while working on the vinegar RFC. No code logic change intended, changes should be purely cosmetic.

Review by commit.

### DIFF
--- a/kimchi/src/proof.rs
+++ b/kimchi/src/proof.rs
@@ -620,6 +620,13 @@ pub mod caml {
         CamlF: From<F>,
     {
         fn from(pe: ProofEvaluations<PointEvaluations<Vec<F>>>) -> Self {
+            let first = pe.public.map(|x: PointEvaluations<Vec<F>>| {
+                // map both fields of each evaluation.
+                x.map(&|x: Vec<F>| {
+                    let y: Vec<CamlF> = x.into_iter().map(Into::into).collect();
+                    y
+                })
+            });
             let w = (
                 pe.w[0]
                     .clone()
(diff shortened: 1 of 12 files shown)


kimchi/src/proof.rs
```diff
@@ -735,85 +742,83 @@ pub mod caml {
...
+                    .map(|x| x.map(&|x| x.into_iter().map(Into::into).collect())),
+                range_check_lookup_selector: pe
+                    .range_check_lookup_selector
+                    .map(|x| x.map(&|x| x.into_iter().map(Into::into).collect())),
+                foreign_field_mul_lookup_selector: pe
+                    .foreign_field_mul_lookup_selector
+                    .map(|x| x.map(&|x| x.into_iter().map(Into::into).collect())),
+            };
+
+            (first, second)
```
querolita, line 821: Not sure I like this distinction

kimchi/src/prover.rs
```diff
@@ -1101,7 +1101,7 @@ where
 
         //~ 1. Evaluate the same polynomials without chunking them
         //~    (so that each polynomial should correspond to a single value this time).
-        let evals = {
+        let evals: ProofEvaluations<PointEvaluations<G::ScalarField>> = {
```
querolita, line 1104: Why is adding the explicit type needed now?

poly-commitment/src/evaluation_proof.rs
```diff
@@ -419,6 +446,7 @@ pub struct OpeningProof<G: AffineCurve> {
     pub z1: G::ScalarField,
     #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub z2: G::ScalarField,
+    /// A final folded commitment base
```
querolita, line 449: Folded?

kimchi/src/proof.rs
volhovm: Let me explain: 
1. The original function's `return` statement was `80` lines, including nested expressions and a lot of commas, so it was very confusing as to what is the return type at all. I like it when you can fast-forward to the end of the function and see what it does.
2. The `From::` conversions in this whole function are, arguably, quite non-intuitive -- it took me quite some time to understand how `map` (being a custom function! and not a generic iterator) is implemented. Figuring this out is further complicated by `From::` instances being picked implicitly, so it takes time to find them. This is why I explicitly typed `first` so that one can see by example what's happening here.

TLDR I just spent 15 minutes trying to figure out what's going on here and thought this might help...

kimchi/src/prover.rs
volhovm: I guess readability. `evals` sounds like it can be any evaluations -- `ProofEvaluations<G::ScalarField>` or `PointEvaluations<somethingelse>`. Ultimately I don't know -- I probably added it because I found it confusing... But I don't think explicit typing hurts here, does it?

poly-commitment/src/evaluation_proof.rs
volhovm: Yeah, but not in the instance-folding way. At each (1/log N) step of bulletproofs we have several vectors that get "folded in half", so that's what I meant. I can say "log-folded" or "IPA-folded" if this is more clear?

querolita: ah yeah, much more

querolita (approved):

PR #1697, opened 2024-01-16
Split hash function into individual steps inside the main loop

querolita:
This PR partially addresses issue https://github.com/o1-labs/proof-systems/issues/1695. 

Before this PR, the syscalls used to run the hash for the whole preimage in one go. But this approach overwrites individual Keccak steps witness data after every step update. Conversely, this PR solves that problem essentially thanks to moving parts of existing code around, so that the loop inside the old `hash()` function is executed as a sub-loop inside the halting condition loop inside `main.rs`. That way, a subsequent PR will be able to propagate the witness of each individual step, populating the folded instance for Keccak. In addition, now the function `request_preimage_write()` will only create an "empty" but non-`None` instance of `KeccakEnv` using as input the hash index and the preimage data. The fact that now `env.keccak_env.is_some()` will be used inside the loop to trigger subsequent Keccak steps (meaning, the `KeccakEnv::new()` function now returns a `KeccakEnv` containing the initialized values that used to be performed at the beginning of the old `hash()` function before the while loop).

### DIFF
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -1,5 +1,4 @@
 use crate::cannon::{Page, State};
-use crate::keccak::interpreter::KeccakInterpreter;
 use crate::keccak::lookups::Lookups;
 use crate::keccak::ArithOps;
 use crate::mips::interpreter::{Lookup, LookupTable};
--- a/optimism/src/mips/witness.rs
+++ b/optimism/src/mips/witness.rs
@@ -624,7 +624,7 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
             let mut keccak_env = KeccakEnv::<Fp>::new(self.hash_count);
             keccak_env.hash(self.preimage.as_ref().unwrap());
 
-            // Write preimage bytes to the communication channel
+            // COMMUNICATION CHANNEL: Write preimage bytes
             let preimage = self.preimage.as_ref().unwrap();
             for (i, byte) in preimage.iter().enumerate() {
                 keccak_env.add_lookup(Lookup::write_one(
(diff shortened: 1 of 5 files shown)


optimism/src/mips/witness.rs
```diff
@@ -654,9 +654,12 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
             }
             self.keccak_env = Some(keccak_env);
         }
-        // Reset Keccak environment
-        self.preimage_bytes_read = Some(0);
+        // Reset environment
```
dannywillems, line 657: Should not happen here. It should be reset in the previous scope.

optimism/src/mips/witness.rs
```diff
@@ -654,9 +654,12 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        self.preimage_bytes_read = None;
+        self.preimage = None;
+        self.preimage_key = None;
         self.hash_count += 1;
```
dannywillems, line 661: hash count should be increase in the previous scope, when we read the whole preimage.

optimism/src/mips/witness.rs
```diff
@@ -655,8 +655,7 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
             self.keccak_env = Some(keccak_env);
 
             // Reset environment
-            self.preimage_bytes_read = None;
-            self.preimage = None;
+            self.preimage_bytes_read = Some(0);
```
dannywillems, line 658: Why is it an option and not `0`?

dannywillems: Opening https://github.com/o1-labs/proof-systems/issues/1702

dannywillems (approved):
I run the vm on my laptop, doesn't break anything.

optimism/src/mips/witness.rs
querolita: Yeah good point... Actually it makes more sense that preimage_bytes_read gets initialized to Some(0) when reading the preimage key, and then here this becomes a None... Otherwise there's no difference with just an int.

PR #1706, opened 2024-01-18
[easy] Create `KeccakProof` struct

querolita:
This PR creates the struct `KeccakProof`, replicating the struct `Proof` inside the mips module. 

Disclaimer: we can probably avoid code duplication by making the contents of `mips/proof.rs` more generic. But for now, let's create an independent one inside the keccak module, make sure it works, and refactor later. Created issue to track it https://github.com/o1-labs/proof-systems/issues/1705.


dannywillems (approved):
Approving to go further. Let's see the upcoming PRs.

PR #1708, opened 2024-01-18
Implement iterators for `KeccakColumns`

querolita:
This PR replicates what is being done in `mips` for `WitnessColumns`.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -12,8 +12,12 @@ use kimchi::circuits::polynomials::keccak::constants::{
     THETA_REMAINDER_C_LEN, THETA_REMAINDER_C_OFF, THETA_SHIFTS_C_LEN, THETA_SHIFTS_C_OFF,
     THETA_STATE_A_LEN, THETA_STATE_A_OFF,
 };
+use rayon::iter::{FromParallelIterator, IntoParallelIterator, ParallelIterator};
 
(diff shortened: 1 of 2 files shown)


optimism/src/keccak/column.rs
```diff
@@ -274,3 +278,148 @@ impl<T: Clone> IndexMut<KeccakColumn> for KeccakColumns<T> {
...
+}
+
+impl<G: Send + std::fmt::Debug> FromParallelIterator<G> for KeccakColumns<G> {
+    fn from_par_iter<I>(par_iter: I) -> Self
+    where
+        I: IntoParallelIterator<Item = G>,
+    {
+        let mut iter_contents = par_iter.into_par_iter().collect::<Vec<_>>();
+        let next = iter_contents
+            .drain(iter_contents.len() - ZKVM_KECCAK_COLS_NEXT..)
```
dannywillems, line 317: I would guess it is fine as the length is modified. We'll see later with the final code. Don't know if there is a better way to achieve this.

dannywillems (approved):
Approving, but let's see the final code.

optimism/src/keccak/column.rs
querolita: Yeah, that was my impression, `drain()` acts as a repeated `pop()`. But in any case, we will find out in the small test I want to add to this file.

PR #1710, opened 2024-01-18
Implement fold for `KeccakProofInputs`

querolita:
This replicates what's done in `mips` for `ProofInputs`.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -338,7 +338,7 @@ impl<G: Send + std::fmt::Debug> FromParallelIterator<G> for KeccakColumns<G> {
         I: IntoParallelIterator<Item = G>,
     {
         let mut iter_contents = par_iter.into_par_iter().collect::<Vec<_>>();
-        let next: [G; ZKVM_KECCAK_COLS_NEXT] = iter_contents
+        let next = iter_contents
             .drain(iter_contents.len() - ZKVM_KECCAK_COLS_NEXT..)
             .collect::<Vec<G>>()
             .try_into()
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -313,7 +313,7 @@ impl<G: Send + std::fmt::Debug> FromParallelIterator<G> for KeccakColumns<G> {
         I: IntoParallelIterator<Item = G>,
     {
         let mut iter_contents = par_iter.into_par_iter().collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


optimism/src/keccak/column.rs
```diff
@@ -313,7 +338,7 @@ impl<G: Send + std::fmt::Debug> FromParallelIterator<G> for KeccakColumns<G> {
         I: IntoParallelIterator<Item = G>,
     {
         let mut iter_contents = par_iter.into_par_iter().collect::<Vec<_>>();
-        let next = iter_contents
+        let next: [G; ZKVM_KECCAK_COLS_NEXT] = iter_contents
```
dannywillems, line 341: I would suppose the typing annotation is required, therefore you added it?

optimism/src/keccak/proof.rs
```diff
@@ -1,17 +1,27 @@
...
-    _evaluations: KeccakColumns<Vec<G::ScalarField>>,
+    evaluations: KeccakColumns<Vec<G::ScalarField>>,
 }
 
 impl<G: KimchiCurve> Default for KeccakProofInputs<G> {
     fn default() -> Self {
         KeccakProofInputs {
-            _evaluations: KeccakColumns {
+            evaluations: KeccakColumns {
                 hash_index: (0..1 << 15).map(|_| G::ScalarField::zero()).collect(),
```
dannywillems, line 25: Note: we should have a global variable, shared with proof in mips.

optimism/src/keccak/proof.rs
```diff
@@ -49,3 +59,50 @@ pub struct KeccakProof<G: KimchiCurve, OpeningProof: OpenProof<G>> {
     _zeta_omega_evaluations: KeccakColumns<G::ScalarField>,
     _opening_proof: OpeningProof,
 }
+
+pub fn fold<
```
dannywillems, line 63: Can you comment succinctly what the function does please?

optimism/src/keccak/column.rs
querolita: Oh it was probably not needed and I just forgot it was there.

optimism/src/keccak/proof.rs
querolita: At this point this is just aggregating instances together with a random combination. It does not contain folding equations per se, those will be located in the folding part of the project (still to be merged completely).

dannywillems (approved):

PR #1715, opened 2024-01-18
zkVM: share domain size

### LINKED ISSUE #1714: Global variable, shared with proof in mipsfor the domain size.
Note: we should have a global variable, shared with proof in mips.

_Originally posted by @dannywillems in https://github.com/o1-labs/proof-systems/pull/1710#discussion_r1457791755_

dannywillems:
Fix #1714


querolita (approved):
Ah nice yeah

PR #1716, opened 2024-01-18
Mimoo/snarky2

fabrizio-m:


poseidon/src/permutation.rs
```diff
@@ -42,6 +42,23 @@ pub fn full_round<F: Field, SC: SpongeConstants>(
     }
 }
 
+/// Same as full round but works better with arrays.
+pub fn full_round2<F: Field, SC: SpongeConstants>(
```
dannywillems, line 46: Is it really useful?
Also, the same is not very concrete. I would get rid of this if not strictly required.

kimchi/src/snarky/union_find.rs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        };
+
+        self.parent[x_root] = y_root;
+        Some(y_root)
+    }
+}
+
+#[test]
+fn it_works() {
```
dannywillems, line 79: Not sure it is the best test name.

poly-commitment/src/lib.rs
```diff
@@ -83,6 +83,10 @@ pub trait SRS<G: CommitmentCurve> {
         plnm: &Evaluations<G::ScalarField, D<G::ScalarField>>,
         rng: &mut (impl RngCore + CryptoRng),
     ) -> BlindedCommitment<G>;
+    ///for now needed by snarky-rs
```
dannywillems, line 86: Would be nice to add more documentation. It is not enough to say "for now needed by snarky-rs". For which reason it is needed?

dannywillems: Why not exposing simply create_trusted_setup?

poseidon/src/permutation.rs
dannywillems: See https://github.com/o1-labs/proof-systems/pull/1720

poly-commitment/src/lib.rs
dannywillems: Let's do it in a follow-up. We should refactorize the SRS structure/type/traits. We should have different type of SRS. One when there is a toxic waste, one when it is transparent.

dannywillems (approved):
I didn't test snarky as it seems it will only be used for folding, and only a small part of it.
I am not totally happy with the interface nor the code of Snarky, but we have to move on, and it seems we don't have time to polish it now.

See https://github.com/o1-labs/proof-systems/pull/1720 for a CI fix + removing `full_round2`.
I opened some follow-up issue.

poly-commitment/src/lib.rs
dannywillems: See https://github.com/o1-labs/proof-systems/issues/1721

dannywillems:
I cherry-picked two commits: https://github.com/o1-labs/proof-systems/pull/1716/commits/9803da23b80b7cb7d4cb7800d20840e59087d898 and https://github.com/o1-labs/proof-systems/pull/1716/commits/e8b1d0b7dd56889aaa66506e31eb2003917e1016

PR #1719, opened 2024-01-18
Test for Keccak proof

querolita:
This PR adds a test for the Keccak proof.


dannywillems (approved):
This is not constrained. Therefore, it makes sense that random values work.

PR #1722, opened 2024-01-19
[easy] Avoid use of `rc` and use `round_constants` instead

querolita:
As Matthew said, characters are free.


dannywillems (approved):
Long life to free characters.

PR #1724, opened 2024-01-19
Create `Input` and `Output` Keccak columns

querolita:
This PR removes `ThetaStateA`, `IotaStateG`, `SpongeOldState`, and `SpongeXorState`, substituting them for `Input` and `Output`. They correspond to the variables in locations `curr[0..100]` and `next[0..100]`. 

Otherwise, when variables are being used in constraints, the value written in the input and output of consecutive steps will not have the same format making it impossible to fold the instance.

Related to https://github.com/o1-labs/proof-systems/issues/1723

Note: it also renames single-indices in environment variables to `idx` to avoid confusion with `i` (normally used to index only from 0 to 4), so when the above issue is addressed fully it becomes more readable.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -6,11 +6,10 @@ use kimchi::circuits::polynomials::keccak::constants::{
     PIRHO_DENSE_E_OFF, PIRHO_DENSE_ROT_E_LEN, PIRHO_DENSE_ROT_E_OFF, PIRHO_EXPAND_ROT_E_LEN,
     PIRHO_EXPAND_ROT_E_OFF, PIRHO_QUOTIENT_E_LEN, PIRHO_QUOTIENT_E_OFF, PIRHO_REMAINDER_E_LEN,
     PIRHO_REMAINDER_E_OFF, PIRHO_SHIFTS_E_LEN, PIRHO_SHIFTS_E_OFF, QUARTERS, RATE_IN_BYTES,
-    SPONGE_BYTES_OFF, SPONGE_NEW_STATE_OFF, SPONGE_OLD_STATE_OFF, SPONGE_SHIFTS_OFF, STATE_LEN,
-    THETA_DENSE_C_LEN, THETA_DENSE_C_OFF, THETA_DENSE_ROT_C_LEN, THETA_DENSE_ROT_C_OFF,
-    THETA_EXPAND_ROT_C_LEN, THETA_EXPAND_ROT_C_OFF, THETA_QUOTIENT_C_LEN, THETA_QUOTIENT_C_OFF,
-    THETA_REMAINDER_C_LEN, THETA_REMAINDER_C_OFF, THETA_SHIFTS_C_LEN, THETA_SHIFTS_C_OFF,
-    THETA_STATE_A_LEN, THETA_STATE_A_OFF,
+    SPONGE_BYTES_OFF, SPONGE_NEW_STATE_OFF, SPONGE_SHIFTS_OFF, STATE_LEN, THETA_DENSE_C_LEN,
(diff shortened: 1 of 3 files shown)


optimism/src/keccak/environment.rs
```diff
@@ -220,30 +221,30 @@ pub(crate) trait KeccakEnvironment {
 
     fn two_to_pad(&self) -> Self::Variable;
 
-    fn in_padding(&self, i: usize) -> Self::Variable;
+    fn in_padding(&self, idx: usize) -> Self::Variable;
```
dannywillems, line 224: Long life to free characters!

optimism/src/keccak/column.rs
```diff
@@ -34,7 +33,7 @@ pub enum KeccakColumn {
     FlagsBytes(usize),                        // 136 boolean values
     PadSuffix(usize),                         // 5 values with padding suffix
     RoundConstants(usize),                    // Round constants
-    ThetaStateA(usize, usize, usize),         // Round Curr[0..100)
+    Input(usize),                             // Curr[0..100) either ThetaStateA or SpongeOldState
```
dannywillems, line 36: Nit: we will need to merge the Keccak RFC at the end. Only a personal taste: I think updating the RFC while we're updating the code might be good.

optimism/src/keccak/column.rs
```diff
@@ -49,12 +48,10 @@ pub enum KeccakColumn {
     PiRhoExpandRotE(usize, usize, usize),     // Round Curr[1065..1165)
     ChiShiftsB(usize, usize, usize, usize),   // Round Curr[1165..1565)
     ChiShiftsSum(usize, usize, usize, usize), // Round Curr[1565..1965)
-    IotaStateG(usize),                        // Round Next[0..100)
-    SpongeOldState(usize),                    // Sponge Curr[0..100)
     SpongeNewState(usize),                    // Sponge Curr[100..200)
     SpongeBytes(usize),                       // Sponge Curr[200..400)
     SpongeShifts(usize),                      // Sponge Curr[400..800)
-    SpongeXorState(usize),                    // Absorb Next[0..100)
+    Output(usize),                            // Next[0..100) either IotaStateG or SpongeXorState
```
dannywillems, line 54: Note that in [witness](https://github.com/o1-labs/proof-systems/blob/dfd5056d46a5960f8933d5979d4dcb6aaa11afc8/optimism/src/keccak/witness.rs#L147) we could also rename.

optimism/src/keccak/column.rs
```diff
@@ -49,12 +48,10 @@ pub enum KeccakColumn {
     PiRhoExpandRotE(usize, usize, usize),     // Round Curr[1065..1165)
     ChiShiftsB(usize, usize, usize, usize),   // Round Curr[1165..1565)
     ChiShiftsSum(usize, usize, usize, usize), // Round Curr[1565..1965)
-    IotaStateG(usize),                        // Round Next[0..100)
```
dannywillems, line 52: Leaving a thought here.
Keccak is made of a non linear mapping ($\chi$), followed by a linear map ($\theta$), a transposition ($\pi$) and a translation ($\rho$), and ending with the round constants ($\iota$) - see [reference](https://keccak.team/files/Keccak-reference-3.0.pdf).
It is nearly the same structure as other zk-friendly permutations (Poseidon, Anemoi, Griffin, ...) IIUC. I wish we can find a way to uniformize it. I find it confusing to call it with Greek letters, I'm always messing up with it. If you have a way to remember the order, I would be happy to hear it.

dannywillems (commented):
We can also rename in [witness.rs](https://github.com/o1-labs/proof-systems/blob/master/optimism/src/keccak/witness.rs#L193). `state_a` becomes `input`, `state_g` becomes `output`.

dannywillems (approved):
I would like to see some regression tests, being sure that the values are still computing correctly after the changes.

optimism/src/keccak/column.rs
querolita: Agree

PR #1725, opened 2024-01-19
ivc

fabrizio-m:


dannywillems (approved):
I will approve to get it into master and move on. However, I would have appreciated:
- having a link to the paper Nova at the top level. 
- add an example on how to use the code.
- add a test (that can be merged with the example, even though we can have code in a comment and thanks to [rustdoc](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) always be sure it compiles.

This should be addressed in follow-ups.

PR #1728, opened 2024-01-20
Keccak/doc: adding link to "official" pseudo-code

dannywillems:
I often need to go on the website to find it, and I always spent some minutes finding the link.
Better to have it at the top of the file IMO.

### DIFF
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -1,3 +1,6 @@
+/// This file contains the implementation of the witness for the Keccak permutation.
+/// For a pseudo code implementation of Keccap-f, see
+/// https://keccak.team/keccak_specs_summary.html
 use super::{
     column::KeccakColumn,
     environment::KeccakEnv,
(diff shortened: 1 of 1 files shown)


optimism/src/keccak/witness.rs
```diff
@@ -200,6 +203,17 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
         self.prev_block = state_g;
     }
 
+    /// Computing
+    /// ```text
+    /// for x in 0…4
+    ///   C[x] = A[x,0] xor A[x,1] xor \
+    ///          A[x,2] xor A[x,3] xor \
+    ///          A[x,4]
+    /// for x in 0…4
```
querolita, line 212: Here maybe you want to do 0..=4 so it is clearer

querolita (approved):
Good idea, it's been so long that I am working with Keccak that I recall these steps from the top of my head. But I didn't realize it could be hard to review for someone else without comments in the code. Thanks for adding them! 

Nonetheless, maybe reconsider the way you are writing the ranges ;)

PR #1731, opened 2024-01-22
Book: typo foreign field arithmetics

dannywillems:


book/src/kimchi/foreign_field_add.md
```diff
@@ -120,7 +120,7 @@ bits  0..............87|88...........175|176...........263
 
 r  =  (-------r0-------|-------r1-------|-------r2-------)
 +
-g  =  (-------g0-------|-------g1-------|-------g2-------)
+b  =  (-------b0-------|-------b1-------|-------b2-------)
```
querolita, line 123:
```suggestion
g  =  (-------g0-------|-------g1-------|-------g2-------)
```

querolita: We are intentionally using `g` to name $2^{264}$, so I suggest we don't use `b` here as it is misleading. Instead, maybe let's add a clarification of what `g` is meant to be.

querolita (changes requested):

volhovm:
Added quite a few clarifications on the text, but ultimately still want to understand the trick with the `u` range check before merging...

volhovm:
Clarified it. Ready now. cc @querolita

book/src/kimchi/foreign_field_add.md
```diff
@@ -52,15 +52,15 @@ f  =  (-------f0-------|-------f1-------|-------f2-------)
...
 
 In order to perform this operation in parts, we first take a look at the least significant limb, which is the easiest part. This means we want to know how to compute $r_0$. First, if the addition of the bits in $a_0$ and $b_0$ produce a carry (or borrow) bit, then it should propagate to the second limb. That means one has to subtract $2^{88}$ from $a_0 + b_0$, add $1$ to $a_1 + b_1$ and set the low carry flag $c_0$ to 1 (otherwise it is zero).  Thus, the equation for the lowest bit is
 
 $$a_0 + s \cdot b_0 = q \cdot f_0 + r_0 + c_0 \cdot 2^{88}$$
 
 Or put in another way, this is equivalent to saying that $a_0 + b_0 - q \cdot f_0 - r_0$ is a multiple of $2^{88}$ (or, the existence of the carry coefficient $c_0$).
 
-This kind of equation needs an additional check that the carry coefficient $c_0$ is a correct carry value (`0`, `1`, or `-1`). We will use this idea for the remaining limbs as well.
+This kind of equation needs an additional check that the carry coefficient $c_0$ is a correct carry value (belonging to $\{-1,0,1\}$). We will use this idea for the remaining limbs as well.
```
querolita, line 63: Sometimes markdown does not render the latex `\{...\}` notation correctly, so I wouldn't be surprised if some visualizers break this line.

book/src/kimchi/foreign_field_add.md
```diff
@@ -250,7 +255,7 @@ $$add(add(add(a,b),c),d)$$
 | 30..33 | `multi-range-check` | $a+b+c+d$     |
 | 34..37 | `multi-range-check` | bound         |
 
-Nonetheless, such an exhaustive set of checks are not necessary for completeness nor soundness. In particular, only the very final range check for the bound is required. Thus, a shorter gadget that is equally valid and takes $(8*n+4)$ fewer rows could be possible if we can assume that the inputs of each addition are correct foreign field elements. It would follow the next layout (with inclusive ranges):
+Nonetheless, such an exhaustive set of checks are not necessary for completeness nor soundness. In particular, only the very final range check for the bound is required. Thus, a shorter gadget that is equally valid and takes $(8\cdotn+4)$ fewer rows could be possible if we can assume that the inputs of each addition are correct foreign field elements. It would follow the next layout (with inclusive ranges):
```
querolita, line 258: you probably need to add a space between cdot and n

querolita (approved):

book/src/kimchi/foreign_field_add.md
volhovm: It works in my local `mdbook` via `make build; make all` -- pretty sure it doesn't work on github, on github you need to wrap it into ``: `$\{-1,0,1\}$` :)

PR #1738, opened 2024-01-22
Reduce Keccak columns by 145

querolita:
Closes https://github.com/o1-labs/proof-systems/issues/1709.

This PR leverages the fact that plenty of columns are only accessed during sponge rows, which only use 800 of the 1965 available `curr` columns. Storing those coefficients in `curr` saves us 145 columns (meaning additional field elements in the witness) whose consumed space is redundant otherwise.

As a side effect, this reduces the lines of code needed in the proof of Keccak.


dannywillems (approved):
I would add at the top level of the file a comment showing the layout used by both step types.

PR #1749, opened 2024-01-24
[easy] Add docs for `keccak/column.rs`

querolita:
This PR adds some more comments to the `keccak/column.rs` module. 

As part of the documentation and spec process, this PR also introduced a renaming from `KeccakColumns` to `KeccakWitness` for better clarity: now `KeccakColumn` refers to the aliases, used to identify variables, whereas `KeccakWitness` contains the actual field elements composing the witness of the circuit.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -1,5 +1,6 @@
-use std::ops::{Index, IndexMut};
-
+/// This module defines the custom columns used in the Keccak witness, which
+/// are aliases for the actual Keccak witness columns also defined here.
+use super::{ZKVM_KECCAK_COLS_CURR, ZKVM_KECCAK_COLS_NEXT};
 use ark_ff::{One, Zero};
 use kimchi::circuits::polynomials::keccak::constants::{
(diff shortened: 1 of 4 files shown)


optimism/src/keccak/column.rs
```diff
@@ -76,29 +80,29 @@ pub enum KeccakColumn {
...
 /// The columns are shared between the Sponge and Round steps.
-/// The step index and hash index are shared between the Sponge and Round step
-/// type.
+/// The hash and step indices are shared between both modes.
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub struct KeccakColumns<T> {
-    pub hash_index: T,
-    pub step_index: T,
+pub struct KeccakWitness<T> {
+    pub hash_index: T,                        // Which hash this is inside the circuit
```
dannywillems, line 89: Follow-up: I would explain somewhere why the index is important. We did not have it before talking about the syscall communication channel. It can be in the code or in the RFC (and we point to the RFC here).

dannywillems (approved):

optimism/src/keccak/column.rs
querolita: I will address that in the following chunk of docs

PR #1750, opened 2024-01-24
Add constraint count and degree in docs

querolita:
This PR improves the documentation of the `keccak/constraints.rs` module. In particular, it includes the constraint count for better traceability, including their degree.

Only 5 constraints have degree 3, so we shall expect that after quadricization is applied, we will need just a few more witness columns and constraints to make them degree 2.

Alternatively, we can intentionally shift the beginning of the sponge columns (since they have about 1k less used columns than the round), so that the `quotient_c` values are located in witness columns which would be set to zero regardless in sponge steps. That way, we could rephrase the constraint to be instead of `is_round * is_boolean (quotient_c)` to just `is_boolean(quotient_c)`. Opening issue to trace this optimization -> https://github.com/o1-labs/proof-systems/issues/1751


dannywillems (approved):
I didn't check the numbers, lgtm for the information we add to the doc.

PR #1752, opened 2024-01-24
Simplify type annotations

dannywillems:
Not a big change, but ease the reading of and reasoning on the function.

### DIFF
--- a/poly-commitment/src/pairing_proof.rs
+++ b/poly-commitment/src/pairing_proof.rs
@@ -91,23 +91,22 @@ impl<
 {
     type SRS = PairingSRS<Pair>;
 
-    fn open<EFqSponge, RNG, D: EvaluationDomain<<G as AffineCurve>::ScalarField>>(
+    fn open<EFqSponge, RNG, D: EvaluationDomain<F>>(
         srs: &Self::SRS,
         _group_map: &<G as CommitmentCurve>::Map,
         plnms: &[(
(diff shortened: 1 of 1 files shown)


poly-commitment/src/pairing_proof.rs
```diff
@@ -91,23 +91,22 @@ impl<
...
+        elm: &[F],          // vector of evaluation points
+        polyscale: F,       // scaling factor for polynoms
+        _evalscale: F,      // scaling factor for evaluation point powers
+        _sponge: EFqSponge, // sponge
         _rng: &mut RNG,
     ) -> Self
     where
-        EFqSponge:
-            Clone + FqSponge<<G as AffineCurve>::BaseField, G, <G as AffineCurve>::ScalarField>,
+        EFqSponge: Clone + FqSponge<<G as AffineCurve>::BaseField, G, F>,
```
volhovm, line 109: Unless you disagree, can you also please add the constraint `F: <G as AffineCurve>::ScalarField` that was removed? I feel like it was not on purpose if you wanted to simplify things... `F` is definitely better for readability, but unconstraining it to be any field is more confusing since then looking at type definition would make me think "this `F` could be anything".

volhovm (approved):
Looks good!

poly-commitment/src/pairing_proof.rs
volhovm: Ah I see! Makes sense :+1:

PR #1757, opened 2024-01-24
Add docs to interpreter and change location where `PadSuffix` columns were assigned

querolita:
While adding doc comments to the `keccak/interpreter.rs` file, I realized that the `PadSuffix` columns of the witness were being assigned in every absorb sponge, even when they are only used in padding absorbs. This is not really a bug, given that constraints and lookups would still hold. But I thought it made more sense to do it only in padding rows.

### DIFF
--- a/optimism/src/keccak/interpreter.rs
+++ b/optimism/src/keccak/interpreter.rs
@@ -1,3 +1,5 @@
+//! This module defines the Keccak interpreter in charge of triggering the Keccak workflow
+
 /// Variants of Keccak steps available for the interpreter
 #[derive(Clone, Debug, PartialEq, Copy)]
 pub enum KeccakStep {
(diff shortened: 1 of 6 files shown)


optimism/src/keccak/interpreter.rs
```diff
@@ -15,13 +17,13 @@ pub enum Sponge {
 /// Order of absorb steps in the computation depending on the number of blocks to absorb
 #[derive(Clone, Debug, PartialEq, Copy)]
 pub enum Absorb {
-    First,
-    Middle,
-    Last,
-    FirstAndLast,
+    First,        // Also known as the root absorb
```
dannywillems, line 20: Nice, I like it.

optimism/src/keccak/witness.rs
```diff
@@ -179,10 +187,6 @@ impl<Fp: Field> KeccakInterpreter for KeccakEnv<Fp> {
         for (idx, value) in shifts.iter().enumerate() {
             self.write_column(KeccakColumn::SpongeShifts(idx), *value);
         }
-        let pad_blocks = pad_blocks::<Fp>(self.pad_len as usize);
```
dannywillems, line 182: Could have been detected with a unit test :grimacing: :)

querolita: Not really, as I said it makes no difference in the constraints. It's just a change for better readability and coherence.

PR #1747, opened 2024-01-23
[master] Bump up to arkwork 0.4.2 - DO NOT MERGE/REVIEW

dannywillems:
Benchmark:
```shell
cargo criterion -p kimchi --bench proof_criterion
```

Before:
```
Proof creation/proof creation (SRS size 2^{16}, 1014 gates)                                                                          
                        time:   [852.95 ms 859.61 ms 867.67 ms]
                        change: [+0.3236% +1.3483% +2.5489%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Benchmarking Proof creation/proof creation (SRS size 2^{16}, 16374 gates): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 16.9s.
Proof creation/proof creation (SRS size 2^{16}, 16374 gates)                                                                          
                        time:   [1.6794 s 1.6851 s 1.6919 s]
                        change: [-5.7030% -5.2167% -4.7135%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking Proof verification/proof verification (SRS size 2^{16}, 1014 gates): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.9s, or reduce sample count to 70.
Proof verification/proof verification (SRS size 2^{16}, 1014 gates)                                                                            
                        time:   [67.749 ms 68.204 ms 68.678 ms]
                        change: [-7.2764% -6.2600% -5.2345%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking Proof verification/proof verification (SRS size 2^{16}, 16374 gates): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.9s, or reduce sample count to 70.
Proof verification/proof verification (SRS size 2^{16}, 16374 gates)                                                                            
                        time:   [67.834 ms 68.301 ms 68.802 ms]
                        change: [-7.5480% -6.5346% -5.4975%] (p = 0.00 < 0.05)
                        Performance has improved.
```
After (commit 70a1d2325dd2ec9dffeba96acfed6401f4ea42f5):
```
Benchmarking Proof creation/proof creation (SRS size 2^{16}, 1014 gates): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 8.4s.
[...]

PR #1759, opened 2024-01-25
SRS: regression tests for Pallas and Vesta

dannywillems:
While updating to arkworks 0.4.x, I realised the default encoding changed. Adding this as a regression test.
See https://github.com/o1-labs/proof-systems/pull/1747/commits/d680bb4a449bc53a4451cc5a42ff5f5b19ac5ce5

### DIFF
--- a/kimchi/src/precomputed_srs.rs
+++ b/kimchi/src/precomputed_srs.rs
@@ -40,9 +40,42 @@ where
 mod tests {
     use super::*;
 
+    use ark_ec::AffineCurve;
     use ark_ff::PrimeField;
     use ark_serialize::Write;
+    use hex;
     use mina_curves::pasta::{Pallas, Vesta};
+    use std::collections::HashMap;
+
+    fn test_regression_serialization_srs_with_generators<G: AffineCurve>(exp_output: String) {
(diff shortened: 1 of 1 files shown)


kimchi/src/precomputed_srs.rs
```diff
@@ -40,9 +40,40 @@ where
...
+            lagrange_bases,
+        };
+        let srs_bytes = rmp_serde::to_vec(&srs).unwrap();
+        let output = hex::encode(srs_bytes.clone());
+        assert_eq!(output, exp_output)
+    }
+
+    #[test]
+    fn test_regression_serialization_srs_with_generators_vesta() {
+        // This is the same than Pallas as we encode the coordinate x only.
```
volhovm, line 66: "Same **as** Pallas" I assume?

volhovm (approved):
Nice idea! I think maybe it would be useful to add a small comment on how the long magic numbers were obtained...  like a comment with a date? The `arkworks` version? I assume this is for the `arkworks` version upgrade?

kimchi/src/precomputed_srs.rs
dannywillems: Updated.

PR #1762, opened 2024-01-25
Preliminary work for zkVM lookups

querolita:
This PR is mainly a code reorganization of lookup-related code that was already present in other parts of the zkvm code, and aggregate it into one lookup module instead.

Apart from that, it lays the foundations to start keeping track of lookups on the mips side as well (they were only being saved in keccak so far). ~This will be subject to changes, as we decide when and how these lookups should be added -> should this happen in the interpreter/witness side or constraints? Also, there's the question of whether lookups should be a Vec<Lookup<E<Fp>>> (meaning of variables), or directly Vec<Lookup<Fp>>.~

~As a reference for the upcoming PRs, I will be using this draft PR as a baseline, as it contained some code for lookups that was removed from the current demo -> https://github.dev/o1-labs/proof-systems/pull/1332~

After talking to Matthew in a meeting, I could confirm that keeping track of the lookups should happen at the constraints side, not the witness side. Also, confirmed that the content in the lookups should be `E<Fp>` and not just `Fp`.

### DIFF
--- a/optimism/src/lookup.rs
+++ b/optimism/src/lookup.rs
@@ -1,0 +1,97 @@
+use ark_ff::One;
+
+#[derive(Copy, Clone, Debug)]
+pub enum Sign {
+    Pos,
+    Neg,
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum LookupMode {
+    Read,
+    Write,
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum LookupTable {
+    MemoryLookup,
+    RegisterLookup,
+    // Single-column table of 2^16 entries with the sparse representation of all values
+    SparseLookup,
+    // Single-column table of all values in the range [0, 2^16)
+    RangeCheck16Lookup,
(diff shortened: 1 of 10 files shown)


optimism/src/lookup.rs
```diff
@@ -0,0 +1,116 @@
...
+    // Input/Output of Keccak steps
+    KeccakStepLookup,
+    // Syscalls communication channel
+    SyscallLookup,
+}
+
+#[derive(Clone, Debug)]
+pub struct Lookup<Fp> {
+    pub mode: LookupMode,
+    /// The number of times that this lookup value should be added to / subtracted from the lookup accumulator.    pub magnitude_contribution: Fp,
```
dannywillems, line 40: I suppose the field can be removed?

dannywillems (approved):

optimism/src/lookup.rs
querolita: Yeah probably, it's always 1 in our case but it is more general if we include it.

PR #1764, opened 2024-01-25
ec addition and scaling

fabrizio-m:


kimchi/src/snarky/ec.rs
```diff
@@ -0,0 +1,304 @@
...
+            };
+            let x1 = w.read_var(&a0);
+            let x2 = w.read_var(&a1);
+            let y1 = w.read_var(&b0);
+            let y2 = w.read_var(&b1);
+
+            let same_x = to_bool(x1 == x2);
+
+            let x_diff = x2 - x1;
+            let x21_inv = x_diff.inverse().unwrap();
```
volhovm, line 44: So the actual constraint is something like `x21_inv * x_diff = 1`, right?

kimchi/src/snarky/ec.rs
```diff
@@ -0,0 +1,304 @@
...
+
+            let a = GroupAffine::<C>::new(x1, y1, false);
+            let b = GroupAffine::<C>::new(x2, y2, false);
+            assert!(!a.infinity, "can not be infinity");
+            assert!(!b.infinity, "can not be infinity");
+
+            let c = a + b;
+            let inf = to_bool(c.infinity);
+
+            let inf_z = if y1 == y2 {
```
volhovm, line 56: Shouldn't this be `y_1 == -y2 && x1 == x2`?

kimchi/src/snarky/ec.rs
```diff
@@ -0,0 +1,304 @@
...
+                y_diff.inverse().unwrap()
+            } else {
+                F::zero()
+            };
+            [c.x, c.y, inf, same_x, slope, inf_z, x21_inv]
+        })?;
+    let p1 = (a0, a1);
+    let p2 = (b0, b1);
+    let p3 = (p3x.clone(), p3y.clone());
+    let c = [p3x, p3y];
```
volhovm, line 68: This line can be moved to the end of the function.

kimchi/src/snarky/ec.rs
```diff
@@ -0,0 +1,304 @@
...
+            assert!(!a.infinity, "can not be infinity");
+            assert!(!b.infinity, "can not be infinity");
+
+            let c = a + b;
+            let inf = to_bool(c.infinity);
+
+            let inf_z = if y1 == y2 {
+                F::zero()
+            } else if x1 == x2 {
+                y_diff.inverse().unwrap()
```
volhovm, line 59: In the current code it seems that it can happen that `x1 == x2 && y1 == -y2` in which case `y_diff` is zero, how are you going to take its inverse?

kimchi/src/snarky/ec.rs
```diff
@@ -0,0 +1,304 @@
...
+            ),
+        };
+        let out = round.as_ref().unwrap().0.clone();
+        let n_next = round.as_ref().unwrap().1.n_next.clone();
+        *state = Some((out, n_next));
+        let round = round.map(|r| r.1);
+        Some(round)
+    });
+    let rounds: SnarkyResult<Vec<ScaleRound<FieldVar<F>>>> = rounds.collect();
+    let rounds = rounds?;
```
volhovm, line 122: Is this explicit unwrapping (and also here https://github.com/o1-labs/proof-systems/pull/1764/files#diff-280f21b46542317fc5664ff595b11c9a91d03f9b76aa17dc9d6f96d3826dc41bR131) necessary for some type safety reasons? As opposed to just using `?` on the previous line directly.

volhovm (commented):
Looked at EC addition, so half of the PR. Will look at multiplication a bit later.

kimchi/src/snarky/ec.rs
fabrizio-m: slightly more complex, you can see in https://o1-labs.github.io/proof-systems/specs/kimchi.html#elliptic-curve-addition that is used to constrain `same_x`.

kimchi/src/snarky/ec.rs
fabrizio-m: I owe you the explanation right now, but it is based on this: https://github.com/o1-labs/proof-systems/blob/fabrizio-m/snarky-rs-ec/kimchi/src/circuits/polynomials/complete_add.rs#L195-L212

kimchi/src/snarky/ec.rs
fabrizio-m: The idea is that infinity should be neither an input nor an output, but maybe it is better to make the check more explicit.

dannywillems:
Merging master.

fabrizio-m:
Drafting for now as it isn't a priority.

PR #1774, opened 2024-01-29
MIPS side of syscalls lookups

querolita:
Closes https://github.com/o1-labs/proof-systems/issues/1773

Includes lookups, constraints, and extra columns to keep track of the non-deterministic amount of bytes read by the syscalls at each row (either 1, 2, 3 or 4; but eventually, we will need 136 read/write lookups, one for each byte of each absorb).

### DIFF
--- a/optimism/src/mips/constraints.rs
+++ b/optimism/src/mips/constraints.rs
@@ -527,9 +527,9 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
 
             // Constrain booleanity of has_n_bytes
             {
-                for i in 0..MIPS_CHUNK_BYTES_LENGTH {
+                for flag in has_n_bytes.clone() {
                     self.constraints
-                        .push(has_n_bytes[i].clone() * (has_n_bytes[i].clone() - Expr::from(1)));
+                        .push(flag.clone() * (flag.clone() - Expr::from(1)));
                 }
             }
 
--- a/optimism/src/mips/constraints.rs
+++ b/optimism/src/mips/constraints.rs
@@ -33,7 +33,9 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
     /// variable/input of our circuit.
     type Position = MIPSColumn;
 
-    // Allocate a new input of our circuit
+    // Use one of the available columns. It won't
+    // create a new column every time this function is called. The number
+    // of columns is defined upfront by crate::mips::witness::SCRATCH_SIZE.
     fn alloc_scratch(&mut self) -> Self::Position {
         // All columns are implemented using a simple index, and a name is given
         // to the index.
--- a/optimism/src/mips/constraints.rs
+++ b/optimism/src/mips/constraints.rs
@@ -493,16 +493,6 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
 
         // EXTRA CONSTRAINTS
         {
-            // Check that you can only read 1, 2, 3 or 4 bytes
-            {
-                self.constraints.push(
-                    (row_bytes.clone() - Expr::from(1))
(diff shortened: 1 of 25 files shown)


optimism/src/mips/constraints.rs
```diff
@@ -440,10 +450,155 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+                * (row_bytes.clone() - Expr::from(4));
+            let read_2 = (row_bytes.clone() - Expr::from(1))
+                * (row_bytes.clone() - Expr::from(3))
+                * (row_bytes.clone() - Expr::from(4));
+            let read_3 = (row_bytes.clone() - Expr::from(1))
+                * (row_bytes.clone() - Expr::from(2))
+                * (row_bytes.clone() - Expr::from(4));
+            let read_4 = (row_bytes.clone() - Expr::from(1))
+                * (row_bytes.clone() - Expr::from(2))
+                * (row_bytes.clone() - Expr::from(4));
```
dannywillems, line 507:
```suggestion
                * (row_bytes.clone() - Expr::from(3));
```

optimism/src/mips/constraints.rs
```diff
@@ -440,10 +450,155 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
     ) -> Self::Variable {
-        Expr::Atom(ExprInner::Cell(Variable {
+        // The (at most) 4-byte chunk that has been read from the preimage
+        let bytes: [_; MIPS_CHUNK_BYTES_LENGTH] = array::from_fn(|i| {
+            Expr::Atom(ExprInner::Cell(Variable {
+                col: Self::Position::ScratchState(MIPS_PREIMAGE_BYTES_OFFSET + i),
+                row: CurrOrNext::Curr,
+            }))
+        });
+        // Whether the preimage chunk read has at least n bytes (1, 2, 3, or 4)
```
dannywillems, line 460: Can you explain the role of these columns?

dannywillems: Ah, is it for the alignment?

dannywillems (commented):
Left a change suggestion and a question.

optimism/src/mips/constraints.rs
querolita: It is basically for the lookup syscall. The idea was that the mips side would create 1 lookup for each byte read from the preimage. But because this could be any number between 1 and 4 bytes in each call, we needed to customize the "structure" of those lookups. For this reason, I created the `has_n_bytes` columns to keep track of whether this instruction had read at least `n` bytes. For example, if the syscall read 3 bytes this time, that means that it has 1 byte, also has 2 bytes, and also has 3 bytes, but not 4 bytes. This means that the lookups that need to be created are the ones corresponding to byte 1, byte 2, and byte 3, leaving the one for the 4th byte set to zero (because `has_n_bytes[3]` will be 0 used in the condition for the write interface). I guess this will have to change when we create different instances for each opcode, we will have to think about how to adapt it.

optimism/src/mips/constraints.rs
```diff
@@ -440,10 +450,173 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         _len: &Self::Variable,
         pos: Self::Position,
     ) -> Self::Variable {
-        Expr::Atom(ExprInner::Cell(Variable {
+        // The (at most) 4-byte chunk that has been read from the preimage
```
dannywillems, line 453: 1. The first 8 bytes are the length of the total preimage.
2. We have to write into the memory on the constraint side. On the witness side, we do update the linear memory. However, on the constraint side, we must add a lookup into `MemoryLookup`.

Can be done in a follow-up.

dannywillems (approved):

PR #1776, opened 2024-01-30
Documentation regarding constraints and expressions

dannywillems:


optimism/src/mips/mod.rs
```diff
@@ -26,4 +26,16 @@ pub mod proof;
...
+/// Type to represent a constraint on the individual columns of the execution
+/// trace.
+/// As a reminder, a constraint can be formally defined as a multi-variate
+/// polynomial over a finite field. The variables of the polynomial are defined
+/// as `crate::column::MIPSColumn`.
+/// The `expression` framework defined in `kimchi::circuits::expr` is used to
+/// describe the multi-variate polynomials.
+/// For instance, a vanilla 3-wires PlonK constraint can be defined using the
+/// multi-variate polynomial of degree 2
+/// `P(X, Y, Z) = q_x X + q_y Y + q_m X Y + q_o Z + q_c`
```
querolita, line 38: Oh so you're explaining constraints as multivariate polynomials where each variable is a column? 

I find this a bit confusing, given that we have 15 columns/polynomials in kimchi, but the variable X is always the same and it refers to the row index. So basically the above should actually be:

$$P(X) = q_L * p_L(X) + q_R * p_R(X) + q_M * p_L(X) * p_R(X) + q_O * p_O(X) + p_C $$

querolita (approved):
I would prefer to avoid the multivariateness, but I am fine to have this as docs for now either way

PR #1778, opened 2024-01-31
`RangeCheck16` lookup table

querolita:

### DIFF
--- a/optimism/src/lookup.rs
+++ b/optimism/src/lookup.rs
@@ -13,25 +13,28 @@ pub enum LookupMode {
 }
 
 #[derive(Copy, Clone, Debug)]
-pub enum LookupTables {
-    MemoryLookup,
-    RegisterLookup,
-    // Single-column table of 2^16 entries with the sparse representation of all values
-    SparseLookup,
-    // Single-column table of all values in the range [0, 2^16)
-    RangeCheck16Lookup,
-    // Dual-column table of all values in the range [0, 2^16) and their sparse representation
-    ResetLookup,
-    // 24-row table with all possible values for round and their round constant in expanded form
-    RoundConstantsLookup,
-    // All [0..136] values of possible padding lengths, the value 2^len, and the 5 corresponding pad suffixes with the 10*1 rule
-    PadLookup,
-    // All values that can be stored in a byte (amortized table, better than model as RangeCheck16 (x and scaled x)
-    ByteLookup,
-    // Input/Output of Keccak steps
-    KeccakStepLookup,
-    // Syscalls communication channel
-    SyscallLookup,
+pub enum LookupTableIDs {
+    // RAM Tables
+    MemoryLookup = 0,
+    RegisterLookup = 1,
+    /// Syscalls communication channel
+    SyscallLookup = 2,
+    /// Input/Output of Keccak steps
+    KeccakStepLookup = 3,
+
+    // Read Tables
+    /// Single-column table of 2^16 entries with the sparse representation of all values
+    SparseLookup = 4,
+    /// Single-column table of all values in the range [0, 2^16)
+    RangeCheck16Lookup = 5,
+    /// Dual-column table of all values in the range [0, 2^16) and their sparse representation
+    ResetLookup = 6,
+    /// 24-row table with all possible values for round and their round constant in expanded form (in big endian)
+    RoundConstantsLookup = 7,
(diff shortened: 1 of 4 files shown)


optimism/src/lookup.rs
```diff
@@ -13,7 +13,7 @@ pub enum LookupMode {
 }
 
 #[derive(Copy, Clone, Debug)]
-pub enum LookupTable {
+pub enum LookupTables {
```
volhovm, line 16: Maybe `VMLookupTableType`? Or `ConcreteLookupTable`?

optimism/src/lookup.rs
```diff
@@ -39,7 +39,7 @@ pub struct Lookup<Fp> {
     pub mode: LookupMode,
     /// The number of times that this lookup value should be added to / subtracted from the lookup accumulator.    pub magnitude_contribution: Fp,
     pub magnitude: Fp,
-    pub table_id: LookupTable,
+    pub table_id: LookupTables,
```
volhovm, line 42: Maybe it's better to rename `table_id` into `table_type`? I assume you can extract the id from the enum, but it's looking confusing as it is.

optimism/src/lookup.rs
```diff
@@ -114,3 +114,27 @@ pub trait Lookups {
...
 }
+
+/// A table of values that can be used for a lookup, along with the ID for the table.
+#[derive(Debug, Clone)]
+pub struct LookupTable<F> {
+    /// The table is a vector of write lookups with the same table ID
+    _table: Vec<Lookup<F>>,
+}
+
+const _TWO_TO_16_UPPERBOUND: u32 = 1 << 16;
```
volhovm, line 125: I'd suggest removing this unless it's actually used in several places. Other options:
- Giving it a name that would imply what it's used for. Like `MAX_SRS_SIZE` or `DOMAIN_SIZE_LIMIT`. Not sure why it is a constant actually.
- Moving it into the `LookupTable` implementation.

volhovm: EDIT: I see that in different PRs you actually use this value again. I guess I'd ask to clarify its meaning/rename. Is it the maximum allowed size of the lookup table? Shouldn't there be a constant for that already somewhere on the higher level?

optimism/src/lookup.rs
```diff
@@ -114,3 +114,27 @@ pub trait Lookups {
     /// Adds all lookups of Self to the environment
     fn lookups(&mut self);
 }
+
+/// A table of values that can be used for a lookup, along with the ID for the table.
```
volhovm, line 118: There's no ID in this struct... is there supposed to be?

optimism/src/lookup.rs
```diff
@@ -114,3 +114,27 @@ pub trait Lookups {
...
+#[derive(Debug, Clone)]
+pub struct LookupTable<F> {
+    /// The table is a vector of write lookups with the same table ID
+    _table: Vec<Lookup<F>>,
```
volhovm, line 122: `lookup_requests` or `lookup_values`?

optimism/src/lookup.rs
```diff
@@ -114,3 +114,27 @@ pub trait Lookups {
     /// Adds all lookups of Self to the environment
     fn lookups(&mut self);
 }
+
+/// A table of values that can be used for a lookup, along with the ID for the table.
+#[derive(Debug, Clone)]
+pub struct LookupTable<F> {
```
volhovm, line 120: This coincides with the `LookupTable` from kimchi... I'm afraid it can create quite a bit of confusion if one looks at both codebases in a comparative way. But it's not a bad name, I guess this struct is the most similar one to the kimchi's `LookupTable` struct anyway.

volhovm (approved):
Looks good! A lot of nits though, sorry for that :)

optimism/src/lookup.rs
querolita: Oh that's an old comment

optimism/src/lookup.rs
querolita: The difference is that in Kimchi the table is transposed. Meaning, the first index refers to the number of columns, whereas here it is the row.

optimism/src/lookup.rs
querolita: requests would make sense for `read` types of entries, but these are (for now) always `write` lookups (meaning, they already exist and we will just consume a few of them throughout the circuit execution).

PR #1780, opened 2024-01-31
`Sparse` lookup table

querolita:

### DIFF
--- a/optimism/src/lookup.rs
+++ b/optimism/src/lookup.rs
@@ -150,4 +150,19 @@ impl<F: Field> LookupTable<F> {
                 .collect(),
         }
     }
+
+    fn _table_sparse() -> Self {
+        Self {
+            _table: (0.._TWO_TO_16_UPPERBOUND)
+                .map(|i| Lookup {
+                    mode: LookupMode::Write,
+                    magnitude: F::one(),
+                    table_id: LookupTables::SparseLookup,
+                    value: vec![F::from(
+                        u64::from_str_radix(&format!("{:b}", i), 16).unwrap(),
+                    )],
+                })
+                .collect(),
+        }
+    }
 }
--- a/optimism/src/lookup.rs
+++ b/optimism/src/lookup.rs
@@ -141,6 +141,20 @@ impl<F: Field> LookupTable<F> {
     }
 
     #[allow(dead_code)]
+    fn table_sparse() -> Self {
+        Self {
+            table_id: LookupTableIDs::SparseLookup,
+            entries: (0..TWO_TO_16_UPPERBOUND)
+                .map(|i| {
+                    vec![F::from(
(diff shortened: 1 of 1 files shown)


optimism/src/lookup.rs
```diff
@@ -150,4 +150,19 @@ impl<F: Field> LookupTable<F> {
...
+
+    fn _table_sparse() -> Self {
+        Self {
+            _table: (0.._TWO_TO_16_UPPERBOUND)
+                .map(|i| Lookup {
+                    mode: LookupMode::Write,
+                    magnitude: F::one(),
+                    table_id: LookupTables::SparseLookup,
+                    value: vec![F::from(
+                        u64::from_str_radix(&format!("{:b}", i), 16).unwrap(),
```
volhovm, line 162: Maybe this table needs a comment. IIRC you convert a 16 bit integer $x$ into bit representation $\{x_i\}_{i=0}^{15}$, after which you treat every single digit as if it's a hex value, so in the end you get $\sum\limits_{i=0}^{15} x_i \cdot 16^{i}$. Which means the maximum value that you store is `2^{256}-1`... 

Are you sure it fits into the curve field? Not sure which curve you're using here, but both Pallas and Vesta are 255 bits, and BN254 is 254 bits...

volhovm (commented):
I have this concern about the field size...

optimism/src/lookup.rs
querolita: This is only meant to represent the expansion of 16-bits at a time to make them 64-bits, not full 64bits becoming a 256-bit

querolita: Okay I am reading this again to understand your concern better.

The idea is not to treat every single digit as if it is an hex value. What the line is doing is represent each `i` between `0` and `2^16-1` as a binary string. That means, at most 16 characters that are either 0 or 1. And then the `from_str_radix` function takes those 16 characters and "reads" them as if they were hex values. This means, whereas before 8 characters represented 1 byte, here every 2 characters you have 1 byte (you "divide" the length by 4). This means that the 16-character long string becomes a u64 integer, without overflows, and that fits in one field element always.

volhovm: Ah yes, I got confused. So in this sparse representation each bit of `i` translates into 4 bits, which means the resulting value is at most 4 * 16 = 64 bits indeed. Sounds good.

volhovm (approved):

optimism/src/lookup.rs
querolita: Basically, reading the binary expression as if it was an hex representation, gives you the expansion. If we inserted 2 bits instead of 3, then the trick would work but base 8. If we inserted 4 bits, then base 32 😉

volhovm: I think I understand the trick, but this is not the same as understanding why you need these sparse representations. But next time :)

PR #1781, opened 2024-01-31
`Reset` lookup table

querolita:

### DIFF
--- a/optimism/src/lookup.rs
+++ b/optimism/src/lookup.rs
@@ -165,4 +165,20 @@ impl<F: Field> LookupTable<F> {
                 .collect(),
         }
     }
+
+    fn _table_reset() -> Self {
+        Self {
+            _table: (0.._TWO_TO_16_UPPERBOUND)
+                .map(|i| Lookup {
+                    mode: LookupMode::Write,
+                    magnitude: F::one(),
+                    table_id: LookupTables::ResetLookup,
+                    value: vec![
+                        F::from(i),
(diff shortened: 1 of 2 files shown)


optimism/src/lookup.rs
```diff
@@ -165,4 +165,20 @@ impl<F: Field> LookupTable<F> {
...
+    fn _table_reset() -> Self {
+        Self {
+            _table: (0.._TWO_TO_16_UPPERBOUND)
+                .map(|i| Lookup {
+                    mode: LookupMode::Write,
+                    magnitude: F::one(),
+                    table_id: LookupTables::ResetLookup,
+                    value: vec![
+                        F::from(i),
+                        F::from(u64::from_str_radix(&format!("{:b}", i), 16).unwrap()),
```
volhovm, line 178: I'm wondering why this table needs to be separate from the `Sparse` one... I guess for efficiency if we want to just check that the value is in the "sparse" range, but we don't want to explicitly show an index `i`. Is that correct?

volhovm (approved):

optimism/src/lookup.rs
querolita: Yeah exactly. Sometimes we just want to check if a value is a correct sparse term, and we don't care about what was the "original" value (or even, we don't have it anywhere in the witness). Until we have the means to look at single columns inside a multi-column table, we need to have separate tables for each column we might want to access independently 😞

PR #1783, opened 2024-01-31
`Pad` lookup table

querolita:
Closes https://github.com/o1-labs/proof-systems/issues/1735

### DIFF
--- a/optimism/src/lookup.rs
+++ b/optimism/src/lookup.rs
@@ -1,6 +1,9 @@
 use ark_ff::{Field, One};
 use kimchi::{
-    circuits::polynomials::keccak::{constants::ROUNDS, Keccak, RC},
+    circuits::polynomials::keccak::{
+        constants::{RATE_IN_BYTES, ROUNDS},
+        Keccak, RC,
+    },
     o1_utils::Two,
 };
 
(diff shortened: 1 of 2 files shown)


optimism/src/lookup.rs
```diff
@@ -201,4 +206,28 @@ impl<F: Field> LookupTable<F> {
                 .collect(),
         }
     }
+
+    fn _table_pad() -> Self {
+        Self {
+            _table: (1..=136)
```
volhovm, line 212: This `136` probably needs to be a constant (with a comment?), looks very magic :) Also the `0` value we explicitly don't need because zero paddings are zero sized?

optimism/src/lookup.rs
```diff
@@ -201,4 +206,28 @@ impl<F: Field> LookupTable<F> {
...
+                .map(|i| {
+                    let suffix = pad_blocks(i as usize);
+                    Lookup {
+                        mode: LookupMode::Write,
+                        magnitude: F::one(),
+                        table_id: LookupTables::PadLookup,
+                        value: vec![
+                            F::from(i),
+                            F::two_pow(i),
```
volhovm, line 221: Why do you need `F::two_pow(i)` here? I assume you need it for the application logic to restrain `y` to be `2^i` for given (`i`,`y`)? And creating a separate table is just more expensive?

volhovm (approved):
As far as I can tell, the lookup table is consistent with `pad_blocks` (except that I don't understand the magic constant 136 and if it can be a valid input for `pad_blocks`). 

I don't understand the semantics of `pad_blocks` but it does not seem to be as important.

optimism/src/lookup.rs
querolita: Oh, it should definitely be a constant. Actually `RATE_IN_BYTES`, I just forgot to import it and use it! 

And the fact that we don't include the zero entry is that the lookups for the padding are only triggered when the padding is at least 1 byte. So there's no need to include the trivial and unused entry when there's no padding to be done (0 bytes).

optimism/src/lookup.rs
querolita: Exactly, in some constraints to check that the padding is performed at the end of the bytestring I need the value `2^i`. Creating another table just for this would be more expensive (essentially 135 more rows in the circuit). Unlike with the sparse representation, in this case I will be able to access the whole entry at once, no need to look at individual columns. So for that reason, there's no need to include it in a separate table.

IF at some point we needed the powers of two for whatever other reason, then it would make sense to have it separate somewhere else. But for now, it's fine to reuse the same table.

PR #1670, opened 2024-01-03
[berkeley] Update curves to arkworks 0.4.2

mrmr1993:
This is https://github.com/o1-labs/proof-systems/pull/1669 merged into the berkeley branch

PR #1788, opened 2024-02-01
cargo update + use workspace deps

emberian:
keeps all the dependencies in one place, makes updating versions easier.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,76 @@ members = [
 ]
 resolver = "2"
 
+[workspace.dependencies]
+ark-algebra-test-templates = "0.3.0"
+ark-bn254 = { version = "0.3.0" }
+ark-ec = { version = "0.3.0", features = ["parallel"] }
+ark-ff = { version = "0.3.0", features = ["parallel", "asm"] }
+ark-poly = { version = "0.3.0", features = ["parallel"] }
+ark-serialize = "0.3.0"
+ark-std = "0.3.0"
+base64 = "0.21.5"
+bitvec = "1.0.0"
+blake2 = "0.10.0"
+bs58 = "0.5.0"
+clap = "4.4.6"
+colored = "2.0.0"
+command-fds = "0.3"
+convert_case = "0.6.0"
+criterion = "0.5"
+elf = "0.7.2"
+env_logger = "0.11.1"
+hex = "0.4"
+iai = "0.1"
+itertools = "0.12.1"
+kimchi = { path = "./kimchi", version = "0.1.0", features = ["bn254"] }
+libc = "0.2.62"
+libflate = "2"
+log = "0.4.20"
+num-bigint = { version = "0.4.3", features = ["rand", "serde"] }
+num-derive = "0.4"
+num-integer = "0.1.45"
+num-traits = "0.2"
+ocaml = { version = "0.22.2" }
+ocaml-gen = { version = "0.1.5" }
+once_cell = "1.10.0"
+os_pipe = { version = "1.1.4", features = ["io_safety"] }
+proc-macro2 = "1.0.43"
+proptest = "1.0.0"
+proptest-derive = "0.4.0"
+quote = "1.0.21"
+rand = { version = "0.8.0", features = ["std_rng"] }
+rand_chacha = { version = "0.3.0" }
+rand_core = { version = "0.6.0" }
+rayon = "1"
+regex = "1.10.2"
+rmp-serde = "1.1.1"
+secp256k1 = "0.28.2"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0.79"
+serde_with = "3.6.0"
+sha2 = "0.10.0"
+snarky-deriver = { path = "./kimchi/snarky-deriver", version = "0.1.0" }
+strum = "0.26.1"
+strum_macros = "0.26.1"
+syn = { version = "1.0.109", features = ["full"] }
+thiserror = "1.0.30"
+tinytemplate = "1.1"
+wasm-bindgen = "=0.2.90"
+
+o1-utils = { path = "./utils", version = "0.1.0" }
+mina-curves = { path = "./curves", version = "0.1.0" }
+groupmap = { path = "./groupmap", version = "0.1.0" }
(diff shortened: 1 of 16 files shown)


utils/Cargo.toml
```diff
@@ -28,7 +28,7 @@ rand = "0.8.0"
 rand_core = "0.6.3"
 
 [dev-dependencies]
-ark-ec = { version = "0.3.0", features = [ "parallel" ] }
+ark-ec = { version = "0.4.0", features = [ "parallel" ] }
```
volhovm, line 31: I'm a bit surprised to see it here, had an impression upgrading to `0.4.0` is quite nontrivial.

Also, do utilities have their separate set of versions?..

Cargo.toml
```diff
@@ -18,6 +18,76 @@ members = [
...
+colored = "2.0.0"
+command-fds = "0.3"
+convert_case = "0.6.0"
+criterion = "0.5"
+elf = "0.7.2"
+env_logger = "0.11.1"
+hex = "0.4"
+iai = "0.1"
+itertools = "0.12.1"
+kimchi = { path = "./kimchi", version = "0.1.0", features = ["bn254"] }
```
volhovm, line 43: This should be in the second group of imports?

Cargo.toml
```diff
@@ -18,6 +18,76 @@ members = [
...
+rand_core = { version = "0.6.0" }
+rayon = "1"
+regex = "1.10.2"
+rmp-serde = "1.1.1"
+secp256k1 = "0.28.2"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0.79"
+serde_with = "3.6.0"
+sha2 = "0.10.0"
+snarky-deriver = { path = "./kimchi/snarky-deriver", version = "0.1.0" }
```
volhovm, line 70: This one also belongs to the next block?

Cargo.toml
```diff
@@ -18,6 +18,76 @@ members = [
 ]
 resolver = "2"
 
+[workspace.dependencies]
+ark-algebra-test-templates = "0.3.0"
+ark-bn254 = { version = "0.3.0" }
+ark-ec = { version = "0.3.0", features = ["parallel"] }
```
volhovm, line 24: I'm not sure if declaring features in `Cargo.toml` is a good idea -- if I understand correctly, they are additive with features requested in any of the sub-packages (e.g. kimchi). I guess it's fine for these libraries since we pretty much want these features (parallel, asm) in all of them?

optimism/Cargo.toml
```diff
@@ -17,31 +17,31 @@ name = "test_optimism_preimage_read"
...
-strum = "0.24.0"
-strum_macros = "0.24.0"
-log = "0.4.20"
-env_logger = "0.10.0"
-command-fds = "0.2.3"
-os_pipe = { version = "1.1.4", features = [ "io_safety" ] }
-rand = "0.8.5"
-libc = "0.2.62"
-rayon = "1.5.0"
+ark-bn254.workspace = true
```
volhovm, line 20: It's a bit annoying one can't just specify that all dependencies are imported "by default" from the workspace IIUC.

optimism/src/preimage_oracle.rs
```diff
@@ -116,19 +115,19 @@ impl PreImageOracle {
         // We need to map 3, 4, 5, 6 in the child process
         cmd.fd_mappings(vec![
             FdMapping {
-                parent_fd: hint_server.0.writer.as_raw_fd(),
+                parent_fd: unsafe { OwnedFd::from_raw_fd(hint_server.0.writer.as_raw_fd()) },
```
volhovm, line 118: I guess this is because some `optimism` library was upgraded? But why is it an unsafe call now/is it sound?

volhovm (commented):
I like the idea! A few comments which I'm not sure about before approving :+1:

Cargo.toml
emberian: the way the workspace resolver works, if any dep in the workspace requests a feature, every use of the crate will have that feature. this centralizing of features into the toplevel config basically encodes the status quo that was represented in the lockfile

emberian: one detail i'm not sure about is how this interacts with cargo publish: if certain crates truly are agnostic towards the feature maybe we should override the workspace features so that the computed package manifest doesn't unnecessarily include the features even when its usage in a typical workspace will have them anyway

optimism/src/preimage_oracle.rs
emberian: yes, a dependency upgradaded its public api to require OwnedFd, modeling as a rype that the code already took ownership of the fd. the unsafety represents that the fd is going to be closed on drop, making the original raw fd dangling

utils/Cargo.toml
emberian: I seem to have tinkered with this crate first, realized upgrading ark was out of scope, and forgot to ever return to it when I thought workspace deps were a good idea- nice catch, fixing

optimism/Cargo.toml
emberian: I think this limitation derives from the top-down nature of how cargo computes its build plan statically from the set of manifests. There isn't a backchannel that lets rustc ask cargo "by the way I need this crate now" when it encounters a name resolution error that seems like it could be fixed by an extra --extern argument. The remaining alternative would be to always build all workspace deps and pass them to all crates, which feels worse to me.

utils/Cargo.toml
volhovm: @emberian makes sense (including in other comments), happy to merge after it's fixed!

mrmr1993 (changes requested):
We don't want to upgrade arkworks. See [here](https://github.com/o1-labs/proof-systems/pull/1670).

volhovm:
@mrmr1993 I don't think there was an intention to upgrade `arkworks`, see the comment above (https://github.com/o1-labs/proof-systems/pull/1788#discussion_r1491247610).

volhovm (approved):
LGTM, cc @mrmr1993 -- no arkworks upgrade.

mrmr1993 (approved):

PR #1804, opened 2024-02-06
Implement generic function to commit to witness columns

dannywillems:


msm/src/proof.rs
```diff
@@ -38,6 +38,17 @@ impl<G: KimchiCurve> Witness<G> {
     }
 }
 
+/// Commit to each individual polynomial
+pub fn commit_witness_columns<G: KimchiCurve, OpeningProof: OpenProof<G>>(
```
mrmr1993, line 42: It should be possible to generalise this, and at the same time to use some parallelism. Consider e.g. [the MIPS prover](https://github.com/o1-labs/proof-systems/blob/b40e6dde58a5de03a62e8eee9b5379099985f232/optimism/src/mips/proof.rs#L153-L164), where the code can be reused by any type implementing `par_iter` and `from_par_iter` (as [here](https://github.com/o1-labs/proof-systems/blob/b40e6dde58a5de03a62e8eee9b5379099985f232/optimism/src/mips/proof.rs#L61-L107)).

Even without using parallelism, implementing `iter` and `FromIterator` would make this a one-liner:
```rust
polynomials.iter().map(|poly| srs.commit_non_hiding(poly, 1, None)).collect::<WitnessColumns<_>>()
```
or IMO even clearer
```rust
let commitments: WitnessColumns<_> =
    polynomials.iter().map(|poly| srs.commit_non_hiding(poly, 1, None)).collect();
```
which avoids the need for this function altogether.

mrmr1993: For added flavour: this would also allow absorbing the witness columns using a fold over `commitments.iter()`, which is IMO better than the manual abstraction leak below.

dannywillems: Far better... See last changes. I was struggling with the iterator, lifetimes and borrow (which is, in other words, Rust...).

volhovm (approved):

PR #1806, opened 2024-02-06
Mina curves: adding grumpkin and bn254

dannywillems:
Pairing parameters will be added later.


volhovm:
@dannywillems please tell me if this requires a review.

I'm not exactly sure we go forward with grumpkin, but we should probably merge existing work anyway?

dannywillems:
Converting into draft as we stay with bn254 as an external library atm.

dannywillems:
Closing as we won't be using it.

PR #1807, opened 2024-02-07
Generic MVLookup - MSM

dannywillems:
Current bug (expected): does not handle multiplicity different than one. Coming in a following PR with a test.
Also support only one table. Change coming in another PR, it is straightforward to generalize.

Fix https://github.com/MinaProtocol/mina/issues/15089

It is not sound, yet. Missing constraints.

TODO:
- [x] support empty list of lookups
- [x] support more than one lookup table in the proof (done in https://github.com/o1-labs/proof-systems/pull/1827)
- [ ] handle multiplicity

### DIFF
--- a/msm/README.md
+++ b/msm/README.md
@@ -1,0 +1,51 @@
+## MSM circuit
+
+This is a specialised circuit to compute a large MSM where the elements are
+defined over a foreign field, like the one used by the Pickles verifier.
+For simplicity, we suppose the foreign field is on maximum 256 bits.
+
+The witness is splitted in two different parts:
+- Columns: this is the column of the circuit.
+- Lookups: this contain a list of lookups per row.
+  [MVLookup](https://eprint.iacr.org/2022/1530.pdf) is used. It contains
+  the individual polynomials $f_i(X)$ containing the looked up values, the
+  table/polynomial $t(X)$ the polynomials $f_i(X)$ look into and the
+  multiplicity polynomial $m(X)$ as defined in the paper.
+
+The individual columns of the circuit are defined in `src/column.rs`.
+The polynomials can be expressed using the expression framework, instantiated in `src/column.rs`, with
+```rust
+type E<Fp> = Expr<ConstantExpr<F>, MSMColumn>
+```
+
+The proof structure is defined in `src/proof.rs`. A proof consists of
+commitments to the circuit columns and the lookup polynomials, in addition to
+their evaluations, the evaluation points and the opening proofs.
+The prover/verifier are respectively in `src/prover.rs` and `src/verifier.rs`.
+Note that the prover/verifier implementation is generic enough to be used to
+implement other circuits.
+
+The protocol used is a Plonk-ish arithmetisations without the permutation
+argument. The circuit is wide enough (i.e. has enough
+columns) to handle one elliptic curve addition on one row.
+
+The foreign field elements are defined with 16 limbs of 16bits. Limbs-wise
+elementary addition and multiplication is used, and range checks on 16 bits is
+used.
+An addition of two field elements $a = (a_{1}, ..., a_{16})$ and $b = (b_{1}, ...,
+b_{16})$ will be performed on one row, using one power of alpha for each (i.e one
+constraint for each).
+
+The elliptic curve points are represented in affine coordinates.
+As a reminder, the equations for addition are:
+
+```math
+\lambda = \frac{y_2 - y_1}{x_2 - x_1}
+```
+
+```math
+\begin{align}
+x_3 & = \lambda^2 - x_{1} - x_{2} \\
+y_3 & = \lambda (x_{1} - x_{3}) - y_{1}
+\end{align}
+```
--- a/msm/src/mvlookup.rs
+++ b/msm/src/mvlookup.rs
@@ -1,0 +1,171 @@
+//! Implement the protocol MVLookup <https://eprint.iacr.org/2022/1530.pdf>
+
+use std::iter;
+
+use ark_ff::{FftField, Field};
+use kimchi::circuits::domains::EvaluationDomains;
+use rand::{seq::SliceRandom, thread_rng, Rng};
+
+// TODO: Add more built-in lookup tables
+#[derive(Copy, Clone, Debug)]
+pub enum LookupTable {
+    RangeCheck16,
+    /// Custom lookup table
+    /// The index of the table is used as the ID, padded with the number of
+    /// built-in tables.
+    Custom(usize),
+}
+
+impl LookupTable {
+    /// Assign a unique ID to the lookup tables.
+    pub fn into_field<F: Field>(self) -> F {
+        match self {
+            LookupTable::RangeCheck16 => F::one(),
+            LookupTable::Custom(id) => F::from(id as u64) + F::one(),
+        }
+    }
+}
+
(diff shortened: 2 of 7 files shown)


msm/src/mvlookup.rs
```diff
@@ -0,0 +1,171 @@
...
+
+use std::iter;
+
+use ark_ff::{FftField, Field};
+use kimchi::circuits::domains::EvaluationDomains;
+use rand::{seq::SliceRandom, thread_rng, Rng};
+
+// TODO: Add more built-in lookup tables
+#[derive(Copy, Clone, Debug)]
+pub enum LookupTable {
```
volhovm, line 11: Potential suggestion for the future: if possible let's make it generic, e.g. instead of `enum` implement a trait and then pass concretely used separately defined tables into the prover.

msm/src/mvlookup.rs
```diff
@@ -0,0 +1,171 @@
...
+    #[allow(dead_code)]
+    pub(crate) t: Vec<Lookup<F>>,
+    /// The multiplicity polynomial
+    #[allow(dead_code)]
+    pub(crate) m: Vec<F>,
+}
+
+// This should be used only for testing purposes.
+// It is not only in the test API because it is used at the moment in the
+// main.rs. It should be moved to the test API when main.rs is replaced with
```
volhovm, line 62: Agree, good to have it though. Maybe could be a `Random` instance instead of just an `impl` method.

msm/src/proof.rs
```diff
@@ -105,14 +107,20 @@ impl<G: KimchiCurve> Witness<G> {
...
         }
     }
 }
 
 #[derive(Debug, Clone)]
 pub struct Proof<G: KimchiCurve, OpeningProof: OpenProof<G>> {
+    // Columns/PlonK argument
     pub(crate) commitments: WitnessColumns<PolyComm<G>>,
     pub(crate) zeta_evaluations: WitnessColumns<G::ScalarField>,
     pub(crate) zeta_omega_evaluations: WitnessColumns<G::ScalarField>,
```
volhovm, line 120: I honestly like this so much more than `kimchi`'s `evaluations` which contains `WitnessColumns` of a pair. Slightly verbose stuff is easily understandable, and we're not even close to the verbosity level of Java so I wouldn't worry :)

msm/src/prover.rs
```diff
@@ -57,6 +58,179 @@ where
...
+    let vector_lookup_value_combiner = fq_sponge.challenge();
+
+    // Coin an evaluation point for the rational functions
+    let beta = fq_sponge.challenge();
+
+    let n = f.len();
+    let lookup_terms_evals: Vec<G::ScalarField> = {
+        // We compute first the denominators of all f_i and t. We gather them in
+        // a vector to perform a batch inversion.
+        // We include t in the denominator, therefore n + 1
```
volhovm, line 98: Nice comments :+1:

msm/src/prover.rs
```diff
@@ -76,13 +251,24 @@ where
         };
         (evals(&zeta), evals(&zeta_omega))
     };
+    // TODO: Parallelize
+    let (mvlookup_zeta_evaluations, mvlookup_zeta_omega_evaluations) = {
+        let evals = |point| {
+            let comm = |poly: &DensePolynomial<G::ScalarField>| poly.evaluate(point);
```
volhovm, line 257: Why is it called `comm` and not smth like `eval`? It's not a commitment, right? IIRC it's scalar element.

volhovm (approved):

msm/src/prover.rs
dannywillems: Oops. Copy paste. Fixed.

msm/src/mvlookup.rs
dannywillems: Opened follow-up https://github.com/o1-labs/proof-systems/issues/1824

msm/src/mvlookup.rs
dannywillems: Opened follow-up https://github.com/o1-labs/proof-systems/issues/1825

msm/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+The proof structure is defined in `src/proof.rs`. A proof consists of
+commitments to the circuit columns and the lookup polynomials, in addition to
+their evaluations, the evaluation points and the opening proofs.
+The prover/verifier are respectively in `src/prover.rs` and `src/verifier.rs`.
+Note that the prover/verifier implementation is generic enough to be used to
+implement other circuits.
+
+The protocol used is a Plonk-ish arithmetisations without the permutation
```
volhovm, line 28: I'm not actually sure that we will use plonkish constraint (https://twitter.com/feministPLT/status/1413815927704014850) at all, but it's not that important.

msm/src/mvlookup.rs
```diff
@@ -0,0 +1,403 @@
...
+        LookupWitness {
+            f: vec![f_evals],
+            t: t_evals,
+            m,
+        }
+    }
+}
+
+/// Represents the proof of the lookup argument
+/// It is parametrized by the type `T` which can be either:
```
volhovm, line 145: What do you think about using explicit fields instead? `m_com, h_com, sum_com, m_eval, h_eval, sum_eval`? Or two structs even, `LookupProofComs` and `LookupProofEvals`?

msm/src/mvlookup.rs
```diff
@@ -0,0 +1,403 @@
...
+/// `table_id`.
+/// The structure represents the individual fraction of the sum described in the
+/// MVLookup protocol (for instance Eq. 8).
+/// The table ID is added to the random linear combination formed with the
+/// values. The combiner for the random linear combination is coined during the
+/// proving phase by the prover.
+#[derive(Debug, Clone)]
+pub struct Lookup<F> {
+    pub(crate) table_id: LookupTable,
+    pub(crate) numerator: F,
```
volhovm, line 39: Would be nice to have a comment that `numerator` is multiplicity.

msm/src/mvlookup.rs
```diff
@@ -0,0 +1,403 @@
...
+                let mut acc = G::ScalarField::zero();
+                for i in 0..domain.d1.size as usize {
+                    // phi(1) = 0
+                    evals.push(acc);
+                    // FIXME: only one lookup for now.
+                    acc += lookup_terms_evals_d1[i]
+                }
+                // Sanity check to verify that the accumulator ends up being zero.
+                // FIXME: This should be removed from runtime, and a constraint
+                // should be added. For now, the verifier accepts any proof.
```
volhovm, line 369: We don't need a constraint that `acc = 0` in the end, only the one that the whole thing is circular. Can start with any nonzero constant. Maybe this comment is outdated.

volhovm (approved):
Looks good, feel free to merge

msm/src/mvlookup.rs
dannywillems: Yes, it can be anything other than 0. Should be mentioned somewhere.

msm/src/mvlookup.rs
dannywillems: I'm in favor of keeping low the number of different structures, and hide the differences in the type. Maybe we can rename the structure.

msm/README.md
dannywillems: Yeah, we should simply use multi-variate polynomial constraints :)

PR #1809, opened 2024-02-07
Generic prover: use iterator for witness columns

dannywillems:


volhovm (approved):
Similar to the previous PR. The previous PR and this one could be perhaps batched together, and maybe even into the target branch PR.

PR #1810, opened 2024-02-07
Kimchi MSM: introduce README

dannywillems:
The document will evolve in the future

### DIFF
--- a/msm/README.md
+++ b/msm/README.md
@@ -1,0 +1,51 @@
+## MSM circuit
+
+This is a specialised circuit to compute a large MSM where the elements are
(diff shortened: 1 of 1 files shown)


volhovm (approved):
Very good to have docs, although I'm not 100% it's better to have it here (and not in rust comments/mina book/smth else?). Auto-generated kimchi spec in mina book is quite good IMO. We can decide later though.

dannywillems:
> Very good to have docs, although I'm not 100% it's better to have it here (and not in rust comments/mina book/smth else?). Auto-generated kimchi spec in mina book is quite good IMO. We can decide later though.

Yes, later is better.

PR #1811, opened 2024-02-08
ec gates for snarkyrs

fabrizio-m:
This is #1764 but directly merging on master, the comments from #1764 still need to be addressed but the interface should remain the same.


volhovm:
@fabrizio-m If we decide to go forward, I suggest to close this PR, because the other one (https://github.com/o1-labs/proof-systems/pull/1764) already targets `master` and it has comments, while this is a clone. What do you think?

PR #1828, opened 2024-02-17
MSM/prover: parallelize more

dannywillems:


volhovm (approved):
I'm not sure if there's any harm in swapping every `iter` by `par_iter` -- hopefully `rayon` is smart enough to figure out how much exactly to parallelise under the bonnet.

PR #1829, opened 2024-02-17
Kimchi/verifier: stop relying on iterator on option

dannywillems:
commitments.lookup is an option. By using iter followed by for_each, the reader might suppose it is a list.
IMO, it seems more readable to check if lookups are used, unwrap the lookup commitments, without verifying if an error is raised as it should never happened, and get the commitment to the lookup aggregator.

### DIFF
--- a/kimchi/src/verifier.rs
+++ b/kimchi/src/verifier.rs
@@ -214,7 +214,11 @@ where
         //~ 1. If using lookup, absorb the commitment to the aggregation lookup polynomial.
         if index.lookup_index.is_some() {
             // Should not fail, as the lookup index is present
-            let lookup_commits = self.commitments.lookup.as_ref()..ok_or(VerifyError::LookupCommitmentMissing)?;
+            let lookup_commits = self
+                .commitments
+                .lookup
+                .as_ref()
+                .ok_or(VerifyError::LookupCommitmentMissing)?;
             absorb_commitment(&mut fq_sponge, &lookup_commits.aggreg);
         }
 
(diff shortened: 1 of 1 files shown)


kimchi/src/verifier.rs
```diff
@@ -212,9 +212,11 @@ where
...
 
         //~ 1. If using lookup, absorb the commitment to the aggregation lookup polynomial.
-        self.commitments.lookup.iter().for_each(|l| {
-            absorb_commitment(&mut fq_sponge, &l.aggreg);
-        });
+        if index.lookup_index.is_some() {
+            // Should not fail, as the lookup index is present
+            let lookup_commits = self.commitments.lookup.as_ref().unwrap();
+            absorb_commitment(&mut fq_sponge, &lookup_commits.aggreg);
+        }
```
rbonichon, line 219: Tiny improvement: use the same pattern as in L195.

```suggestion
        if index.lookup_index.is_some() {
            // Should not fail, as the lookup index is present
            let lookup_commits = self.commitments.lookup.as_ref()..ok_or(VerifyError::LookupCommitmentMissing)?;
            absorb_commitment(&mut fq_sponge, &lookup_commits.aggreg);
        }
```

rbonichon (approved):
It makes the code a little bit more coherent.

volhovm (approved):
:+1: `for_each` in rust is so much confusing than e.g. `fmap` in haskell without its explicitly applicative semantics...

PR #1840, opened 2024-02-22
Adding the circuit structures to compute the polynomial coefficients

dannywillems:


dannywillems:
Closing it as it seems there is a better method.

PR #1842, opened 2024-02-22
Unify same proof file in Keccak and MIPS

querolita:
This PR leverages the generic witness definition for the zkvm to reuse the same code for the prove, fold, and verify algorithms.

Addresses https://github.com/o1-labs/proof-systems/issues/1705

I will talk to @fabrizio-m to learn how to unify these further with folding, so that different width instances can be aggregated together in one single proof (because for now, it is generating two proofs, one per instance).


dannywillems (approved):
NICE!

PR #1847, opened 2024-02-26
Implement FoldingColumnTrait for MIPS and Keccak

querolita:
Here all columns are witnesses. Perhaps when circuits are split into multiple instances, some of them might be non-witnesses, but that would come later.


dannywillems (approved):
I will accept it, even though I don't see it used atm.

PR #1854, opened 2024-02-27
Merge develop into master

volhovm:
Merge conflicts were related to shifting. More than half of them were cosmetic (comments), but some places needed to use the updated interface without shifting. Nothing hard, merge should be very OK.


mrmr1993:
Why are we doing this? In general, we should not.

mrmr1993 (changes requested):
Don't do this

mrmr1993 (approved):
My mistake :man_facepalming: LGTM, thanks!

PR #1861, opened 2024-02-28
Start dummy implementation of InterpreterEnv for Env in witness

dannywillems:


msm/src/serialization/witness.rs
```diff
@@ -16,6 +18,18 @@ pub struct Env<const N: usize, Fp> {
     pub intermediate_limbs: [Fp; 19],
 }
 
+impl<const N: usize, Fp: Field> InterpreterEnv for Env<N, Fp> {
+    type Position = Column;
+
+    // FIXME: is u128 ok? I think so, we only have 15 bits, 88 bits and 4 bits
```
volhovm, line 24: I feel it should be OK, indeed all our limbs are strictly smaller... And whatever FFMul we do it'll be on 15 bits

volhovm (approved):

PR #1866, opened 2024-02-28
Fix tuple cvars bad split index

rickwebiii:
When using a tuples as a `PublicInput` type on an `impl SnarkyCircuit`, the cvars get split in the wrong place resulting in an assertion failure. Fix that.


volhovm (approved):
Thanks @rickwebiii, this seems reasonable!

rickwebiii:
Obviated by 97a999f040.

PR #1839, opened 2024-02-22
MSM: Bring in quotient polynomial and constraints 

volhovm:
Solves #1790: 
1. Moves constraint creating logic into MSM prover/verifier
2. Hopefully wiring it up with extra simple circuit to prove it works

PR #1868, opened 2024-02-29
Serialization/MSM: create the main file

dannywillems:
This PR introduces the main binary for the serialization. This can be moved later in a test if we notice we don't need a binary.
The prover and verifier will be run once the constraints are added and https://github.com/o1-labs/proof-systems/pull/1839 is done and merged.

The witness file has been picked from the zkVM. As said offline, we will move to this structure later.
There is still a change to make the prover/verifier generic. Atm, cols[i] must be extended after each execution of the interpreter. It will be made generic soon.

### DIFF
--- a/msm/src/witness.rs
+++ b/msm/src/witness.rs
@@ -1,0 +1,92 @@
+use ark_ff::Zero;
+use rayon::iter::{FromParallelIterator, IntoParallelIterator, ParallelIterator};
+
+/// The witness columns used by a gate of the MSM circuits.
+/// It is generic over the number of columns, N, and the type of the witness, T.
+/// It can be used to represent the different subcircuits used by the project.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Witness<const N: usize, T> {
+    /// A witness row is represented by an array of N witness columns
+    /// When T is a vector, then the witness describes the rows of the circuit.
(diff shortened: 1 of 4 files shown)


msm/src/witness.rs
```diff
@@ -0,0 +1,92 @@
+use ark_ff::Zero;
+use rayon::iter::{FromParallelIterator, IntoParallelIterator, ParallelIterator};
+
+/// The witness columns used by a gate of the MSM circuits.
+/// It is generic over the number of columns, N, and the type of the witness, T.
+/// It can be used to represent the different subcircuits used by the project.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Witness<const N: usize, T> {
```
volhovm, line 8: Nit: could you please inverse template argument order? Then it matches `[T;N]` and in practice `N` be a long constant name, so we'll have `Witness<Fp, MSM_SERIALIZATION_COLUMNS_NUM>` which is IMHO more readable than `Witness<MSM_SERIALIZATION_COLUMNS_NUM, Fp>`

volhovm (approved):

msm/src/witness.rs
dannywillems: Noted for https://github.com/o1-labs/proof-systems/pull/1869

PR #1869, opened 2024-02-29
WIP - MSM: use constant size witness and rename old witness in ProofI…

dannywillems:
…nputs


msm/src/verifier.rs
```diff
@@ -50,25 +51,24 @@ pub fn verify<
     let omega = domain.d1.group_gen;
     let zeta_omega = zeta * omega;
 
-    let mut es: Vec<_> = zeta_evaluations
+    let mut es: Vec<Vec<Vec<G::ScalarField>>> = zeta_evaluations
```
dannywillems, line 54: We will change the type later. But it does imply changes outside the scope of this PR

msm/src/lib.rs
```diff
@@ -29,6 +29,11 @@ pub type BN254 = ark_ec::bn::Bn<ark_bn254::Parameters>;
 pub type BN254G1Affine = <BN254 as ark_ec::PairingEngine>::G1Affine;
 pub type BN254G2Affine = <BN254 as ark_ec::PairingEngine>::G2Affine;
 
+/// Number of columns
+/// FIXME: we must move it into the subdirectory of the
+/// foreign field addition circuit
+pub const N: usize = 3 * LIMBS_NUM;
```
volhovm, line 35: `MSM_COLUMN_N`? Can we at least keep single letters for type parameters? There are two `N` in this file...

volhovm (approved):
A lot of this stuff is very directly conflicting with my PR :( Would be nice to avoid code duplication in the future.

msm/src/main.rs
```diff
@@ -10,7 +10,7 @@ use kimchi_msm::precomputed_srs::get_bn254_srs;
 use kimchi_msm::prover::prove;
 use kimchi_msm::verifier::verify;
 use kimchi_msm::{
-    BN254G1Affine, BaseSponge, Ff1, Fp, OpeningProof, ScalarSponge, BN254, DOMAIN_SIZE,
+    BN254G1Affine, BaseSponge, Ff1, Fp, OpeningProof, ScalarSponge, BN254, DOMAIN_SIZE, N,
```
querolita, line 13: I would probably call this `N` something different here. For example, in kimchi we had something like `COLUMNS` or `KIMCHI_COLS` so it is more meaningful. Maybe `N` can be the name for the generic constant in the traits/functions, but the concrete value of it for MSM can be called `MSM_COLS` or similar.

msm/src/serialization/main.rs
```diff
@@ -3,21 +3,27 @@ use kimchi_msm::serialization::witness;
...
+use kimchi_msm::columns::Column;
 use kimchi_msm::precomputed_srs::get_bn254_srs;
+use kimchi_msm::proof::ProofInputs;
+use kimchi_msm::prover::prove;
 use kimchi_msm::serialization::witness::deserialize_field_element;
-use kimchi_msm::{Fp, BN254, DOMAIN_SIZE, LIMBS_NUM};
+use kimchi_msm::verifier::verify;
+use kimchi_msm::{BaseSponge, Fp, OpeningProof, ScalarSponge, BN254, DOMAIN_SIZE, LIMBS_NUM};
+
+const N: usize = 3 + 19 + LIMBS_NUM;
```
querolita, line 14: similar here with N

querolita (approved):
Except for minor rename, looks very nice!

msm/src/main.rs
dannywillems: Follow-up https://github.com/o1-labs/proof-systems/issues/1881

msm/src/serialization/main.rs
dannywillems: Follow-up https://github.com/o1-labs/proof-systems/issues/1881

PR #1871, opened 2024-02-29
RAMLookups initial interfaces

querolita:
This PR includes some initial interfaces to deal with RAMLookups, built on top of the MVLookups from the kimchi-msm crate. It provides a converter from RAMLookups into MVLookups, with the idea of being able to reuse the argument being built for MSM. Nonetheless, if we realize that the protocol is not general enough, further PRs will build on top of these changes to adapt the existing argument and make it more flexible.

### DIFF
--- a/msm/src/mvlookup.rs
+++ b/msm/src/mvlookup.rs
@@ -16,6 +16,22 @@ pub struct Lookup<F, ID: LookupTableID + Send + Sync + Copy> {
     pub(crate) value: Vec<F>,
 }
 
+/// Basic trait for MVLookups
+impl<F, ID> Lookup<F, ID>
+where
+    F: Clone,
+    ID: LookupTableID + Send + Sync + Copy,
+{
+    /// Creates a new MVLookup
+    pub fn new(table_id: ID, numerator: F, value: &[F]) -> Self {
+        Self {
+            table_id,
+            numerator,
+            value: value.to_vec(),
+        }
+    }
+}
+
 /// Trait for lookup table variants
 pub trait LookupTableID {
     /// Assign a unique ID to the lookup tables.
--- a/msm/src/mvlookup.rs
+++ b/msm/src/mvlookup.rs
@@ -38,6 +38,17 @@ pub trait LookupTableID {
     fn into_field<F: Field>(self) -> F;
 }
 
+/// A table of values that can be used for a lookup, along with the ID for the table.
+#[derive(Debug, Clone)]
+pub struct LookupTable<F, ID: LookupTableID + Send + Sync + Copy> {
+    /// Table ID corresponding to this table
+    #[allow(dead_code)]
+    pub table_id: ID,
+    /// Vector of values inside each entry of the table
+    #[allow(dead_code)]
+    pub entries: Vec<Vec<F>>,
+}
+
 /// Represents a witness of one instance of the lookup argument
 #[derive(Debug)]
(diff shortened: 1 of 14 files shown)


optimism/src/ramlookup.rs
```diff
@@ -0,0 +1,119 @@
...
+            table_id,
+            magnitude,
+            value: value.to_vec(),
+        }
+    }
+
+    /// Returns the numerator corresponding to this lookup in the MVLookup argument
+    pub fn numerator(&self) -> T {
+        match self.mode {
+            LookupMode::Read => T::zero() - self.magnitude.clone(),
```
volhovm, line 48: Doesn't `-self.magnitude.clone()` work?

optimism/src/lookups.rs
```diff
@@ -127,32 +67,17 @@ pub trait Lookups {
...
-#[derive(Debug, Clone)]
-pub struct LookupTable<F> {
-    /// Table ID corresponding to this table
-    #[allow(dead_code)]
-    table_id: LookupTableIDs,
-    /// Vector of values inside each entry of the table
-    #[allow(dead_code)]
-    entries: Vec<Vec<F>>,
+/// Trait that creates all the fixed lookup tables used in the VM
+pub(crate) trait FixedLookupTables<F> {
```
volhovm, line 71: Subjective: If the trait is not going to have more than one instance, one can just make these functions top-level citizens and put them in a separate file, used like `concrete_tables::table_reset()` and so on.

volhovm (approved):

msm/src/mvlookup.rs
```diff
@@ -16,12 +16,39 @@ pub struct MVLookup<F, ID: LookupTableID + Send + Sync + Copy> {
...
     fn into_field<F: Field>(self) -> F;
 }
 
+/// A table of values that can be used for a lookup, along with the ID for the table.
+#[derive(Debug, Clone)]
+pub struct LookupTable<F, ID: LookupTableID + Send + Sync + Copy> {
+    /// Table ID corresponding to this table
+    #[allow(dead_code)]
+    pub table_id: ID,
+    /// Vector of values inside each entry of the table
```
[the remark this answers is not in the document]
querolita: It actually doesn't matter how you think of them. It is just a list of values (witness variables) that contribute to the actual field entry corresponding to that lookup. I tend to imagine them as columns, but it doesn't matter.

optimism/src/ramlookup.rs
querolita: Unluckily not because that would mean that Self::Variable in the Env traits implements Neg, but that is not possible when Variable = u32 as it is the case in the witness of mips for example.

optimism/src/lookups.rs
querolita: Oh, they will probably contain more functions as we start using them in the protocol and realize we need extra functionalities 😃

msm/src/mvlookup.rs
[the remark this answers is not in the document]
volhovm: I guess because RAMLookup entires are all commutative? But we will eventually need to commit to this stuff, and that we need to do consistently. Anyway I see what you mean.

optimism/src/lookups.rs
volhovm: Nothing wrong with many top-level functions :shrug: If there are many of them it's probably good not to duplicate it twice

PR #1875, opened 2024-02-29
MSM/Serialization: add constraints file and basis environment

dannywillems:


msm/src/serialization/constraints.rs
```diff
@@ -0,0 +1,7 @@
+use kimchi::circuits::expr::{ConstantExpr, Expr};
+
+use crate::columns::Column;
+
+pub struct Env<Fp> {
```
volhovm, line 5: Also are you sure we really need an environment-typed struct for this? We will literally have one constraint per row, nothing to really build. Maybe the struct can be called `SerializationConstraint` and it contains just this vector of expressions and one `build()` constructor.

volhovm (approved):

PR #1877, opened 2024-02-29
MSM/serialization: start implementing env for constraints

dannywillems:


msm/src/serialization/constraints.rs
```diff
@@ -1,7 +1,54 @@
...
 
 use crate::columns::Column;
 
+use super::interpreter::InterpreterEnv;
+
 pub struct Env<Fp> {
     pub constraints: Vec<Expr<ConstantExpr<Fp>, Column>>,
 }
+
+impl<F: Field> InterpreterEnv for Env<F> {
```
volhovm, line 12: I'm really not sure I understand this design... `InterpreterEnv` is just like .. a bunch of common functions that can be used for building both constraints and witness? Because for building witness one actually needs some values, while for building constraints one needs merely size parameters and no real values. confused :c

volhovm (approved):

PR #1883, opened 2024-02-29
MSM/Serialization: implement bitmask_be for constraints env

dannywillems:


msm/src/serialization/constraints.rs
```diff
@@ -60,8 +60,13 @@ impl<F: Field> InterpreterEnv for Env<F> {
         _x: &Self::Variable,
         _highest_bit: u32,
         _lowest_bit: u32,
-        _position: Self::Position,
+        position: Self::Position,
     ) -> Self::Variable {
-        unimplemented!()
+        // No constraint added. It is supposed that the caller will constraint
```
volhovm, line 65: But do you plan to implement something else in this function? Like do we actually plan to have constraints/range checks here later? 

It seems quite counterintuitive to me that a function with quite a few arguments ignores most of them and just creates a new variable...

dannywillems: Yes. There will be range checks later. We build step by step.

PR #1884, opened 2024-02-29
MSM/Serialization: introduce var for intermediate limbs

dannywillems:


msm/src/serialization/main.rs
```diff
@@ -11,7 +11,7 @@ use kimchi_msm::serialization::witness::deserialize_field_element;
 use kimchi_msm::verifier::verify;
 use kimchi_msm::{BaseSponge, Fp, OpeningProof, ScalarSponge, BN254, DOMAIN_SIZE, LIMBS_NUM};
 
-const SERIALIZATION_N_COLUMNS: usize = 3 + 19 + LIMBS_NUM;
+const SERIALIZATION_N_COLUMNS: usize = 3 + N_INTERMEDIATE_LIMBS + LIMBS_NUM;
```
volhovm, line 14: I kind of like this naming guide, we could also rename `LIMBS_NUM` into `N_LIMBS`?

volhovm (approved):

msm/src/serialization/main.rs
dannywillems: Open follow-up issue: https://github.com/o1-labs/proof-systems/issues/1889

PR #1886, opened 2024-02-29
MSM/Serialization: implement first constraint

dannywillems:
Checking the decomposition in base 16 of the last limbs

### DIFF
--- a/msm/src/serialization/witness.rs
+++ b/msm/src/serialization/witness.rs
@@ -132,12 +132,18 @@ pub fn deserialize_field_element<Env: InterpreterEnv>(env: &mut Env, limbs: [u12
 
     env.copy(&input_limb0, kimchi_limbs0);
     env.copy(&input_limb1, kimchi_limbs1);
-    env.copy(&input_limb2, kimchi_limbs2);
+    let limb2_var = env.copy(&input_limb2, kimchi_limbs2);
 
     // Compute individual 4 bits limbs of b2
-    for j in 0..N_INTERMEDIATE_LIMBS {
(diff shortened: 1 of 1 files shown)


msm/src/serialization/witness.rs
```diff
@@ -132,12 +132,18 @@ pub fn deserialize_field_element<Env: InterpreterEnv>(env: &mut Env, limbs: [u12
...
-        env.bitmask_be(&input_limb2, 4 * (j + 1) as u32, 4 * j as u32, position);
+    {
+        let mut constraint = limb2_var.clone();
+        for j in 0..N_INTERMEDIATE_LIMBS {
+            let position = Env::get_column_for_intermediate_limb(j);
+            let var = env.bitmask_be(&input_limb2, 4 * (j + 1) as u32, 4 * j as u32, position);
+            let pow = Env::constant(1 << (4 * j));
+            constraint = constraint - var * pow;
+        }
+        env.add_constraint(constraint)
```
volhovm, line 146: It's quite misleading IMO that witness generation code uses the language/interface for constraint generation although it literally generates no constraints. Unless I'm missing how this code can be reused in `constraints.rs`? Do you have in mind moving it to the `interpreter.rs` later?

volhovm (approved):

msm/src/serialization/witness.rs
dannywillems: Sure. This function should be moved into interpreter.rs

dannywillems: Open follow-up issue https://github.com/o1-labs/proof-systems/issues/1890

PR #1887, opened 2024-02-29
MSM/Serialization: implement second constraint

dannywillems:
Review https://github.com/o1-labs/proof-systems/pull/1888 first

### DIFF
--- a/msm/src/serialization/witness.rs
+++ b/msm/src/serialization/witness.rs
@@ -138,14 +138,18 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
     assert!(limbs[1] < 2u128.pow(88));
     assert!(limbs[2] < 2u128.pow(79));
 
-    env.copy(&input_limb0, kimchi_limbs0);
-    env.copy(&input_limb1, kimchi_limbs1);
+    let limb0_var = env.copy(&input_limb0, kimchi_limbs0);
+    let limb1_var = env.copy(&input_limb1, kimchi_limbs1);
     let limb2_var = env.copy(&input_limb2, kimchi_limbs2);
 
+    let limb2_0_var = {
+        let limb2_0 = Env::get_column_for_intermediate_limb(0);
+        env.bitmask_be(&input_limb2, 4, 0, limb2_0)
+    };
     // Compute individual 4 bits limbs of b2
     {
-        let mut constraint = limb2_var.clone();
-        for j in 0..N_INTERMEDIATE_LIMBS {
+        let mut constraint = limb2_var.clone() - limb2_0_var.clone();
+        for j in 1..N_INTERMEDIATE_LIMBS {
             let position = Env::get_column_for_intermediate_limb(j);
             let var = env.bitmask_be(&input_limb2, 4 * (j + 1) as u32, 4 * j as u32, position);
             let pow: u128 = 1 << (4 * j);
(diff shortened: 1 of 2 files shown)


msm/src/serialization/witness.rs
```diff
@@ -138,14 +138,18 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
     assert!(limbs[1] < 2u128.pow(88));
     assert!(limbs[2] < 2u128.pow(79));
 
-    env.copy(&input_limb0, kimchi_limbs0);
-    env.copy(&input_limb1, kimchi_limbs1);
+    let limb0_var = env.copy(&input_limb0, kimchi_limbs0);
```
querolita, line 141: I am writing this here because I cannot write comments in non-diff lines.

I wonder why the input becomes constant variables. I would have imagined that whatever input to the msm you have, this will be included in a column of the gate. Constants are normally used for "real" constants like fixed numbers in equations, coefficients that are shared across all constraints, etc.

msm/src/serialization/witness.rs
```diff
@@ -138,14 +138,18 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
+    let limb1_var = env.copy(&input_limb1, kimchi_limbs1);
     let limb2_var = env.copy(&input_limb2, kimchi_limbs2);
 
+    let limb2_0_var = {
```
querolita, line 145: Why this change? It looks equivalent to me at first sight

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
-
-    env.bitmask_be(&input_limb0, 15, 0, c0);
-    env.bitmask_be(&input_limb0, 30, 15, c1);
-    env.bitmask_be(&input_limb0, 45, 30, c2);
-    env.bitmask_be(&input_limb0, 60, 45, c3);
-    env.bitmask_be(&input_limb0, 75, 60, c4);
-    {
+    let c0_var = {
+        let c0 = Env::get_column_for_msm_limb(0);
+        env.bitmask_be(&input_limb0, 15, 0, c0)
```
querolita, line 165: I was checking inside the code to understand if we're sure it was an open interval (meaning, the 15th is not included, but it goes from 0 to 14). But this was not clarified in the doc. I would suggest to add a comment clarifying that this is the case.

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
+    };
+    let c3_var = {
+        let c3 = Env::get_column_for_msm_limb(3);
+        env.bitmask_be(&input_limb0, 60, 45, c3)
+    };
+    let c4_var = {
+        let c4 = Env::get_column_for_msm_limb(4);
+        env.bitmask_be(&input_limb0, 75, 60, c4)
+    };
+    let c5_var = {
```
querolita, line 183: So the 5th is inlined because 88 is not perfectly divisible by chunks of 15bits?

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
+        let c5 = Env::get_column_for_msm_limb(5);
         let res = (limbs[0] >> 75) & ((1 << (88 - 75)) - 1);
         let res_prime = limbs[1] & ((1 << 2) - 1);
         let res = res + (res_prime << (15 - 2));
-        let res = Env::constant(res.into());
-        env.copy(&res, c5);
-    }
+        let res = Env::constant(Fp::from(res));
```
querolita, line 188: Again, I find the use of constant inaccurate for this use case...

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
-    let c9 = Env::get_column_for_msm_limb(9);
-    let c10 = Env::get_column_for_msm_limb(10);
-    let c11 = Env::get_column_for_msm_limb(11);
-    env.bitmask_be(&input_limb1, 17, 2, c6);
-    env.bitmask_be(&input_limb1, 32, 17, c7);
-    env.bitmask_be(&input_limb1, 47, 32, c8);
-    env.bitmask_be(&input_limb1, 62, 47, c9);
-    env.bitmask_be(&input_limb1, 77, 62, c10);
-    {
+    let c6_var = {
```
querolita, line 194: I am thinking about a better interface where you could just bitmask certain bit positions from a biguint or foreign field element instead of the limbs for less error-proneness. Specially thinking about these "middle" steps where you have to get bits from two different limbs.

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
         let res = res + (res_prime << (15 - 4));
         let res = Env::constant(res.into());
-        env.copy(&res, c11);
-    }
+        env.copy(&res, c11)
+    };
+
+    // Unfolding for readability.
+    // IMPROVEME using fold later.
+    // Shift left by 88
```
querolita, line 225: So this is 2^88 no?

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
+    let shl_88 = Fp::from(1u128 << 88u128);
+    let mut shl_88_var = Env::constant(shl_88);
```
querolita, line 227: Okay this looks like a good place to use `Env::constant()`, because 2^88 really is a constant for all instances.

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
+
+    // b0 + b1 * 2^88
+    let mut constraint = limb0_var + limb1_var * shl_88_var.clone();
+    // Compute shift by 176
+    shl_88_var = shl_88_var * Env::constant(shl_88);
```
querolita, line 232: Also here, good place for it.

msm/src/serialization/witness.rs
```diff
@@ -156,46 +160,119 @@ pub fn deserialize_field_element<Fp: Field, Env: InterpreterEnv<Fp>>(
...
+    cst = cst * Env::constant(Fp::from((1 << 15) as u64));
+    constraint = constraint - c8_var * cst.clone();
+
+    cst = cst * Env::constant(Fp::from((1 << 15) as u64));
+    constraint = constraint - c9_var * cst.clone();
+
+    cst = cst * Env::constant(Fp::from((1 << 15) as u64));
+    constraint = constraint - c10_var * cst.clone();
+
+    cst = cst * Env::constant(Fp::from((1 << 15) as u64));
```
querolita, line 272: you probably want to loop this

querolita (commented):
Left a few comments with questions and suggestions

msm/src/serialization/witness.rs
dannywillems: Yes

msm/src/serialization/witness.rs
dannywillems: Yes.

msm/src/serialization/witness.rs
dannywillems: `constant` is not the most appropriate function name I guess. It is to build a variable from a constant.

msm/src/serialization/witness.rs
dannywillems: It is equivalent. It was simply to use limb2_0_var below. Certainly a useless change. Let me remove it.

msm/src/serialization/witness.rs
querolita: But shouldn't you have that variable in your witness as a column already? I mean, I get it if you are in the witness code and just want to create the witness columns from witness data. But here it felt like this was constraints code, so the witness should already be created somewhere else, no?

msm/src/serialization/witness.rs
dannywillems: Opening follow-up issue https://github.com/o1-labs/proof-systems/issues/1899. Will be done directly after merging other PRs to avoid conflicts if it is fine for you.

msm/src/serialization/witness.rs
dannywillems: Opening https://github.com/o1-labs/proof-systems/issues/1900.
I think it must be included in a more generalized way of handing the input of the function.

msm/src/serialization/witness.rs
dannywillems: Added a3e039fa7cb91dd31a16427a2b570bc0bb0c612a. Feel free to close this thread if it is fine for you.

msm/src/serialization/witness.rs
dannywillems: I consider the input of the witness builder (`deserialize_field_element` in this case) being “circuit agnostic”. The function must populate the appropriate “input” columns. The first 3 columns will be populated with the inputs. We might need to change this design later. I don't know yet how we will gather all circuits.

msm/src/serialization/witness.rs
dannywillems: If it is fine for you, I will remove it after https://github.com/o1-labs/proof-systems/pull/1892 is merged. Otherwise, it will create a conflict.

msm/src/serialization/witness.rs
dannywillems: I suppose it is related to [this thread](https://github.com/o1-labs/proof-systems/pull/1887#discussion_r1511668568)

msm/src/serialization/witness.rs
dannywillems: I suppose it is more a comment and I can close this thread?

msm/src/serialization/witness.rs
dannywillems: Same than https://github.com/o1-labs/proof-systems/pull/1887#discussion_r1511694513?

msm/src/serialization/witness.rs
dannywillems: Lemme know if it is better for you https://github.com/o1-labs/proof-systems/pull/1901

querolita (approved):

PR #1891, opened 2024-03-04
MSM: Add FFA Addition gate

volhovm:


volhovm:
Closing in favour of https://github.com/o1-labs/proof-systems/pull/1904 which does the same but with micro-PR tactics.

PR #1896, opened 2024-03-04
MSM/Serialization: split range checks implementations + add multiplicities

dannywillems:
Asserting for now on the witness side.

The multiplicities are computing while we are computing the witness. In the next PRs, we will build the lookup witness from it.

~Note: box is added in 68ae3581439feff26f312272077be5615e2b3432 (next PR)~. Added now for CI

### DIFF
--- a/msm/src/serialization/witness.rs
+++ b/msm/src/serialization/witness.rs
@@ -17,8 +17,9 @@ pub struct Env<Fp> {
     /// field Kimchi gate
     pub intermediate_limbs: [Fp; N_INTERMEDIATE_LIMBS],
 
-    pub lookup_multiplicities_rangecheck4: [Fp; 1 << 4],
-    pub lookup_multiplicities_rangecheck15: [Fp; 1 << 15],
+    // Boxing to avoid stack overflow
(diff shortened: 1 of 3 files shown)


msm/src/serialization/witness.rs
```diff
@@ -15,6 +16,10 @@ pub struct Env<Fp> {
     /// Used for the decomposition in base 4 of the last limb of the foreign
     /// field Kimchi gate
     pub intermediate_limbs: [Fp; N_INTERMEDIATE_LIMBS],
+
+    // Boxing to avoid stack overflow
```
volhovm, line 20: Just curious: what's the problem and how boxing solves it? I thought `Box` is just a pointer?

volhovm (approved):

msm/src/serialization/witness.rs
dannywillems: It is allocating on the heap the vector, and use a pointer in the stack to this heap memory allocation

volhovm: But does it not /always/ do that with vectors? 

(ah, I guess with vectors but not with slices?)

Issue #1790, opened 2024-02-05
Bring expression framework to Large MSMs, implement a simple test circuit


volhovm:
Create an MSMEnv similar to how it's done in zkVM, make sure commitments are created and evaluated properly. Make sure it works with lookups.

dannywillems:
Depends on https://github.com/o1-labs/proof-systems/pull/1807 for lookups.

dannywillems:
Based on our Friday meeting, @volhovm wanted to take this task, after we went through the code together, and I explained how it should be added.

PR #1904, opened 2024-03-05
MSM FFA: Create msm/ffa folder & structure

volhovm:
Creates a directory structure for having `/ffa` code. 

Parent branch PR: https://github.com/o1-labs/proof-systems/pull/1839

Part of #1790.


msm/src/main.rs
```diff
@@ -41,7 +42,7 @@ pub fn main() {
 
     let circuit_env = generate_random_msm_witness();
     let proof_inputs = circuit_env.get_witness();
-    let constraint_exprs = circuit_env.get_exprs_add();
```
dannywillems, line 44: For a follow-up: we can move the main file in the subdirectory I think

msm/src/expr.rs
```diff
@@ -0,0 +1,45 @@
...
+///     ExprInner::<Operations<ConstantExprInner<Fp>>, Column>::Cell(Variable {
+///         col: Column::X(1),
+///         row: CurrOrNext::Curr,
+///     }),
+/// );
+/// let constraint = x3 - x1 - x2;
+/// ```
+/// A list of such constraints is used to represent the entire circuit and will
+/// be used to build the quotient polynomial.
+pub type MSMExpr<F> = Expr<ConstantExpr<F>, Column>;
```
dannywillems, line 45: For a follow-up: I think `expr.rs` can be deleted. We only need the generalized definition of  `Column`.

dannywillems (approved):
LGTM. Need to merge master into the branch.

msm/src/main.rs
volhovm: Will do

PR #1905, opened 2024-03-05
MSM: Adding simple concrete multiplication circuit to tests

volhovm:
The MSM tests were not using concrete circuits before. Now they use `a * b = c` in both completeness and soundness tests.


msm/src/lib.rs
```diff
@@ -180,7 +211,7 @@ mod tests {
         {
             let mut proof_clone = proof.clone();
             proof_clone.proof_evals.witness_evals = proof_prime.proof_evals.witness_evals;
-            let verifies = verify::<_, OpeningProof, BaseSponge, ScalarSponge, N>(
+            let verifies = verify::<_, OpeningProof, BaseSponge, ScalarSponge, MSM_FFADD_N_COLUMNS>(
```
dannywillems, line 214: We should add the reason the verification does not pass: for a follow-up.
It can be for multiple reasons that the verification does not pass, not only because of the constraints.

msm/src/lib.rs
```diff
@@ -59,19 +59,38 @@ pub type OpeningProof = PairingProof<BN254>;
...
     };
     use ark_ff::UniformRand;
     use kimchi::circuits::domains::EvaluationDomains;
     use poly_commitment::pairing_proof::PairingSRS;
+    use rand::{thread_rng, Rng};
 
     // Number of columns
     const N: usize = 10;
 
+    // Creates a test witness for a * b = c constraint.
```
dannywillems, line 77: I suppose it will disappear in next PRs when we will have FFA and FFM implemented, or simply the serialization.

dannywillems (approved):
LGTM, but I think it does not do anything new. I would have kept the previous version. I think we should get rid of the generic completeness and soundness tests and move them into the subcircuit examples. Or maybe have a generic example with the envs

msm/src/lib.rs
volhovm: Partially, yes. One still needs to decide on a /particular/ instance we're testing even though the circuit/witness builder env is in place. Maybe it can be `random_witness` in `serialization/test.rs` or similarly in `ffa/`.

Even more generally, I thought we could just have a specific easy circuit just for testing, and avoid using `ffa/` and ` serialization/` altogether here.

volhovm:
The soundness test is perhaps not more valuable than it was before, but completeness wise that we actually test the quotient polynomial actually using the constraints is definitely of value -- at the very least we are sure that we have a proper full proof running end to end. Later we can make test circuits more involved to cover different degrees of the expressions, etc. 

I'm not seeing this test suite's purpose being testing a particular circuit, but rather testing the prover/verifier, maybe that's the difference...

Issue #1879, opened 2024-02-29
Expand the specs in folding


querolita:
As an outsider of the folding module trying to integrate it into the zkVM project, I realized that the public interface is missing some important documentation to better understand what the traits are meant to represent. The [`Mock` example](https://github.com/o1-labs/proof-systems/blob/a6f1c76eed78ccf44ae82ffec9f4e73867451340/kimchi/src/folding/test.rs#L16) is useful as a skeleton, but given that the functions are empty inside (`todo!()`), there's no much concrete information one can extract from it. For that reason, I am opening an issue with the points where I felt like we could add more specs, making the folding module a much more accessible code:

Let's start with [`FoldingConfig`](https://github.com/o1-labs/proof-systems/blob/a6f1c76eed78ccf44ae82ffec9f4e73867451340/kimchi/src/folding/mod.rs#L25), as this seems to be the "parent" trait that involves all of the remaining needed ones:

```rust
pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
    type Column: FoldingColumnTrait + Debug + Eq + Hash;
    type Challenge: Clone + Copy + Debug + Eq + Hash;
    type Curve: CommitmentCurve;
    type Srs: SRS<Self::Curve>;
    type Sponge: Sponge<Self::Curve>;
    type Instance: Instance<Self::Curve>;
    type Witness: Witness<Self::Curve>;
    type Structure;
    type Env: FoldingEnv<
        <Self::Curve as AffineCurve>::ScalarField,
        Self::Instance,
        Self::Witness,
        Self::Column,
        Self::Challenge,
        Structure = Self::Structure,
    >;
    fn rows() -> usize;
}
```

## rows() 

From the signature, it seems to suggest that it refers to the number of rows of some circuit. Is it perhaps the domain size? Is it the number of elements so far pushed to some vector of witness rows? 

## Column 

This is a type that implements the [`FoldingColumnTrait`](https://github.com/o1-labs/proof-systems/blob/a6f1c76eed78ccf44ae82ffec9f4e73867451340/kimchi/src/folding/expressions.rs#L13). This is probably the clearest trait I found, but we should still include documentation for it.
```rust
pub trait FoldingColumnTrait: Copy + Clone {
    fn is_witness(&self) -> bool;
    fn degree(&self) -> Degree {
        match self.is_witness() {
            true => Degree::One,
            false => Degree::Zero,
        }
    }
}
```
[...]

PR #1821, opened 2024-02-14
Document folding + real world example

dannywillems:

PR #1906, opened 2024-03-05
Improve folding docs

querolita:
Closes https://github.com/o1-labs/proof-systems/issues/1879

For now, the full test in `mod.rs` is commented out (and still incomplete). This PR gives now an example implementation of the folding traits, as well as some doc comments for each of them.

This branch is based on that in https://github.com/o1-labs/proof-systems/pull/1821

### DIFF
--- a/kimchi/src/folding/error_term.rs
+++ b/kimchi/src/folding/error_term.rs
@@ -13,8 +13,8 @@ use poly_commitment::SRS;
 
 #[derive(Clone, Copy)]
 pub enum Side {
-    Left,
-    Right,
+    Left = 0,
+    Right = 1,
 }
 
 impl Side {
--- a/kimchi/src/folding/error_term.rs
+++ b/kimchi/src/folding/error_term.rs
@@ -11,6 +11,7 @@ use ark_ff::{Field, One};
 use ark_poly::{Evaluations, Radix2EvaluationDomain};
 use poly_commitment::SRS;
 
+/// This type refers to the two instances to be folded
 #[derive(Clone, Copy)]
 pub enum Side {
     Left = 0,
--- a/kimchi/src/folding/mod.rs
+++ b/kimchi/src/folding/mod.rs
@@ -40,7 +40,7 @@ pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
     type Instance: Instance<Self::Curve>;
 
     /// For PlonK, it will be the polynomials in evaluation form that we commit to, i.e. the columns.
-    /// In the generic prover/verifier, it would be WitnessColumns.
+    /// In the generic prover/verifier, it would be [kimchi_msm::witness::Witness].
     type Witness: Witness<Self::Curve>;
 
     type Structure;
--- a/kimchi/src/folding/mod.rs
+++ b/kimchi/src/folding/mod.rs
@@ -40,7 +40,7 @@ pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
     type Instance: Instance<Self::Curve>;
 
     /// For PlonK, it will be the polynomials in evaluation form that we commit to, i.e. the columns.
-    /// In the generic prover/verifier, it would be [kimchi_msm::witness::Witness].
+    /// In the generic prover/verifier, it would be kimchi_msm::witness::Witness.
     type Witness: Witness<Self::Curve>;
 
     type Structure;
--- a/kimchi/src/folding/mod.rs
+++ b/kimchi/src/folding/mod.rs
@@ -23,7 +23,8 @@ type ScalarField<C> = <<C as FoldingConfig>::Curve as AffineCurve>::ScalarField;
 pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
     type Column: FoldingColumnTrait + Debug + Eq + Hash;
 
-    /// The type of challenge that the Sponge returns
+    /// The type of an abstract challenge that can be found in the expressions
+    /// provided as constraints.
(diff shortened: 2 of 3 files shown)


kimchi/src/folding/mod.rs
```diff
@@ -22,13 +22,28 @@ type ScalarField<C> = <<C as FoldingConfig>::Curve as AffineCurve>::ScalarField;
 
 pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
     type Column: FoldingColumnTrait + Debug + Eq + Hash;
+
+    /// The type of challenge that the Sponge returns
```
fabrizio-m, line 26: This is not completely correct, it is the type of an abstract challenge that can be found in the expressions you provide as your constraints, but it does not represent an actual value like you can get from the sponge.

kimchi/src/folding/mod.rs
```diff
@@ -22,13 +22,28 @@ type ScalarField<C> = <<C as FoldingConfig>::Curve as AffineCurve>::ScalarField;
...
     type Srs: SRS<Self::Curve>;
+
+    /// FIXME: use Sponge from kimchi
+    /// The sponge used to create challenges
     type Sponge: Sponge<Self::Curve>;
+
+    /// FIXME: ??
     type Instance: Instance<Self::Curve>;
+
+    /// For PlonK, it will be the polynomials we commit to, i.e. the columns.
```
fabrizio-m, line 41: just mention that it will be the polynomials specifically in evaluation form.

kimchi/src/folding/mod.rs
```diff
@@ -161,18 +178,45 @@ impl<'a, F: Clone> EvalLeaf<'a, F> {
...
+    // TODO: move into `FoldingConfig`
+    // FIXME: when we move this to `FoldingConfig` it will be general for all impls as:
+    // vec![F::zero(); Self::rows()]
+    /// Returns a vector of zeros with the same length as the number of rows in the circuit.
     fn zero_vec(&self) -> Vec<F>;
+
+    /// Returns the evaluations of a given column witness at omega or zeta*omega.
     fn col(&self, col: Col, curr_or_next: CurrOrNext, side: Side) -> &Vec<F>;
+
+    // TODO: could be shared across circuits of the same type
```
fabrizio-m, line 204: That can be done already, like storing an `Rc` in the structure.

kimchi/src/folding/mod.rs
querolita: Thanks for your clarification. I guess a full test would help visualize what each part corresponds to. Right now I commented out a preliminary test inside `mod.rs` to check one single constraint, circuits of just one row. But there's still a lot to parse -> initialize the srs, domain, commit to the witnesses, fold the instances, pass the constraints, and finally prove it. Are you planning to write any test of this type?

dannywillems (approved):
Good for a first step

dannywillems:
Closing https://github.com/o1-labs/proof-systems/pull/1821 in favor of this one

kimchi/src/folding/mod.rs
```diff
@@ -22,13 +22,29 @@ type ScalarField<C> = <<C as FoldingConfig>::Curve as AffineCurve>::ScalarField;
...
+
+    /// FIXME: use Sponge from kimchi
+    /// The sponge used to create challenges
     type Sponge: Sponge<Self::Curve>;
+
+    /// For Plonk, it will be the commitments to the polynomials and the challenges
     type Instance: Instance<Self::Curve>;
+
+    /// For PlonK, it will be the polynomials in evaluation form that we commit to, i.e. the columns.
+    /// In the generic prover/verifier, it would be WitnessColumns.
```
dannywillems, line 43: Note: it changes. It is now `Witness`.
```suggestion
    /// In the generic prover/verifier, it would be Witness.
```

It might require some annotations like `kimchi_msm::witness::Witness`.

kimchi/src/folding/mod.rs
```diff
@@ -161,18 +179,45 @@ impl<'a, F: Clone> EvalLeaf<'a, F> {
...
+    // FIXME: when we move this to `FoldingConfig` it will be general for all impls as:
+    // vec![F::zero(); Self::rows()]
+    /// Returns a vector of zeros with the same length as the number of rows in the circuit.
     fn zero_vec(&self) -> Vec<F>;
+
+    /// Returns the evaluations of a given column witness at omega or zeta*omega.
     fn col(&self, col: Col, curr_or_next: CurrOrNext, side: Side) -> &Vec<F>;
+
+    // TODO: could be shared across circuits of the same type
+    /// Returns the evaluations of the i-th lagrangian term.
```
dannywillems, line 206:
```suggestion
    /// Returns the evaluations of the i-th Lagrangian term.
```
Neat: I suppose we should use capital `L`?

dannywillems: Even maybe simply `Lagrange`? Not sure we can make it an adjective from the name :).

kimchi/src/folding/error_term.rs
```diff
@@ -13,8 +13,8 @@ use poly_commitment::SRS;
 
 #[derive(Clone, Copy)]
 pub enum Side {
```
dannywillems, line 15: For a follow-up: describe what `Side` means.

kimchi/src/folding/mod.rs
querolita: Looks like capital letter is the way to go, but Lagrangian does exist indeed -> https://en.wikipedia.org/wiki/Lagrangian

PR #1907, opened 2024-03-05
MSM FFA Addition: Rewrite test circuit using interpreter pattern

### LINKED ISSUE #1791: Implement FFA addition
Implementing a simple bigint-based addition for the foreign field.

Write a test.

volhovm:
Resolves #1791 . 

This PR splits existing test `a * b - c` and `a + b - c` circuit (together with witness generation) into interpreter-witness-circuit pattern. Does not change the circuit itself, only how it's represented. Moving some code around.

Previous part: https://github.com/o1-labs/proof-systems/pull/1904

Next part will implement the actual FFA addition circuit (or part of it, depending on the diff size).

### DIFF
--- a/msm/src/ffa/columns.rs
+++ b/msm/src/ffa/columns.rs
@@ -29,3 +29,22 @@ impl ColumnIndexer for FFAColumnIndexer {
         }
     }
 }
+
+impl FFAColumnIndexer {
+    pub fn column_to_ix(col: Column) -> Self {
+        let Column::X(pos) = col;
+        let upper_bound = |i: usize| (i + 1) * N_LIMBS;
+        let pos_map = |i: usize| pos - upper_bound(i - 1);
+        if pos < upper_bound(0) {
+            FFAColumnIndexer::A(pos_map(0))
+        } else if pos < upper_bound(1) {
+            FFAColumnIndexer::B(pos_map(1))
+        } else if pos < upper_bound(2) {
+            FFAColumnIndexer::C(pos_map(2))
+        } else if pos < upper_bound(3) {
+            FFAColumnIndexer::D(pos_map(3))
+        } else {
+            panic!("column_to_ix: Invalid column index")
+        }
+    }
+}
--- a/msm/src/ffa/constraint.rs
+++ b/msm/src/ffa/constraint.rs
@@ -10,17 +10,17 @@ use kimchi::circuits::{
 };
 
 /// Contains constraints for just one row.
-pub struct ConstraintBuilder<F> {
+pub struct ConstraintBuilderEnv<F> {
     pub constraints: Vec<MSMExpr<F>>,
 }
 
-impl<F: PrimeField> FFAInterpreterEnv<F> for ConstraintBuilder<F> {
+impl<F: PrimeField> FFAInterpreterEnv<F> for ConstraintBuilderEnv<F> {
     type Position = Column;
 
     type Variable = MSMExpr<F>;
 
     fn empty() -> Self {
-        ConstraintBuilder {
+        ConstraintBuilderEnv {
             constraints: vec![],
         }
     }
--- a/msm/src/ffa/constraint.rs
+++ b/msm/src/ffa/constraint.rs
@@ -1,62 +1,61 @@
 use crate::{
     columns::{Column, ColumnIndexer},
     expr::MSMExpr,
-    ffa::columns::FFAColumnIndexer,
-    {Fp, N_LIMBS},
+    ffa::{columns::FFAColumnIndexer, interpreter::FFAInterpreterEnv},
 };
+use ark_ff::PrimeField;
 use kimchi::circuits::{
-    expr::{ConstantExprInner, ExprInner, Operations, Variable},
+    expr::{ConstantExpr, ConstantTerm, Expr, ExprInner, Variable},
     gate::CurrOrNext,
 };
 
-/// Access exprs for addition
-pub fn get_exprs_add() -> Vec<MSMExpr<Fp>> {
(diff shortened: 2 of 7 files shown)


msm/src/ffa/columns.rs
```diff
@@ -29,3 +29,22 @@ impl ColumnIndexer for FFAColumnIndexer {
         }
     }
 }
+
+impl FFAColumnIndexer {
```
dannywillems, line 33: Do you plan to use the same structure than the one in the interpreter?

msm/src/ffa/constraint.rs
```diff
@@ -1,62 +1,61 @@
...
-                col: FFAColumnIndexer::D(i).ix_to_column(),
-                row: CurrOrNext::Curr,
-            }));
-            a_i * b_i - d_i
-        };
-        limb_exprs.push(limb_constraint);
+impl<F: PrimeField> FFAInterpreterEnv<F> for ConstraintBuilderEnv<F> {
+    type Position = Column;
+
+    type Variable = MSMExpr<F>;
```
dannywillems, line 20: Personal taste: I would simply use `E<ConstrantExpr<F>, Column>` to have something generic.

msm/src/ffa/columns.rs
volhovm: Hm, I'm not exactly sure what the final structure will look like. We can chat tomorrow about it perhaps. 

Some top of the head questions (more for writing down now / discussing later):
1. Both instances of the trait use `Column` to instantiate `Position`. I'm actually not sure if any other thing makes sense, like if it even makes sense to have this parameter.  
2. If we use this parameter, maybe it makes sense to use the Indexer-type thing instead of the raw column. Just to abstract away the "flat" witness representation. Maybe we don't need `column_pos(...)` because the position could be directly expressed by the indexer.
3. These questions generalize it seems: I also want to think a bit about the "very generic interpreter trait" that can be reused across pretty much all the projects (with very minimal methods), as opposed to the "interpreter for concrete set of columns". I think we're leaning towards (2) but then some parts of the abstraction are a signs of (1) which we don't plan to do it seems, and are just rudimentary.

volhovm: On the pragmatic side I'm trying to keep kind of somewhat close to the interpreter in serialization, but thinking if some alternative way could be better, so that we can talk about both / unify them later. If an alternative is not written down it's hard to talk about it :)

msm/src/ffa/constraint.rs
volhovm: Yeah, maybe it's better. I want to kind of highlight that we're using `Column` that is an MSM type (generic, and not particular), and thus the expression is not a generic expression, but a particular over this generic column. Not sure maybe it's already clear though...

volhovm: `Expr<ConstantExpr<F>,crate::column::Column>` is a good alternative too.

msm/src/ffa/interpreter.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// TODO use more foreign_field.rs with from/to bigint conversion
+fn limb_decompose<F: PrimeField>(input: &Ff1) -> [F; N_LIMBS] {
+    let input_bi: BigUint = FieldHelpers::to_biguint(input);
+    let ff_el: ForeignElement<F, N_LIMBS> = ForeignElement::from_biguint(input_bi);
+    ff_el.limbs
+}
+
+/// Reads values from limbs A and B, returns resulting value in C.
```
dannywillems, line 38: I suppose it is not the appropriate place for this?

msm/src/ffa/main.rs
```diff
@@ -1,48 +1,48 @@
...
 
     let srs: PairingSRS<BN254> = get_bn254_srs(domain);
 
-    let circuit_env = generate_random_msm_witness();
-    let proof_inputs = circuit_env.get_witness();
-    let constraint_exprs = get_exprs_add();
+    let mut witness_env = FFAWitnessBuilderEnv::<Fp>::empty();
+    let mut constraint_env = FFAConstraintBuilderEnv::<Fp>::empty();
+
+    ffa_interpreter::constrain_multiplication(&mut constraint_env);
```
dannywillems, line 32: Is it addition or multiplication?

dannywillems (approved):
It seems that the directory contains code unrelatd to foreign field addition.
Please remove. Approving in the meantime to get in master and merge other PRs.

Would be nice to have tests for the witness environment like [here](https://github.com/o1-labs/proof-systems/blob/master/msm/src/serialization/witness.rs#L147).

msm/src/ffa/interpreter.rs
volhovm: Why isn't it? I mean the comment is not descriptive I agree :) But it's technically correct? I wanted to say what's the interface for this, that it's a constraint (thus function name) + it returns some values.

msm/src/ffa/main.rs
volhovm: Per-row `a * b - c`, so multiplication, yes.

volhovm:
@dannywillems Sure, by the unrelated code you mean these "test" additions/multiplications? If yes then we're on the same page:
- I plan to write a test for actual ffa addition (basically convert ffa/main.rs to a test).
- And then I will move everything that is not ffa into `test/` perhaps so that we have simple test circuits too for testing/debugging.

msm/src/ffa/interpreter.rs
dannywillems: I meant the function.

PR #1909, opened 2024-03-06
Add ProverError to prover.rs

volhovm:
Closes #1708. Adds `ProverError` that can be used instead of panicking in `prover.rs`.


volhovm:
@dannywillems the task took more than 10min so I created an issue that is closed from the commit. I expect the issue to be automatically closed when PR is merged? let's see how it works...

dannywillems (approved):

PR #1910, opened 2024-03-06
Make MSM dependencies use workspace pattern as other crates

volhovm:
Makes MSM project use top level workspace dependencies instead of individual one, as other crates in the repository do unless necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1432,7 +1432,7 @@ dependencies = [
  "rmp-serde",
  "serde",
  "serde_json",
- "serde_with 1.14.0",
+ "serde_with 3.6.0",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1432,7 +1432,7 @@ dependencies = [
  "rmp-serde",
  "serde",
  "serde_json",
- "serde_with 1.14.0",
+ "serde_with 3.6.0",
```
dannywillems, line 1435: Expected?

volhovm: That's the version used across all the other crates (except us?), so I guess it's expected yes.

PR #1919, opened 2024-03-06
Generic prover: add public input size

dannywillems:
Will be used later

### DIFF
--- a/msm/src/proof.rs
+++ b/msm/src/proof.rs
@@ -20,6 +20,9 @@ pub struct ProofInputs<const N: usize, G: KimchiCurve, ID: LookupTableID + Send 
     /// evaluations of polynomial P_w that interpolates w_i.
     pub evaluations: Witness<N, Vec<G::ScalarField>>,
     pub mvlookups: Vec<MVLookupWitness<G::ScalarField, ID>>,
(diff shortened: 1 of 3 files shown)


msm/src/proof.rs
```diff
@@ -20,6 +20,9 @@ pub struct ProofInputs<const N: usize, G: KimchiCurve, ID: LookupTableID + Send
     /// evaluations of polynomial P_w that interpolates w_i.
     pub evaluations: Witness<N, Vec<G::ScalarField>>,
     pub mvlookups: Vec<MVLookupWitness<G::ScalarField, ID>>,
+    /// The number of public input. It is supposed that the first
```
querolita, line 23:
```suggestion
    /// The number of public inputs. It is supposed that the first
```

querolita (approved):
Where will it be used? I guess it is not used in the zkvm project... So in that case it will remain as 0 I suppose.

dannywillems:
> Where will it be used? I guess it is not used in the zkvm project... So in that case it will remain as 0 I suppose.

I am using public inputs at the moment for the serialization subcircuit, to load the kimchi limbs in the circuit. We could also use lookup tables.

PR #1920, opened 2024-03-07
Expr: add some documentations

dannywillems:


querolita (approved):
I wonder if you forgot about `get_domain` and `vanishes_on_zero_knowledge_and_previous_rows` or if this was intentional left for later.

PR #1926, opened 2024-03-07
add range check support for snarkyrs

fabrizio-m:
Adding support to use range checks in snarkyrs


dannywillems:
@fabrizio-m can you make the CI happy please? :)

kimchi/src/snarky/range_checks.rs
```diff
@@ -0,0 +1,268 @@
...
+            let v: FieldVar<Fp> = sys.compute(loc!(), |_| *private.unwrap())?;
+
+            sys.range_check(loc!(), v.clone(), v.clone(), v)?;
+
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn test_bits() {
```
dannywillems, line 203: Is it a test?

dannywillems: I would remove it as it is not a test and it is not related to range_checks.

kimchi/src/snarky/constraint_system.rs
```diff
@@ -1649,6 +1651,38 @@ impl<Field: PrimeField> SnarkyConstraintSystem<Field> {
                     self.add_row(labels, loc, vars, GateType::EndoMulScalar, vec![]);
                 }
             }
+            KimchiConstraint::RangeCheck(rows) => {
+                let rows: Result<[Vec<Cvar>; 4], _> = rows.try_into();
+                let rows: Result<[[Cvar; 15]; 4], _> = rows.map(|rows| {
```
dannywillems, line 1656: Suggestion: I would use the constant from Kimchi for the number of columns.

volhovm (approved):

PR #1931, opened 2024-03-07
MSM/Serialization: test completeness

dannywillems:
Move main file into test, to have it in the CI


msm/Cargo.toml
```diff
@@ -16,10 +16,6 @@ path = "src/lib.rs"
 name = "ffa"
 path = "src/ffa/main.rs"
 
-[[bin]]
-name = "serialization"
-path = "src/serialization/main.rs"
```
volhovm, line 21: I have a feeling we /might/ actually want to have a separate executable to test a separate circuit, maybe for debugging. But accepting if you find it unnecessary for now.

volhovm (approved):

PR #1932, opened 2024-03-07
Generalize ForeignElement to be parameterized over LIMB_BITSIZE

volhovm:
Indirect part of #1791 -- I was using these functions for computing limbs, until I realised they were actually using hardcoded `LIMBS_NUM = 88`. 

This PR generalises the library a bit, basically adding some code for the case `B = LIMBS_NUM != 88`. I did not touch the code for the cases where `B % 8 == 0`, which is more efficient, but I added some code for when `B % 8 != 0` (such as `B = 15` as in MSM), where implementations use bit decomposition (less efficient, c'est la vie).

I also modified the tests, according to which it all works.


dannywillems:
Can you fix the conflicts please? @volhovm

utils/src/foreign_field.rs
```diff
@@ -182,10 +218,10 @@ pub trait BigUintForeignFieldHelpers {
     /// 2
     fn two() -> Self;
 
-    /// 2^{LIMB_SIZE}
+    /// 2^{LIMB_BITS}
```
dannywillems, line 221: The names are now misleading as some functions use `LIMB_BITS = 88`, defined at the top level of the module
I would rename (or gather in a `kimchi` nested module) the functions depending on the 88-bits limb encoding.

utils/src/foreign_field.rs
```diff
@@ -413,33 +449,55 @@ mod tests {
         let big = secp256k1_modulus();
         let bytes = big.to_bytes_be();
         assert_eq!(
-            ForeignElement::<BaseField, 3>::from_be(&bytes),
-            ForeignElement::<BaseField, 3>::from_biguint(big)
+            ForeignElement::<BaseField, LIMB_BITS, 3>::from_be(&bytes),
```
dannywillems, line 452: (we cannot comment on unchanged lines)
Not for this PR, but I would use a randomized seed, see above:
```
    const RNG_SEED: [u8; 32] = [
        12, 31, 143, 75, 29, 255, 206, 26, 67, 193, 86, 160, 1, 90, 131, 221, 86, 168, 4, 95, 50,
        48, 89, 29, 13, 250, 215, 172, 130, 24, 164, 162,
    ];
```

It is a small change I would suggest to do as a quick follow-up after.

utils/src/foreign_field.rs
```diff
@@ -413,33 +449,55 @@ mod tests {
...
+            ForeignElement::<BaseField, LIMB_BITS, 3>::from_biguint(big.clone())
+        );
+        assert_eq!(
+            ForeignElement::<BaseField, 15, 18>::from_be(&bytes),
```
dannywillems, line 456: I would suggest making the tests even more generic, to be sure we do not only test cases we want.
For instance, we could take some random decompositions in two numbers of the field modulus.

I would then get rid of `LIMB_BITS = 88` (as it is now misleading as described above), and define a nested module `kimchi` or smth close to it.

utils/src/foreign_field.rs
```diff
@@ -98,47 +98,83 @@ impl<F: Field, const N: usize> ForeignElement<F, N> {
...
+            for i in 0..bits.len() {
+                bytes[i / 8] |= u8::from(bits[i]) << (i % 8);
+            }
         }
         BigUint::from_bytes_le(&bytes)
     }
 
-    /// Split a foreign field element into a vector of `LIMB_BITS` bits field elements of type `F` in little-endian.
+    /// Split a foreign field element into a vector of `B` bits field elements of type `F` in little-endian.
     /// Right now it is written so that it gives `LIMB_COUNT` limbs, even if it fits in less bits.
```
dannywillems, line 129: This comment is out-dated now I think.

utils/src/foreign_field.rs
```diff
@@ -413,33 +449,55 @@ mod tests {
         let big = secp256k1_modulus();
```
dannywillems, line 449: A bit unrelated to this PR, but still good to have a for a quick follow-up: generalizing the foreign field, and take a random prime field.

utils/src/foreign_field.rs
volhovm: Hmmm I think it's actually good to do it in this PR too to make sure that the tests /actually/ work and nothing breaks?

dannywillems: I'm not against :)

volhovm: I did it here: https://github.com/o1-labs/proof-systems/pull/1942

volhovm:
@dannywillems Let's merge this PR? I created a generalisation follow-up here: https://github.com/o1-labs/proof-systems/pull/1951

dannywillems (approved):

PR #1933, opened 2024-03-07
MSM/MVLookup: add challenges into column environment

dannywillems:
The generation of the challenges should be at the same round for both the prover and the verifier.
We do not have a typed sponge unfortunately to detect that we coin at the right time. Maybe later.
I am integrating lookups for the serialization subcircuits. We will have tests when this is done.

### DIFF
--- a/msm/src/prover.rs
+++ b/msm/src/prover.rs
@@ -153,22 +153,13 @@ where
 
     let zk_rows = 0;
     let column_env = {
-        let challenges = if let Some(lookup_env) = lookup_env.as_ref() {
-            Challenges {
-                alpha,
-                // NB: as there is on permutation argument, we do use the beta
-                // field instead of a new one for the evaluation point.
-                beta: lookup_env.beta,
-                gamma: G::ScalarField::zero(),
-                joint_combiner: Some(lookup_env.joint_combiner),
-            }
-        } else {
-            Challenges {
-                alpha,
-                beta: G::ScalarField::zero(),
-                gamma: G::ScalarField::zero(),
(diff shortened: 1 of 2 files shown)


msm/src/prover.rs
```diff
@@ -152,23 +152,37 @@ where
...
-        domain,
+    let column_env = {
+        let challenges = if let Some(lookup_env) = lookup_env.as_ref() {
+            Challenges {
+                alpha,
+                // NB: as there is on permutation argument, we do use the beta
+                // field instead of a new one for the evaluation point.
+                beta: lookup_env.beta,
+                gamma: G::ScalarField::zero(),
+                joint_combiner: Some(lookup_env.joint_combiner),
```
volhovm, line 163: Suggestion: keeping just one set of Challenges but doing:
`lookup_env.map(|x| x.joint_combiner)` 
or
`lookup_env.as_ref().map(|x| x.joint_combiner)`

and same with `beta`

volhovm (approved):

msm/src/prover.rs
dannywillems: Can not be used for beta, as it is not a map on an option

volhovm: Ah yes indeed. Smth like `lookup_env.map(|x| x.beta).unwrap_or(F::zero())`?

dannywillems: Done in https://github.com/o1-labs/proof-systems/pull/1933/commits/095d2b23a00a4e3d5260b0eb3038fc77d8d8de7f

PR #1935, opened 2024-03-07
MSM: introduce different lookup tables used by the serialization

dannywillems:


msm/src/serialization/mod.rs
```diff
@@ -6,6 +10,20 @@ pub mod constraints;
 pub mod interpreter;
 pub mod witness;
 
+pub enum LookupTable {
```
volhovm, line 13: I would also want to discuss these interfaces at some point, they're all different in kimchi/mips and not always intuitive IMO. Like I would assume `LookupTable` is the actual /table/, like the data. What this is is more of a... enum of concrete types? `SerializationLookupTables`? just `LookupTables`? `LookupTableIDs`? I don't have a better name :(

volhovm (approved):

msm/src/serialization/mod.rs
dannywillems: I also want to change later. Let's keep it in mind for when all circuits are defined and working.

PR #1936, opened 2024-03-07
MSM/Serialization: use domain of 1 << 15

dannywillems:


msm/src/serialization/mod.rs
```diff
@@ -47,23 +47,34 @@ mod tests {
     #[test]
     fn test_completeness() {
         let mut rng = o1_utils::tests::make_test_rng();
-        const DOMAIN_SIZE: usize = 1 << 5;
+        // Must be at least 1 << 15 to support rangecheck15
```
volhovm, line 50: This will slow things quite a bit... either we need cached SRS (preferred by me), or we could make the gate generic and run tests with `LIMB_SIZE = 7` for example.

msm/src/serialization/mod.rs
```diff
@@ -47,23 +47,34 @@ mod tests {
...
-        ]; DOMAIN_SIZE];
+        });
+
+        // Boxing to avoid stack overflow
+        let mut field_elements = vec![];
+        // FIXME: we do use always the same values here, because we have a
+        // constant check (X - c), different for each row. And there is no
+        // constant support/public input yet in the quotient polynomial.
+        let (x, y, z) = (
+            rng.gen_range(0..1000000),
```
volhovm, line 71: If I understand correctly `10^6 > 2^15` so several small 15-bit limbs are involved, but then the test would be more complete with actually full-domain 88-bit inputs (and the last limb a bit smaller). Just for the future.

volhovm (approved):

PR #1937, opened 2024-03-08
Keccak bug fix: Include column for block index

querolita:
I realized that the syscalls lookups in the Keccak side were using a value that was being stored as part of the environment but did not have a corresponding column for the block index (used to give a byte index for each byte read from the preimage). Then, this PR adds the alias for the block index, increases by 1 the witness columns width, adds a function to access that variable, and updates the witness code to initialize this value in absorb steps.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -16,8 +16,10 @@ use std::ops::{Index, IndexMut};
 
 /// The total number of witness columns used by the Keccak circuit.
 pub const ZKVM_KECCAK_COLS: usize =
-    ZKVM_KECCAK_COLS_CURR + ZKVM_KECCAK_COLS_NEXT + MODE_FLAGS_COLS_LEN + 2;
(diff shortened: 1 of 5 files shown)


optimism/src/keccak/column.rs
```diff
@@ -16,8 +16,10 @@ use std::ops::{Index, IndexMut};
 
 /// The total number of witness columns used by the Keccak circuit.
 pub const ZKVM_KECCAK_COLS: usize =
-    ZKVM_KECCAK_COLS_CURR + ZKVM_KECCAK_COLS_NEXT + MODE_FLAGS_COLS_LEN + 2;
+    ZKVM_KECCAK_COLS_CURR + ZKVM_KECCAK_COLS_NEXT + MODE_FLAGS_COLS_LEN + STATUS_FLAGS_LEN;
 
+// The number of columns used by the Keccak circuit to represent the status flags.
+const STATUS_FLAGS_LEN: usize = 3;
```
dannywillems, line 22: Curious that the compiler does not complain that the variable is defined after we use it :thinking:

dannywillems (approved):

PR #1944, opened 2024-03-08
MSM/Serialization: start lookup

dannywillems:


msm/src/serialization/mod.rs
```diff
@@ -44,6 +48,40 @@ mod tests {
...
+        fn into_lookup_vector<F: FftField + PrimeField + Field>(
+            self,
+            domain: EvaluationDomains<F>,
+        ) -> Vec<Lookup<F>> {
+            assert!(domain.d1.size >= (1 << 15));
+            match self {
+                Self::RangeCheck15 => (0..(1 << 15))
+                    .map(|i| Lookup {
+                        table_id: LookupTable::RangeCheck15,
+                        numerator: -F::one(),
```
volhovm, line 63: Why are these `-1` by default? I'd assume that by default we are not using any of the elements, so they should have `0`?

msm/src/serialization/mod.rs
```diff
@@ -102,11 +145,83 @@ mod tests {
                     constraints.push(cst.clone())
                 }
             }
+
+            for (j, lookup) in witness_env.rangecheck4_lookups.iter().enumerate() {
```
volhovm, line 149: Probably next step, but this code that collects something from witness and passes to MVLookup can be arguably moved to `witness.rs` so that `witness_env.get_inputs` would give you the inputs directly

msm/src/serialization/witness.rs
```diff
@@ -17,9 +20,26 @@ pub struct Env<Fp> {
...
     // Boxing to avoid stack overflow
     pub lookup_multiplicities_rangecheck4: Box<[Fp; 1 << 4]>,
+
+    /// Keep track of the RangeCheck4 table multiplicities.
+    /// The value `0` is used as a (repeated) dummy value.
+    // Boxing to avoid stack overflow
+    pub lookup_t_multiplicities_rangecheck4: Box<[Fp; 1 << 4]>,
+
+    /// Keep track of the RangeCheck15 lookup multiplicities
+    /// No t multiplicities as we do suppose we have a domain of
```
volhovm, line 33: How does this work? You still need to construct the multiplicities for `t`, I guess you wanted to say that it's easier in the case of 2^15 domain size, but why? :thinking:

msm/src/serialization/witness.rs
```diff
@@ -112,6 +140,43 @@ impl<Fp: PrimeField> Env<Fp> {
...
+            .collect();
+        m.extend(repeated_dummy_value);
+        m
+    }
+
+    pub fn get_rangecheck15_normalized_multipliticies(
+        &self,
+        _domain: EvaluationDomains<Fp>,
+    ) -> Vec<Fp> {
+        self.lookup_multiplicities_rangecheck15.to_vec()
```
volhovm, line 179: Ah OK, answered my own question about why 15bit case does not have `t`. The multiplicities are /exactly/ equal to the "called" vector that we build in the witness, so they are much easier to compute.

msm/src/serialization/witness.rs
```diff
@@ -112,6 +140,44 @@ impl<Fp: PrimeField> Env<Fp> {
...
+    pub fn get_rangecheck4_normalized_multipliticies(
+        &self,
+        domain: EvaluationDomains<Fp>,
+    ) -> Vec<Fp> {
+        let mut m = vec![Fp::zero(); 1 << 4];
+        self.lookup_multiplicities_rangecheck4
+            .into_iter()
+            .zip(self.lookup_t_multiplicities_rangecheck4.iter())
+            .enumerate()
+            .for_each(|(i, (m_f, m_t))| m[i] = m_f / m_t);
```
volhovm, line 166: Why do we need to do this (field? or is it integer?) division here?  Is it related to the dummy values?..

msm/src/serialization/mod.rs
```diff
@@ -102,11 +145,83 @@ mod tests {
...
+                rangecheck4[j].push(lookup.clone())
+            }
+
+            for (j, lookup) in witness_env.rangecheck15_lookups.iter().enumerate() {
+                rangecheck15[j].push(lookup.clone())
+            }
+
+            witness_env.add_rangecheck4_table_value(i);
```
volhovm, line 157: I feel like this needs to be called somewhere from `range_check4()` function on the witness `impl InterpreterEnv` side, no?

msm/src/serialization/witness.rs
```diff
@@ -112,6 +140,44 @@ impl<Fp: PrimeField> Env<Fp> {
             }
         }
     }
+
+    pub fn add_rangecheck4_table_value(&mut self, i: usize) {
+        if i < (1 << 4) {
+            self.lookup_t_multiplicities_rangecheck4[i] += Fp::one();
+        } else {
+            self.lookup_t_multiplicities_rangecheck4[0] += Fp::one();
```
volhovm, line 148: Is this dummy value handling? Wondering if it's not bringing any soundness issues, since in the first `if` case you also add to the `[0]` element, right?

volhovm (approved):

msm/src/serialization/mod.rs
dannywillems: See equation (15): https://eprint.iacr.org/2022/1530.pdf.

msm/src/serialization/witness.rs
dannywillems: See equation (14): https://eprint.iacr.org/2022/1530.pdf

msm/src/serialization/witness.rs
dannywillems: See equation 14: https://eprint.iacr.org/2022/1530.pdf

msm/src/serialization/witness.rs
dannywillems: See equation 14: https://eprint.iacr.org/2022/1530.pdf

msm/src/serialization/mod.rs
dannywillems: I would like to move it somewhere else, yes. The table should be built outside of the interpreter as it is a fixed table. We should have a way to handle multiple domain size.

msm/src/serialization/mod.rs
volhovm: Ah, ok, I think I misunderstood what this function does. Please tell me if I'm right: what you're doing is given a fixed table you're building a set of queries for that table covering every element of this table?

msm/src/serialization/witness.rs
volhovm: I thought (14) does not apply to our case because our `t(x)` is fully injective, since every element in the range table is unique?

PR #1650, opened 2023-12-30
Add Rust 1.75

dannywillems:
Require some changes to maek clippy happy.

PR #1945, opened 2024-03-08
CI: update rust toolchains

dannywillems:
See https://github.com/o1-labs/proof-systems/pull/1650#discussion_r1473679152

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,7 +15,7 @@ jobs:
   run_checks:
     strategy:
       matrix:
-        rust_toolchain_version: ["1.71", "1.72", "1.73", "1.74"]
+        rust_toolchain_version: ["1.71", "1.72", "1.73", "1.74", "1.75", "1.76", "beta"]
         # FIXME: currently not available for 5.0.0.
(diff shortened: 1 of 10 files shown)


.github/workflows/rust.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   run_checks:
     strategy:
       matrix:
-        rust_toolchain_version: ["1.71", "1.72", "1.73", "1.74"]
+        rust_toolchain_version: ["1.71", "1.72", "1.73", "1.74", "1.75", "1.76", "beta"]
```
volhovm, line 18: Are you sure about `beta` though?..

volhovm (approved):
New clippy -- new life! :tada:

.github/workflows/rust.yml
volhovm: + do we have to support 1.71 still? CI might not like building 75% more versions than it does now.

dannywillems:
Oh no, the doc does not build...
```
error: internal compiler error: compiler/rustc_middle/src/ty/generic_args.rs:184:43: expected a const, but found another kind
```

dannywillems:
Gonna close it as old.

PR #1957, opened 2024-03-08
Move constraints into Keccak interpreter

querolita:
This moves the constraints into the interpreter. Note that it has no different implementation for witness/constraints thanks to the specialization of the trait. This means, when it is run on constraints env, it will actually create constraints with expressions. But with witness env, it checks assert_equals of the field values. This will allow for easy testing in a future PR.


dannywillems (approved):
nice

PR #1959, opened 2024-03-08
folding example

fabrizio-m:


querolita (approved):
I found this really helpful for [this PR](https://github.com/o1-labs/proof-systems/pull/2015), thanks!

I think this will also be key when I integrate the folding/proving part into zkvm.

PR #1961, opened 2024-03-11
MSM: Add basic public input support

volhovm:
This adds very basic public input support, namely that the verifier can explicitly specify some columns of the witness (its prefix) instead of fully relying on the commitments in the proof.


volhovm:
@dannywillems rebased this now, should be an easy PR.

dannywillems (approved):

PR #1963, opened 2024-03-12
ASCII Keccak doc comment + remove redundancy

querolita:
This PR addresses a follow-up issue brought up by Danny, who was suggesting to include a more ASCII-like type of comment to better visualize the content of the Keccak columns. 

While doing so, I realized that 4 columns were being reserved for the round constants both inside `mode_flags` and inside `curr`. I removed this redundancy to only be stored inside the former. That makes the column count 2220. This "fix" was not really a bug, but rather a non-necessary duplicate. Thus, tests worked both with it and without it.

TODO: when the instances are split into types (probably just round and sponge), we will be able to split the columns further and use less columns for sponge instances, and run fewer constraints on them.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -96,9 +96,52 @@ impl FoldingColumnTrait for Column {
 /// The row is split into the following entries:
 /// - hash_index: Which hash this is inside the circuit
 /// - step_index: Which step this is inside the hash
-/// - mode_flags: Round, Absorb, Squeeze
-/// - curr: Contains 1969 witnesses used in the current step including Input and RoundConstants
+/// - mode_flags: Round, Absorb, Squeeze, Root, PadLength, InvPadLength, TwoToPad, PadBytesFlags, PadSuffix, RoundConstants
+/// - curr: Contains 1969 witnesses used in the current step including Input
 /// - next: Contains the Output
+///
+///   Keccak Witness Columns: KeccakWitness.cols
+///  ----------------------------------------------
+/// | 0 | 1 | 2 | 3..154 | 155..2119 | 2120..2219 |
+///  ----------------------------------------------
+///   0 -> hash_index
+///   1 -> block_index
+///   2 -> step_index
+///   3..154 -> mode_flags
+///          -> 3: FlagRound
+///          -> 4: FlagAbsorb
+///          -> 5: FlagSqueeze
+///          -> 6: FlagRoot
+///          -> 7..142: PadBytesFlags
+///          -> 143: PadLength
+///          -> 144: InvPadLength
+///          -> 145: TwoToPad
+///          -> 146..150: PadSuffix
+///          -> 151..154: RoundConstants
+///   155..2123 -> curr
+///         155                                                                        2119
+///          <--------------------------------if_round<---------------------------------->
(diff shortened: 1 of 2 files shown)


optimism/src/keccak/column.rs
```diff
@@ -96,9 +96,52 @@ impl FoldingColumnTrait for Column {
...
+///                                              | -> 400..419: ThetaExpandRotC
+///                                              | -> 420..819: PiRhoShiftsE
+///                                              | -> 820..919: PiRhoDenseE
+///                                              | -> 920..1019: PiRhoQuotientE
+///                                              | -> 1020..1119: PiRhoRemainderE
+///                                              | -> 1120..1219: PiRhoDenseRotE
+///                                              | -> 1220..1319: PiRhoExpandRotE
+///                                              | -> 1320..1719: ChiShiftsB
+///                                              | -> 1720..2119: ChiShiftsSum
+///   2124..2223 -> next
```
dannywillems, line 142: 2219 or 2223? :thinking:

optimism/src/keccak/column.rs
```diff
@@ -96,9 +96,52 @@ impl FoldingColumnTrait for Column {
...
+///          -> 151..154: RoundConstants
+///   155..2123 -> curr
+///         155                                                                        2119
+///          <--------------------------------if_round<---------------------------------->
+///          <-------------if_sponge------------->    
+///         155                                 954           
+///          -> SPONGE:                          | -> ROUND:    
+///         -> 155..254: Input == SpongeOldState | -> 155..254: Input == ThetaStateA    
+///         -> 255..354: SpongeNewState          | -> 255..334: ThetaShiftsC           
+///                    : 323..354 -> SpongeZeros | -> 335..354: ThetaDenseC
```
volhovm, line 129: Why is this `SpongeZero` part aligned differently?

optimism/src/keccak/column.rs
```diff
@@ -96,9 +96,52 @@ impl FoldingColumnTrait for Column {
...
 /// - step_index: Which step this is inside the hash
-/// - mode_flags: Round, Absorb, Squeeze
-/// - curr: Contains 1969 witnesses used in the current step including Input and RoundConstants
+/// - mode_flags: Round, Absorb, Squeeze, Root, PadLength, InvPadLength, TwoToPad, PadBytesFlags, PadSuffix, RoundConstants
+/// - curr: Contains 1969 witnesses used in the current step including Input
 /// - next: Contains the Output
+///
+///   Keccak Witness Columns: KeccakWitness.cols
+///  ----------------------------------------------
+/// | 0 | 1 | 2 | 3..154 | 155..2119 | 2120..2219 |
```
volhovm, line 105: `2119` here versus `2123` below.

volhovm (approved):
_Potentially inconsistent_ as Danny also noticed + in the previous `2119` index. Approving, but please address it before merging.

optimism/src/keccak/column.rs
querolita: 2219 😅

optimism/src/keccak/column.rs
querolita: It refers to the last 32 bytes of the new state which are supposed to be zeros.

PR #1966, opened 2024-03-12
Improve lookup support

dannywillems:
TODO:
- [ ] integrate with serialization
- [ ] implement the lookup env for the quotient polynomial
- [ ] fix domain for evaluation in mvlookup.rs

### DIFF
--- a/msm/src/ffa/constraint.rs
+++ b/msm/src/ffa/constraint.rs
@@ -1,6 +1,6 @@
 use crate::{
     columns::{Column, ColumnIndexer},
-    expr::MSMExpr,
+    expr::E,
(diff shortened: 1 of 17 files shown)


msm/src/ffa/constraint.rs
```diff
@@ -1,6 +1,6 @@
 use crate::{
     columns::{Column, ColumnIndexer},
-    expr::MSMExpr,
+    expr::E,
```
volhovm, line 3: I'm happy with removing `MSM` from `MSMExpr`, but can we at least have `Expr` instead of just `E`? One can always do `expr::Expr as E` to save up three letters (although I think it decreases readability significantly, similarly to `D` sometimes used alongside `Domain`).

dannywillems:
Closing as it is done in smaller PRs.

PR #1968, opened 2024-03-13
[easy] Fine-grain negative tests for Keccak constraints

querolita:
This PR adds negative tests for the Keccak constraints with a fine-grain design.

In particular, I created an error enum to keep track of which constraint failed throughout execution. This way, when the constraints (either all, or just a subset of them) are run over a witness row, they keep track of the errors accumulated so far (if any) inside the `errors` variable in the witness environment. 

This way, each constraint is now paired with a tag with a name. That is the identifier used in the tests to assert that the expected errors were produced.

### DIFF
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -7,7 +7,9 @@
 //! For a pseudo code implementation of Keccap-f, see
 //! <https://keccak.team/keccak_specs_summary.html>
 use crate::{
-    keccak::{column::KeccakWitness, interpreter::KeccakInterpreter, KeccakColumn, KeccakError},
+    keccak::{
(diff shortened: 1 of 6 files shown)


optimism/src/keccak/witness.rs
```diff
@@ -60,14 +62,10 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
     }
 
     /// Assert that the input is zero
```
dannywillems, line 64: The comment is not correct anymore I think.

dannywillems (approved):
LGTM. However, I didn't check the correctness of all the tests.

Tests can even be more generalized (but it is fine atm).
For instance, any non zero value could be generated for `FlagAbsorb` when checking Boolean constraints.

Tests could be separated to use their own witness environment (by using scope and cloning a "reference" environment and provide more fine-grained scenarii.

PR #1969, opened 2024-03-13
[no-op] Move impl of lookup functions inside Keccak interpreter

querolita:
This PR simply moves the implementation of the lookup-related functions from `constraints.rs` to `interpreter.rs` because they will be shared across witness and constraints environments.

The only difference will be in `add_lookups()`, which will behave differently for each type of environment. This change will introduce multiplicities in a coming PR.


dannywillems (approved):
LGTM.

PR #1973, opened 2024-03-13
Compute multiplicities of Keccak lookups in the witness environment

querolita:
This PR addresses the `todo!()` inside `add_lookup()` of the witness environment of Keccak.  Here, when the lookup is a "fixed table lookup" (as opposed to read/write lookups, where it is a no-op for the witness), and the condition is satisfied (example, `is_round()` or `is_sponge()`), then it checks that the desired value is indeed in the given table. If this is not the case, the function pushes a lookup error to the list of errors in the environment. This latter behavior will be used in a coming PR in negative tests.

I added one simple positive test just to show that it works. But for now, it seems to be too slow when the tables and the variables are F. Perhaps I will create a follow-up PR where `Self::Variable = i128` instead for witnesses, so it is faster. No priority for now, as this only has an effect in tests speed.

### DIFF
--- a/optimism/src/keccak/mod.rs
+++ b/optimism/src/keccak/mod.rs
@@ -1,10 +1,9 @@
+use crate::{keccak::column::Column as KeccakColumn, lookups::LookupTableIDs};
 use ark_ff::Field;
 use kimchi::circuits::{
     expr::{ConstantExpr, Expr},
     polynomials::keccak::constants::{DIM, KECCAK_COLS, QUARTERS, RATE_IN_BYTES, STATE_LEN},
 };
-
-use self::column::Column as KeccakColumn;
 
 pub mod column;
 pub mod constraints;
@@ -34,7 +33,7 @@ pub(crate) type E<F> = Expr<ConstantExpr<F>, KeccakColumn>;
 pub enum Error {
     Constraint(Constraint),
     #[allow(dead_code)]
-    Lookup(usize),
+    Lookup(LookupTableIDs),
 }
 
 /// All the names for constraints involved in the Keccak circuit
--- a/optimism/src/keccak/mod.rs
+++ b/optimism/src/keccak/mod.rs
@@ -32,7 +32,6 @@ pub(crate) type E<F> = Expr<ConstantExpr<F>, KeccakColumn>;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Error {
     Constraint(Constraint),
-    #[allow(dead_code)]
     Lookup(LookupTableIDs),
 }
 
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -21,7 +21,7 @@ use kimchi_msm::LookupTableID;
 pub struct Env<Fp> {
     /// The full state of the Keccak gate (witness)
     pub witness: KeccakWitness<Fp>,
-    /// The multiplicities of each lookup entry
+    /// The multiplicities of each lookup entry. Should not be cleared between steps.
     pub multiplicities: Vec<Vec<u32>>,
     /// If any, an error that occurred during the execution of the constraints, to help with debugging
     pub(crate) errors: Vec<Error>,
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -70,7 +70,7 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
         self.witness[column]
     }
 
-    /// Assert that the input is zero
+    /// Checks the constraint `tag` by checking that the input `x` is zero
     fn constrain(&mut self, tag: Constraint, x: Self::Variable) {
         if x != F::zero() {
             self.errors.push(Error::Constraint(tag));
@@ -88,14 +88,14 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
             | ByteLookup => {
                 if lookup.magnitude == Self::one() {
                     // Check that the lookup value is in the table
-                    if let Some(idx) = LookupTable::in_table(lookup.table_id, lookup.value) {
+                    if let Some(idx) = LookupTable::is_in_table(lookup.table_id, lookup.value) {
                         self.multiplicities[lookup.table_id as usize][idx] += 1;
                     } else {
                         self.errors.push(Error::Lookup(lookup.table_id));
                     }
                 }
             }
-            _ => (),
+            MemoryLookup | RegisterLookup | SyscallLookup | KeccakStepLookup => (),
         }
     }
 }
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -10,7 +10,7 @@ use crate::{
     keccak::{
         column::KeccakWitness, interpreter::KeccakInterpreter, Constraint, Error, KeccakColumn,
     },
-    lookups::{Lookup, LookupTableIDs::*},
+    lookups::{FixedLookupTables, Lookup, LookupTable, LookupTableIDs::*},
 };
 use ark_ff::Field;
 use kimchi::o1_utils::Two;
@@ -88,7 +88,11 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
             | ByteLookup => {
                 if lookup.magnitude == Self::Variable::one() {
                     // Check that the lookup value is in the table
-                    //self.multiplicities[lookup.table_id as usize][entry] += 1;
+                    if let Some(idx) = LookupTable::in_table(lookup.table_id, lookup.value) {
+                        self.multiplicities[lookup.table_id as usize][idx] += 1;
+                    } else {
+                        self.errors.push(Error::Lookup(lookup.table_id));
+                    }
                 }
             }
             _ => (),
--- a/optimism/src/keccak/witness.rs
+++ b/optimism/src/keccak/witness.rs
@@ -81,21 +81,18 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
     // LOOKUPS OPERATIONS //
     ////////////////////////
 
-    fn add_lookup(&mut self, _lookup: Lookup<Self::Variable>) {
+    fn add_lookup(&mut self, lookup: Lookup<Self::Variable>) {
         // Keep track of multiplicities for fixed lookups
-        todo!()
-    }
-
-    fn lookup_syscall_preimage(&mut self) {
-        // No-op for witness
-    }
-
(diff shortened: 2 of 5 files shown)


optimism/src/keccak/mod.rs
```diff
@@ -34,7 +33,7 @@ pub(crate) type E<F> = Expr<ConstantExpr<F>, KeccakColumn>;
 pub enum Error {
     Constraint(Constraint),
     #[allow(dead_code)]
```
dannywillems, line 35: Is it still required?

optimism/src/keccak/witness.rs
```diff
@@ -72,67 +81,21 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
...
+                if lookup.magnitude == Self::one() {
+                    // Check that the lookup value is in the table
+                    if let Some(idx) = LookupTable::in_table(lookup.table_id, lookup.value) {
+                        self.multiplicities[lookup.table_id as usize][idx] += 1;
+                    } else {
+                        self.errors.push(Error::Lookup(lookup.table_id));
+                    }
+                }
+            }
+            _ => (),
```
dannywillems, line 98: (it is a personal taste, but it has been shown to be useful in the past)
I would prefer to list all the other possible variants. This way, if one day we add a variant to the type, the compiler will force us to handle the case manually, and we will have to give a thought on the value to return.

optimism/src/keccak/witness.rs
```diff
@@ -72,67 +81,21 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
...
-        if flag == Self::one() {
-            // TODO: keep track of multiplicity of range check 16 entry
-            // TODO: check that [value] is in the RoundConstantsLookup
+        match lookup.table_id {
+            RangeCheck16Lookup | SparseLookup | ResetLookup | RoundConstantsLookup | PadLookup
+            | ByteLookup => {
+                if lookup.magnitude == Self::one() {
+                    // Check that the lookup value is in the table
+                    if let Some(idx) = LookupTable::in_table(lookup.table_id, lookup.value) {
+                        self.multiplicities[lookup.table_id as usize][idx] += 1;
```
dannywillems, line 92: Nice. Can be used to generalized later.

optimism/src/lookups.rs
```diff
@@ -71,7 +72,54 @@ pub(crate) trait FixedLookupTables<F> {
...
+impl<F: Field> FixedLookupTables<F, LookupTableIDs> for LookupTable<F> {
+    fn in_table(id: LookupTableIDs, value: Vec<F>) -> Option<usize> {
+        let table = match id {
+            LookupTableIDs::RangeCheck16Lookup => Self::table_range_check_16().entries,
+            LookupTableIDs::SparseLookup => Self::table_sparse().entries,
+            LookupTableIDs::ResetLookup => Self::table_reset().entries,
+            LookupTableIDs::RoundConstantsLookup => Self::table_round_constants().entries,
+            LookupTableIDs::PadLookup => Self::table_pad().entries,
+            LookupTableIDs::ByteLookup => Self::table_byte().entries,
+            _ => return None,
```
dannywillems, line 84: Same than [here](https://github.com/o1-labs/proof-systems/pull/1973/files#r1524471706)

optimism/src/lookups.rs
```diff
@@ -71,7 +72,54 @@ pub(crate) trait FixedLookupTables<F> {
...
+            }
+            LookupTableIDs::PadLookup => {
+                // Because this table starts with entry 1
+                if table[idx - 1] == value {
+                    Some(idx - 1)
+                } else {
+                    None
+                }
+            }
+            _ => None,
```
dannywillems, line 119: Same than [here](https://github.com/o1-labs/proof-systems/pull/1973/files#r1524471706)

optimism/src/lookups.rs
```diff
@@ -71,7 +72,54 @@ pub(crate) trait FixedLookupTables<F> {
...
+        let table = match id {
+            LookupTableIDs::RangeCheck16Lookup => Self::table_range_check_16().entries,
+            LookupTableIDs::SparseLookup => Self::table_sparse().entries,
+            LookupTableIDs::ResetLookup => Self::table_reset().entries,
+            LookupTableIDs::RoundConstantsLookup => Self::table_round_constants().entries,
+            LookupTableIDs::PadLookup => Self::table_pad().entries,
+            LookupTableIDs::ByteLookup => Self::table_byte().entries,
+            _ => return None,
+        };
+        let idx = value[0]
```
dannywillems, line 86: There is an assumption that the index is the first element (which is true). But it might be worth to keep this invariant somewhere in a type or make it clear somewhere.

optimism/src/lookups.rs
```diff
@@ -71,7 +72,54 @@ pub(crate) trait FixedLookupTables<F> {
...
+                    Some(idx - 1)
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+
     #[allow(dead_code)]
```
dannywillems, line 123: The annotation can now be removed.

optimism/src/lookups.rs
```diff
@@ -71,7 +72,54 @@ pub(crate) trait FixedLookupTables<F> {
...
+            .iter()
+            .rev()
+            .fold(0u64, |acc, &x| acc * 256 + x as u64) as usize;
+
+        match id {
+            LookupTableIDs::RangeCheck16Lookup
+            | LookupTableIDs::ResetLookup
+            | LookupTableIDs::ByteLookup
+            | LookupTableIDs::RoundConstantsLookup => {
+                if table[idx] == value {
```
dannywillems, line 97: So the table must contain also the index as a value? We are using the equality on vectors here. Just for security, in case `==` is not implemented as the equality on values on each element of the vector, I would use something like:
```
table[idx].into_iter().zip(value.into_iter().for_all(|x, y| F::eq(x, y))
```

A small test is welcome :).

optimism/src/lookups.rs
```diff
@@ -71,7 +72,54 @@ pub(crate) trait FixedLookupTables<F> {
     fn table_byte() -> LookupTable<F>;
 }
 
-impl<F: Field> FixedLookupTables<F> for LookupTable<F> {
+impl<F: Field> FixedLookupTables<F, LookupTableIDs> for LookupTable<F> {
+    fn in_table(id: LookupTableIDs, value: Vec<F>) -> Option<usize> {
```
dannywillems, line 76: I would rename it into `is_in_table`, and add a small line of documentation.

optimism/src/lookups.rs
dannywillems: I think this design can be generalized later.

optimism/src/lookups.rs
```diff
@@ -7,34 +7,34 @@ use kimchi::{
...
-    /// Syscalls communication channel
-    SyscallLookup = 2,
-    /// Input/Output of Keccak steps
-    KeccakStepLookup = 3,
-
-    // Read Tables
+    // Read tables come first to allow indexing with the table ID for the multiplicities
     /// Single-column table of all values in the range [0, 2^16)
-    RangeCheck16Lookup = 4,
+    RangeCheck16Lookup = 0,
```
dannywillems, line 19: Is it dangerous to have a table with ID 0 which contains the value 0?

optimism/src/keccak/witness.rs
```diff
@@ -10,18 +10,19 @@ use crate::{
...
 
 /// This struct contains all that needs to be kept track of during the execution of the Keccak step interpreter
 #[derive(Clone, Debug)]
 pub struct Env<Fp> {
     /// The full state of the Keccak gate (witness)
     pub witness: KeccakWitness<Fp>,
-    // The multiplicities of each lookup table
-    // TODO
+    /// The multiplicities of each lookup entry
+    pub multiplicities: Vec<Vec<u32>>,
```
dannywillems, line 25: For further generalization: using a hashmap where the key is the ID of the table. From there, we can remove the assumptions made below that the table ID is the index in this vector.

optimism/src/lookups.rs
```diff
@@ -7,34 +7,34 @@ use kimchi::{
...
-    // RAM Tables
-    MemoryLookup = 0,
-    RegisterLookup = 1,
-    /// Syscalls communication channel
-    SyscallLookup = 2,
-    /// Input/Output of Keccak steps
-    KeccakStepLookup = 3,
-
-    // Read Tables
+    // Read tables come first to allow indexing with the table ID for the multiplicities
```
dannywillems, line 17: See comment https://github.com/o1-labs/proof-systems/pull/1973/files#r1524534949

optimism/src/keccak/witness.rs
dannywillems: And, also from there, we can append any table, and make the code generic for other circuits.

optimism/src/keccak/mod.rs
querolita: Oh! Probably not anymore, I will fix it!

optimism/src/lookups.rs
querolita: Yeah, this holds for these tables in particular, but it could be untrue for others. I will add one comment. Another "more general" option would be to just look inside the whole table, but that seemed too inefficient considering the table lengths if one can know where exactly each entry should sit on.

optimism/src/lookups.rs
querolita: > So the table must contain also the index as a value?

Well, it doesn't have to, but it is convenient for these tables in particular. Otherwise you need to look into the whole table and that could be costlier. 

Agree that I should add some tests for the `in_table()` function, just to make sure that equality in vectors of field elements works as expected.

optimism/src/lookups.rs
querolita: Oh sure, forgot about that!

optimism/src/keccak/witness.rs
```diff
@@ -30,6 +31,14 @@ impl<F: Field> Default for Env<F> {
     fn default() -> Self {
         Self {
             witness: KeccakWitness::default(),
+            multiplicities: vec![
```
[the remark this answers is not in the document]
querolita: Yeah right, planning to add tests in a follow-up pr

optimism/src/lookups.rs
querolita: You mean because no matter what the joint combiner could be, this entry will always be zero? That's an interesting point! I guess a simple fix for this is just reorder the ids so that the first entry of the 0 table contains a nonzero value 👍🏻

dannywillems (approved):

PR #1974, opened 2024-03-13
MISC: add some table ID converters

dannywillems:
Part of https://github.com/o1-labs/proof-systems/pull/1966/

### DIFF
--- a/msm/src/lookups.rs
+++ b/msm/src/lookups.rs
@@ -1,7 +1,7 @@
 //! Instantiate the MVLookup protocol for the MSM project.
 
 use crate::mvlookup::{LookupTableID, MVLookup, MVLookupWitness};
-use ark_ff::{FftField, Field};
+use ark_ff::FftField;
 use kimchi::circuits::domains::EvaluationDomains;
 use rand::{seq::SliceRandom, thread_rng, Rng};
 use std::iter;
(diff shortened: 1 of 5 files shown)


msm/src/lookups.rs
```diff
@@ -14,14 +14,14 @@ pub enum LookupTableIDs {
     /// Custom lookup table
     /// The index of the table is used as the ID, padded with the number of
     /// built-in tables.
-    Custom(usize),
+    Custom(u32),
```
querolita, line 17: What's the reason that motivates you to change it to u32?

msm/src/lookups.rs
```diff
@@ -14,14 +14,14 @@ pub enum LookupTableIDs {
...
 }
 
 impl LookupTableID for LookupTableIDs {
-    fn into_field<F: Field>(self) -> F {
+    fn to_u32(&self) -> u32 {
         match self {
-            LookupTableIDs::RangeCheck16 => F::one(),
-            LookupTableIDs::Custom(id) => F::from(id as u64) + F::one(),
+            LookupTableIDs::RangeCheck16 => 1_u32,
```
querolita, line 23: I suppose this is equivalent to `1u32`?

querolita (approved):

msm/src/lookups.rs
dannywillems: Yes

msm/src/lookups.rs
dannywillems: To use `Fp::from(id)` :)

dannywillems: And I think u32 is way enough :)

PR #1976, opened 2024-03-13
MSM: add rangecheck4 and rangecheck15 lookups in constraints

dannywillems:
Cherry-pick from https://github.com/o1-labs/proof-systems/pull/1966

### DIFF
--- a/msm/src/serialization/constraints.rs
+++ b/msm/src/serialization/constraints.rs
@@ -6,7 +6,8 @@ use kimchi::circuits::{
 
 use crate::{columns::Column, expr::E, serialization::N_INTERMEDIATE_LIMBS, N_LIMBS};
 
-use super::interpreter::InterpreterEnv;
+use super::Lookup;
+use super::{interpreter::InterpreterEnv, LookupTable};
 
 pub struct Env<Fp> {
     /// An indexed set of constraints.
@@ -15,6 +16,8 @@ pub struct Env<Fp> {
     /// folding for instance.
     pub constraints: Vec<(usize, Expr<ConstantExpr<Fp>, Column>)>,
     pub constrain_index: usize,
+    pub rangecheck4_lookups: Vec<Lookup<E<Fp>>>,
+    pub rangecheck15_lookups: Vec<Lookup<E<Fp>>>,
 }
 
 impl<Fp: PrimeField> Env<Fp> {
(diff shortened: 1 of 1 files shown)


msm/src/serialization/constraints.rs
```diff
@@ -15,13 +16,17 @@ pub struct Env<Fp> {
     /// folding for instance.
     pub constraints: Vec<(usize, Expr<ConstantExpr<Fp>, Column>)>,
     pub constrain_index: usize,
+    pub rangecheck4_lookups: Vec<Lookup<E<Fp>>>,
```
querolita, line 19: I wonder if it is really a good idea to have the lookups split into different variable names. Looks more general to have just a single lookups vector, and then all of them get pushed into it.

msm/src/serialization/constraints.rs
```diff
@@ -63,12 +68,22 @@ impl<F: PrimeField> InterpreterEnv<F> for Env<F> {
         Column::X(3 + j)
     }
 
-    fn range_check15(&mut self, _value: &Self::Variable) {
-        // TODO
+    fn range_check15(&mut self, value: &Self::Variable) {
+        let one = ConstantExpr::from(ConstantTerm::Literal(F::one()));
+        self.rangecheck15_lookups.push(Lookup {
+            table_id: LookupTable::RangeCheck15,
+            numerator: Expr::Atom(ExprInner::Constant(one)),
```
querolita, line 75: I must assume you only have "static" range checks? Like, they are always added, regardless of the values of other conditionals?

msm/src/serialization/constraints.rs
querolita: Yeah, probably we should think of a way to generalize them to just have one. And then if you really need the distinction of different tables, think of a different way to split them but "inside" the `lookup` type/vector/struct/whatever. Maybe for a follow up.

querolita (approved):

msm/src/serialization/constraints.rs
dannywillems: Yes

msm/src/serialization/constraints.rs
dannywillems: Yes, agreed. This is ugly atm, and we must generalize. Let's do it quickly after everything works.

PR #1979, opened 2024-03-13
Update poly-commitment.md

skaunov:
A typo correction


querolita:
That name was intentional. Bulletproof would refer just to https://eprint.iacr.org/2017/1066 but with "bootle-proof" we're referring to a wider type of proofs derived from https://eprint.iacr.org/2016/263, among which one could sit the Bulletproofs paper.

skaunov:
That's fun! I guess you don't want to have such a note/clarification there... 
I wouldn't know without you, thanks!

PR #1980, opened 2024-03-14
Logup: prepare the column environment to add the lookup

dannywillems:
No-op. Only reorganizing the method `get_column` to handle case by case the lookup columns.

### DIFF
--- a/msm/src/column_env.rs
+++ b/msm/src/column_env.rs
@@ -25,8 +25,9 @@ pub struct ColumnEnvironment<'a, F: FftField> {
     /// The domains used in the PLONK argument.
     pub domain: EvaluationDomains<F>,
     /// Lookup specific polynomials
-    // TODO define MVlookup one
     pub lookup: Option<()>,
+    // TODO: rename in additive lookup or "logup"
+    // We do not use multi-variate lookups, only the additive part
 }
 
 impl<'a, F: FftField> TColumnEnvironment<'a, F> for ColumnEnvironment<'a, F> {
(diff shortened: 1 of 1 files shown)


msm/src/column_env.rs
```diff
@@ -25,8 +25,9 @@ pub struct ColumnEnvironment<'a, F: FftField> {
     /// The domains used in the PLONK argument.
     pub domain: EvaluationDomains<F>,
     /// Lookup specific polynomials
-    // TODO define MVlookup one
     pub lookup: Option<()>,
+    // TODO: rename in additive lookup or "logup"
+    // We do not use multi-variate lookups, only the additive part
```
querolita, line 30: fair point

msm/src/column_env.rs
```diff
@@ -36,19 +37,28 @@ impl<'a, F: FftField> TColumnEnvironment<'a, F> for ColumnEnvironment<'a, F> {
...
+        let coefficients_length: usize = self.coefficients.len();
+        match *col {
+            // Handling the "relation columns"
+            Self::Column::X(i) => {
+                if i < witness_length {
+                    let res = &self.witness[i];
+                    Some(res)
+                } else if i < witness_length + coefficients_length {
+                    // FIXME and add a test
+                    Some(&self.coefficients[i])
```
querolita, line 50: Not sure but, wouldn't this be `self.coefficients[i-witness_length]` then?

dannywillems: Yes... It must be fixed. Even though it is obvious to change the index, I want to fix it later. it is not used atm.

dannywillems: I simply renamed var and reorg the code. The logic is the same than before.

querolita: okay, let's not forget then

querolita (approved):
approving, but let's keep in mind the index mismatch to be fixed later

msm/src/column_env.rs
dannywillems: Open https://github.com/o1-labs/proof-systems/issues/1988

PR #1981, opened 2024-03-14
Define constant for the max supported degree

dannywillems:
It will be used to define the quotient polynomial and split the looked-up values


dannywillems:
Note that in a following PR, it is merged in lib.rs. Sorry for the too simple commit.

PR #1982, opened 2024-03-14
Logup: merge looked-up values with fixed table

dannywillems:


msm/src/mvlookup.rs
```diff
@@ -206,15 +209,17 @@ pub mod prover {
...
+            // Contain the evalations of the h_i. We divide the looked-up values
+            // in chunks of (MAX_SUPPORTED_DEGREE - 2)
             let lookup_terms_evals: Vec<Vec<G::ScalarField>> = lookups
                 .into_iter()
                 .map(|lookup| {
-                    let MVLookupWitness { f, t, m: _ } = lookup;
+                    let MVLookupWitness { f, m: _ } = lookup;
+                    // The number of functions to look up, including the fixed table.
                     let n = f.len();
                     // We compute first the denominators of all f_i and t. We gather them in
```
querolita, line 220: When you say `t` like only one, you mean if you only have one table no? But I thought you had one for rangecheck4 and another one for rangecheck15

dannywillems: It will be a different `LookupWitness` structure.

PR #1990, opened 2024-03-17
Move generic_column into ColumnEnvironment

dannywillems:


volhovm (approved):
I'm not 100% convinced this is what `kimchi` wanted (because there the column is non-generic, so it /does/ have a degree), however the design was confusing, and it's less confusing now. And `ColumnEnvironment` needs to be implemented for a non-generic `kimchi` column anyway... cool!

PR #1995, opened 2024-03-17
Generic prover/verifier: implement tests for degree 1, 2 and 3

dannywillems:


msm/src/prover.rs
```diff
@@ -118,6 +118,18 @@ where
...
 
+    let max_degree = {
+        if lookup_env.is_none() {
+            constraints
+                .iter()
+                .map(|expr| expr.degree(1, 0))
+                .max()
+                .unwrap_or(0)
+        } else {
+            8
```
dannywillems, line 129: This will come later.

msm/src/prover.rs
```diff
@@ -236,20 +238,14 @@ where
...
-        for _ in 0..dummies_n {
-            t_comm.elems.push(G::zero());
-        }
-        t_comm
+        let num_chunks: usize = if max_degree == 1 {
+            1
+        } else {
+            (max_degree - 1) as usize
+        };
+        srs.commit_non_hiding(&quotient_poly, num_chunks)
```
volhovm, line 248: I'm not sure I understand chunking enough to say if it's a correct way to implement things... The code above with `dummies_n` was copied from mips demo, and it handled mvlookup somehow? Maybe not.

It feels reasonable, but I want to understand this better ideally.

msm/src/test/mod.rs
```diff
@@ -2,3 +2,228 @@ pub mod columns;
...
+            witness.clone(),
+            domain_size,
+            &mut rng,
+        );
+        // FIXME: we would want to allow the prover to make a proof, but the verification must fail.
+        // TODO: Refactorize code in prover to handle a degug or add an adversarial prover.
+        // test_soundness_generic(constraints, witness, domain_size, &mut rng);
+    }
+
+    // Test degree a degree one constraint:
```
volhovm, line 185: "three"?

msm/src/test/mod.rs
```diff
@@ -2,3 +2,228 @@ pub mod columns;
...
+    //   X_{0} * X_{0} * X_{0} \
+    // - 42 * X_{1} * X_{2} \
+    // + X_{3}
+    #[test]
+    fn test_completeness_degree_three() {
```
volhovm, line 190: I think all these tests for different constant degree (2,3,6,7) have a lot of commonly shared boilerplate and can be probably made much more concise later on

volhovm (approved):
Approving, but let's discuss the chunking for t part later.

msm/src/test/mod.rs
dannywillems: Agreed. It is planned when we can use the generic interpreter, see comment.

msm/src/prover.rs
dannywillems: Constraints for logup will be added in a following PR.

msm/src/test/mod.rs
dannywillems: FIxed.

msm/src/prover.rs
dannywillems: The only issue I see at the moment is constraints of degree 1.

querolita: What I don't know here is if it should be non hiding or not. Maybe are you using non-hiding because they don't want zk?

msm/src/test/mod.rs
```diff
@@ -2,3 +2,228 @@ pub mod columns;
...
+        };
+
+        test_completeness_generic::<N, _>(
+            constraints.clone(),
+            witness.clone(),
+            domain_size,
+            &mut rng,
+        );
+        // FIXME: we would want to allow the prover to make a proof, but the verification must fail.
+        // TODO: Refactorize code in prover to handle a degug or add an adversarial prover.
```
querolita, line 145: degug?

querolita (approved):
Please read concern

msm/src/test/mod.rs
dannywillems: Oopsie, typo, debug*.

PR #2000, opened 2024-03-17
Environment/Lookup: evaluations on d1 are not necessary

dannywillems:
We always use d8 (as we always suppose degree 7 for the lookup constraints).


querolita (commented):
Aren't they used to compute each poly_d1?

dannywillems:
> Aren't they used to compute each poly_d1?

They are not used to compute the quotient polynomial, therefore not required in the lookup environment.

PR #2007, opened 2024-03-18
Generic prover: use correct number of chunks for Z_H(X) * t(X)

dannywillems:
Only one chunk was computed, and we were padding the others with zero.


querolita (approved):
Oh great catch

volhovm (approved):
Why didn't completeness test catch it? :thinking:

PR #2010, opened 2024-03-18
Utils: ensuring that the requested number of chunks is returned

dannywillems:
And fixing tests accordingly.

### DIFF
--- a/poly-commitment/src/evaluation_proof.rs
+++ b/poly-commitment/src/evaluation_proof.rs
@@ -119,7 +119,11 @@ pub fn combine_polys<G: CommitmentCurve, D: EvaluationDomain<G::ScalarField>>(
     if !plnm_evals_part.is_empty() {
         let n = plnm_evals_part.len();
         let max_poly_size = srs_length;
-        let num_chunks = n / max_poly_size;
+        let num_chunks = if n % max_poly_size == 0 {
+            n / max_poly_size
+        } else {
+            n / max_poly_size + 1
+        };
         plnm += &Evaluations::from_vec_and_domain(plnm_evals_part, D::new(n).unwrap())
             .interpolate()
             .to_chunked_polynomial(num_chunks, max_poly_size)
--- a/poly-commitment/src/tests/batch_15_wires.rs
+++ b/poly-commitment/src/tests/batch_15_wires.rs
@@ -30,8 +30,6 @@ where
     let size = 1 << 7;
     let srs = SRS::<Vesta>::create(size);
 
-    let num_chunks = 1;
-
     let group_map = <Vesta as CommitmentCurve>::Map::setup();
 
     let sponge = DefaultFqSponge::<VestaParameters, SC>::new(
@@ -83,10 +81,19 @@ where
             let mut start = Instant::now();
             let comm = (0..a.len())
                 .map(|i| {
+                    let n = a[i].len();
+                    let num_chunks = if n % srs.g.len() == 0 {
+                        n / size
+                    } else {
+                        n / size + 1
+                    };
                     (
                         srs.commit(&a[i].clone(), num_chunks, rng),
                         x.iter()
-                            .map(|xx| a[i].to_chunked_polynomial(1, size).evaluate_chunks(*xx))
+                            .map(|xx| {
+                                a[i].to_chunked_polynomial(num_chunks, size)
+                                    .evaluate_chunks(*xx)
+                            })
                             .collect::<Vec<_>>(),
                         bounds[i],
                     )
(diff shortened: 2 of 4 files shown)


poly-commitment/src/evaluation_proof.rs
```diff
@@ -119,7 +119,11 @@ pub fn combine_polys<G: CommitmentCurve, D: EvaluationDomain<G::ScalarField>>(
     if !plnm_evals_part.is_empty() {
         let n = plnm_evals_part.len();
         let max_poly_size = srs_length;
-        let num_chunks = n / max_poly_size;
+        let num_chunks = if n % max_poly_size == 0 {
```
emberian, line 122: Maybe `n.div_ceil(max_poly_size)` is equivalent?

emberian: Nevermind, this API is still unstable ...

emberian: Ah, stable for unsigned integers though!

dannywillems: Nice. Didn't know. Updating.

dannywillems: Yes, unstable unfortunately. Not working for usize :(. Good to remember though.

poly-commitment/src/tests/batch_15_wires.rs
```diff
@@ -83,10 +81,19 @@ where
             let mut start = Instant::now();
             let comm = (0..a.len())
                 .map(|i| {
+                    let n = a[i].len();
+                    let num_chunks = if n % srs.g.len() == 0 {
```
volhovm, line 85: @dannywillems could you please give a bit of context on what's happening with the tests? Primarily, were they committing to less chunks than necessary, and if yes, does it mean that the tests were not sound -- they were passing but not doing what was originally expected? 

In this particular case, it seems that `a[i].len()` is always `1`, isn't it?..

poly-commitment/src/tests/commitment.rs
```diff
@@ -150,8 +150,14 @@ fn test_randomised<RNG: Rng + CryptoRng>(mut rng: &mut RNG) {
...
             for point in eval_points.clone() {
+                let n = poly.len();
+                let num_chunks = if n % srs.g.len() == 0 {
+                    n / srs.g.len()
+                } else {
+                    n / srs.g.len() + 1
+                };
                 chunked_evals.push(
-                    poly.to_chunked_polynomial(1, srs.g.len())
+                    poly.to_chunked_polynomial(num_chunks, srs.g.len())
```
volhovm, line 160: If I understand this correctly, `num_chunks` is always `1` here... So you just generalised it?

utils/src/dense_polynomial.rs
```diff
@@ -57,6 +57,10 @@ impl<F: Field> ExtendedDensePolynomial<F> for DensePolynomial<F> {
             chunk_polys.push(DensePolynomial::from_coefficients_vec(vec![]));
         }
 
+        // Ensuring that the number of chunks is the one requested, following
+        // trait documentation
+        assert_eq!(chunk_polys.len(), num_chunks);
```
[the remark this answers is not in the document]
dannywillems: No, it is not. See the commit message content.

poly-commitment/src/tests/batch_15_wires.rs
dannywillems: [This](https://github.com/o1-labs/proof-systems/pull/2010#discussion_r1529949745) should answer your question

poly-commitment/src/tests/commitment.rs
dannywillems: See https://github.com/o1-labs/proof-systems/pull/2010#discussion_r1529950740

querolita (approved):
Looks safer to me

utils/src/dense_polynomial.rs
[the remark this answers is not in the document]
volhovm: On line 51 you push some `N` elements, and then on line 57 you push `num_chunks - N` elements, so in the end you will have `N` elements anyway. I don't see how it's related to line 50 with vector initiation, you could've had `vec![]` there and it would've worked semantically in the same way. Am I missing smth?

poly-commitment/src/tests/batch_15_wires.rs
volhovm: That seems to be very orthogonal, my question here was about the test:

The length of the vector in `a` is random up to `s = length_i`. Elements in `length_i` are up to `500`. So unless the SRS is less than `2^8` we will need to create one chunk, and so this code would behave exactly as before. But the SRS is `2^7` above, so was the test actually /failing/ if the SRS was smaller? We created some commitments, they are not correct, but does the test check it after construction-time later at all?

PR #2011, opened 2024-03-19
Improve the negative padding test of Keccak

querolita:
Danny got a failing assert in one of the Keccak tests yesterday. I found out that with that seed, the byte position that I was trying to nullify in the negative test was already 0, so instead of 2 errors, only 1 should occur. 

This type of "problem" could arise in other negative tests. I would need to spend more time looking into them to know if others need some special treatment as well. For now I am fixing this one.

### DIFF
--- a/optimism/src/keccak/tests.rs
+++ b/optimism/src/keccak/tests.rs
@@ -199,15 +199,37 @@ fn test_keccak_fake_witness_wont_satisfy_constraints() {
     // NEGATIVIZE THE WITNESS
 
     // Break padding constraints
-    witness_env[0].witness[KeccakColumn::PadBytesFlags(0)] = Fp::from(1u32);
-    witness_env[0].constrain_padding();
+    assert_eq!(witness_env[0].is_pad(), Fp::one());
+    // Padding can only occur in suffix[3] and suffix[4] because length is 100 bytes
+    assert_eq!(witness_env[0].pad_suffix(0), Fp::zero());
+    assert_eq!(witness_env[0].pad_suffix(1), Fp::zero());
+    assert_eq!(witness_env[0].pad_suffix(2), Fp::zero());
+    // Check that the padding blocks are corrrect
+    assert_eq!(witness_env[0].block_in_padding(0), Fp::zero());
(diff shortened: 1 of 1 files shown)


optimism/src/keccak/tests.rs
```diff
@@ -199,15 +199,37 @@ fn test_keccak_fake_witness_wont_satisfy_constraints() {
...
-        ]
+        witness_env[0].witness[KeccakColumn::PadBytesFlags(0)],
+        Fp::zero()
     );
+    witness_env[0].witness[KeccakColumn::PadBytesFlags(0)] = Fp::from(1u32);
+    // Now that PadBytesFlags(0) is 1, then block_in_padding(0) should be 0b10*
+    println!("{:?}", witness_env[0].block_in_padding(0).to_hex());
+    witness_env[0].constrain_padding();
+    // When the byte(0) is different than 0 then the padding suffix constraint also fails
+    if witness_env[0].sponge_bytes()[0] != Fp::zero() {
```
volhovm, line 221: I assume that is the main change, that in the case when it's actually /already/ zero, we expect only `PadAtEnd`.

optimism/src/keccak/tests.rs
```diff
@@ -199,15 +199,37 @@ fn test_keccak_fake_witness_wont_satisfy_constraints() {
     // NEGATIVIZE THE WITNESS
 
     // Break padding constraints
-    witness_env[0].witness[KeccakColumn::PadBytesFlags(0)] = Fp::from(1u32);
-    witness_env[0].constrain_padding();
+    assert_eq!(witness_env[0].is_pad(), Fp::one());
+    // Padding can only occur in suffix[3] and suffix[4] because length is 100 bytes
+    assert_eq!(witness_env[0].pad_suffix(0), Fp::zero());
```
volhovm, line 204: And these assertions are not directly related to the issue, just some extra checks?

volhovm (approved):
Looks consistent to me, but not merging it if you want another pair of eyes on this. I don't understand the logic of this very well.

optimism/src/keccak/tests.rs
querolita: correct

PR #2016, opened 2024-03-20
MSM: Store Boxed array in Witness

volhovm:
Problem: When dealing with long columns (200+), msm/prover.rs was failing with stack overflow while creating the first witness evaluations. The tests were passing if stack size was increased via an environment variable (`RUST_MIN_STACK` it was I think).

Solution: Insead of storing a potentially long array directly in the witness, storing just a reference to it (Box) solves the problem.

### DIFF
--- a/msm/src/ffa/witness.rs
+++ b/msm/src/ffa/witness.rs
@@ -29,7 +29,7 @@ impl<F: PrimeField> FFAInterpreterEnv<F> for WitnessBuilderEnv<F> {
     fn empty() -> Self {
         WitnessBuilderEnv {
             witness: vec![Witness {
-                cols: [Zero::zero(); FFA_N_COLUMNS],
+                cols: Box::new([Zero::zero(); FFA_N_COLUMNS]),
             }],
         }
     }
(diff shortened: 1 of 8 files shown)


querolita (approved):
Oh! This was indeed a problem when I was testing the first version of Keccak inside Kimchi and bypassed it by increasing the stack size indeed. I am so glad these boxes solve the problem that easily!

volhovm:
It solves a concrete problem for me, but maybe there are other places that we need to optimise similarly too...

volhovm:
In particular, my code was throwing stack overflow at the moment the first `into_par_iter()` was computed on `Witness` values. This code:
```
    let witness_evals: Witness<N, Evaluations<G::ScalarField, R2D<G::ScalarField>>> = inputs
        .evaluations
        .into_par_iter()
        .map(|evals| {
            Evaluations::<G::ScalarField, R2D<G::ScalarField>>::from_vec_and_domain(
                evals, domain.d1,
            )
        })
        .collect::<Witness<N, Evaluations<G::ScalarField, R2D<G::ScalarField>>>>();
```

RN we repack witness into a vector, which looks inefficient (didn't immediately find how to optimise it since the parallel iterator instance is not available for arrays? maybe I wasn't looking right), but I'm not sure that's the actual culprit.

Issue #1843, opened 2024-02-26
Implement folding traits for optimism crate


querolita:

querolita:
I increased from 2 to 3 because I learnt that the conversions between kimchi expressions and folding expressions was undone.

PR #2019, opened 2024-03-20
Map Kimchi expressions to folding expressions

querolita:
I recently learnt that the usual expressions that we have been using throughout the zkvm project to define our variables and constraints are not directly compatible with the ones expected by the folding library. For that reason, as part of integration, I added `From` implementations for the expressions in the Expr framework that should work in particular for the generalized expressions over columns (so it is compatible with Keccak columns, MIPS columns, Test columns as in the test provided, and other column definitions we could have in MSM). 

I created a "silly" test using "toy" column types to make sure that it works. In particular, for this to work, one needs to implement a `From` between your project's challenges (the ones for folding) and the ones in Kimchi, and make sure that no alphas are being included in the constraints. That's because folding will be the one that includes them for the linearization. 

Closes https://github.com/o1-labs/proof-systems/issues/1843


kimchi/src/folding/expressions.rs
```diff
@@ -500,3 +502,67 @@ pub fn folding_expression<C: FoldingConfig>(
...
+    <Config as FoldingConfig>::Challenge: From<ChallengeTerm>,
+{
+    fn from(expr: ConstantExprInner<F>) -> Self {
+        match expr {
+            ConstantExprInner::Challenge(chal) => {
+                FoldingCompatibleExprInner::Challenge(chal.into())
+            }
+            ConstantExprInner::Constant(c) => match c {
+                ConstantTerm::Literal(f) => FoldingCompatibleExprInner::Constant(f),
+                _ => panic!("ConstantExprInner not supported in folding expressions"),
```
fabrizio-m, line 518: it isn't completely unsupported, you can replace the other variants by the field element they represent and use `FoldingCompatibleExprInner::Constant(f)` for all. Maybe we can add an associated type to the config to allow more complex constant in expressions, but you would have to then resolve it in the environment, and I think it will end being even more work.

fabrizio-m (approved):

kimchi/src/folding/expressions.rs
querolita: Oooh I see, I will work a bit on it and if it ends up being a small change I will add one commit for it. Otherwise, open a follow up pr.

PR #2023, opened 2024-03-21
Logup: adding a method to identify fixed and RAMLookup

dannywillems:


querolita (approved):
Looks like a good idea for now, it will integrate easily with the hash maps for multiplicities etc. Awesome!

dannywillems:
> Looks like a good idea for now, it will integrate easily with the hash maps for multiplicities etc. Awesome!

I'm exactly on this!

PR #2024, opened 2024-03-21
Small changes - comments/remove unused code.

dannywillems:


optimism/src/keccak/column.rs
```diff
@@ -126,11 +126,11 @@ impl FoldingColumnTrait for Column {
 ///   155..2123 -> curr
 ///         155                                                                        2119
 ///          <--------------------------------if_round<---------------------------------->
-///          <-------------if_sponge------------->
```
volhovm, line 129: @querolita Not sure which editor you're using, but it'd help to make it automatic...

https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

volhovm (approved):

optimism/src/keccak/column.rs
volhovm: (sorry if this was already discussed somewhere else, I realise this change already got into master before?)

dannywillems:
@volhovm requested a new approval, I pushed on the same branch :grimacing:

volhovm (approved):
Sometimes I can push after approval and it doesn't erase the approval... sometimes it doesn't. It's a bit of a mystery...

optimism/src/keccak/column.rs
querolita: I am using VSCode, so that link should help!

querolita: > did this change get into master before?)

This line was merged in this PR https://github.com/o1-labs/proof-systems/pull/1963, Misha approved so that's why you might not recall where it came from

querolita (approved):

volhovm:
@querolita Now looking back at that PR I'm actually surprised github UI does not highlight trailing whitespaces :thinking: You literally don't see them on the "Files changed" panel. Weird.

PR #2028, opened 2024-03-21
Logup: add table ID as a parameter for proofs

dannywillems:
I want to enforce at the type level that the proof is only for a set of tables.
Also, coming in a following PR, we use a hashmap for the multiplicities and the fixed lookup tables, as discussed in previous PRs.

### DIFF
--- a/msm/src/proof.rs
+++ b/msm/src/proof.rs
@@ -49,17 +49,23 @@ impl<const N: usize, G: KimchiCurve> ProofInputs<N, G, LookupTableIDs> {
 }
 
 #[derive(Debug, Clone)]
-pub struct ProofEvaluations<const N: usize, F> {
-    /// Witness evaluations, including public input
+pub struct ProofEvaluations<const N: usize, F, ID: LookupTableID> {
+    /// Witness evaluations, including public inputs
     pub(crate) witness_evals: Witness<N, PointEvaluations<F>>,
     /// MVLookup argument evaluations
-    pub(crate) mvlookup_evals: Option<LookupProof<PointEvaluations<F>>>,
-    /// Evaluation of Z_H(xi) (t_0(X) + xi^n t_1(X) + ...) at xi.
+    pub(crate) mvlookup_evals: Option<LookupProof<PointEvaluations<F>, ID>>,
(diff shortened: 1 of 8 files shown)


msm/src/proof.rs
```diff
@@ -42,17 +42,23 @@ impl<const N: usize, G: KimchiCurve> ProofInputs<N, G, LookupTableIDs> {
...
+    pub(crate) mvlookup_evals: Option<LookupProof<PointEvaluations<F>, ID>>,
     /// Evaluation of Z_H(ζ) (t_0(X) + ζ^n t_1(X) + ...) at ζω.
     pub(crate) ft_eval1: F,
 }
 
-impl<const N: usize, F: Clone> ColumnEvaluations<F> for ProofEvaluations<N, F> {
+/// The trait ColumnEvaluations is used by the verifier.
+/// It will return the evaluation of the corresponding column at the
+/// evaluation points coined by the verifier during the protocol.
+impl<const N: usize, F: Clone, ID: LookupTableID> ColumnEvaluations<F>
```
volhovm, line 57: I'd go for `LID` or something more concrete

msm/src/proof.rs
```diff
@@ -66,21 +72,21 @@ impl<const N: usize, F: Clone> ColumnEvaluations<F> for ProofEvaluations<N, F> {
...
     /// Commitments to the quotient polynomial.
     /// The value contains the chunked polynomials.
     pub(crate) t_comm: PolyComm<G>,
 }
 
 #[derive(Debug, Clone)]
-pub struct Proof<const N: usize, G: KimchiCurve, OpeningProof: OpenProof<G>> {
-    pub(crate) proof_comms: ProofCommitments<N, G>,
-    pub(crate) proof_evals: ProofEvaluations<N, G::ScalarField>,
+pub struct Proof<const N: usize, G: KimchiCurve, OpeningProof: OpenProof<G>, ID: LookupTableID> {
```
volhovm, line 88: This all starts to become quite heavy. I'm thinking we could do a trait `ProofParams` with a bunch of associated types instead, and just pass that around. Similar to how arkworks does for a `Circuit` if I remember correctly?

volhovm (approved):

PR #2031, opened 2024-03-22
MSM/Proof: implement ColumnEvaluation for lookups

dannywillems:
Still missing fixed lookup tables


msm/src/proof.rs
```diff
@@ -60,14 +60,44 @@ impl<const N: usize, F: Clone, ID: LookupTableID> ColumnEvaluations<F>
     type Column = crate::columns::Column;
 
     fn evaluate(&self, col: Self::Column) -> Result<PointEvaluations<F>, ExprError<Self::Column>> {
-        let crate::columns::Column::X(i) = col else {
-            todo!()
+        let res = match col {
+            Self::Column::X(i) => {
```
volhovm, line 64: I think it'd be nice to have a separate `struct WitnessColumn(usize)` for many purposes where we don't actually care about lookup columns (like writing witness values in the interpreter), and this `X` then should be definitely renamed into something ... meaningful. What do you think?

PR #2033, opened 2024-03-22
Continue integrating generic additive lookup

dannywillems:
Built on top of https://github.com/o1-labs/proof-systems/pull/2065

- [x] The fixed tables must also be committed and absorbed.
- [x] column env for lookup must include the fixed lookup tables
- [x] Add fixed lookup tables in evals and polys for the opening proof.
- [x] Add lookup constraints automatically
- [ ] Include lookup constraints in the argument
- [x] multiplicities must be a hashmap.
- [x] use btreemap in serialization circuit
- [ ] make serialization circuit works with lookup constraints


msm/src/mvlookup.rs
```diff
@@ -215,6 +220,46 @@ pub fn combine_lookups<F: Field, ID: LookupTableID>(
...
+pub fn constraint_lookups<F: Field, ID: LookupTableID>(
+    lookups_map: &HashMap<ID, Vec<MVLookup<E<F>, ID>>>,
+) -> Vec<E<F>> {
+    let mut constraints: Vec<E<F>> = vec![];
+    let mut idx_partial_sum = 0;
+    lookups_map.iter().for_each(|(id, lookups)| {
+        let table_lookup = MVLookup {
+            table_id: *id,
+            numerator: -curr_cell(Column::LookupMultiplicity(id.to_u32())),
+            value: vec![curr_cell(Column::LookupFixedTable(id.to_u32()))],
```
querolita, line 229: probably, this will only have an actual value once the joint combiner is chosen right?

msm/src/mvlookup.rs
```diff
@@ -215,6 +220,46 @@ pub fn combine_lookups<F: Field, ID: LookupTableID>(
...
+            ));
+            idx_partial_sum += 1;
+        });
+    });
+
+    // Generic code over the partial sum
+    // Compute φ(ωX) - φ(X) - \sum_{i = 1}^{N} h_i(X)
+    {
+        let constraint =
+            next_cell(Column::LookupAggregation) - curr_cell(Column::LookupAggregation);
```
querolita, line 248: how does `next_cell` work when it's the last row?

msm/src/serialization/constraints.rs
```diff
@@ -16,17 +17,15 @@ pub struct Env<Fp> {
...
 }
 
 impl<Fp: PrimeField> Env<Fp> {
     pub fn create() -> Self {
         Self {
             constraints: vec![],
             constrain_index: 0,
-            rangecheck4_lookups: vec![],
-            rangecheck15_lookups: vec![],
+            lookups: HashMap::new(),
```
querolita: nicer

msm/src/serialization/constraints.rs
```diff
@@ -43,13 +43,12 @@ impl<F: PrimeField> InterpreterEnv<F> for Env<F> {
...
 
-    fn copy(&mut self, x: &Self::Variable, position: Self::Position) -> Self::Variable {
-        let y = Expr::Atom(ExprInner::Cell(Variable {
+    fn copy(&mut self, _x: &Self::Variable, position: Self::Position) -> Self::Variable {
+        // No-op in constraints, witness only
+        Expr::Atom(ExprInner::Cell(Variable {
             col: position,
             row: CurrOrNext::Curr,
-        }));
-        self.add_constraint(y.clone() - x.clone());
```
querolita, line 125: why is this constraint being removed?

msm/src/serialization/constraints.rs
```diff
@@ -125,3 +125,25 @@ impl<F: PrimeField> InterpreterEnv<F> for Env<F> {
...
+            .iter()
+            .map(|(_, cst)| cst.clone())
+            .collect();
+        constraints.extend(relation_constraints);
+
+        assert!(self.lookups[&LookupTable::RangeCheck15].len() == 17);
+        assert!(self.lookups[&LookupTable::RangeCheck4].len() == 20);
+
+        let _lookup_constraint = constraint_lookups(&self.lookups);
+        // FIXME: LookupMultiplicity must still be correctly implemented in column_env.
```
querolita, line 139: I wonder if the same idea in zkvm could be useful for you here to compute the multiplicities in the witness side of the env.

msm/src/mvlookup.rs
```diff
@@ -113,13 +113,13 @@ pub struct MVLookupWitness<F, ID: LookupTableID> {
 #[derive(Debug, Clone)]
 pub struct LookupProof<T, ID> {
     /// The multiplicity polynomials
-    pub(crate) m: Vec<T>,
+    pub(crate) m: BTreeMap<ID, T>,
```
querolita, line 116: I wonder if having bit tree maps instead of hash maps means we no longer need the `from_u32` function for table ids?

msm/src/serialization/constraints.rs
```diff
@@ -130,6 +112,16 @@ impl<F: PrimeField> InterpreterEnv<F> for Env<F> {
 }
 
 impl<F: PrimeField> Env<F> {
+    fn add_lookup(&mut self, table_id: LookupTable, value: &E<F>) {
```
querolita, line 115: Lovely.

querolita (commented):
Leaving a bunch of comments for now. It is looking great to me so far.

msm/src/serialization/constraints.rs
dannywillems: Fixed. I removed it for testing. Good catch.

msm/src/serialization/constraints.rs
dannywillems: Yes, we should merge how the multiplicities are computed.

msm/src/column_env.rs
```diff
@@ -71,9 +71,9 @@ impl<'a, const N: usize, F: FftField, ID: LookupTableID> TColumnEnvironment<'a,
                     panic!("No lookup provided")
                 }
             }
-            Self::Column::LookupMultiplicity(i) => {
+            Self::Column::LookupMultiplicity(id) => {
                 if let Some(ref lookup) = self.lookup {
-                    Some(&lookup.lookup_counters_evals_d8[i as usize])
+                    Some(&lookup.lookup_counters_evals_d8[&ID::from_u32(id)])
```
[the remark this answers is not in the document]
dannywillems: In the witness

dannywillems:
I set this as ready. However, there is a bunch of FIXME that really needed to be fixed.
- constraints are not satisfied for the partial sums. Still working on it, I have a branch called dw/multiplicities-debug.
- still some designs I'd like to change.
- still needs to handle the inputs of the serialization circuit using lookup, and remove the FIXME related to the generation of the witness.

Making it ready so we can move forward.

msm/src/mvlookup.rs
querolita: I meant btree maps, just autocorrect typo

msm/src/mvlookup.rs
```diff
@@ -121,15 +130,154 @@ impl<'lt, G, ID: LookupTableID> IntoIterator for &'lt LookupProof<G, ID> {
...
+    lookups_map.iter().for_each(|(id, lookups)| {
+        let table_lookup = MVLookup {
+            table_id: *id,
+            numerator: curr_cell(Column::LookupMultiplicity(id.to_u32())),
+            value: vec![curr_cell(Column::LookupFixedTable(id.to_u32()))],
+        };
+        // FIXME: do not clone
+        let mut lookups = lookups.clone();
+        lookups.push(table_lookup);
+        // We split in chunks of 6 (MAX_SUPPORTED_DEGREE - 2)
```
querolita, line 258: I believe Matthew told me something about the inverses and the fact that we might not want to split in chunks. Did he probably refer just to the implementation in folding?

msm/src/mvlookup.rs
```diff
@@ -260,20 +436,31 @@ pub mod prover {
...
-                                }) * vector_lookup_combiner
-                                    + table_id.to_field::<G::ScalarField>();
+                                value.iter().rev().fold(G::ScalarField::zero(), |acc, y| {
+                                    acc * vector_lookup_combiner + y
+                                }) * vector_lookup_combiner;
+                            // add table id
+                            let combined_value =
+                                combined_value + table_id.to_field::<G::ScalarField>();
+
+                            // If last element and fixed lookup tables, we keep
```
querolita, line 454: I don't think I get this step

msm/src/mvlookup.rs
```diff
@@ -318,6 +508,11 @@ pub mod prover {
                 .iter()
                 .for_each(|evals| assert_eq!(evals.len(), domain.d1.size as usize));
 
+            // Sanity check to verify that we have all the evaluations for the fixed lookup tables
+            fixed_lookup_tables
+                .values()
+                .for_each(|evals| assert_eq!(evals.len(), domain.d1.size as usize));
```
querolita, line 514: I suppose it is a repeated dummy value in the rows at the end to pad up to domain size if they are shorter?

msm/src/serialization/witness.rs
```diff
@@ -175,62 +175,45 @@ impl<Fp: PrimeField> Env<Fp> {
...
-        self.lookup_multiplicities_rangecheck4
-            .into_iter()
-            .zip(self.lookup_t_multiplicities_rangecheck4.iter())
-            .enumerate()
-            .for_each(|(i, (m_f, m_t))| m[i] = m_f / m_t);
-        let repeated_dummy_value: Vec<Fp> = iter::repeat(m[0])
+    pub fn get_rangecheck4_multipliticies(&self, domain: EvaluationDomains<Fp>) -> Vec<Fp> {
+        let mut m = Vec::with_capacity(domain.d1.size as usize);
+        m.extend(self.lookup_multiplicities[&LookupTable::RangeCheck4].to_vec());
+        let repeated_dummy_value: Vec<Fp> = iter::repeat(-Fp::zero())
```
querolita, line 186: I see, so it is indeed a dummy repeated value, in this case the first entry which turns out to be zero.

querolita (approved):

msm/src/mvlookup.rs
dannywillems: It is a way to save the fixed lookup tables

msm/src/mvlookup.rs
```diff
@@ -332,26 +365,74 @@ pub mod prover {
...
+
             let lookup_terms_comms_d1: Vec<PolyComm<G>> = (&lookup_terms_evals_d1)
                 .into_par_iter()
                 .map(|lte| srs.commit_evaluations_non_hiding(domain.d1, lte))
                 .collect::<Vec<_>>();
 
+            let fixed_lookup_tables_comms_d1: HashMap<ID, PolyComm<G>> =
+                (&fixed_lookup_tables_evals_d1)
+                    .into_par_iter()
+                    .map(|(id, evals)| (*id, srs.commit_evaluations_non_hiding(domain.d1, evals)))
```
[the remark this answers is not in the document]
volhovm: Also in the MSM / EC addition there's [-2^15,2^15] check equivalently, so 16 bit range check is definitely nice to have, but I could store a sign bit separately and just do 15 bit range check if it's not available.

msm/src/mvlookup.rs
```diff
@@ -328,40 +523,79 @@ pub mod prover {
...
 
             // -- start computing the running sum in lookup_aggregation
-            // The running sum, \phi, is defined recursively over the subgroup as followed:
-            // - phi(1) = 0
-            // - phi(\omega^{j + 1}) = \phi(\omega^j) + \
-            //                         \sum_{i = 1}^{n} (1 / \beta + f_i(\omega^{j + 1})) - \
-            //                         (m(\omega^{j + 1}) / beta + t(\omega^{j + 1}))
-            // - phi(\omega^n) = 0
+            // The running sum, φ, is defined recursively over the subgroup as followed:
+            // - φ(1) = 0
```
volhovm, line 589: Nice comment. Comment for myself: for the trans-circuit MVLookup we'll have to set these 1/n values to previous and next accumulator.

volhovm (commented):
Made a pass. Surprisingly, I had some pending comments from week ago which I forgot to submit. Whoopsie.

Issue #2034, opened 2024-03-22
allow special constants in folding expressions


querolita:
it isn't completely unsupported, you can replace the other variants by the field element they represent and use `FoldingCompatibleExprInner::Constant(f)` for all. Maybe we can add an associated type to the config to allow more complex constant in expressions, but you would have to then resolve it in the environment, and I think it will end being even more work.

_Originally posted by @fabrizio-m in https://github.com/o1-labs/proof-systems/pull/2019#discussion_r1534373334_

PR #2035, opened 2024-03-22
Improve `Literal` so that folding expressions can take special constants as well

querolita:
This PR adds a `as_literal()` function inside the `Literal` trait of the `Expr` framework which is meant to transform expressions with special constants (meaning, EndoCoefficient, MDS, etc) to be represented as plain literal constants using as field value the content of a `Constants` variable passed as argument. 

This way, when using the folding library, one can first "literize" your expressions before converting them into folding compatible expressions. Otherwise, `from` will panic when reaching any special constant.

Closes https://github.com/o1-labs/proof-systems/issues/2034

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -71,6 +71,7 @@ pub struct Challenges<F> {
 }
 
 /// The collection of constants required to evaluate an `Expr`.
+#[derive(Clone)]
 pub struct Constants<F: 'static> {
     /// The endomorphism coefficient
     pub endo_coefficient: F,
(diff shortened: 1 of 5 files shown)


kimchi/src/circuits/expr.rs
```diff
@@ -288,11 +289,12 @@ pub enum ConstantTerm<F> {
     Literal(F),
 }
 
-pub trait Literal: Sized {
+pub trait Literal: Sized + Clone {
```
volhovm, line 292: This is such a weird trait. Very unclear what it's doing without docs. The type suggests it's just some kind of combination of `To + TryTo + From`, but `Result<Self::F, Self>` is definitely not fitting the pattern then :thinking: .

optimism/src/folding.rs
```diff
@@ -276,17 +289,30 @@ mod tests {
...
+                Box::new(x_f.clone()),
                 Box::new(y_f),
             )),
             Box::new(z_f),
         );
-
         assert_eq!(FoldingCompatibleExpr::<TestConfig>::from(xyz), xyz_f);
+
+        let x_endo = x + endo;
+        let x_endo_f = FoldingCompatibleExpr::<TestConfig>::BinOp(
```
volhovm, line 305: So we either add it directly as a constant, or we use the hardcoded `EndoCoefficient`, and checking if these are the same. And the `as_literal` is just for convenience of conversion?

volhovm (approved):

optimism/src/folding.rs
querolita: Yes

PR #2036, opened 2024-03-22
POC Lens-based Gadget Composition

volhovm:
This introduces a mechanism to compose our gadgets (not currently composable) using simple Getter-like lenses. 

Addresses #2037


dannywillems (changes requested):
I won't accept the PR as it is mostly PoC, with `Bla`, `Foo`, `Kek`, ... Not really something to push into a master branch in a main file. It will be dead code/unmaintained.

volhovm:
@dannywillems sure, not an issue, we can integrate it this way or another -- any comments on the content though? I've split the trait into three different capabilities (access, write, lookups) since the last time we talked. cc @querolita that's what we discussed the last time.

PR #2038, opened 2024-03-23
kimchi-visu: custom filepath, bundle assets

emberian:


dannywillems:
However, I think introducing less external dependencies would be great.

dannywillems:
I'm closing it as there is no more activity.

PR #2042, opened 2024-03-26
Use flat memory

dannywillems:


optimism/src/mips/witness.rs
```diff
@@ -819,64 +787,8 @@ impl<Fp: Field> Env<Fp> {
...
-        for (i, (page_index, _memory_write_index)) in self.memory_write_index.iter_mut().enumerate()
-        {
-            if *page_index == page {
-                self.update_last_memory_write_index_access(i);
-                return i;
-            }
-        }
-
-        // Memory not found; dynamically allocate
-        let memory_write_index = vec![0u64; PAGE_SIZE as usize];
```
dannywillems, line 868: We should reallocate here.

dannywillems:
Do not review, it is for later.

PR #2045, opened 2024-03-27
Fresh state are now u64. No need to use Fp

dannywillems:


querolita:
Done in https://github.com/o1-labs/proof-systems/pull/2047/ but merging into parent branch in any case

PR #2040, opened 2024-03-26
[experiment] Speed things up

mrmr1993:

PR #2047, opened 2024-03-27
[Experiment] Batch inversions of registers accesses done before folding

querolita:
This is an exploratory PR trying to fix https://github.com/o1-labs/proof-systems/pull/2040/ , where inverses involved in register accesses were substituted with a dummy term (making the constraints unsatisfiable). 

Here, I am proposing a new kind of `Column` called `InverseState` with at most 9 elements which will keep track of the values that were supposed to be inverted throughout the step execution. Instead of doing this per-step, we are batching these inversions at the end, before folding the witness. At that time, the witness gets mapped from `u64`'s to actual `Fp` elements, and the columns corresponding to inverses are mapped to their inverse value in the field with a single batch inversion for the whole 9 columns. 

Given that most instructions won't use the 9 columns for inversions, these columns are set to 1 by default (reason why when we reset the `self.inverse_state` after every step, we initialize them with 1's instead of 0's as for the scratch state). 

This is just an idea, so please feel free to point out any possible detail I might be missing. I am pointing this out because at first I thought it would be a very complicated change, but the diff right now is not that huge (so I am afraid I might be forgetting about something important).


optimism/src/mips/column.rs
```diff
@@ -86,6 +93,7 @@ impl<T: Clone> IndexMut<Column> for MIPSWitness<T> {
     fn index_mut(&mut self, index: Column) -> &mut Self::Output {
         match index {
             Column::ScratchState(i) => &mut self.cols[i],
+            Column::InverseState(i) => &mut self.cols[SCRATCH_SIZE + i],
             Column::InstructionCounter => &mut self.cols[SCRATCH_SIZE],
```
dannywillems, line 97: :eyes:

optimism/src/mips/witness.rs
```diff
@@ -808,13 +829,19 @@ impl<Fp: Field> Env<Fp> {
...
+        self.inverse_state = fresh_inverse_state();
     }
 
-    pub fn write_field_column(&mut self, column: Column, value: Fp) {
+    pub fn write_column(&mut self, column: Column, value: u64) {
         match column {
             Column::ScratchState(idx) => self.scratch_state[idx] = value,
+            Column::InverseState(idx) => {
+                // So that we won't try to invert the zero value
+                assert_ne!(value, 0);
```
dannywillems, line 842:
```
FILENAME=2023-12-13-15-02-00-op-program-data-log.sh bash run-code.sh
```
raises an error here.

dannywillems:
Idea: not using batch_inverse from arkworks, but implementing the batch inversion ourself, ignoring zero
```
acc = 1
for i, x in witness.iter().enumerate():
  acc *= x if x != 0
acc_inv = acc^-1
for (i, x) in witness_backwards.iter().enumerate():
  inverse_state[i] = acc_inv * x if x!= 0 else 0
  ```

optimism/src/mips/witness.rs
querolita: That's because I am allocating inverse state even when sometimes x is zero, will have to fix this. thanks!

querolita:
Idea from Matthew after our 1:1 to avoid costly u64->Fp transformations:

Perhaps we don't need the Montgomery representation until the very end of the folding, and in between we can instead work with BigUint. Even compute the batch inverses as a BigUint modulo the field size.

querolita:
I made this a draft because performance is still not as fast as desired (in my machine, Matthew's idea runs at 1.3 Mhz, but this one is about 5 times slower). I believe it is the castings from u64 to F what is causing it. Will go back to it in May after finishing the main goal for part 1.

querolita:
> Idea: not using batch_inverse from arkworks, but implementing the batch inversion ourself, ignoring zero

@dannywillems actually the `batch_inverse` from arkworks already ignores zeros, so that's not a problem anymore

dannywillems:
Closing as https://github.com/o1-labs/proof-systems/pull/2813 has been merged, and reuse the same idea than this PR.

PR #2050, opened 2024-04-01
Keccak: add regression tests on the number of constraints and lookups

dannywillems:


querolita (approved):
Great idea, I completely miscounted the degree of `is_round()` and `in_padding()`. It is clear now that we need to have selectors to avoid these unnecessary high degrees in constraints and lookups.

Issue #1751, opened 2024-01-24
Optimize sponge witness offset to avoid degree 3 constraints


querolita:
> Only 5 constraints have degree 3, so we shall expect that after quadricization is applied, we will need just a few more witness columns and constraints to make them degree 2.

> Alternatively, we can intentionally shift the beginning of the sponge columns (since they have about 1k less used columns than the round), so that the `quotient_c` values are located in witness columns which would be set to zero regardless in sponge steps. That way, we could rephrase the constraint to be instead of `is_round * is_boolean (quotient_c)` to just `is_boolean(quotient_c)`. Opening issue to trace this optimization

PR #2051, opened 2024-04-02
Split Keccak circuit into steps with specialized constraints

querolita:
This PR manages to split the all-purpose Keccak circuit into separate specialized steps so that each row can be made part of an independent instance for folding. In particular, it now has 6 different modes of operation: `Round`, `Sponge::Squeeze`, `Sponge::Absorb(First)`, `Sponge::Absorb(Middle)`, `Sponge::Absorb(Last)`, and `Sponge::Absorb(Only)` (they reuse the `Step` enum used previously).

Now, less constraints and actual columns are used in comparison with `master`. In particular, 2079 columns instead of 2219 (140 columns less), and the sub-circuit with more constraint is the `Sponge::Absorb(Only)`, which uses 474 constraints instead of 887. Also, constraints checking the correct composition of the flags are no longer needed, because now there will be a "public" selector for each of the variants independently. This means some constraints are removed (booleanity and mutual exclusivity), and they have been replaced with simple checks on the witness environment side, just to check that they were configured correctly. 

Now, for each step of the execution, one witness environment will be instantiated with a single selector set to 1 (and the rest are 0). Note that the folding column trait has been adapted so that `is_witness()` returns `false` for selector columns. Apart from this, constraints for specific sub-circuits can be generated upon a constraint environment which has the `.step` field set to the desired selector. These will be shared across the whole sub-circuit for that type.

The regression test has been adapted accordingly as well, to bring further visibility to the concrete changes that have occurred on the constraints side. Special mentioning that the degrees of the constraints have been reduced drastically by introducing these "fine-grain" selectors. In particular, most constraints are degree 1 and just a few are degree 2, whereas the old design had many degree 3 and 4 constraints (which would have resulted in a non-negligible increase of folding columns).

I recommend reviewing this PR in a per-commit basis, where I tried to keep the changes to a minimum. Do not try to compile each individual commit, because most changes done throughout the implementation of this PR were breaking other parts of the code. Compilation was fixed only at the very last few commits of the history. 

Closes https://github.com/MinaProtocol/mina/issues/15163 and https://github.com/o1-labs/proof-systems/issues/1751 as a side effect. 

NOTE: this still DOES NOT split lookups into sub-circuits. That means, in this PR all lookups are added, no matter the step type. I am addressing this in an upcoming PR because I didn't want this one to be that long. That was a non-breaking stop point, convenient for a PR switch.

TODO: I would like to treat the selector columns separately from the `Witness::cols` field, so we can address commitments to these columns in a special way (meaning, they might only become real columns in folding, so there are no duplicates.

### DIFF
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -26,7 +26,7 @@ const FLAG_ROUND_OFF: usize = 0; // Offset of the Round selector inside the mode
 const FLAG_ROOT_OFF: usize = 1; // Offset of the Root selector inside the mode flags
 const FLAG_ABSORB_OFF: usize = 2; // Offset of the Absorb selector inside the mode flags
 const FLAG_PAD_OFF: usize = 3; // Offset of the Pad selector  inside the mode flags
-const FLAG_PADROOT_OFF: usize = 4; // Offset of the PadRoot selector  inside the mode flags
+const FLAG_ROOTPAD_OFF: usize = 4; // Offset of the RootPad selector  inside the mode flags
 const FLAG_SQUEEZE_OFF: usize = 5; // Offset of the Squeeze selector inside the mode flags
 
 const STATUS_IDXS_OFF: usize = MODE_LEN; // The offset of the columns reserved for the status indices
@@ -113,12 +113,12 @@ pub enum Flag {
 }
 
 /// The columns used by the Keccak circuit.
-/// The Keccak circuit is split into two main modes: Round and Sponge (split into Root, Absorb, Pad, PadRoot, Squeeze).
+/// The Keccak circuit is split into two main modes: Round and Sponge (split into Root, Absorb, Pad, RootPad, Squeeze).
 /// The columns are shared between the Sponge and Round steps
 /// (the total number of columns refers to the maximum of columns used by each mode)
 /// The hash, block, and step indices are shared between both modes.
 /// The row is split into the following entries:
-/// - mode_flags: what kind of mode is running: round, root, absorb, pad, padroot, squeeze. Only 1 of them can be active.
+/// - mode_flags: what kind of mode is running: round, root, absorb, pad, rootpad, squeeze. Only 1 of them can be active.
 /// - hash_index: Which hash this is inside the circuit
 /// - block_index: Which block this is inside the hash
 /// - step_index: Which step this is inside the hash
@@ -272,7 +272,7 @@ impl<T: Clone> Index<Column> for KeccakWitness<T> {
             Column::Selector(Flag::Root) => &self.mode_flags()[FLAG_ROOT_OFF],
             Column::Selector(Flag::Absorb) => &self.mode_flags()[FLAG_ABSORB_OFF],
             Column::Selector(Flag::Pad) => &self.mode_flags()[FLAG_PAD_OFF],
-            Column::Selector(Flag::PadRoot) => &self.mode_flags()[FLAG_PADROOT_OFF],
+            Column::Selector(Flag::PadRoot) => &self.mode_flags()[FLAG_ROOTPAD_OFF],
             Column::Selector(Flag::Squeeze) => &self.mode_flags()[FLAG_SQUEEZE_OFF],
 
             Column::HashIndex => self.hash_index(),
@@ -389,7 +389,7 @@ impl<T: Clone> IndexMut<Column> for KeccakWitness<T> {
             Column::Selector(Flag::Root) => &mut self.mode_flags_mut()[FLAG_ROOT_OFF],
             Column::Selector(Flag::Absorb) => &mut self.mode_flags_mut()[FLAG_ABSORB_OFF],
             Column::Selector(Flag::Pad) => &mut self.mode_flags_mut()[FLAG_PAD_OFF],
-            Column::Selector(Flag::PadRoot) => &mut self.mode_flags_mut()[FLAG_PADROOT_OFF],
+            Column::Selector(Flag::RootPad) => &mut self.mode_flags_mut()[FLAG_ROOTPAD_OFF],
             Column::Selector(Flag::Squeeze) => &mut self.mode_flags_mut()[FLAG_SQUEEZE_OFF],
 
             Column::HashIndex => &mut self.cols[STATUS_IDXS_OFF],
--- a/optimism/src/keccak/column.rs
+++ b/optimism/src/keccak/column.rs
@@ -1,5 +1,6 @@
 //! This module defines the custom columns used in the Keccak witness, which
 //! are aliases for the actual Keccak witness columns also defined here.
+use self::Sponge::*;
 use crate::keccak::{ZKVM_KECCAK_COLS_CURR, ZKVM_KECCAK_COLS_NEXT};
 use kimchi::circuits::polynomials::keccak::constants::{
     CHI_SHIFTS_B_LEN, CHI_SHIFTS_B_OFF, CHI_SHIFTS_SUM_LEN, CHI_SHIFTS_SUM_OFF, PIRHO_DENSE_E_LEN,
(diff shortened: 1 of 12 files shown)


optimism/src/keccak/interpreter.rs
```diff
@@ -791,19 +791,19 @@ pub trait KeccakInterpreter<F: One + Debug + Zero> {
...
     /// Returns a variable that encodes whether the current step is the first absorb sponge (1 = yes)
     fn is_root(&self) -> Self::Variable {
-        self.variable(KeccakColumn::FlagRoot)
+        self.variable(KeccakColumn::Selector(Root))
     }
-    /// Returns a degree-2 variable that encodes whether the current step is the last absorb sponge (1 = yes)
+    /// Returns a degree-1 variable that encodes whether the current step is the last absorb sponge (1 = yes)
     fn is_pad(&self) -> Self::Variable {
-        self.pad_length() * self.variable(KeccakColumn::InvPadLength)
+        self.variable(KeccakColumn::Selector(Pad))
```
volhovm: IIRC you added an extra selector here, right? Instead of storing length and inv_length

optimism/src/keccak/folding.rs
```diff
@@ -29,8 +29,10 @@ pub(crate) struct KeccakConfig;
 
 impl FoldingColumnTrait for KeccakColumn {
     fn is_witness(&self) -> bool {
-        // All Keccak columns are witness columns
-        true
+        match self {
+            KeccakColumn::Selector(_) => false,
```
volhovm: What's the semantics of this trait, what's exactly a "witness" for folding? Selectors are part of the computation, so why are they not part of the witness? Because they're public?..

optimism/src/keccak/interpreter.rs
```diff
@@ -115,7 +115,7 @@ pub trait KeccakInterpreter<F: One + Debug + Zero> {
     fn variable(&self, column: KeccakColumn) -> Self::Variable;
 
     /// Adds one KeccakConstraint to the environment if the selector holds
-    fn constrain(&mut self, tag: Constraint, flag: Self::Variable, x: Self::Variable);
+    fn constrain(&mut self, tag: Constraint, if_true: Self::Variable, x: Self::Variable);
```
volhovm, line 124: `if_true` is a variable expressing the selector result? (I see, yeah.) Both names are not exactly directly intuitive IMHO...

optimism/src/keccak/interpreter.rs
```diff
@@ -165,107 +180,118 @@ pub trait KeccakInterpreter<F: One + Debug + Zero> {
...
-                        - Self::from_shifts(&sponge_shifts, Some(i), None, None, None)),
+                self.is_squeeze(),
+                (self.old_state(i).clone()
+                    - Self::from_shifts(&sponge_shifts, Some(i), None, None, None)),
             );
         }
     }
 
     /// Constrains 389 checks of round steps
     /// - 384 constraints of degree 3
```
volhovm, line 276: Do you need to update these comments, or the degree stays the same?

optimism/src/keccak/interpreter.rs
```diff
@@ -130,6 +136,8 @@ pub trait KeccakInterpreter<F: One + Debug + Zero> {
     /// So:
     /// - At most, 474 constraints are added per row
     fn constraints(&mut self) {
+        self.check_selectors();
```
volhovm: How is `check_selectors` related to `checks`? And how are these related to `check`? It seems that they are all very similar in their implementation with quite generic naming.

optimism/src/keccak/witness.rs
```diff
@@ -159,19 +174,19 @@ impl<F: Field> KeccakInterpreter<F> for Env<F> {
...
     }
     fn mode_root(&self) -> Self::Variable {
-        self.variable(KeccakColumn::Selector(Root))
+        self.variable(KeccakColumn::Selector(Sponge(Absorb(First))))
     }
     fn mode_pad(&self) -> Self::Variable {
-        self.variable(KeccakColumn::Selector(Pad))
+        self.variable(KeccakColumn::Selector(Sponge(Absorb(Last))))
     }
     fn mode_rootpad(&self) -> Self::Variable {
```
volhovm, line 192: Is this renamed into `mode_only` in some following commits?

optimism/src/keccak/environment.rs
```diff
@@ -47,18 +53,16 @@ pub struct KeccakEnv<F> {
     /// Byte-length of the 10*1 pad (<=136)
     pub(crate) pad_len: u64,
 
-    /// Precomputed 2^pad_len (including 2^0 = 1)
-    two_to_pad: [F; RATE_IN_BYTES + 1],
-    /// Precomputed inverses of the padding lengths (including dummy 0)
-    inv_pad_len: [F; RATE_IN_BYTES + 1],
```
volhovm, line 53: Inverses are removed permanently, right?

optimism/src/keccak/environment.rs
```diff
@@ -47,18 +53,16 @@ pub struct KeccakEnv<F> {
...
+    /// Precomputed 2^pad_len
+    two_to_pad: [F; RATE_IN_BYTES],
     /// Precomputed suffixes for the padding blocks
-    pad_suffixes: [[F; PAD_SUFFIX_LEN]; RATE_IN_BYTES + 1],
```
volhovm, line 55: Why do `pad_suffixes` and `two_to_pad` get one less element than they had before? `RATE_IN_BYTES` instead of `RATE_IN_BYTES + 1`?

volhovm (approved):

optimism/src/keccak/column.rs
```diff
@@ -76,11 +79,21 @@ pub enum Column {
...
+    SpongeShifts(usize),   // Sponge Curr[400..800)
+}
+
+/// These selectors determine the specific behaviour so that Keccak steps
+/// can be split into different instances for folding
+pub enum Flag {
+    IsRound,        // Current step performs a round of the permutation
+    IsAbsorbFirst,  // Current step performs a root absorb
+    IsAbsorbMiddle, // Current step performs an absorb of the sponge
+    IsAbsorbLast,   // Current step performs the padding absorb
```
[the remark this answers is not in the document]
querolita: FlagAbsorb was any absorb: so either the first one, the middle one, the last one, or the only one

optimism/src/keccak/interpreter.rs
querolita: Actually if the Last absorb and the Only absorb have their own selectors for themselves, then there's no need of storing the inverse lengths anymore to check whether it is nonzero.

optimism/src/keccak/folding.rs
querolita: Well, they will be considered aside so as to avoid duplicates. But that will come after optimizations.

optimism/src/keccak/interpreter.rs
querolita: oh you're right, they are outdated in those comments. Will push a commit

optimism/src/keccak/interpreter.rs
querolita: `check_selectors` are no longer used in later commits. There will only be `check` to check one thing, and `checks` which adds all the checks that could be relevant for the witness but which do not require constraints for them. Basically, just to check that I am configuring the selectors correctly in the witness side.

optimism/src/keccak/witness.rs
querolita: No, those modes remain as rootpad, pad, root, etc.

optimism/src/keccak/environment.rs
querolita: Yes

optimism/src/keccak/environment.rs
querolita: Because before, the constraints for the dummy witness set to all zeros should also hold. But now that you can select to run them or not, you can be fine with simply the "real" values. The one with dummy entries will never be used now.

PR #2053, opened 2024-04-03
FF addition snarkyrs support

fabrizio-m:
Mostly done, remains for it to pass the test, drafting for now as it isn't a priority.


dannywillems:
o1rs is being rewritten. I guess we can close this.

PR #2056, opened 2024-04-03
Add selector aliases in MIPS columns

querolita:
This PR creates some common background to be used in upcoming PRs trying to split the MIPS circuit into separate sub-witnesses depending on the instruction type. ~It also creates a wrapper called `variable` inside the interpreter that can be used inside `constraints.rs` to reduce the number of lines of code used before every time a variable was being returned.~ <- This was added independently [here](https://github.com/o1-labs/proof-systems/pull/2058). 

This still does not address constraint nor lookup separation. It just creates some selectors that will be used later.

TODO: probably, the Keccak Syscall instruction will need some separate selectors for finer grain treatment.


querolita:
This is probably not needed thanks to the way the MIPS interpreter is currently designed.

PR #2058, opened 2024-04-08
Create `variable` function inside the MIPS interpreter to improve readability

querolita:
This PR reates a wrapper called `variable` inside the interpreter that can be used inside `constraints.rs` to reduce the number of lines of code used before every time a variable was being returned.

### DIFF
--- a/optimism/src/mips/constraints.rs
+++ b/optimism/src/mips/constraints.rs
@@ -48,6 +48,13 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
 
     type Variable = Expr<ConstantExpr<Fp>, Column>;
 
+    fn variable(&self, column: Self::Position) -> Self::Variable {
+        Expr::Atom(ExprInner::Cell(Variable {
+            col: column,
+            row: CurrOrNext::Curr,
+        }))
+    }
+
     fn add_constraint(&mut self, assert_equals_zero: Self::Variable) {
         self.constraints.push(assert_equals_zero)
     }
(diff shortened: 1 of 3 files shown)


optimism/src/mips/witness.rs
```diff
@@ -124,6 +124,24 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
 
     type Variable = u64;
 
+    // TODO: update variable() function once scratch_state is [u64; SCRATCH_SIZE]
```
volhovm, line 127: I assume this is intended for this PR at least?

optimism/src/mips/constraints.rs
```diff
@@ -48,6 +48,13 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
 
     type Variable = Expr<ConstantExpr<Fp>, Column>;
 
+    fn variable(&self, column: Self::Position) -> Self::Variable {
```
volhovm, line 51: Technically could be a static function, used as `Env::variable(column)`?

volhovm (approved):

optimism/src/mips/constraints.rs
querolita: Yeah, it's just used to improve readability of the content of the functions inside the Env of constraints.

optimism/src/mips/witness.rs
querolita: Yeah, because this is an implementation of the `InterpreterEnv` trait, I needed to create the corresponding `variable` function inside `witness.rs` as well. But for now, I cannot return a `Self::Variable` in that case because `Self::Variable` is a native type but the witness data is a field type. But when we move to Matthew's proposal (https://github.com/o1-labs/proof-systems/pull/2047) and the witness is computed in native types, then I will be able to remove the `todo!()` with the code comented below.

PR #2059, opened 2024-04-08
Using generic prover inside zkVM

querolita:
This PR removes the old prove/verify algorithms inside the zkVM project and instead uses the types and functions in the generic proof inside the MSM project. This means the proofs generated in `main()` use that prover/verifier pair instead. But because inside the loop the circuits are being folded in a dummy way, and creating no error terms whatsoever, running the actual constraints on the folded witnesses would not be satisfiable. For that reason, the proof is just being created with empty vector of constraints for now, and no lookups.

It also adds one test in the Keccak folder that does use the constraints in a non-folded circuit run. I had to increase the stack size for now, given the long witness. But the test is still failing and I am working to figure out why, given that the witness satisfies constraints individually. Will keep this description updated.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,6 +10,8 @@ env:
   CARGO_INCREMENTAL: 1
   # https://nexte.st/book/pre-built-binaries.html#using-nextest-in-github-actions
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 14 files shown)


querolita:
Closing because this was addressing the old circuit format where Keccak instructions were not yet split. The branch `zkvm/split/generic-prover` does it instead.

PR #2063, opened 2024-04-08
Move berkeley related expressions into berkeley_column.rs

dannywillems:
Planning to implement FormattedOutput for the generic prover


mrmr1993 (approved):
Very nice :)

PR #2067, opened 2024-04-10
gate example with bn254 feature

fabrizio-m:


querolita (commented):
Isn't there need to include the cfg also around here? https://github.com/o1-labs/proof-systems/blob/842d0a9584c072507c1f0a774f6d4738f5271399/optimism/src/folding.rs#L154

fabrizio-m:
indeed, as it using `BaseSponge`

querolita (approved):

PR #2070, opened 2024-04-11
MSM: Integrate range lookups into existing MSM circuits

volhovm:
This PR:
* Touches only `src/serialization` code
* Adds support for `4abs` and `ffhighest` range checks.
* Adds serialization ECADD circuit to the `serialization/` completeness test.
* Generalises some range check related code to handle 4 existing tables uniformly without copy-paste.


msm/src/serialization/witness.rs
```diff
@@ -210,13 +213,29 @@ impl<F: PrimeField> Env<F> {
             lookups,
         }
     }
+
+    // Commonly used by range checking functions.
+    fn record_range_check(
```
dannywillems, line 218: This doesn't seem to be anything related to range checks. It is simply a lookup, right?

msm/src/serialization/witness.rs
```diff
@@ -169,37 +163,46 @@ impl<F: PrimeField> Env<F> {
...
+        if table_id.length() < (domain.d1.size as usize) {
+            let n_repeated_dummy_value: usize = (domain.d1.size as usize) - table_id.length() - 1;
+            let repeated_dummy_value: Vec<F> = iter::repeat(-F::one())
+                .take(n_repeated_dummy_value)
+                .collect();
+            m.extend(repeated_dummy_value);
+            // TODO @volhovm Why do we push the last index as a last
+            // element?
+            //
+            // FIXME this makes the logup unsound: we add dummy values
```
dannywillems, line 189: This is not true.

msm/src/serialization/witness.rs
```diff
@@ -169,37 +163,46 @@ impl<F: PrimeField> Env<F> {
...
-        *self.lookups.get_mut(&LookupTable::RangeCheck15).unwrap() = Vec::new();
+        for table in self.lookups.values_mut() {
+            *table = Vec::new();
+        }
+        // TODO do we need to reset multiplicities?
     }
 
-    pub fn get_rangecheck4_multipliticies(&self, domain: EvaluationDomains<F>) -> Vec<F> {
+    /// Getting multiplicities for range check tables less or equal than 15 bits.
+    pub fn get_rangecheck_multiplicities(
```
dannywillems, line 173: I would rename it in `get_lookup_multiplicities`.

msm/src/serialization/mod.rs
```diff
@@ -8,24 +12,30 @@ pub mod witness;
...
 pub const N_INTERMEDIATE_LIMBS: usize = 20;
 
-#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
-pub enum LookupTable {
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Ord, PartialOrd, EnumIter)]
+pub enum LookupTable<Ff> {
     RangeCheck15,
     RangeCheck4,
+    RangeCheck4Abs,
+    RangeCheckFfHighest(PhantomData<Ff>),
```
dannywillems, line 20: Can you add a comment describing the goal of the table, please?
Nit: I would mention what `Ff` means in a comment.

msm/src/serialization/interpreter.rs
```diff
@@ -83,23 +89,23 @@ pub trait InterpreterEnv<F: PrimeField> {
 /// ```
 /// And we can ignore the last 10 bits (i.e. `limbs2[78..87]`) as a field element
 /// is 254bits long.
-pub fn deserialize_field_element<F: PrimeField, Env: InterpreterEnv<F>>(
+pub fn deserialize_field_element<F: PrimeField, Ff: PrimeField, Env: InterpreterEnv<F, Ff>>(
     env: &mut Env,
-    limbs: [u128; 3],
+    limbs: [BigUint; 3],
```
dannywillems, line 94: Any reason to change?

msm/src/serialization/constraints.rs
```diff
@@ -132,8 +135,14 @@ impl<F: PrimeField> Env<F> {
             .collect();
         constraints.extend(relation_constraints);
 
-        assert!(self.lookups[&LookupTable::RangeCheck15].len() == 17);
+        // @volhovm The numbers here depend on the circuit, so these
```
dannywillems, line 138: I don't understand the reason of this comment. Is it useful?

msm/src/serialization/interpreter.rs
volhovm: The problem is that there is no `From` converting a field element to `u128` which I needed in several places. Generally it seems `BigUint` is not adding any significant overhead over `u128`, but it has significantly better interoperability with `Field` because we have `FieldHelpers` etc in `utils`.

volhovm (commented):
I hate these delayed comments. I wrote them hours ago :( :( :( :( :(

volhovm:
@dannywillems ready for review now. Importantly, the two new tables work + the extra piece of code that does FFEC addition. 

I'm not happy about interfaces, but will improve them in the next PR to keep this one more lightweight.

dannywillems:
> @dannywillems ready for review now. Importantly, the two new tables work + the extra piece of code that does FFEC addition.
> 
> I'm not happy about interfaces, but will improve them in the next PR to keep this one more lightweight.

Does not seem to be ready, cf CI.

volhovm:
Trying again. Funnily enough this passes perfectly well on my machine. Rust version mismatch?...

dannywillems (approved):

PR #2071, opened 2024-04-11
Split MIPS and Keccak instances in `main` and proof with constraints

querolita:
This PR updates the `main.rs` file (and thus the zkvm demo) to split the instructions/steps into separate instances for MIPS and Keccak, as a prior step to folding. On top of that, the generic prover is being used with the actual constraints for both MIPS and Keccak. In order for this to work, I am creating a dummier folding algorithm which does not aggregate instances as a linear combination of the witnesses, but instead just keeps the last witness chunk. This for sure, will need to be modified when the folding library is fully in place, so that folded instances contain information about all the previous witnesses. The reason for this tweak now is that if I randomly aggregated witnesses without actually obtaining the corresponding error terms, the constraints would not be satisfied on the randomized witnesses.

TODO: Lookups are still not being proven in this PR. Only witness constraints.

Closes: https://github.com/MinaProtocol/mina/issues/15166 https://github.com/MinaProtocol/mina/issues/15164

NOTE: while testing this I realized that some MIPS constraints are not passing. Will investigate more.

### DIFF
--- a/optimism/src/keccak/environment.rs
+++ b/optimism/src/keccak/environment.rs
@@ -85,8 +85,10 @@ impl<F: Field> KeccakEnv<F> {
     /// Starts a new Keccak environment for a given hash index and bytestring of preimage data
     pub fn new(hash_idx: u64, preimage: &[u8]) -> Self {
         // Must update the flag type at each step from the witness interpretation
-        let mut env = KeccakEnv::default();
-        env.hash_idx = hash_idx;
+        let mut env = KeccakEnv::<F> {
+            hash_idx,
+            ..Default::default()
+        };
 
         // Store hash index in the witness
         env.write_column(KeccakColumn::HashIndex, env.hash_idx);
--- a/optimism/src/keccak/environment.rs
+++ b/optimism/src/keccak/environment.rs
@@ -134,6 +134,15 @@ impl<F: Field> KeccakEnv<F> {
         self.constraints_env.lookups = vec![];
     }
 
+    /// Returns the selector of the current step
+    pub fn selector(&self) -> Steps {
+        let mut step = self.constraints_env.step.unwrap();
+        if let Round(_) = step {
+            step = Round(0);
+        }
+        step
+    }
+
     /// Entrypoint for the interpreter. It executes one step of the Keccak circuit (one row),
     /// and updates the environment accordingly (including the witness and inter-step lookups).
     /// When it finishes, it updates the value of the current step, so that the next call to
--- a/optimism/src/keccak/environment.rs
+++ b/optimism/src/keccak/environment.rs
@@ -63,24 +63,30 @@ pub struct KeccakEnv<F> {
     pad_suffixes: [[F; PAD_SUFFIX_LEN]; RATE_IN_BYTES],
 }
 
-impl<F: Field> KeccakEnv<F> {
-    /// Starts a new Keccak environment for a given hash index and bytestring of preimage data
-    pub fn new(hash_idx: u64, preimage: &[u8]) -> Self {
-        let mut env = Self {
-            // Must update the flag type at each step from the witness interpretation
+impl<F: Field> Default for KeccakEnv<F> {
+    fn default() -> Self {
+        Self {
             constraints_env: ConstraintsEnv::default(),
             witness_env: WitnessEnv::default(),
-            hash_idx,
+            hash_idx: 0,
             step_idx: 0,
             block_idx: 0,
             prev_block: vec![],
             blocks_left_to_absorb: 0,
             padded: vec![],
             pad_len: 0,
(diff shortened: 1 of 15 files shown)


optimism/src/keccak/environment.rs
```diff
@@ -63,23 +63,31 @@ pub struct KeccakEnv<F> {
     pad_suffixes: [[F; PAD_SUFFIX_LEN]; RATE_IN_BYTES],
 }
 
-impl<F: Field> KeccakEnv<F> {
-    /// Starts a new Keccak environment for a given hash index and bytestring of preimage data
-    pub fn new(hash_idx: u64, preimage: &[u8]) -> Self {
```
volhovm, line 68: Was `preimage` unused before?

optimism/src/keccak/mod.rs
```diff
@@ -131,23 +132,35 @@ impl<F: Field> KeccakCircuit<F> {
...
-            Sponge(Absorb(Middle)),
-            Sponge(Absorb(Last)),
-            Sponge(Absorb(Only)),
-            Sponge(Squeeze),
-        ] {
-            let rows_left =
-                self.witness[&step].cols[0].capacity() - self.witness[&step].cols[0].len();
-            for _ in 0..rows_left {
-                self.push_row(step, &[F::zero(); ZKVM_KECCAK_COLS]);
+    fn pad(&mut self, step: Steps) -> bool {
```
volhovm, line 135:
```suggestion
    /// Pads unused rows for a particular keccak step if needed. Returns true if padding was performed, false otherwise.
    fn pad(&mut self, step: Steps) -> bool {
```

optimism/src/lib.rs
```diff
@@ -50,11 +50,30 @@ pub use ramlookup::{LookupMode as RAMLookupMode, RAMLookup};
...
+    pub witness: HashMap<SELECTOR, Witness<N, Vec<F>>>,
     /// The vector of constraints for a given selector
-    pub(crate) constraints: HashMap<SELECTOR, Vec<E<F>>>,
+    pub constraints: HashMap<SELECTOR, Vec<E<F>>>,
     /// The vector of lookups for a given selector
-    pub(crate) lookups: HashMap<SELECTOR, Vec<Lookup<E<F>>>>,
+    pub lookups: HashMap<SELECTOR, Vec<Lookup<E<F>>>>,
+}
+
+pub trait CircuitTrait<const N: usize, SELECTOR, F, Env> {
```
volhovm, line 64: As I mentioned in the previous PR, this to me looks much more like a generic interface for a `WitnessBuilder`. In `msm/src/ffadd/` all these functions are part of the `impl WitnessBuilderEnv`. Generally a useful abstraction, I just feel the name is confusing.

optimism/src/mips/mod.rs
```diff
@@ -15,9 +15,176 @@
...
 pub mod folding;
 pub mod interpreter;
 pub mod registers;
 pub mod witness;
+
+#[allow(dead_code)]
+/// The Keccak circuit
+pub type MIPSCircuit<F> = Circuit<MIPS_COLUMNS, Instruction, F>;
+
+pub const INSTRUCTIONS: [Instruction;
```
volhovm, line 51: Was this created from scratch or moved from somewhere?...

optimism/src/mips/mod.rs
```diff
@@ -15,9 +15,176 @@
...
+        self.witness.entry(instr).and_modify(|wit| {
+            for (i, value) in row.iter().enumerate() {
+                if wit.cols[i].len() < wit.cols[i].capacity() {
+                    wit.cols[i].push(*value);
+                }
+            }
+        });
+    }
+
+    fn pad(&mut self, instr: Instruction) -> bool {
```
volhovm, line 161: A lot of these functions will be quite similar for many implementations... I'm wondering if it makes sense to abstract this.

volhovm (approved):

optimism/src/keccak/environment.rs
querolita: Not really, it is used from the witness env of the mips side to create a new keccak env after finding a syscall instruction.

optimism/src/lib.rs
querolita: Opened an issue to keep track of it in a follow-up https://github.com/o1-labs/proof-systems/issues/2073

optimism/src/mips/mod.rs
querolita: From scratch by enumerating all the variants for the Instruction enum. If you know of a method that automatically iterates over them all please let me know! This is awful 🥲

volhovm: Nice, didn't know about this mechanism, looks quite elegant!

PR #2077, opened 2024-04-16
Separate functionalities of Keccak interpreter

querolita:
After our chat about a general-purpose interpreter, I worked on this PR aiming to split the functionalities in the Keccak interpreter across different traits with separate scopes. In particular, I have split the old `KeccakInterpreter` into 5 traits: `Interpreter` (the most general type of interpreter I could think of for now), `KeccakInterpreter` (which contains Keccak-related functionalities for both witness and constraints environments), `ArithHelpers` (containing arithmetic operations), `BoolHelpers` (containing boolean operations), and `LookupHelpers` (containing lookup helpers for the tables used in the zkvm project).

Note that there is no need to parameterize these traits with any generic `Env` in particular. But instead, the use of the clause `where Self: Interpreter<F> + ArithHelpers<F> + BoolHelpers<F> + LookupHelpers<F>` makes the diff super small (no need to change the `self.fun()` invocations nor change the type of `Self::Variable`) in comparison. And of course, it is still possible to instantiate it for a witness or constraint environment, just setting the `Self` to be either of them.

### DIFF
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -6,7 +6,7 @@ use crate::{
             Sponges::*,
             Steps::{self, *},
         },
-        helpers::{ArithHelpers, BoolHelpers, LookupHelpers},
+        helpers::{ArithHelpers, BoolHelpers, LogupHelpers},
         interpreter::{Interpreter, KeccakInterpreter},
         Constraint, KeccakColumn, Selector,
     },
@@ -52,7 +52,7 @@ impl<F: Field> ArithHelpers<F> for Env<F> {
 
 impl<F: Field> BoolHelpers<F> for Env<F> {}
 
-impl<F: Field> LookupHelpers<F> for Env<F> {}
+impl<F: Field> LogupHelpers<F> for Env<F> {}
 
 impl<F: Field> Interpreter<F> for Env<F> {
     type Variable = E<F>;
--- a/optimism/src/keccak/constraints.rs
+++ b/optimism/src/keccak/constraints.rs
@@ -6,6 +6,8 @@ use crate::{
             Sponges::*,
             Steps::{self, *},
         },
+        helpers::{ArithHelpers, BoolHelpers, LookupHelpers},
+        interpreter::{Interpreter, KeccakInterpreter},
         Constraint, KeccakColumn, Selector,
     },
     lookups::Lookup,
@@ -20,8 +22,6 @@ use kimchi::{
     o1_utils::Two,
 };
 use kimchi_msm::columns::ColumnIndexer;
-
-use super::interpreter::KeccakInterpreter;
 
 /// This struct contains all that needs to be kept track of during the execution of the Keccak step interpreter
 #[derive(Clone, Debug)]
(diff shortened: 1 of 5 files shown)


optimism/src/keccak/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.add_lookup(flag, Lookup::read_one(SyscallLookup, value));
+    }
+
+    fn write_syscall(&mut self, flag: Self::Variable, value: Vec<Self::Variable>) {
+        self.add_lookup(flag, Lookup::write_one(SyscallLookup, value));
+    }
+}
+
+/// This trait contains helper functions for boolean operations used in the Keccak circuit
+pub trait BoolHelpers<F: One + Debug + Zero>
```
volhovm, line 59: I don't know what's the good practice, but in general I find it very counterintuitive to define actual functions within traits. If `BoolHelpers` is not going to be instantiated especially... I kind of think every trait should have several instantiations, but `BoolHelpers` does not need any. Weird.

However I don't see a good alternative for now. One could declare these functions (`is_boolean` and so on) on the top level, but it seems to be inelegant in a different way. So I probably agree to go forward with this technique.

optimism/src/keccak/witness.rs
```diff
@@ -61,11 +62,18 @@ impl<F: Field> Default for Env<F> {
...
 }
 
-impl<F: Field> KeccakInterpreter<F> for Env<F> {
+impl<F: Field> ArithHelpers<F> for Env<F> {
+    fn two_pow(x: u64) -> <Env<F> as Interpreter<F>>::Variable {
+        Self::constant_field(F::two_pow(x))
+    }
+}
+
+impl<F: Field> BoolHelpers<F> for Env<F> {}
```
volhovm, line 71: Yes, so this seems weird. If these functions were declared on the top level, one wouldn't need these declarations. But maybe it's easier like this. Dunno.

optimism/src/keccak/constraints.rs
```diff
@@ -21,8 +23,6 @@ use kimchi::{
 };
 use kimchi_msm::columns::ColumnIndexer;
 
-use super::interpreter::KeccakInterpreter;
-
 /// This struct contains all that needs to be kept track of during the execution of the Keccak step interpreter
 #[derive(Clone, Debug)]
 pub struct Env<Fp> {
```
volhovm, line 28: What do you think about calling this one `ConstraintEnv` and the other `WitnessEnv`?

volhovm (approved):

volhovm:
I spent like 30 minutes thinking of is there any middle way between these traits and a module with top-level function. Really puzzles me. Like I have my 
```
pub trait MyTrait {
    fn core_function(&self);
}
```
And I can't declare in the other crate/module/package something like
```
impl MyTrait {
    fn helper_function(&self) {
        self.core_function();
    }
}
```
But I can only do
```
fn helper_function<T: MyTrait>(my_trait: T) {
    my_trait.core_function();
}
```
???????

This seems to me like there should be a better way. But maybe there's none.

dannywillems (approved):

optimism/src/keccak/constraints.rs
querolita: it is called like that from outside when accessing each of them (main), I just followed what appears in MIPS

PR #2080, opened 2024-04-17
MSM/witness: rename Fp in F, solving TODO

dannywillems:
I don't like leaving small TODO's like this :).


volhovm:
Added a few more `Fp -> F` renamings, hopefully found all of them at least in the `msm/` folder.

volhovm (approved):

querolita (approved):

PR #2084, opened 2024-04-17
Logup: more documentation

dannywillems:
Run `RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps`.

### DIFF
--- a/msm/src/logup.rs
+++ b/msm/src/logup.rs
@@ -1,11 +1,133 @@
-//! Implement a variant of the logarithmic lookups based on the equations
-//! described in the paper [Multivariate lookups based on logarithmic
-//! derivatives](https://eprint.iacr.org/2022/1530.pdf)
+//! Implement a variant of the logarithmic derivative lookups based on the
+//! equations described in the paper ["Multivariate lookups based on logarithmic
+//! derivatives"](https://eprint.iacr.org/2022/1530.pdf).
+//!
 //! The variant is mostly based on the observation that the polynomial
 //! identities can be verified using the "Idealised low-degree protocols"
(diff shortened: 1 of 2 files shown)


querolita (approved):
Looks great to me

msm/src/logup.rs
```diff
@@ -1,11 +1,133 @@
...
+//! derivatives"](https://eprint.iacr.org/2022/1530.pdf) uses the sumcheck
+//! protocol to compute the partial sums (equations 16 and 17). However, we use
+//! the PlonK polynomial IOP and therefore, we will use the quotient polynomial,
+//! and the computation of the partial sums will be translated into a constraint
+//! in a new power of alpha.
+//!
+//! Note that the inner sum h(X) can be constrainted as followed:
+//! ```text
+//!         k                   k  /          k                 \
+//! h(X) *  ᴨ  (β + f_{i}(X)) = ∑  | m_{i} *  ᴨ  (β + f_{j}(X)) |     (5)
```
volhovm, line 95: Here it should be `h(X) * \prod (\beta + \phi(i)(X))`, that is the product should include the `t`. Or you can say that `t` is just a zero (or k+1)th element alternatively.

volhovm (approved):
Equation 5 needs to be fixed, otherwise looks great!

volhovm:
Oh I was too late :|

msm/src/logup.rs
dannywillems: You're right, I noticed it and I was sure someone would see it too :D. I will modify it in https://github.com/o1-labs/proof-systems/pull/2087.

PR #2087, opened 2024-04-17
Logup constraints

dannywillems:


msm/src/serialization/witness.rs
```diff
@@ -181,10 +176,12 @@ impl<F: PrimeField> Env<F> {
     pub fn get_rangecheck4_multipliticies(&self, domain: EvaluationDomains<F>) -> Vec<F> {
         let mut m = Vec::with_capacity(domain.d1.size as usize);
         m.extend(self.lookup_multiplicities[&LookupTable::RangeCheck4].to_vec());
-        let repeated_dummy_value: Vec<F> = iter::repeat(-F::zero())
-            .take((domain.d1.size - (1 << 4)) as usize)
+        let n_repeated_dummy_value = domain.d1.size - (1 << 4) - 1;
```
volhovm, line 179: If I understand correctly you're using the last value as a dummy. Why not use the zero value as a dummy since it's anyway present in the range?

msm/src/logup.rs
```diff
@@ -561,115 +573,120 @@ pub mod prover {
...
+                assert!(denominators.len() == n * domain.d1.size as usize);
+
+                ark_ff::fields::batch_inversion(&mut denominators);
+
+                // Evals is the sum on the individual columns for each row
+                let mut denominator_index = 0;
+
+                // We only need to add the numerator now
+                for j in 0..domain.d1.size {
+                    let mut partial_sum_idx = 0;
```
volhovm, line 637: Highlighting for myself/posterity: indexing each `h`-chunk separately.

msm/src/logup.rs
```diff
@@ -561,115 +573,120 @@ pub mod prover {
...
+                            partial_sum_idx += 1;
                         }
                     }
-                    partial_sums
-                })
-                .collect::<Vec<_>>();
-
-            let lookup_terms_evals: Vec<Vec<G::ScalarField>> =
-                lookup_terms_evals.into_iter().flatten().collect();
+                    if n % (MAX_SUPPORTED_DEGREE - 2) != 0 {
```
volhovm, line 656:
```suggestion
                    // Whatever leftover in `row_acc` left in the end of the iteration, we write it into
                    // `partial_sums` too. This is only done in case `n % (MAX_SUPPORTED_DEGREE - 2) != 0`
                    // which means that the similar addition to `partial_sums` a few lines above won't be triggered.
                    // So we have this wrapping up call instead.
                    if n % (MAX_SUPPORTED_DEGREE - 2) != 0 {
```

volhovm (approved):

msm/src/serialization/witness.rs
dannywillems: Can be anything.

Issue #2074, opened 2024-04-15
Fix Keccak failing constraints


querolita:
I saw that the `PadAtEnd` and `ChiShiftsSum` constraints are not passing in the prover. Figure out why and fix it.

Issue #2075, opened 2024-04-15
Fix MIPS failing constraints


querolita:
When running the `main.rs` with the generic prover I could observe some MIPS constraints not passing. Figure out why and which and fix it.

querolita:
I have been trying to identify the instructions and concrete constraints that are failing:

- Inside `is_zero()` the constraint fails
```
self.add_constraint(x.clone() * x_inv_or_zero.clone() + res.clone() - Self::constant(1));
```

- `RType(JumpRegister)` kills the prover process 

- `RType(JumpAndLinkRegister)` kills the prover and fails in 
```
env.write_register(&rd, instruction_pointer + Env::constant(8u32));
```

- `RType(SyscallExitGroup)` kills the prover process 

- `RType(SyscallReadOther)` aborts and these constraints fail
```
let v0 = other_fd.clone() * Env::constant(0xFFFFFFFF);
let v1 = other_fd * Env::constant(0x9); // EBADF

env.write_register(&Env::constant(2), v0);
env.write_register(&Env::constant(7), v1);
```

- `RType(SyscallWriteOther)` kills the prover and fails in
```
let v0 = known_fd * write_length + other_fd.clone() * Env::constant(0xFFFFFFFF);
let v1 = other_fd * Env::constant(0x9); // EBADF

env.write_register(&Env::constant(2), v0);
env.write_register(&Env::constant(7), v1);
```

- `RType(SyscallFcntl)` fails in 
```
let v0 = is_getfl.clone()
          * (is_write.clone()
             + (Env::constant(1) - is_read.clone() - is_write.clone())
             * Env::constant(0xFFFFFFFF))
           + (Env::constant(1) - is_getfl.clone()) * Env::constant(0xFFFFFFFF);
let v1 =
         is_getfl.clone() * (Env::constant(1) - is_read - is_write.clone())
         * Env::constant(0x9) /* EBADF */
         + (Env::constant(1) - is_getfl.clone()) * Env::constant(0x16) /* EINVAL */;

env.write_register(&Env::constant(2), v0);
env.write_register(&Env::constant(7), v1);
```

- `RType(Sync)` kills the prover

- `RType(MoveToHi)` kills the prover

- `JType(Jump)` kills the prover

- `JType(Jump)` kills the prover

- `JType(JumpAndLink)` fails with
```
[...]

PR #2093, opened 2024-04-19
Use non-zero dummy rows to pad witnesses for testing

querolita:
After chatting with Matthew about the failing constraints on MIPS and Keccak, we narrowed down the issue to the padding rows. Basically, when there were explicit selectors multiplying each constraint to switch them on or off, it was clear that rows would have an heterogeneous behavior. And in particular, any extra row needed to pad the witness could just be all zeros because in any case the selector multiplicands would cancel out all constraints. But now that the witness is split into different circuits, each of them for the same behavior, these multiplicands have been removed (also decreasing degree of constraints as a nice consequence). However, I had forgotten about the padding rows. In this case now, if there's any constraint involving constants being added, then a zero row will not counteract those constants, thus making the quotient polynomial in the prover fail (not divisible by vanishing poly). 

This PR partially fixes that issue by substituting the type of padded rows: instead of just being zero rows, now these are replicating the first non-zero row of the witness. Meaning, as soon as you have an instance for one step type, the trait just copies the first row until reaching the desired domain size. As a consequence, I was able to uncomment the Keccak constraints that were failing (PadAtEnd and ChiSumShiftsB), both of them involving constants.

After looking again at https://github.com/o1-labs/proof-systems/issues/2075#issuecomment-2059745490, one can see that the same problem happens in those failing constraints -> there is a constant coefficient involved in the constraints. This PR should remove some of those problems. 

Closes https://github.com/o1-labs/proof-systems/issues/2074 and partially https://github.com/o1-labs/proof-systems/issues/2075

TODO: padding the witness for the constraint side is "easy". Also for fixed-table lookups. But as soon as there's any row performing actual RAM lookups, it won't be so trivial. That is because we don't want to add extra reads/writes that could break the equation. For that reason, upcoming PRs along these lines will follow.

NOTE: the diff might look too large, but it's mainly because I renamed the `Circuit` struct to something less confusing like `Trace`. So basically some file renames make the diff become larger than desired.


optimism/src/keccak/trace.rs
```diff
@@ -56,7 +56,26 @@ impl<F: Field> CircuitTrait<ZKVM_KECCAK_COLS, Steps, F, KeccakEnv<F>> for Keccak
         });
     }
 
-    fn pad(&mut self, step: Steps) -> bool {
+    fn pad_with_row(&mut self, step: Steps, row: &[F; ZKVM_KECCAK_COLS]) -> bool {
```
dannywillems, line 59: For push_row, line 52:
I would say what happens when the index is bigger. If it has to fail with an exception, it must be explicit.

optimism/src/keccak/trace.rs
```diff
@@ -56,7 +56,26 @@ impl<F: Field> CircuitTrait<ZKVM_KECCAK_COLS, Steps, F, KeccakEnv<F>> for Keccak
...
+        let mut step = step;
+        if let Round(_) = step {
+            step = Round(0);
+        }
+        let rows_left = self.domain_size - self.witness[&step].cols[0].len();
+        if rows_left == 0 {
```
dannywillems, line 65: Just in case as not explicit for the runtime here: what about `rows_left < 0`.

optimism/src/keccak/trace.rs
dannywillems: Could you also comments what these functions are supposed to do, please? The name, IMO, is not explicit enough. In particular, the condition `if let Round(_) = step` is not obvious.

optimism/src/trace.rs
```diff
@@ -0,0 +1,62 @@
+use crate::{lookups::Lookup, E};
+use ark_ff::Field;
+use kimchi_msm::witness::Witness;
+use std::{collections::HashMap, hash::Hash};
+
+pub struct Trace<const N: usize, SELECTOR, F> {
+    /// The domain size of the circuit
+    pub domain_size: usize,
+    /// The witness for a given selector
+    pub witness: HashMap<SELECTOR, Witness<N, Vec<F>>>,
```
dannywillems, line 10: Note: if you want to process in order the keys (and therefore the witness), you need to use `BTreeMap`. `keys` and `values` are not deterministic on `HashMap`, but are on `BTreeMap`

optimism/src/keccak/trace.rs
querolita: Never mind, I will add more comments and improve the readability of it.

optimism/src/trace.rs
```diff
@@ -0,0 +1,62 @@
...
+        self.witness[&step].cols[0].is_empty()
+    }
+
+    /// Resets the witness after folding
+    pub fn reset(&mut self, step: SELECTOR) {
+        self.witness.insert(
+            step,
+            Witness {
+                cols: Box::new(std::array::from_fn(|_| {
+                    Vec::with_capacity(self.domain_size)
```
dannywillems, line 29: Is there not a function to reset the vector instead of reallocating?

optimism/src/trace.rs
```diff
@@ -0,0 +1,62 @@
...
+                })),
+            },
+        );
+    }
+}
+
+pub trait Tracer<const N: usize, SELECTOR, F, Env> {
```
dannywillems, line 36: Might be obvious, but is `N` the number of columns, for the given selector? If yes, can you add a small comment please?

dannywillems: Would it be possible to document each type parameter please? A single line should be enough like:
```
A (F-)tracer builds traces for some program executions.
The type parameter `F` is the type the data points in the trace are encoded into. It can be a field or a native type (u64).
The type `SELECTOR` encodes the information of the kind of information the trace encodes. For instance, if `SELECTOR` is somehow `MIPS_ADD`, the trace could be interpreted as an encoding of the execution of the CPU instruction of a MIPS machine.
The type parameter `N` is defined as the number of columns/"registers" the trace can use per row.
...
```

optimism/src/trace.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Add a witness row to the circuit
+    fn push_row(&mut self, selector: SELECTOR, row: &[F; N]);
+
+    /// Pad the rows of one selector with the given row until
+    /// reaching the domain size if needed.
+    /// Returns true if padding was performed, false otherwise.
+    fn pad_with_row(&mut self, step: SELECTOR, row: &[F; N]) -> bool;
+
+    /// Pads the rows of one selector with zero rows until
```
dannywillems, line 48: Nit: should `F` be constrained to have a `zero` in its inhabitants? I know it might not be easy as it must be generic over fields and mative types like u64. But worth mentioning in the Tracer top-level description?

optimism/src/keccak/trace.rs
querolita: No need because it is a usize 😉

optimism/src/trace.rs
```diff
@@ -0,0 +1,62 @@
...
+
+pub trait Tracer<const N: usize, SELECTOR, F, Env> {
+    /// Create a new circuit
+    fn new(domain_size: usize, env: &mut Env) -> Self;
+
+    /// Add a witness row to the circuit
+    fn push_row(&mut self, selector: SELECTOR, row: &[F; N]);
+
+    /// Pad the rows of one selector with the given row until
+    /// reaching the domain size if needed.
```
dannywillems, line 44: Wondering if the domain size should be included in the type, in case we want the type checker verifying we do not "concatenate" circuits of different "lengths".

optimism/src/trace.rs
querolita: For now I don't care about the order because the main is creating one proof for each different witness type. But I agree with you that when real folding is involved and there's just one single final proof, then yes BTreeMap will be a better structure. But for now I wanted to keep it simple as hash maps, because this struct will surely be subject to change as soon as I try to integrate the interface for folding.

optimism/src/trace.rs
querolita: Mmmm probably there's a `clear`, will check that out

optimism/src/trace.rs
querolita: Yes, that's it. Probably given that SELECTOR is verbose, I can also rename N for COLUMNS instead.

optimism/src/trace.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn pad_with_row(&mut self, step: SELECTOR, row: &[F; N]) -> bool;
+
+    /// Pads the rows of one selector with zero rows until
+    /// reaching the domain size if needed.
+    /// Returns true if padding was performed, false otherwise.
+    fn pad_with_zeros(&mut self, step: SELECTOR) -> bool;
+
+    /// Pad the rows of one selector with the first row until
+    /// reaching the domain size if needed.
+    /// Returns true if padding was performed, false otherwise.
```
dannywillems, line 55: I would be more explicit in "was performed". What does it mean?
IIUC, it is performed when there are some "empty" rows.
If it is *only* filling the empty rows, maybe worth returning the number of actual rows that have been filled, the type being a usize, `0` being `false` in this case.

optimism/src/trace.rs
querolita: Mmmm okay well it won't hurt anybody if I configure the trait definition to force `F: Zero` in any case.

optimism/src/trace.rs
querolita: Oh, I hadn't thought of that... Well for now they are not being concatenated because they are being used independently to generate one separate proof for each variant. But I agree that it could be a good idea to keep in mind when it is concatenated for folding. Will open a follow-up issue.

querolita: https://github.com/o1-labs/proof-systems/issues/2096

optimism/src/keccak/tests.rs
```diff
@@ -499,28 +497,28 @@ fn test_keccak_prover() {
...
+        let mut keccak_circuit = KeccakTrace::<Fp>::new(domain_size, &mut keccak_env);
 
-        // We want to use domain_size rows, even if that is an incomplete Keccak execution
         while keccak_env.constraints_env.step.is_some() {
             let step = keccak_env.constraints_env.step.unwrap();
+
             // Run the interpreter, which sets the witness columns
             keccak_env.step();
-            // Check witness satisfies constraints
-            keccak_env.witness_env.constraints();
```
dannywillems, line 510: Why do you remove?

querolita: Because I have other tests checking the witness data satisfies constraints, and it seemed unnecessary to do it again. Here the important part is checking the prover can generate a proof for the constraints instead.

optimism/src/keccak/trace.rs
dannywillems: You need, even if it is a usize. https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=c6c1c1dafe3cffe462cf019c2216ecdc

optimism/src/keccak/trace.rs
```diff
@@ -56,7 +56,26 @@ impl<F: Field> CircuitTrait<ZKVM_KECCAK_COLS, Steps, F, KeccakEnv<F>> for Keccak
         });
     }
 
-    fn pad(&mut self, step: Steps) -> bool {
+    fn pad_with_row(&mut self, step: Steps, row: &[F; ZKVM_KECCAK_COLS]) -> bool {
+        let mut step = step;
+        if let Round(_) = step {
+            step = Round(0);
```
dannywillems, line 62: I don't understand this condition. Can you explain, please?

dannywillems (commented):
Minor comments. LGTM otherwise.

optimism/src/keccak/trace.rs
querolita: Added a comment, will push in a few min

dannywillems (approved):

PR #2095, opened 2024-04-19
IVC: Implement wide Poseidon circuit

fabrizio-m:


ivc/src/poseidon/constraint.rs
```diff
@@ -0,0 +1,71 @@
...
+            col,
+            row: CurrOrNext::Curr,
+        }))
+    }
+
+    fn constant(value: F) -> Self::Variable {
+        Exp::Atom(ExprInner::Constant(value))
+    }
+
+    fn mds(&self) -> &[[Self::Variable; S]; S] {
```
volhovm, line 60: I have a very different interpreter design in mind, in which all these constants are passed as a struct to the functions over the interpreter. Instead of being functions /of/ interpreter. With this design you have there's unnecessary code duplication instead, like you have literally the same stuff inside two instances of the interpreter. What do you think about the idea of moving them to `interpreter.rs`?

ivc/src/poseidon/interpreter.rs
```diff
@@ -0,0 +1,97 @@
...
+    for (i, p) in preimage.iter().enumerate() {
+        env.write(p, Column::Input(i));
+    }
+    poseidon_row(env)
+}
+
+/// creates the constraints and computes most of the witness for a row
+/// for witness it assumes the preimage is already written to the respective
+/// columns
+pub fn poseidon_row<F, const S: usize, const R: usize, E>(env: &mut E) -> [E::Variable; S]
```
volhovm, line 60: A test (including some non-random test asserting the computation matches concrete inputs-outputs) is very necessary I think. Check out this function for example: https://github.com/o1-labs/proof-systems/blob/master/msm/src/fec/mod.rs#L52

volhovm (commented):
Not blocking, but I think this a test (definitely), some docs (~definitely), and a slightly better interface (my personal preference, optional if you disagree).

dannywillems (changes requested):
See messages on Slack.

dannywillems (approved):

PR #2112, opened 2024-04-23
Change Poseidon circuit to map generic interpreter

dannywillems:
Change Poseidon circuit to use the generic interpreter.

- [x]  Instantiate with bn254
- [x] check parameters for bn254
  - [x] nb full rounds
  - [x] constants
  - [x] MDS
- [x] add test vectors for bn254
- [x] use generic interpreter


volhovm (approved):
Looks great! 

@dannywillems I pushed one commit that (1) moves code populating the wires into `interpreter.rs` and (2) adding a separate smaller/faster test for just spec consistency. Not merging myself thus so you can take a look at my changes in case you don't like them.

PR #2114, opened 2024-04-23
Generic interpreter: Split into capabilities & move into separate folder

volhovm:
* Creates a folder `circuit_designer` that now contains all "generic" code for designing circuits
* Splits the mono-trait "interpreter" into several "capabilities"
    * `ColAccessCap`, `ColWriteCap`, `LookupCap` for now. 
    * Future: Think also `RAMLookupCap`, `TransWireCommunicationCap` etc. The idea is that whenever one adds a "feature" (capability) to circuit building, it's a separate trait on a "interpreter environment". This makes things highly composable and (hopefully) future-proof.
* Also shows how "hacky" (IMO) helpers can be added for backwards compatibility -- example `bitmask_be` and `copy`.

Small bonus/side diff: I shrunk `deserialize_field_element` a bit. It was making things a bit easier.

### DIFF
--- a/msm/src/circuit_design/capabilities.rs
+++ b/msm/src/circuit_design/capabilities.rs
@@ -1,0 +1,47 @@
+/// Provides generic environment traits that allow manipulating columns and
+/// requesting lookups.
+///
+/// Every trait implies two categories of implementations:
+/// constraint ones (that operate over expressions, building a
+/// circuit), and witness ones (that operate over values, building
+/// values for the circuit).
+use crate::{columns::ColumnIndexer, logup::LookupTableID};
+use ark_ff::PrimeField;
+
+/// Environment capability for accessing and reading columns. This is necessary for
+/// building constraints.
+pub trait ColAccessCap<F: PrimeField, CIx: ColumnIndexer> {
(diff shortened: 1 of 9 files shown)


msm/src/circuit_design/capabilities.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn assert_zero(&mut self, cst: Self::Variable);
+
+    /// Reads value from a column position.
+    fn read_column(&self, ix: CIx) -> Self::Variable;
+
+    /// Turns a constant value into a variable.
+    fn constant(value: F) -> Self::Variable;
+}
+
+/// Environment capability similar to `ColAcessT` but for /also
```
dannywillems, line 32:
```suggestion
/// Environment capability similar to `ColAcessCap` but for /also
```

dannywillems (approved):
Interested in seeing where this goes after the other PRs. Nice.

msm/src/circuit_design/capabilities.rs
dannywillems: No opinion for me. My change was to stay consistent with the trait above.

PR #2115, opened 2024-04-24
Generic interpreter: add composition (lenses)

volhovm:
* Adds `composition.rs` which defines mechanics needed for calling one interpreter from another
* See https://github.com/o1-labs/proof-systems/pull/2036 for general POC end-to-end sketch
* Not adding "demo" as requested by @dannywillems https://github.com/o1-labs/proof-systems/pull/2036#pullrequestreview-2011009849
    * Can write some `README.md` later. The "demo" is still in that old branch, so can be restored any minute.
    * We'll start composing gadgets today anyway, which will be a real-world application to illustrate what's happening.

### DIFF
--- a/msm/src/circuit_design/composition.rs
+++ b/msm/src/circuit_design/composition.rs
@@ -1,0 +1,249 @@
+/// Tools to compose different circit designers.
+use crate::{
+    circuit_design::capabilities::{ColAccessCap, ColWriteCap, LookupCap},
+    columns::ColumnIndexer,
+    logup::LookupTableID,
+};
+use ark_ff::PrimeField;
+
+/// `MPrism` allows one to Something like a Prism, but for Maybe and not just any Applicative.
+///
+/// See
+/// - <https://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Prism.html>
+/// - <https://hackage.haskell.org/package/lens-tutorial-1.0.4/docs/Control-Lens-Tutorial.html>
+pub trait MPrism {
+    /// The lens source type, i.e., the object containing the field.
+    type Source;
+
+    /// The lens target type, i.e., the field to be accessed or modified.
+    type Target;
+
+    fn traverse(&self, source: Self::Source) -> Option<Self::Target>;
+
+    fn re_get(&self, target: Self::Target) -> Self::Source;
+}
+
+pub struct ComposedMPrism<LHS, RHS> {
+    /// The left-hand side of the composition.
+    lhs: LHS,
+
+    /// The right-hand side of the composition.
+    rhs: RHS,
+}
+
+impl<LHS, RHS> ComposedMPrism<LHS, RHS>
+where
+    LHS: MPrism,
+    LHS::Target: 'static,
+    RHS: MPrism<Source = LHS::Target>,
+{
+    pub fn compose(lhs: LHS, rhs: RHS) -> ComposedMPrism<LHS, RHS> {
+        ComposedMPrism { lhs, rhs }
+    }
+}
+
+impl<LHS, RHS> MPrism for ComposedMPrism<LHS, RHS>
+where
+    LHS: MPrism,
+    LHS::Target: 'static,
+    RHS: MPrism<Source = LHS::Target>,
+{
+    type Source = LHS::Source;
+    type Target = RHS::Target;
+
+    fn traverse(&self, source: Self::Source) -> Option<Self::Target> {
+        let r1: Option<_> = self.lhs.traverse(source);
+        let r2: Option<_> = r1.and_then(|x| self.rhs.traverse(x));
+        r2
+    }
+
+    fn re_get(&self, target: Self::Target) -> Self::Source {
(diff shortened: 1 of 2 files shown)


msm/src/circuit_design/composition.rs
```diff
@@ -0,0 +1,249 @@
...
+////////////////////////////////////////////////////////////////////////////
+// Trait implementations
+////////////////////////////////////////////////////////////////////////////
+
+impl<'a, F: PrimeField, CIx1: ColumnIndexer, Env1: ColAccessCap<F, CIx1>, L>
+    SubEnv<'a, F, CIx1, Env1, L>
+{
+    pub fn new(env: &'a mut Env1, lens: L) -> Self {
+        SubEnv {
+            env,
```
dannywillems, line 96: I'm worried about memory overhead (copying).

msm/src/circuit_design/composition.rs
```diff
@@ -0,0 +1,249 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+// Generic sub-environment struct. Internal object to avoid copy-paste.
+struct SubEnv<'a, F: PrimeField, CIx1: ColumnIndexer, Env1: ColAccessCap<F, CIx1>, L> {
+    env: &'a mut Env1,
+    lens: L,
+    phantom: core::marker::PhantomData<(F, CIx1)>,
+}
+
+/// Sub environment with a lens that is mapping lookup tables.
```
dannywillems, line 76: mapping lookup tables of columns?

msm/src/circuit_design/composition.rs
```diff
@@ -0,0 +1,249 @@
+/// Tools to compose different circit designers.
```
dannywillems, line 1:
```suggestion
/// Tools to compose different circuit designers.
```

msm/src/circuit_design/composition.rs
```diff
@@ -0,0 +1,249 @@
...
+    lhs: LHS,
+
+    /// The right-hand side of the composition.
+    rhs: RHS,
+}
+
+impl<LHS, RHS> ComposedMPrism<LHS, RHS>
+where
+    LHS: MPrism,
+    LHS::Target: 'static,
```
dannywillems, line 37: Wondering if it might cause an issue. Depending on the data living in Target, we might want to deallocate the value earlier than at the end of the program.

dannywillems (approved):
I would appreciate some drawings that define the different structures/categories and the links between them, just for a high-level understanding and to avoid having to go through the (nice) code every time to remember.

msm/src/circuit_design/composition.rs
```diff
@@ -0,0 +1,249 @@
...
+        'a,
+        F: PrimeField,
+        CIx1: ColumnIndexer,
+        CIx2: ColumnIndexer,
+        Env1: ColWriteCap<F, CIx1>,
+        L: MPrism<Source = CIx1, Target = CIx2>,
+    > ColWriteCap<F, CIx2> for SubEnv<'a, F, CIx1, Env1, L>
+{
+    fn write_column(&mut self, ix: CIx2, value: &Self::Variable) {
+        self.env.write_column(self.lens.re_get(ix), value)
```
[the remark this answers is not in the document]
dannywillems: I suppose yes, as under the hod, ColumnIndexer maps to columns `X`.

msm/src/circuit_design/composition.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/// Sub environment with a lens that is mapping lookup tables.
+/// Can't use `SubEnv` directly because rust is not idris.
+pub struct SubEnvColumn<'a, F: PrimeField, CIx1: ColumnIndexer, Env1: ColAccessCap<F, CIx1>, L>(
+    SubEnv<'a, F, CIx1, Env1, L>,
+);
+
+/// Sub environment with a lens that is mapping lookup tables.
+pub struct SubEnvLookup<'a, F: PrimeField, CIx1: ColumnIndexer, Env1: ColAccessCap<F, CIx1>, L>(
```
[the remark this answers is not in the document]
volhovm: Yes :) Rust is stupid -- ideally I would use just `SubEnv`, but because of trait limitations I can only capture one type of "lens" in one type of `SubEnv`. So there are two `SubEnv`s that are the same on the data level, but have different trait implementations.

msm/src/circuit_design/composition.rs
volhovm: Columns. Will fix.

msm/src/circuit_design/composition.rs
volhovm: I realised just now: there's no copying overhead, we already store a mutable reference in `SubEnv` if that's what you meant?

msm/src/circuit_design/composition.rs
volhovm: Hm. I'll add a comment, but can't comment otherwise now. My hope is that building a circuit will not require more than a (dozen at most!) lenses, so this should be super cheap.

PR #2121, opened 2024-04-24
Folding: Move to a separate crate

fabrizio-m:


folding/src/expressions.rs
```diff
@@ -68,7 +66,18 @@ pub enum ExpExtension {
 }
 
 ///Internal expression used for folding, simplified for that purpose
-pub type FoldingExp<C> = Operations<ExtendedFoldingColumn<C>>;
```
volhovm, line 71: IIUC this change is not because of necessity..? You could've left it as it was, but decided to change for better future compatibility?

volhovm (approved):
Super nice!

volhovm:
Needs to be rebased indeed :(

volhovm:
@fabrizio-m I rebased. It seems to compile & pass tests, so let's merge & clean up smaller details after the merge?

volhovm:
@fabrizio-m @querolita I hope that I didn't break anything with this -- could you please double check? 

I'm also a bit confused that `kimchi/src/snarky/folding/instance.rs` is still in `kimchi`?..

fabrizio-m:
`kimchi/src/snarky/folding/instance.rs` is the old ivc implementation based on cycles of curves.

PR #2122, opened 2024-04-24
Allow circuits to not have any lookup

dannywillems:


volhovm (approved):
Nice! Actually some code was failing before when a table was declared but not used. Not sure if it's still present, would have to check.

PR #2123, opened 2024-04-24
ColWriteCap: implement copy

dannywillems:


volhovm (changes requested):
Can we discuss this during today's call? I was thinking about `copy` for a while and I'm not exactly sure it belongs to any of these traits. It can perhaps be in a different trait? Ultimately I want to avoid having methods that are "no-op" in one of the two impls. Right now my plan was to get rid of it as much as possible, but let's discuss.

volhovm:
Also: there is already a trait that implements `copy`, it's in `serialization/interpreter.rs` and called something like `SerializationHelpers`.

dannywillems:
Setting in draft for now. We will discuss later, and @volhovm has to convince me that is problematic (:beer: on bet).

dannywillems:
Copy should be replaced by read and write.

volhovm:
Thanks Danny :) sure. My claim is: using copy is legit, but it doesn't belong in any of the three existing traits, and style of writing circuits with copy is "orthogonal" to the style that is currently written. You can do either read/write or you can do copy. For now (until next meeting) I suggest to avoid copy, and write `ColAccessCap` functions separately from `ColWriteCap` functions where possible. I'm also fine with it being a separate trait if one wants to experiment.

volhovm (approved):
Accepting my own edits..........

dannywillems:
Thanks for noticing it.
It seems that `copy` is kind of `unsafe`, as it might break some invariants/does not fit in the general semantics.
I agree with the separate trait. What do you think about renaming `hcopy` into `copy`, but flag it as `unsafe` using `unsafe fn copy(...)`, and describe the "unsafe" behavior in a comment using your example?

dannywillems:
If you don't like `copy` as a name, we could have `unsafe_copy`, and still flag with the Rust keyword `unsafe`.

dannywillems:
Merging to continue working on Poseidon. Let's follow up in https://github.com/o1-labs/proof-systems/issues/2131

volhovm:
>What do you think about renaming hcopy into copy, but flag it as unsafe using unsafe fn copy(...), and describe the "unsafe" behavior in a comment using your example?

I feel like this approach does not quite capture the point of the method, or I don't understand why you think `hcopy` is "unsafe". I think it's not unsafe, it's just a "parallel hierarchy" to the Write trait. Methods which semantics is not "for sure" but "optional". So I kind of liked "hybrid" as a word. 

"Unsafe" to me seems somehow much more scary, like that calling this method might break something very deep on the compiler level/segfault or anything. I think `hcopy` is quite alright, and can be used, and is safe -- the only precaution is that people don't mentally think of it as "copy" because it's not, it's hybrid. Renaming it into "passthrough" or "link_two" might do the same job.

I'm a bit allured by "unsafe" though, maybe if you can explain why you think it applies to `hcopy` I'd be more happy with the name :)

PR #2124, opened 2024-04-25
MSM FEC: Use labelled columns (easy)

volhovm:
Before (for experimentation reasons) the columns in the FEC module were integer-based, so direct memory access by i:usize. Now the alias FECColumn explicitly names each column, which improves readability and safety.

+ it adds some helpers for convenience.

### DIFF
--- a/msm/src/circuit_design/capabilities.rs
+++ b/msm/src/circuit_design/capabilities.rs
@@ -51,16 +51,16 @@ where
 ////////////////////////////////////////////////////////////////////////////
 
 /// Write an array of values simultaneously.
-pub fn read_column_array<F, Env, const ARR_N: usize, CIx: ColumnIndexer, Func>(
+pub fn read_column_array<F, Env, const ARR_N: usize, CIx: ColumnIndexer, ColMap>(
     env: &mut Env,
-    column_foo: Func,
+    column_map: ColMap,
 ) -> [Env::Variable; ARR_N]
 where
     F: PrimeField,
     Env: ColAccessCap<F, CIx>,
-    Func: Fn(usize) -> CIx,
+    ColMap: Fn(usize) -> CIx,
 {
-    core::array::from_fn(|i| env.read_column(column_foo(i)))
+    core::array::from_fn(|i| env.read_column(column_map(i)))
 }
 
 /// Write a field element directly as a constant.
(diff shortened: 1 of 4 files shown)


msm/src/circuit_design/capabilities.rs
```diff
@@ -45,3 +45,59 @@ where
 {
     fn lookup(&mut self, lookup_id: LT, value: &Self::Variable);
 }
+
+////////////////////////////////////////////////////////////////////////////
+// Helpers
+////////////////////////////////////////////////////////////////////////////
+
+/// Write an array of values simultaneously.
+pub fn read_column_array<F, Env, const ARR_N: usize, CIx: ColumnIndexer, Func>(
```
dannywillems, line 54: Should it not be in the trait `ColAccessCap`? Same for the functions below (into `ColWriteCap`).

msm/src/circuit_design/capabilities.rs
```diff
@@ -45,3 +45,59 @@ where
...
+    env: &mut Env,
+    column_foo: Func,
```
dannywillems, line 56: I suppose `column_mapping`?

msm/src/fec/interpreter.rs
```diff
@@ -583,11 +511,22 @@ pub fn ec_add_circuit<
             F::from_biguint(&carry_f).unwrap()
         };
 
-        let assign_carry = |env: &mut Env, newcarry: F, carryvar: &mut F, extra_offset: usize| {
+        fn assign_carry<F, Env, Foo>(
```
dannywillems, line 514: Foo -> `ColumnMapping`? or something close to the description?

dannywillems (approved):
I approve, but I think it is worth discussing the comments.

msm/src/circuit_design/capabilities.rs
volhovm: Big question. I prefer this style for now because then people outside of our library can define these methods. It seems generally more extendable -- you don't have to modify the trait to add another helper. Maybe they can be in a `capabilities::caphelpers` module?

msm/src/circuit_design/capabilities.rs
volhovm: Renamed into `column_map` here and elsewhere.

msm/src/fec/interpreter.rs
volhovm: Yep, renamed into `ColMap`, much better than `Foo` :)

msm/src/circuit_design/capabilities.rs
volhovm: https://stackoverflow.com/questions/58157367/how-can-i-add-extension-methods-to-a-trait-with-associated-types-lying-in-a-diff

Issue #2039, opened 2024-03-26
I think it'd be nice to have a separate `struct WitnessColumn(usize)` for many purposes where we don't actually care about lookup columns (like writing witness values in the interpreter), and this `X` then should be definitely renamed into something ... meaningful. What do you think?


dannywillems:
I think it'd be nice to have a separate `struct WitnessColumn(usize)` for many purposes where we don't actually care about lookup columns (like writing witness values in the interpreter), and this `X` then should be definitely renamed into something ... meaningful. What do you think?

_Originally posted by @volhovm in https://github.com/o1-labs/proof-systems/pull/2031#discussion_r1535993380_

volhovm:
Woah this is big.

volhovm:
@dannywillems wait, but unless I missed something, how is it closed? I see `Column::Relation(usize)` but it's not a separate struct. One still has to do this:
```
        let Column::Relation(i) = ix.to_column() else {
            todo!()
        };
```

PR #2126, opened 2024-04-25
Separate selector columns from aliases in MIPS and Keccak

querolita:
This PR is meant to separate the selector columns completely from the column aliases in MIPS and Keccak. As part of this task, plenty of indexer-styled traits had to be modified so that the mapping from alias/selector to witness column and generic column is performed correctly. 

A major modification implied by this PR is the split of the `X` generic function into `Relation` and `DynamicSelector`. The reason for doing this is mainly inspired by the folding example by Fabrizio, where it seems to indicate that actual witness columns and selectors should be given as separate realities, even if the underlying witness does have space for all of them. A consequence of this was having to pass not just one but three generic constants to some types of the generic prover. This is meant to allow these functions to distinguish between the relation columns and the dynamic selector columns (when non-literal constant generics are supported, this should be 2 instead and then N is just the sum of both).

I modified the tracer object for testing in a way that throughout computation, only relation columns are updated. And then only at the very end, when the circuit is going to be passed to folding, the selector columns get instantiated as columns of all zeros or ones. 

As a side effect, I had to change the way the Keccak current step was being stored. That is because before, the witness columns for selectors were being created in the execution of the witness itself (with the `step()` function). That made me remove the field `step` inside the constraints environment and move it to the Keccak environment itself. Together with that change, I had to pass the desired step to the functions in charge of returning the lists of constraints and lookups, for a specific step (this makes a big diff, but just a refactor). This was not a problem in the MIPS side because the columns for selectors were not being saved in execution in the first place.

I recommend a per-commit review, as I tried to explain in the message what each change is meant to provide. 

As a bonus, this PR closes https://github.com/o1-labs/proof-systems/issues/2039


dannywillems:
Taking care of merging master and fixing conflicts now.

dannywillems:
> Taking care of merging master and fixing conflicts now.

Done. Reviewing code.

dannywillems (approved):
I see that we assign a single boolean value for each selector, and we interpret them as columns. We could have a single integer representing the selectors, and reason on this selector.
We can also assign a prime number to each selector, and when we want to compose two selectors, we can use the product of it to have a unique representation of the decomposition. It reduces the the memory footprint.

dannywillems (approved):
I approve and merge as we can always simplify by using N_SEL = 1.

dannywillems:
Will merge after CI, had to solve conflicts by merging master.

querolita:
> We can also assign a prime number to each selector, and when we want to compose two selectors, we can use the product of it to have a unique representation of the decomposition. It reduces the the memory footprint.

Actually I intentionally created a separate selector for each different behavior in this iteration. Some weeks ago, "modes" used to be a composable type coming from the product of multiple more basic selectors. But the problem of that approach is that I believe it would increase the degree significantly.

PR #2128, opened 2024-04-25
Decomposable folding test with Keccak steps

querolita:
Work in progress, the test does nothing for now.

As part of this PR I realized of a few things that needed to be implemented for the tests to succeed with our current design:

- Further `From` implementations needed to happen so that our zkVM expressions (`E<F>` which are actually `Operations<ExprInner<Operations<ConstantExprInner<F>>, Column>>`
- If we want to keep using generic columns inside the constraints returned by the zkVM interpreters (which we did in order to be able to use the generic prover) then we must adapt the folding structs accordingly so that they work over the generic columns instead of the aliases for each circuit. This means I had to implement some other traits for it.


dannywillems:
Note:
We have `Operations<ExprInner<Operations<ConstantExprInner<F>>, Column>>` because we define:
`type E<F> = Operations<ExprInner<ConstantExpr<F>, Column>>`
And `ConstantExpr<F>` is also defined using the Operations type, because we can make operations on constant terms like add, mul, etc.

dannywillems:
I'm making this PR ready if you agree @querolita. I will merge this PR https://github.com/o1-labs/proof-systems/pull/2154/ into this one. I need some of your changes for https://github.com/o1-labs/proof-systems/pull/2146

PR #2044, opened 2024-03-27
Cargo fmt: Automatic imports sorting

volhovm:
This adds rustfmt for the repo + `imports_granularity = "Crate"`

Since we build our project w.r.t. a stable rust release, you need to rust `rustfmt` with nightly explicitly in your editor.
- On emacs, do this: `(setq rustic-rustfmt-args "+nightly")`
- On VSCode: modify `rustFmtArgs` https://github.com/rust-lang/rust-analyzer/issues/3627

PR #2129, opened 2024-04-26
Quadriticization fixes

fabrizio-m:
Solves a few with quadraticization, also simplifies some code and removes generics from `fold_instance_witness_pair`, now it can be called with just inference.


volhovm:
@fabrizio-m I'd suggest you add `cargo +nightly fmt` into your VSCode automatic settings (https://github.com/o1-labs/proof-systems/pull/2044), it should take <1min.

folding/src/quadraticization.rs
```diff
@@ -90,11 +90,23 @@ fn lower_degree_to_1<C: FoldingConfig>(
     let degree = exp.degree();
     match degree {
         1 => exp,
-        _ => {
-            let exp = lower_degree_to_2(exp, rec);
-            let id = rec.get_id(exp);
-            FoldingExp::Atom(ExtendedFoldingColumn::WitnessExtended(id))
-        }
+        _ => match exp {
+            FoldingExp::Add(e1, e2) => FoldingExp::Add(
```
dannywillems, line 94: Do you have any example testing the output of quadraticization of simple expressions, like `X1 * X2`, `X1 * X2 * (X3 + X4)`, etc?

folding/src/expressions.rs
```diff
@@ -56,6 +57,67 @@ pub enum FoldingCompatibleExpr<C: FoldingConfig> {
...
+                    c.to_string()
+                }
+                FoldingCompatibleExprInner::Challenge(c) => {
+                    format!("{:?}", c)
+                }
+                FoldingCompatibleExprInner::Cell(cell) => {
+                    let Variable { col, row } = cell;
+                    let next = match row {
+                        CurrOrNext::Curr => "",
+                        CurrOrNext::Next => " * ω",
```
mrmr1993, line 79: This is technically incorrect: we're getting `col(x)` for `Curr` and `col(x * \omega)` for `Next`. Might be easier to write simply `Curr` or `Next`.

dannywillems (approved):
Merging it, and see following PR which adds some documentations and maybe additional tests.

PR #2132, opened 2024-04-29
Generate Poseidon parameters for the scalar field of BN254

dannywillems:


ivc/src/poseidon/params.rs
```diff
@@ -0,0 +1,840 @@
+use ark_bn254::Fr as Fp;
+use mina_poseidon::poseidon::ArithmeticSpongeParams;
+use once_cell::sync::Lazy;
+
+// FIXME: move into mina_poseidon when this code is ready to go into production.
```
volhovm, line 5: Wasn't there already a file like that somewhere else? I can't find it but I have a feeling I already saw a huge file with Poseidon params at some point before in my life...

volhovm (approved):

ivc/src/poseidon/params.rs
dannywillems: Yes, it is already in mina_poseidon for Pasta curves.
I prefer to not move it there yet, as I would like to check twice the parameters before going into prod.

PR #2133, opened 2024-04-29
Regression tests for poseidon over bn254

dannywillems:
Test with `cargo nextest run "poseidon::params::tests::test_poseidon" --release`


volhovm (approved):

volhovm:
(again, I have a feeling I already saw it in some different crate... am I hallucinating or it exists somewhere else, and we could unify the two?)

dannywillems:
> (again, I have a feeling I already saw it in some different crate... am I hallucinating or it exists somewhere else, and we could unify the two?)

Answered in the other PR

PR #2135, opened 2024-04-30
IVC: Sketch column structure

volhovm:


ivc/src/ivc/columns.rs
```diff
@@ -0,0 +1,141 @@
...
+///      |  ...??? empty space?
+///      |     reuse w/ selectors?
+/// 2^15 |---------------
+///```
+///
+/// Counting cells:
+/// - Inputs:              2 * 17 * 3N = 102N
+/// - Inputs repacked 75:  2 * 4 * 3N = 24N
+/// - Inputs repacked 150: 2 * 2 * 3N = 12N
+/// - Hashes:              2 * 165 * 3N = 990N (max 4 * 165 * 3N if we add 165 constants to every call)
```
fabrizio-m, line 40: we also need to hash the challenges, but they are not that significant.

ivc/src/ivc/columns.rs
```diff
@@ -0,0 +1,141 @@
...
+
+// NB: We can reuse hash constants.
+// TODO: Can we pass just one coordinate and sign (x, sign) instead of (x,y) for hashing?
+#[derive(Debug, Clone, PartialEq)]
+pub enum IVCColumn {
+    /// 2*17 15-bit limbs (two base field points)
+    Input(usize),
+    /// 2*4 75-bit limbs
+    InputRepacked75(usize),
+    /// 2*2 150-bit limbs
```
fabrizio-m, line 95: do we need both 75 and 150?

fabrizio-m (approved):

ivc/src/ivc/columns.rs
volhovm: Yep, you're right. I want to move somewhat iteratively. There are lots of things missing.

ivc/src/ivc/columns.rs
volhovm: Yeah. 150 is as much as we can do for hashing, and 75bit is maximum that FEC ADD is working with... but it's cheap to have both, just a few extra rows. Not hundreds :)

PR #2136, opened 2024-04-30
Poseidon: separate generic builder + add a light circuit test

volhovm:
* Moves the code that writes Poseidon circuit wires into a separate `ColWriterCap`-type function
* Creates an additional lighter test just to verify if the circuit is correct without doing full prove/verify

### DIFF
--- a/ivc/src/poseidon/interpreter.rs
+++ b/ivc/src/poseidon/interpreter.rs
@@ -13,7 +13,7 @@
 
 use crate::poseidon::columns::PoseidonColumn;
 use ark_ff::{FpParameters, PrimeField};
-use kimchi_msm::circuit_design::{ColAccessCap, HybridCopyCap};
+use kimchi_msm::circuit_design::{ColAccessCap, ColWriteCap, HybridCopyCap};
 use num_bigint::BigUint;
 use num_integer::Integer;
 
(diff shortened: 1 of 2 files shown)


ivc/src/poseidon/interpreter.rs
```diff
@@ -28,6 +28,48 @@ pub trait Params<F: PrimeField, const STATE_SIZE: usize, const NB_FULL_ROUNDS: u
...
+pub fn poseidon_circuit<
+    F: PrimeField,
+    const STATE_SIZE: usize,
+    const NB_FULL_ROUND: usize,
+    PARAMETERS,
+    Env,
+>(
+    env: &mut Env,
+    param: PARAMETERS,
+    init_state: (F, F, F),
```
dannywillems, line 41: That should be a list of the state size to stay consistent with the whole design.

ivc/src/poseidon/interpreter.rs
```diff
@@ -28,6 +28,48 @@ pub trait Params<F: PrimeField, const STATE_SIZE: usize, const NB_FULL_ROUNDS: u
     fn mds(&self) -> [[F; STATE_SIZE]; STATE_SIZE];
 }
 
+/// Populates and checks one poseidon invocation.
+pub fn poseidon_circuit<
```
dannywillems, line 32: I would call it `poseidon_permutation`.

dannywillems (commented):
See comments. I see it is nearly the same than the other PR. I would not expose poseidon_circuit as said previously, it must not be public. Only `apply_permutation` should be.

dannywillems (approved):

PR #2139, opened 2024-05-01
IVC Circuit (Part 1): Add data repacking

volhovm:
Add IVC circuit code related to populating and verifying input columns in several different limb input types.

PR #2140, opened 2024-05-01
IVC Circuit (Part 2): Use FEC sub-lookup table

volhovm:
All the foreign EC Addition lookups are used in the IVC circuit. It's easy to reuse lookup tables since we now have lenses. 

This PR does this: removes copy-paste lookup table definition, and uses FEC sub-table instead.

It didn't work right away because `IntoEnumIterator` only iterates over the top-level, while we want full iterator, going through every leaf. So I added a method into lookup table interface for that and removed dependency on `IntoEnumIterator`.

Previous PR: https://github.com/o1-labs/proof-systems/pull/2139


dannywillems (approved):

dannywillems:
Good for the future to generalize to logup table libraries

PR #1401, opened 2023-12-05
Remove unit tests from zkVM interpreter

mrmr1993:
The existing tests were brittle. Removing them for now.

This PR builds on https://github.com/o1-labs/proof-systems/pull/1389 so that I don't have to fix a merge conflict once that lands.

PR #2145, opened 2024-05-02
Reintroduce some tests in the zkVM interpreter.

dannywillems:
Ignore atm. Removed in https://github.com/o1-labs/proof-systems/pull/1401. Can be used later to verify true assignments after folding.


fabrizio-m (approved):

dannywillems:
Investigating the error raised. Weird beucase it is an error raised in debug mode normally.

dannywillems:
> Investigating the error raised. Weird beucase it is an error raised in debug mode normally.

I see, it is because the CI executes with the flags `RUSTFLAGS="-Coverflow-checks=y -Cdebug-assertions=y"`

PR #2150, opened 2024-05-02
Vendored depedencies

joseandro:
PR for vendoring Rust dependencies


dannywillems (approved):
Approving the changes. I tested locally, and it does what it is expected.
However, I would not merge this before:
- a full integration in Mina
- keeping in mind (in a follow-up issue maybe) that we must have CI checks that the submodule tracks the same branch (commit must be an ancestor). Maybe this will be solved post hard-fork.

joseandro:
Here's the mina repository's PR with the full integration.

dannywillems (approved):
LGTM. Tested on the Mina branch.

PR #2157, opened 2024-05-03
WIP Folding: use Sponge trait from Kimchi

dannywillems:
Note: might not be that easy as challenges must be continuous and the sponge state as to be propagated through folding instances.


dannywillems:
Closing in favor of https://github.com/o1-labs/proof-systems/pull/2180.

PR #2158, opened 2024-05-03
Folding: reorganize examples + add explanation for the user

dannywillems:
This can improved later. Maybe good already to merge a first iteration?

The initial idea was to get rid of duplicated code and improve my understanding of the implementation. If we want to expose the library later, it is good to have this kind of structure.
It can be mixed later with the generic interpreter/prover/verifier.

### DIFF
--- a/folding/src/examples/example.rs
+++ b/folding/src/examples/example.rs
@@ -8,13 +8,13 @@
 /// ```
 use crate::{
     error_term::Side,
-    examples::generic::{Alphas, BaseSponge, Column, Curve, Fp},
+    examples::generic::{Alphas, BaseSponge, Checker, Column, Curve, Fp, Provide},
     expressions::FoldingCompatibleExprInner,
     ExpExtension, FoldingCompatibleExpr, FoldingConfig, FoldingEnv, Instance, RelaxedInstance,
     RelaxedWitness, Witness,
 };
 use ark_ec::{AffineCurve, ProjectiveCurve};
-use ark_ff::{Field, One, UniformRand, Zero};
+use ark_ff::{One, UniformRand, Zero};
 use ark_poly::{Evaluations, Radix2EvaluationDomain};
 use itertools::Itertools;
 use kimchi::circuits::{expr::Variable, gate::CurrOrNext};
@@ -236,7 +236,7 @@ fn circuit() -> [Vec<Fp>; 2] {
 /// check row by row for a zero result.
 mod checker {
     use super::*;
-    use log::debug;
+
     pub struct Provider {
         structure: TestStructure<Fp>,
         instance: TestInstance,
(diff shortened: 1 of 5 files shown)


folding/src/examples/example.rs
```diff
@@ -196,17 +110,20 @@ impl FoldingEnv<Fp, TestInstance, TestWitness, TestColumn, TestChallenge, ()> fo
...
             CurrOrNext::Curr => &self.curr_witnesses[side as usize],
             CurrOrNext::Next => &self.next_witnesses[side as usize],
         };
         match col {
-            TestColumn::A => &wit[0].evals,
-            TestColumn::B => &wit[1].evals,
-            TestColumn::C => &wit[2].evals,
-            TestColumn::SelecAdd => &self.structure.s_add,
-            TestColumn::SelecMul => &self.structure.s_mul,
+            Column::X(0) => &wit[0].evals,
```
querolita, line 119: Oh I thought it was called `Register` now

folding/src/examples/example.rs
```diff
@@ -432,73 +351,22 @@ mod checker {
...
-            for (i, row) in res.iter().enumerate() {
-                if !row.is_zero() {
-                    panic!("check in row {i} failed, {row} != 0");
-                }
-            }
-        }
-    }
-    impl<T: Provide> Checker for T {}
+    impl Checker<TestFoldingConfig> for Provider {}
+    impl Checker<TestFoldingConfig> for ExtendedProvider {}
```
querolita, line 355: Cool

folding/src/examples/generic.rs
```diff
@@ -0,0 +1,180 @@
...
+pub type BaseSponge = DefaultFqSponge<ark_bn254::g1::Parameters, SpongeParams>;
+
+// 1. We continue by defining a generic type of columns and selectors.
+// The selectors can be seen as additional (public) columns that are not part of
+// the witness.
+// The column must implement the trait [Hash] as it will be used by internal
+// structures of the library.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum Column {
+    X(usize),
```
querolita, line 39: Oh I see that's why

folding/src/examples/generic.rs
```diff
@@ -0,0 +1,180 @@
...
+// 2. We implement the trait [FoldingColumnTrait] that allows to distinguish
+// between the public and private inputs, often called the "instances" and the
+// "witnesses".
+// By default, we consider that the columns are all witness values and selectors
+// are public.
+impl FoldingColumnTrait for Column {
+    fn is_witness(&self) -> bool {
+        match self {
+            Column::X(_) => true,
+            Column::Selector(_) => false,
```
querolita, line 52: Actually according to Fabrizio, the selectors we will be using in the real examples of MIPS and Keccak should have Selectors also set to true in terms of witness, but I guess for these toy examples it is fine to assume they are false.

folding/src/examples/generic.rs
```diff
@@ -0,0 +1,180 @@
...
+// 3. We define the combinators that will be used to fold the constraints,
+// called the "alphas".
+// The alphas are exceptional, their number cannot be known ahead of time as it
+// will be defined by folding.
+// The values will be computed as powers in new instances, but after folding
+// each alpha will be a linear combination of other alphas, instand of a power
+// of other element. This type represents that, allowing to also recognize
+// which case is present.
+#[derive(Debug, Clone)]
+pub enum Alphas {
```
querolita, line 66: Awesome, let's have this somewhere more accessible than before.

querolita (approved):
We needed this, great job

PR #2167, opened 2024-05-06
Folding/Keccak: makes structures public to use it in other crates

dannywillems:
@querolita will certainly need it


querolita (approved):
Thanks 😄

PR #2164, opened 2024-05-03
WIP Fold in o1VM

dannywillems:
Should be generalized for any folding config.
`FoldingInstance`, `FoldingWitness`, etc should be built there. Will be for Monday.

PR #2169, opened 2024-05-06
Implement `Folder` trait for `Trace` to improve compatibility with folding types

querolita:
This could be used inside the `fold()` in `msm/proof.rs` as well as in tests.


dannywillems:
I prefer this idea than the one I was trying in https://github.com/o1-labs/proof-systems/pull/2164. I will continue with these changes @querolita

optimism/src/trace.rs
```diff
@@ -75,6 +83,66 @@ impl<
...
+            .witness
+            .cols
+            .iter()
+            .map(|w| srs.commit_evaluations_non_hiding(domain, w))
+            .map(|c| c.elems[0])
+            .collect_vec()
+            .try_into()
+            .unwrap();
+
+        // FIXME: this would need the sponge instead to obtain the challenges from
```
dannywillems, line 130: I'm taking care of this now. Will be in another PR.

folding/src/checker.rs
```diff
@@ -0,0 +1,342 @@
+//! A kind of pseudo-prover, will compute the expressions over the witness a check row by row
```
dannywillems, line 1: Is it expected in this PR?

optimism/src/trace.rs
dannywillems: It seems that the randomness is not carefully handled in the folding library. It requires to change the library (see conversation on Slack). Taking care of it.

folding/src/checker.rs
querolita: Oh shoot, why is this commit in this PR... No, this is meant to be in https://github.com/o1-labs/proof-systems/pull/2181

optimism/src/trace.rs
querolita: Will fix it, give me a minute

dannywillems (approved):

PR #2171, opened 2024-05-06
Trace/folding: avoid clone + use parallel

dannywillems:
And use abstraction of Witness structure


optimism/src/trace.rs
```diff
@@ -108,14 +108,10 @@ impl<
...
-            witness: Witness {
-                cols: Box::new(
-                    self.witness[&selector]
-                        .cols
-                        .clone()
-                        .map(|col| Evaluations::from_vec_and_domain(col, domain)),
-                ),
-            },
+            witness: (&self.witness[&selector])
+                .into_par_iter()
```
querolita, line 112: I had forgotten we had this one to be honest...

querolita (approved):

PR #2180, opened 2024-05-06
Handle sponge in folding properly

dannywillems:
Note: the PR does not aim to check that all the randomness is handled properly. I only replace the sponge trait by an appropriate instance of a sponge.


querolita (approved):
Awesome

PR #2181, opened 2024-05-06
Generic implementations for `Check` and `Provide`

querolita:
The goal of this PR is to make the testing tooling of folding a bit more generic so it involves less code duplication. 

The main changes proposed here are:
- Single definition of Alphas (will fix the possible conflict with `master` now) 
- Generic definition of `Provider` and `ExtendedProvider` structs
- Generic implementation of `Check` and `Provide` directly in the folding crate.
- Adapt trait bounds accordingly to allow this: 
  - make `Witness` and `Instance` of `FoldingConfig` clonable, 
  - implement `Index` for `FoldingInstance` with `Challenge`
  - include an `alphas()` function inside the `Instance` trait


folding/src/lib.rs
```diff
@@ -262,16 +266,16 @@ impl<'a, CF: FoldingConfig> FoldingScheme<'a, CF> {
 /// of other element. This type represents that, allowing to also recognize
 /// which case is present.
 #[derive(Debug, Clone)]
-pub enum Alphas<F: Field> {
-    Powers(F, Rc<AtomicUsize>),
-    Combinations(Vec<F>),
+pub enum Alphas<G: AffineCurve> {
```
dannywillems, line 269: Why not keeping the field?

optimism/src/keccak/mod.rs
```diff
@@ -10,6 +10,7 @@ use kimchi::circuits::polynomials::keccak::constants::{
 pub mod column;
 pub mod constraints;
 pub mod environment;
+#[cfg(feature = "bn254")]
```
dannywillems, line 13: Can be removed.

folding/src/checker.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn is_witness(&self) -> bool {
+        match self {
+            Column::X(_) => true,
+            Column::Selector(_) => false,
+        }
+    }
+}
+
+// TODO: get rid of trait Sponge in folding, and use the one from kimchi
+impl Sponge<Curve> for BaseSponge {
```
dannywillems, line 63: Note that it will be removed in https://github.com/o1-labs/proof-systems/pull/2180/

dannywillems (commented):
I only have a strong opinion on the Alphas parametrized by the curve.
Otherwise, the code is only touching examples.
In the near future, I see the examples being replaced by the generic prover/verifier. Therefore, I don't mind the code moving.

folding/src/lib.rs
querolita: Oh, that's just a version I had so that the types of the generic checker would match. But now I am trying to switch it for field as it was in your PR and it also seems to compile just fine. So I will set it to field, as it makes more sense that way.

optimism/src/keccak/mod.rs
querolita: But if I remove it then it does not find `checker` in `folding`.

dannywillems: Fixed in 2182

dannywillems (approved):

PR #2182, opened 2024-05-07
Finish decompsable folding Keccak test

querolita:
The goal of this PR is to fold one Keccak instruction in a test. In particular, it is folding `Sponge(Absorb(Only))` circuits together. For this to work, I had to make the following changes to the codebase, which are mostly collateral:
- For the domain size to be correctly spread out to the folding library, the `zero_vec()` vector inside `FoldingEnv` needed to be dynamically configurable. Otherwise, it was taking $2^{15}$ by default as the number of rows. 
  - In particular, I realized that the `rows()` function inside `FoldingConfig` does not talk at all to the `zero_vec()` function, which was producing this mismatch. 
  - Moreover, the `rows()` function was only being used by the examples, and never inside the actual code of folding. Thus, it looked quite superficial and I proceeded to remove it.
  - Instead, I created a `domain_size()` inside `FoldingEnv` that can be used to obtain the number of rows in `zero_vec()`. 
  - This new function works thanks to the use of the structure. It used to be just `()` as we were not using it before. But after talking to @dannywillems, it made sense that we could use the `Trace` inside the optimism folding types. Thanks to that, we can just access the domain size that was used in the first place when initializing the trace used in the tests.
  - Apart from this, I created a `rows()` function inside the `Provider` trait that simply looks at the length of the witness to obtain this value (before, it was hardcoded in the type).
- In order to avoid some clones in the user-facing functions, I made the `Structure` clonable and it is passed as a reference instead in the inner functions of folding.
- I made the `FoldingEnvironment` optimism type to be generic over the folding configuration. That way, I can define the instances and witnesses linking them to the ones associated to that configuration.
  - Consequently, there's a bit of refactors in the types inside the `FoldingEnv` implementation for that type.

### DIFF
--- a/folding/src/lib.rs
+++ b/folding/src/lib.rs
@@ -74,7 +74,7 @@ pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
     type Column: FoldingColumnTrait + Debug + Eq + Hash;
 
     // in case of using docomposable folding, if not it can be just ()
-    type Selector: Clone + Debug + Eq + Hash + Copy;
+    type Selector: Clone + Debug + Eq + Hash + Copy + Ord + PartialOrd;
 
     /// The type of an abstract challenge that can be found in the expressions
     /// provided as constraints.
--- a/folding/src/lib.rs
+++ b/folding/src/lib.rs
@@ -93,7 +93,7 @@ pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
     /// In the generic prover/verifier, it would be `kimchi_msm::witness::Witness`.
     type Witness: Witness<Self::Curve> + Clone;
 
-    type Structure;
+    type Structure: Clone;
 
     type Env: FoldingEnv<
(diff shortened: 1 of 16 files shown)


dannywillems:
I cherry-picked my two commits into 2184 and 2185.

dannywillems (approved):

folding/src/lib.rs
```diff
@@ -117,19 +114,24 @@ pub trait FoldingConfig: Clone + Debug + Eq + Hash + 'static {
...
     type Structure;
 
     /// Creates a new environment storing the structure, instances and witnesses.
     fn new(structure: &Self::Structure, instances: [&I; 2], witnesses: [&W; 2]) -> Self;
 
+    /// Returns the domain size of the circuit.
+    fn domain_size(&self) -> usize;
+
     // TODO: move into `FoldingConfig`
     // FIXME: when we move this to `FoldingConfig` it will be general for all impls as:
```
dannywillems, line 128: Does it mean this comment is old, now?

PR #2187, opened 2024-05-07
Use sponge for challenges

dannywillems:


querolita (approved):
I will build on top of this my PR to use more generic types instead of Fp, Curve, Sponge that is now used inside the folding types of optimism.

dannywillems:
> I will build on top of this my PR to use more generic types instead of Fp, Curve, Sponge that is now used inside the folding types of optimism.

Awesome!

PR #2188, opened 2024-05-07
Using generics for folding types in optimism crate

querolita:
This modifies mainly the `FoldingInstance` and `FoldingWitness` types inside optimism, together with the `Folder` trait for the tracer.

### DIFF
--- a/optimism/src/keccak/tests.rs
+++ b/optimism/src/keccak/tests.rs
@@ -538,7 +538,6 @@ fn test_keccak_prover_constraints() {
 }
 
 #[test]
-#[cfg(feature = "bn254")]
 fn test_keccak_decomposable_folding() {
     use crate::{keccak::folding::KeccakConfig, trace::Folder, BaseSponge, Curve};
     use ark_poly::{EvaluationDomain, Radix2EvaluationDomain as D};
--- a/optimism/src/keccak/tests.rs
+++ b/optimism/src/keccak/tests.rs
@@ -14,7 +14,7 @@ use crate::{
     },
     lookups::{FixedLookupTables, LookupTable, LookupTableIDs::*},
     trace::Tracer,
-    BaseSponge,
+    BaseSponge, Fp,
 };
 use ark_ff::{One, Zero};
 use kimchi::{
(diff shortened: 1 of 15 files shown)


querolita:
I commented out the decomposable folding Keccak test because it is still failing to obtain the right domain size. Planning to address it in https://github.com/o1-labs/proof-systems/pull/2182.

optimism/src/keccak/tests.rs
```diff
@@ -625,18 +618,28 @@ fn test_keccak_decomposable_folding() {
         let (_scheme, _final_constraint) =
             DecomposableFoldingScheme::<KeccakConfig>::new(constraints, vec![], &srs, domain, ());
 
+        /*
         // Fold Sponge(Absorb(Only))
-        {
```
dannywillems, line 629: Note that these should not be in this PR. But, ok to merge to avoid blocking and wait another CI.

optimism/src/trace.rs
```diff
@@ -85,46 +86,50 @@ impl<
...
-impl<
-        const N: usize,
-        const N_REL: usize,
-        const N_SEL: usize,
-        Selector: Eq + Hash + Indexer + Copy + Ord + PartialOrd,
-    > Folder<N, Selector, Fp> for Trace<N, N_REL, N_SEL, Selector, Fp>
+impl<const N: usize, const N_REL: usize, const N_SEL: usize, C: FoldingConfig, Sponge>
+    Folder<N, C, Sponge, ScalarField<C>> for Trace<N, N_REL, N_SEL, C::Selector, ScalarField<C>>
+where
+    C::Selector: Ord + PartialOrd + Indexer,
```
dannywillems, line 108: Wondering if `Ord + PartialOrd` should not be moved by default into folding. It is used by decomposable folding.

dannywillems (approved):
LGTM, minor changes I would not have included in this PR, but not blocking to continue. I prefer to go fast on these small changes :).

optimism/src/trace.rs
querolita: yeah, I was wondering the same in the next PR I was working on... Will probably try that out and see if it all works fine.

PR #2198, opened 2024-05-08
absorb instances in folding

fabrizio-m:


folding/src/instance_witness.rs
```diff
@@ -82,6 +86,19 @@ impl<G: CommitmentCurve, I: Instance<G>> ExtendedInstance<G, I> {
             extended: vec![],
         }
     }
+    /// wraps the inner absorb, adding its own commitments
+    fn to_absorb(&self) -> (Vec<G::ScalarField>, Vec<G::BaseField>) {
+        let mut elements = self.inner.to_absorb();
+        let extended_commitments = self.extended.iter().flat_map(|commit| {
+            assert_eq!(commit.elems.len(), 1);
+            let point = commit.elems[0];
+            //this may need change if we need to support the infinity point
```
dannywillems, line 95: Why this case? Can we not simply use absorb_commitment, please?

folding/src/lib.rs
```diff
@@ -237,6 +237,10 @@ impl<'a, CF: FoldingConfig> FoldingScheme<'a, CF> {
         fq_sponge.absorb_g(&error_commitments[0].elems);
         fq_sponge.absorb_g(&error_commitments[1].elems);
 
+        let to_absorb = env.to_absorb();
```
dannywillems, line 240: Can you indicate the order of the absorbtion in the documentation, please? We need to be consistent for the IVC circuit, and it is good to document it.
We should have a test checking the order is in line with the documentation.

dannywillems: Note that the typed transcript could be used, see https://github.com/o1-labs/proof-systems/pull/2094.
The idea is that every "round" (compute_error, compute_extended, etc) sends a message type, that can be absorb.
Each function returns a type "Message", and the sponge gets to absorb a heterogeneous list of "absorbable messages".
The heterogeneous list can be implemented as `Nil, Cons` constructors.

But I suppose it will be for the future. For now, we can keep it simple, and write in comments the order.

dannywillems (changes requested):
Converting to draft + see comments for the requests regarding changes.
I'm not sure zero_vec/zero_commitments should be parts of the code, and I suppose it makes the code fails.

folding/src/instance_witness.rs
fabrizio-m: we could, but I was not sure about its implementation, and is one extra thing to duplicate in IVC, but I have no issue with changing it if you think is better.

fabrizio-m: another option is to change the signature to allow only one field, and leave it up to the trait implementation to decompose and represent the points in the native field.

dannywillems:
Any update @fabrizio-m?

dannywillems:
back on fixing the conflicts :(

folding/src/instance_witness.rs
```diff
@@ -82,6 +95,16 @@ impl<G: CommitmentCurve, I: Instance<G>> ExtendedInstance<G, I> {
             extended: vec![],
         }
     }
+    /// wraps the inner absorb, adding its own commitments
```
dannywillems, line 98:
```suggestion
    /// Return the elements to be absorbed by the sponge
    ///
    /// The commitments to the additional columns created by quadriticization
    /// are appended to the existing commitments of the initial instance
    /// to be absorbed. The scalars are unchanged.
```

dannywillems (approved):
I will accept for now, for the interface, and the test.
If the number of values to be absorbed is wrong, I suggest fixing after when we add the IVC library.

PR #2200, opened 2024-05-09
MDBook: Fix compilation + add CI check

volhovm:
Problem: 
* (recently added) mdbook linkcheck does not build with nightly.
* Checking that MDBook is not broken was not part of the build script, 

Solution:
* Now we check MDBook in CI (because we want to make sure contributions to the book are not buggy..?)
* Downgraded the book deployment version from `nightly` to `1.72` to be in line with the rust script/our main `rust-toolchain` version. 

Question:
* Why does our CI check the project builds with `1.74` but `rust-toolchain` is `1.72`? If we can upgrade it to `1.74` I can remove `--fixed` from the script, which is a way to make `mdbook` work on `1.72` which is our default version. Because lot of dependencies want `1.74` actually.

### DIFF
--- a/.github/workflows/gh-page.yml
+++ b/.github/workflows/gh-page.yml
@@ -8,7 +8,7 @@ env:
   OCAML_VERSION: "4.14.0"
   # This version has been chosen randomly. It seems that with 2023-11-16, it is
   # broken. The compiler crashes. Feel free to pick any newer working version.
-  RUST_TOOLCHAIN_VERSION: "nightly-2024-01-16"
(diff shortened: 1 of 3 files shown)


dannywillems (approved):

dannywillems:
> Why does our CI check the project builds with 1.74 but rust-toolchain is 1.72? If we can upgrade it to 1.74 I can remove --fixed from the script, which is a way to make mdbook work on 1.72 which is our default version. Because lot of dependencies want 1.74 actually.

Depends on the version used in Mina.

PR #2205, opened 2024-05-09
Remove some type aliases for FoldingWitness/FoldingInstance

dannywillems:
It is an opinionated change. Feel free to refuse.
I split in different commits other changes I would like to see landed in master.


querolita (commented):
Oh well, I thought that having those aliases for types made it cleaner to read, otherwise it takes a lot of space no?

dannywillems:
> Oh well, I thought that having those aliases for types made it cleaner to read, otherwise it takes a lot of space no?

No opinion. Let's refuse then.
My reason to change was to remove a type indirection, but small taste.
Closing.

PR #2206, opened 2024-05-09
o1VM: prefix columns vars with N and suffix with COLS

dannywillems:
To be consistent.


querolita (approved):

volhovm:
Nice! in https://github.com/o1-labs/proof-systems/pull/2186 the current format is as follows:
```
    const N_COL: usize,
    const N_REL: usize,
    const N_DSEL: usize,
    const N_FSEL: usize,
```

dannywillems:
Ah, the PR should not have been merged as it contained changes from other PRs. I was fixing this.
I am reverting in a follow-up PR @volhovm

PR #2207, opened 2024-05-09
o1VM/folding: use prefix name Decomposable when appropriate

dannywillems:


querolita (approved):
I am not a huge fan of the long name, but for now it is ok until we find a more succinct naming.

PR #2208, opened 2024-05-09
o1VM: introduce the term ProvableTrace

dannywillems:


querolita (approved):
All good except for maybe `SingleInstructionTrace` could be called `Trace` and then our old `Trace` could be `DecomposedTrace` or `MultiTrace`.

dannywillems:
> All good except for maybe `SingleInstructionTrace` could be called `Trace` and then our old `Trace` could be `DecomposedTrace` or `MultiTrace`.

Let me change it in a follow-up.

PR #2213, opened 2024-05-09
Sort out MSM tests

volhovm:
MSM tests were quite disorganised. This PR:
* Moves generic prover-verifier tests from `lib.rs` into `tests/proof_system.rs`
* Moves test circuits from `test/...` to `test/test_circuit/...`
* Moves `tests/mod.rs` tests (that were circuit specific) into `test/test_circuit/mod.rs`.
* Moves logup test into `test/logup.rs`


msm/src/circuit_design/witness.rs
```diff
@@ -10,14 +10,24 @@ use crate::{
 use ark_ff::PrimeField;
 use kimchi::circuits::domains::EvaluationDomains;
 use log::debug;
-use std::{collections::BTreeMap, iter};
+use std::{collections::BTreeMap, iter, marker::PhantomData};
 
-/// Witness builder environment. Operates
-pub struct WitnessBuilderEnv<F: PrimeField, const CIX_COL_N: usize, LT: LookupTableID> {
+/// Witness builder environment. Operates on multiple rows at the same
+/// time. `CIx::N_COL` must be equal to `N_COL`; passing these two
```
querolita, line 16: Cannot this be forced in Rust? Something similar to `CIx: ColumnIndexer<N_COL = N_COL>`. I do similar things when telling the traits the equalities that must hold between folding types and o1vm's.

msm/src/circuit_design/witness.rs
```diff
@@ -10,14 +10,24 @@ use crate::{
...
+    const N_REL: usize,
+    const N_DSEL: usize,
+    const N_FSEL: usize,
+    LT: LookupTableID,
+> {
     /// The witness columns that the environment is working with.
     /// Every element of the vector is a row, and the builder is
     /// always processing the last row.
-    pub witness: Vec<Witness<CIX_COL_N, F>>,
+    pub witness: Vec<Witness<N_COL, F>>,
```
querolita, line 30: Like, here maybe you could say instead `Witness<CIx::N_COL,F>` ? Not sure as I haven't tried with constant generics, but maybe worth trying.

msm/src/circuit_design/witness.rs
```diff
@@ -28,10 +38,24 @@ pub struct WitnessBuilderEnv<F: PrimeField, const CIX_COL_N: usize, LT: LookupTa
     /// each row is a map from lookup type to a vector of concrete
     /// lookups requested.
     pub lookups: Vec<BTreeMap<LT, Vec<Logup<F, LT>>>>,
+
+    /// Fixed values for selector columns. `fixed_selectors[i][j]` is the
+    /// value for row #j of the selector #i.
+    pub fixed_selectors: Vec<Vec<F>>,
+
+    // Not strictly needed at this point, but likely will be used later.
+    pub phantom_cix: PhantomData<CIx>,
```
querolita, line 47: Maybe explain what this could potentially be, doesn't bring as much meaning to the struct as the other names `lookups`, `fixed_selectors`, etc.

msm/src/circuit_design/witness.rs
```diff
@@ -212,27 +302,39 @@ impl<F: PrimeField, const CIX_COL_N: usize, LT: LookupTableID> WitnessBuilderEnv
...
         // Boxing to avoid stack overflow
-        let mut witness: Box<Witness<CIX_COL_N, Vec<F>>> = Box::new(Witness {
+        let mut witness: Box<Witness<N_COL, Vec<F>>> = Box::new(Witness {
             cols: Box::new(std::array::from_fn(|_| Vec::with_capacity(domain_size))),
         });
 
         // Filling actually used rows first
         for witness_row in self.witness.iter().take(domain_size) {
-            for j in 0..CIX_COL_N {
+            for j in 0..N_REL {
```
querolita, line 314: Beware that here, before all of `N_COL` was being used in the loop but here only up to `N_REL`. Is this the expected behavior?

msm/src/circuit_design/witness.rs
```diff
@@ -212,27 +302,39 @@ impl<F: PrimeField, const CIX_COL_N: usize, LT: LookupTableID> WitnessBuilderEnv
...
         // FIXME: Maybe this is not always wise, as default instance can be non-zero.
         if self.witness.len() < domain_size {
-            for i in 0..CIX_COL_N {
+            for i in 0..N_REL {
                 witness.cols[i].extend(vec![F::zero(); domain_size - self.witness.len()]);
             }
         }
 
+        // Fill out dynamic selectors.
+        for i in 0..N_DSEL {
```
querolita, line 328: Oh I see, so you continue afterwards with the rest of columns okay.

msm/src/columns.rs
```diff
@@ -58,9 +70,10 @@ impl FormattedOutput for Column {
 
 /// A datatype expressing a generalized column, but with potentially
 /// more convenient interface than a bare column.
-pub trait ColumnIndexer {
+pub trait ColumnIndexer: core::fmt::Debug + Copy + Eq + Ord {
+    // TODO rename into N_COL for consistency
```
querolita, line 74: Maybe delete this todo?

querolita (approved):
Approving but please take a look at the comments before merging 😄

msm/src/circuit_design/witness.rs
volhovm: Yes -- that's because `WitnessBuilderEnv` before assumed `CIX_COL_N = N_REL`. This loop only cares about relation wires. And now that I passed more constants into it this loop needs to only iterate over `N_REL`.

msm/src/circuit_design/witness.rs
volhovm: Will add a comment, but generally `phantom` fields are needed when one wants to parameterise a struct with a type parameter that is not used otherwise. 

E.g. sometimes `struct EmptyFieldTag<T: Field>(tag_ix: usize)` won't compile, and then `struct EmptyFieldTag<T: Field>{ tag_ix: usize, phantom: PhantomData<T> }` does the trick. https://doc.rust-lang.org/std/marker/struct.PhantomData.html

msm/src/circuit_design/witness.rs
volhovm: this does not work at least in stable... which is a very useful feature to have tbh. So, unfortunately.
```
error: generic parameters may not be used in const operations
  --> msm/src/circuit_design/witness.rs:30:32
   |
30 |     pub witness: Vec<Witness<{ CIx::N_COL }, F>>,
   |                                ^^^^^^^^^^ cannot perform const operation using `CIx`
   |
   = note: type parameters may not be used in const expressions
```

volhovm: It will work on `nightly` though, which makes me super envious :( :( :( 

https://stackoverflow.com/questions/65461672/rust-constant-expression-depends-on-a-generic-parameter

msm/src/circuit_design/witness.rs
volhovm: I already spent some time before trying to do it, but just tried again, and my conclusions are negative: it's not easy.
- https://docs.rs/type-equals/latest/type_equals/ doesn't seem to work for `const` equality
- https://docs.rs/type-equalities/latest/type_equalities/index.html seems too experimental
- https://stackoverflow.com/questions/60138397/how-to-test-for-type-equality-in-rust something unstable + something hacky

volhovm: Hope it were possible though...

querolita: Awesome, thanks for trying though!

volhovm:
@querolita addressed here: https://github.com/o1-labs/proof-systems/pull/2213/commits/ce85583b2d5af4ad6b227d6f930109620539b758

msm/src/circuit_design/witness.rs
querolita: Mmmm I see, well let's wait until they allow this in stable 👍🏻

PR #2214, opened 2024-05-09
Make the folding environment working with ProvableTrace

dannywillems:
Allow to define a folding scheme without selectors, i.e. with only the columns that are really used. For MIPS, it is the 93 columsn for the state + the error term and the halt.

The pull request includes functions to build a whole trace for a single MIPS instruction, addiu, and can be replicated for other.


o1vm/src/mips/tests.rs
```diff
@@ -323,4 +335,200 @@ mod folding {
...
+            build_folding_instance(&folding_witness_one, &mut fq_sponge, domain, &srs);
+        let folding_instance_two =
+            build_folding_instance(&folding_witness_two, &mut fq_sponge, domain, &srs);
+
+        let folding_compat_constraints: Vec<FoldingCompatibleExpr<MIPSFoldingConfig>> = constraints
+            .iter()
+            .map(|x| FoldingCompatibleExpr::from(x.clone()))
+            .collect::<Vec<_>>();
+
+        let (folding_scheme, _) = FoldingScheme::<MIPSFoldingConfig>::new(
```
querolita, line 520: I suppose the next step will be to have the same example but using the decomposable scheme instead, right? I am curious to see if it works for ADDIU or not.

o1vm/src/mips/tests.rs
```diff
@@ -323,4 +335,200 @@ mod folding {
...
+        });
+        // sanity check
+        witness
+            .cols
+            .iter()
+            .for_each(|x| assert_eq!(x.len(), domain_size));
+        witness
+    }
+
+    fn build_folding_instance(
```
querolita, line 390: Don't you prefer to implement `Foldable` for the normal `Trace`? I can adapt the signature so it matches better what you have in here.

querolita (approved):

o1vm/src/mips/tests.rs
dannywillems: Yes. Which would be duplicated what you did in Keccak

o1vm/src/mips/tests.rs
dannywillems: It would be nice to merge, yes. I was thinking we could do a better interface in the near future.

dannywillems:
Back to solving conflicts

dannywillems:
I hope it is fine now!!!

PR #2219, opened 2024-05-14
o1VM: use overflowing_add to handle overflow

dannywillems:
It makes the code safer, and we can run it with
RUSTFLAGS="-Coverflow-checks=y -Cdebug-assertions=y"

Also makes the code slightly faster as it might remove some equality testing if the implementation of overflow_add is done using assembly instructions


querolita (approved):
Oh, fishy, great catch!

PR #2220, opened 2024-05-14
Simplify interface of decomposable traces

querolita:
This PR suggests a refactor on top of the decomposable traces to keep track of executions more easily. In the following, I explain what are the main changes in this branch:

- Making `DecomposedTrace` a struct containing just a `domain_size` and `trace` which is a `BTreeMap` of `Trace`. Then, just like before, inside each `Trace` you can find a `domain_size` (which is the same), `witness`, `lookups`, and `constraints`. 
  - Implemented `Index` for `DecomposedTrace` with `C::Selector` so that one can access the trace contents without performing `self.trace[&selector]`. 

- Removal of some generics that were not being used in the type itself. Basically, this reduces the need of passing `N_REL` and `N_SEL` in most cases. 

- An implementation of `Tracer` for both `Trace` and `DecomposedTrace`. The main difference between both is that an internal type of the trait called `Selector` is set to `C::Selector` for `DecomposedTrace` whereas it is set to `()` for `Trace` (as it only refers to one type of instruction). 

- Renaming the `new()` function inside `Tracer` as `init()` to avoid compiler confusion with the `new()` function inside `DecomposableTracer`. 

- Moved `pad_witnesses` from `Tracer` to `DecomposableTracer` as it only makes sense when `Self` has several witness variants.

- Removal of the checks that the step needs to be standardized inside `Tracer` for `KeccakTrace`. Instead, the step is guaranteed to be in standard shape (meaning, any `Round(i)` is mapped to `Round(0)` for convenience) by initializing the step to be `keccak_env.selector()` instead of `keccak_env.step.unwrap()`. 
  - As a side effect, this allows us to implement `Tracer` for any `DecomposedTrace` generically.  

- Created a `folding_constraints()` functions inside `Foldable` trait to obtain the constraints in `BTreeMap` shape with expressions that are already folding-friendly. 

- Included the absorption of the commitments into the sponge inside `to_folding_pair` of `Foldable`, as this was missing at the moment.


dannywillems:
Fixing conflicts

o1vm/src/keccak/trace.rs
```diff
@@ -7,118 +7,106 @@ use strum::IntoEnumIterator;
...
+            domain_size,
+            witness: Witness {
+                cols: Box::new(std::array::from_fn(|_| Vec::with_capacity(domain_size))),
+            },
+            constraints: KeccakEnv::constraints_of(step),
+            lookups: KeccakEnv::lookups_of(step),
+        }
+    }
+
     fn push_row(
```
dannywillems, line 69: Wondering if we can still improve this duplication of code. We might want to discuss again what @volhovm had in mind for these "shared" implementations. For each structure implementing the tracer trait, we will have nearly the same implementation.

dannywillems (approved):

PR #2221, opened 2024-05-15
Reintroduce folding expression conversion test

querolita:
At some point this was removed. Maybe a `mod tests` inside `o1vm/src/folding.rs` is not the perfect location for this, as it should ideally live inside the folding library. But for now let's have them there and then migrate them into `folding` with any additional changes that could be required.


volhovm (approved):
Approving on the assumption of trusting you that this test is now not duplicated anywhere else in the codebase & that it makes sense so I don't have to check its logic :) 

(I think I saw it before at some point, yes...)

PR #2233, opened 2024-05-16
IVC: Add selector columns

volhovm:
Before: `ivc_circuit` was building all /witness/ wires, but /constraints/, e.g. `c_{b0,i}`, were available for every block separately. This means `c_{b0,i}` for block 0 would not work on rows of block 1.

Now:
* There are selectors for each block.
* Every constraint is now working for the /whole circuit/, since it's now multiplied by selector: `sel0 * c_{b0,i}`.
* Completeness test is added with lookups disabled.

Offtop:
* I added the `ID` lens because I wanted to generalise circuit generator to work for two tables simultaneously. This proved to /not/ be useful. But the `ID` lens is useful I think so I keep it.

Next step:
* Reduce stack usage. It overflows if stack is default 128KB, but fine with 2MB. Done in next PR https://github.com/o1-labs/proof-systems/pull/2247.

### DIFF
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -151,6 +151,9 @@ pub type IVCPoseidonColumn = PoseidonColumn<IVC_POSEIDON_STATE_SIZE, IVC_POSEIDO
 ///
 /// Challenges block.
 ///
+///            relaxed
+///                       strict
+///                    (relaxed in-place)
 ///        r   α_{L,i}    α_{R}^i     α_{O,i}
 ///  1    |--|--------|-----------|-------------------|
 ///       |  |        | α_R = h_R |                   |
@@ -212,6 +215,9 @@ pub type IVCPoseidonColumn = PoseidonColumn<IVC_POSEIDON_STATE_SIZE, IVC_POSEIDO
 // TODO: Can we pass just one coordinate and sign (x, sign) instead of (x,y) for hashing?
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum IVCColumn {
+    /// Selector for blocks. Inner usize is ∈ [0,#blocks).
+    BlockSel(usize),
+
     /// 2*17 15-bit limbs (two base field points)
     Block1Input(usize),
     /// 2*4 75-bit limbs
@@ -280,6 +286,9 @@ pub enum IVCColumn {
     Block6UOutput,
 }
 
+/// Number of blocks in the circuit.
+const N_BLOCKS: usize = 6;
+
 impl ColumnIndexer for IVCColumn {
     // This should be
     //   const N_COL: usize = std::cmp::max(IVCPoseidonColumn::N_COL, FECColumn::N_COL);
@@ -288,74 +297,83 @@ impl ColumnIndexer for IVCColumn {
 
     fn to_column(self) -> Column {
         match self {
+            IVCColumn::BlockSel(i) => {
+                assert!(i < 2 * N_BLOCKS);
+                Column::FixedSelector(i)
+            }
+
             IVCColumn::Block1Input(i) => {
                 assert!(i < 2 * N_LIMBS_SMALL);
-                Column::Relation(i)
+                Column::Relation(N_BLOCKS + i)
             }
             IVCColumn::Block1InputRepacked75(i) => {
                 assert!(i < 2 * N_LIMBS_LARGE);
-                Column::Relation(2 * N_LIMBS_SMALL + i)
+                Column::Relation(N_BLOCKS + 2 * N_LIMBS_SMALL + i)
             }
             IVCColumn::Block1InputRepacked150(i) => {
                 assert!(i < 2 * N_LIMBS_XLARGE);
-                Column::Relation(2 * N_LIMBS_SMALL + 2 * N_LIMBS_LARGE + i)
+                Column::Relation(N_BLOCKS + 2 * N_LIMBS_SMALL + 2 * N_LIMBS_LARGE + i)
             }
 
             IVCColumn::Block2Hash(poseidon_col) => poseidon_col.to_column(),
 
-            IVCColumn::Block3ConstPhi => Column::Relation(0),
-            IVCColumn::Block3ConstR => Column::Relation(1),
-            IVCColumn::Block3PhiPow => Column::Relation(2),
-            IVCColumn::Block3PhiPowR => Column::Relation(3),
-            IVCColumn::Block3PhiPowR2 => Column::Relation(4),
-            IVCColumn::Block3PhiPowR3 => Column::Relation(5),
+            IVCColumn::Block3ConstPhi => Column::Relation(N_BLOCKS),
+            IVCColumn::Block3ConstR => Column::Relation(N_BLOCKS + 1),
+            IVCColumn::Block3PhiPow => Column::Relation(N_BLOCKS + 2),
+            IVCColumn::Block3PhiPowR => Column::Relation(N_BLOCKS + 3),
+            IVCColumn::Block3PhiPowR2 => Column::Relation(N_BLOCKS + 4),
+            IVCColumn::Block3PhiPowR3 => Column::Relation(N_BLOCKS + 5),
             IVCColumn::Block3PhiPowLimbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(6 + i)
+                Column::Relation(N_BLOCKS + 6 + i)
             }
             IVCColumn::Block3PhiPowRLimbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(6 + N_LIMBS_SMALL + i)
+                Column::Relation(N_BLOCKS + 6 + N_LIMBS_SMALL + i)
             }
             IVCColumn::Block3PhiPowR2Limbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(6 + 2 * N_LIMBS_SMALL + i)
+                Column::Relation(N_BLOCKS + 6 + 2 * N_LIMBS_SMALL + i)
             }
             IVCColumn::Block3PhiPowR3Limbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(6 + 3 * N_LIMBS_SMALL + i)
+                Column::Relation(N_BLOCKS + 6 + 3 * N_LIMBS_SMALL + i)
             }
 
             IVCColumn::Block4Input1(i) => {
                 assert!(i < 2 * N_LIMBS_LARGE);
-                Column::Relation(i)
+                Column::Relation(N_BLOCKS + i)
             }
-            IVCColumn::Block4Coeff => Column::Relation(8),
-            IVCColumn::Block4Input2AccessTime => Column::Relation(9),
+            IVCColumn::Block4Coeff => Column::Relation(N_BLOCKS + 8),
+            IVCColumn::Block4Input2AccessTime => Column::Relation(N_BLOCKS + 9),
             IVCColumn::Block4Input2(i) => {
                 assert!(i < 2 * N_LIMBS_LARGE);
-                Column::Relation(10 + i)
+                Column::Relation(N_BLOCKS + 10 + i)
             }
-            IVCColumn::Block4ECAddInter(fec_inter) => fec_inter.to_column().add_rel_offset(18),
+            IVCColumn::Block4ECAddInter(fec_inter) => {
+                fec_inter.to_column().add_rel_offset(N_BLOCKS + 18)
+            }
             IVCColumn::Block4OutputRaw(fec_output) => fec_output
                 .to_column()
-                .add_rel_offset(18 + FECColumnInter::N_COL),
+                .add_rel_offset(N_BLOCKS + 18 + FECColumnInter::N_COL),
             IVCColumn::Block4OutputAccessTime => {
-                Column::Relation(18 + FECColumnInter::N_COL + FECColumnOutput::N_COL)
+                Column::Relation(N_BLOCKS + 18 + FECColumnInter::N_COL + FECColumnOutput::N_COL)
             }
             IVCColumn::Block4OutputRepacked(i) => {
                 assert!(i < 2 * N_LIMBS_LARGE);
-                Column::Relation(18 + FECColumnInter::N_COL + FECColumnOutput::N_COL + 1 + i)
+                Column::Relation(
+                    N_BLOCKS + 18 + FECColumnInter::N_COL + FECColumnOutput::N_COL + 1 + i,
+                )
             }
-            IVCColumn::Block5ConstHr => Column::Relation(0),
-            IVCColumn::Block5ConstR => Column::Relation(1),
-            IVCColumn::Block5ChalLeft => Column::Relation(2),
-            IVCColumn::Block5ChalRight => Column::Relation(3),
-            IVCColumn::Block5ChalOutput => Column::Relation(4),
+            IVCColumn::Block5ConstHr => Column::Relation(N_BLOCKS),
+            IVCColumn::Block5ConstR => Column::Relation(N_BLOCKS + 1),
+            IVCColumn::Block5ChalLeft => Column::Relation(N_BLOCKS + 2),
+            IVCColumn::Block5ChalRight => Column::Relation(N_BLOCKS + 3),
+            IVCColumn::Block5ChalOutput => Column::Relation(N_BLOCKS + 4),
 
-            IVCColumn::Block6ConstR => Column::Relation(0),
-            IVCColumn::Block6ULeft => Column::Relation(1),
-            IVCColumn::Block6UOutput => Column::Relation(2),
+            IVCColumn::Block6ConstR => Column::Relation(N_BLOCKS),
+            IVCColumn::Block6ULeft => Column::Relation(N_BLOCKS + 1),
+            IVCColumn::Block6UOutput => Column::Relation(N_BLOCKS + 2),
         }
     }
 }
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -293,7 +293,8 @@ impl ColumnIndexer for IVCColumn {
     // This should be
     //   const N_COL: usize = std::cmp::max(IVCPoseidonColumn::N_COL, FECColumn::N_COL);
     // which is runtime-only expression..?
-    const N_COL: usize = IVCPoseidonColumn::N_COL;
+    // 333 is not enough
+    const N_COL: usize = 400;
 
     fn to_column(self) -> Column {
         match self {
@@ -315,7 +316,9 @@ impl ColumnIndexer for IVCColumn {
                 Column::Relation(N_BLOCKS + 2 * N_LIMBS_SMALL + 2 * N_LIMBS_LARGE + i)
             }
 
-            IVCColumn::Block2Hash(poseidon_col) => poseidon_col.to_column(),
+            IVCColumn::Block2Hash(poseidon_col) => {
+                poseidon_col.to_column().add_rel_offset(N_BLOCKS)
+            }
 
             IVCColumn::Block3ConstPhi => Column::Relation(N_BLOCKS),
             IVCColumn::Block3ConstR => Column::Relation(N_BLOCKS + 1),
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -299,84 +299,79 @@ impl ColumnIndexer for IVCColumn {
     fn to_column(self) -> Column {
         match self {
             IVCColumn::BlockSel(i) => {
-                assert!(i < 2 * N_BLOCKS);
+                assert!(i < N_BLOCKS);
                 Column::FixedSelector(i)
             }
 
             IVCColumn::Block1Input(i) => {
                 assert!(i < 2 * N_LIMBS_SMALL);
-                Column::Relation(N_BLOCKS + i)
+                Column::Relation(i)
             }
             IVCColumn::Block1InputRepacked75(i) => {
                 assert!(i < 2 * N_LIMBS_LARGE);
-                Column::Relation(N_BLOCKS + 2 * N_LIMBS_SMALL + i)
+                Column::Relation(2 * N_LIMBS_SMALL + i)
             }
             IVCColumn::Block1InputRepacked150(i) => {
                 assert!(i < 2 * N_LIMBS_XLARGE);
-                Column::Relation(N_BLOCKS + 2 * N_LIMBS_SMALL + 2 * N_LIMBS_LARGE + i)
+                Column::Relation(2 * N_LIMBS_SMALL + 2 * N_LIMBS_LARGE + i)
             }
 
-            IVCColumn::Block2Hash(poseidon_col) => {
-                poseidon_col.to_column().add_rel_offset(N_BLOCKS)
-            }
+            IVCColumn::Block2Hash(poseidon_col) => poseidon_col.to_column(),
 
-            IVCColumn::Block3ConstPhi => Column::Relation(N_BLOCKS),
-            IVCColumn::Block3ConstR => Column::Relation(N_BLOCKS + 1),
-            IVCColumn::Block3PhiPow => Column::Relation(N_BLOCKS + 2),
-            IVCColumn::Block3PhiPowR => Column::Relation(N_BLOCKS + 3),
-            IVCColumn::Block3PhiPowR2 => Column::Relation(N_BLOCKS + 4),
-            IVCColumn::Block3PhiPowR3 => Column::Relation(N_BLOCKS + 5),
+            IVCColumn::Block3ConstPhi => Column::Relation(0),
+            IVCColumn::Block3ConstR => Column::Relation(1),
+            IVCColumn::Block3PhiPow => Column::Relation(2),
+            IVCColumn::Block3PhiPowR => Column::Relation(3),
+            IVCColumn::Block3PhiPowR2 => Column::Relation(4),
+            IVCColumn::Block3PhiPowR3 => Column::Relation(5),
             IVCColumn::Block3PhiPowLimbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(N_BLOCKS + 6 + i)
+                Column::Relation(6 + i)
             }
             IVCColumn::Block3PhiPowRLimbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(N_BLOCKS + 6 + N_LIMBS_SMALL + i)
+                Column::Relation(6 + N_LIMBS_SMALL + i)
             }
             IVCColumn::Block3PhiPowR2Limbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
-                Column::Relation(N_BLOCKS + 6 + 2 * N_LIMBS_SMALL + i)
+                Column::Relation(6 + 2 * N_LIMBS_SMALL + i)
             }
             IVCColumn::Block3PhiPowR3Limbs(i) => {
                 assert!(i < N_LIMBS_SMALL);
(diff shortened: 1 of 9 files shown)


dannywillems:
Starting having a look now.

ivc/src/ivc/columns.rs
```diff
@@ -212,6 +218,9 @@ pub type IVCPoseidonColumn = PoseidonColumn<IVC_POSEIDON_STATE_SIZE, IVC_POSEIDO
 // TODO: Can we pass just one coordinate and sign (x, sign) instead of (x,y) for hashing?
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum IVCColumn {
+    /// Selector for blocks. Inner usize is ∈ [0,#blocks).
+    BlockSel(usize),
```
querolita, line 222: Just to make sure, this is not a row selector (as in Kimchi) but rather a selector for a chunk of columns within the same row, no?

ivc/src/ivc/columns.rs
```diff
@@ -284,78 +293,90 @@ impl ColumnIndexer for IVCColumn {
     // This should be
     //   const N_COL: usize = std::cmp::max(IVCPoseidonColumn::N_COL, FECColumn::N_COL);
     // which is runtime-only expression..?
-    const N_COL: usize = IVCPoseidonColumn::N_COL;
+    // 333 is not enough
+    const N_COL: usize = 400;
```
querolita, line 297: Where does this number come from?

ivc/src/ivc/columns.rs
```diff
@@ -284,78 +293,90 @@ impl ColumnIndexer for IVCColumn {
...
         match self {
+            IVCColumn::BlockSel(i) => {
+                assert!(i < 2 * N_BLOCKS);
+                Column::FixedSelector(i)
+            }
+
             IVCColumn::Block1Input(i) => {
                 assert!(i < 2 * N_LIMBS_SMALL);
-                Column::Relation(i)
+                Column::Relation(N_BLOCKS + i)
```
querolita, line 308: What's in `Column::Relation(0..N_BLOCKS)`?

ivc/src/ivc/columns.rs
```diff
@@ -284,78 +293,90 @@ impl ColumnIndexer for IVCColumn {
...
+            }
 
-            IVCColumn::Block3ConstPhi => Column::Relation(0),
-            IVCColumn::Block3ConstR => Column::Relation(1),
-            IVCColumn::Block3PhiPow => Column::Relation(2),
-            IVCColumn::Block3PhiPowR => Column::Relation(3),
-            IVCColumn::Block3PhiPowR2 => Column::Relation(4),
-            IVCColumn::Block3PhiPowR3 => Column::Relation(5),
+            IVCColumn::Block3ConstPhi => Column::Relation(N_BLOCKS),
+            IVCColumn::Block3ConstR => Column::Relation(N_BLOCKS + 1),
```
querolita, line 324: To me it looks like `IVCColumn::Block1Input(i)` and `IVCColumn::Block3ConstR` could collide for example when `i=1`, no?

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +916,101 @@ where
...
+pub fn build_selectors<F, const N_COL_TOTAL: usize, const N_CHALS: usize>(
+    domain_size: usize,
+) -> Vec<Vec<F>>
+where
+    F: PrimeField,
+{
+    // 3*N + 6*N+2 + N+1 + 35*N + 5 + 1 + N_CHALS =
+    // 45N + 9 + N_CHALS
+    let mut selectors: Vec<Vec<F>> = vec![vec![F::zero(); domain_size]; N_BLOCKS];
+    let mut cur_row = 0;
```
querolita, line 929: Just a tiny nit, in this repo we normally use 4 letters for current, meaning we would normally name this `curr_row`, in case you want to make it all look uniform with the rest.

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +916,101 @@ where
...
 }
 
+/// Builds selectors for the IVC circuit.
+pub fn build_selectors<F, const N_COL_TOTAL: usize, const N_CHALS: usize>(
+    domain_size: usize,
+) -> Vec<Vec<F>>
+where
+    F: PrimeField,
+{
+    // 3*N + 6*N+2 + N+1 + 35*N + 5 + 1 + N_CHALS =
```
querolita, line 926: Maybe change the order to `N_CHALS + 1` so that it matches the order of the for-loops below.

ivc/src/ivc/columns.rs
```diff
@@ -284,78 +293,90 @@ impl ColumnIndexer for IVCColumn {
...
     //   const N_COL: usize = std::cmp::max(IVCPoseidonColumn::N_COL, FECColumn::N_COL);
     // which is runtime-only expression..?
-    const N_COL: usize = IVCPoseidonColumn::N_COL;
+    // 333 is not enough
+    const N_COL: usize = 400;
 
     fn to_column(self) -> Column {
         match self {
+            IVCColumn::BlockSel(i) => {
+                assert!(i < 2 * N_BLOCKS);
```
dannywillems, line 302: Why `2 * N_BLOCKS`?

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +916,101 @@ where
...
+/// This function generates constraitns for the whole IVC circuit.
+pub fn constrain_selectors<F, Env>(env: &mut Env)
+where
+    F: PrimeField,
+    Env: ColAccessCap<F, IVCColumn>,
+{
+    for i in 0..N_BLOCKS {
+        // Each selector must have value either 0 or 1.
+        let sel = env.read_column(IVCColumn::BlockSel(i));
+        env.assert_zero(sel.clone() * (sel.clone() - Env::constant(F::from(1u64))));
```
querolita, line 967: Don't you have `Env::one()`? Otherwise, try with `Env::constant(F::one())` importing `ark_ff::One`.

querolita, line 967: Also not sure if I am a huge fan of `assert_zero` as it reads to me as if this was not an actual constraint but just a witness check.

ivc/src/ivc/columns.rs
volhovm: Typo! will fix.

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +916,101 @@ where
...
+pub fn constrain_ivc<F, Ff, Env>(env: &mut Env)
+where
+    F: PrimeField,
+    Ff: PrimeField,
+    Env: ColAccessCap<F, IVCColumn> + LookupCap<F, IVCColumn, IVCLookupTable<Ff>>,
+{
+    constrain_selectors(env);
+
+    let s0 = env.read_column(IVCColumn::BlockSel(0));
+    env.set_assert_mapper(Box::new(move |x| s0.clone() * x));
```
querolita, line 981: I would appreciate some comments explaining what these constraints are doing because I cannot judge whether it looks okay to me or not.

msm/src/circuit_design/capabilities.rs
```diff
@@ -11,17 +11,25 @@ use ark_ff::PrimeField;
 /// Environment capability for accessing and reading columns. This is necessary for
 /// building constraints.
 pub trait ColAccessCap<F: PrimeField, CIx: ColumnIndexer> {
+    // NB: 'static here means that `Variable` does not contain any
+    // references witha a lifetime less than 'static. Which is true in
```
querolita, line 15:
```suggestion
    // references with a lifetime less than 'static. Which is true in
```

msm/src/circuit_design/composition.rs
```diff
@@ -235,6 +265,10 @@ impl<
         self.0.assert_zero(cst);
     }
 
+    fn set_assert_mapper(&mut self, mapper: Box<dyn Fn(Self::Variable) -> Self::Variable>) {
+        self.0.set_assert_mapper(mapper);
```
querolita, line 269: I am not sure I like the `self.0` notation. Is it a tuple? If so, wouldn't it make sense to make it a struct instead?

ivc/src/ivc/columns.rs
dannywillems, line 308: Are you sure about the offset?
Looking at column_env.rs, I am not sure if the mapping we need is the one you're suggesting :thinking:.

msm/src/circuit_design/constraints.rs
```diff
@@ -91,10 +93,7 @@ impl<F: PrimeField, CIx: ColumnIndexer, LT: LookupTableID> LookupCap<F, CIx, LT>
 impl<F: PrimeField, LT: LookupTableID> ConstraintBuilderEnv<F, LT> {
     /// Get constraints related to the application logic itself.
     pub fn get_relation_constraints(&self) -> Vec<E<F>> {
-        self.constraints
-            .iter()
-            .map(|(_, cst)| cst.clone())
-            .collect()
+        self.constraints.clone()
```
dannywillems, line 96: Yeah, finally!

msm/src/circuit_design/capabilities.rs
dannywillems, line 15:
```suggestion
    // references with a lifetime less than 'static. Which is true in
```

ivc/src/ivc/columns.rs
dannywillems: I don't see it modified in https://github.com/o1-labs/proof-systems/pull/2249. I think I am not wrong, thinking the indexing is not correctly happening?

ivc/src/ivc/columns.rs
dannywillems: It makes sense to collide. It is like selectors in vanilla plonk. The columns are shared.

ivc/src/ivc/columns.rs
```diff
@@ -284,78 +293,90 @@ impl ColumnIndexer for IVCColumn {
...
+                poseidon_col.to_column().add_rel_offset(N_BLOCKS)
+            }
 
-            IVCColumn::Block3ConstPhi => Column::Relation(0),
-            IVCColumn::Block3ConstR => Column::Relation(1),
-            IVCColumn::Block3PhiPow => Column::Relation(2),
-            IVCColumn::Block3PhiPowR => Column::Relation(3),
-            IVCColumn::Block3PhiPowR2 => Column::Relation(4),
-            IVCColumn::Block3PhiPowR3 => Column::Relation(5),
+            IVCColumn::Block3ConstPhi => Column::Relation(N_BLOCKS),
```
dannywillems, line 323: I think it must be 0.

dannywillems: Like for all the relation columns.

ivc/src/ivc/mod.rs
```diff
@@ -7,32 +7,51 @@ mod tests {
...
+        witness::Witness,
+        BaseSponge, Ff1, Fp, OpeningProof, ScalarSponge, BN254,
+    };
+    use poly_commitment::pairing_proof::PairingSRS;
     use rand::{CryptoRng, RngCore};
 
     // Test number
-    pub const TEST_N_COL_TOTAL: usize = 50;
+    pub const TEST_N_COL_TOTAL: usize = IVCColumn::N_COL + 50;
+    // Absolutely no idea.
```
dannywillems, line 37: :joy:

ivc/src/ivc/mod.rs
```diff
@@ -7,32 +7,51 @@ mod tests {
...
         Fp,
         IVCColumn,
         { <IVCColumn as ColumnIndexer>::N_COL },
         { <IVCColumn as ColumnIndexer>::N_COL },
         0,
         0,
-        IVCLookupTable<Ff1>,
+        LT,
     >;
+    //type IVCWitnessBuilderEnv = IVCWitnessBuilderEnvRaw<IVCLookupTable<Ff1>>;
```
dannywillems, line 53: Nit: should we remove this? Or it is used later in a following PR?

ivc/src/ivc/interpreter.rs
```diff
@@ -286,20 +287,22 @@ where
...
-                    .into_iter()
-                    .map(|subvec: Vec<_>| subvec.try_into().unwrap())
-                    .collect::<Vec<_>>()
-                    .try_into()
-                    .unwrap()
-            })
-            .collect::<Vec<_>>()
-            .try_into()
-            .unwrap()
+    ) -> Box<[[[F; TWO_LIMB_SIZE]; N_COL_TOTAL]; 3]> {
```
dannywillems, line 290: Wondering if we could get rid of `Vec<Vec<F>>` and always use fixed size array and/or flat array. This code adds noise for the reader.

ivc/src/ivc/columns.rs
volhovm: You're right, will fix. I started with making everything `Relation` but forgot to remove this offset. Just wastes space indeed.

ivc/src/ivc/columns.rs
volhovm: It's a stub. I don't have the exact number right now, but as the comment above says the number of columns for IVC should be maximum of Poseidon and FEC Addition.

ivc/src/ivc/columns.rs
volhovm: Will fix!

ivc/src/ivc/interpreter.rs
volhovm: Fixed!

ivc/src/ivc/interpreter.rs
volhovm: We can discuss it at some other point :) I think it means "assert that constraint is zero" as in "asert that the constraint polynomial is zero" so I'm kind of happy :)

msm/src/circuit_design/capabilities.rs
volhovm: Fixed

msm/src/circuit_design/composition.rs
volhovm: That's the standard way to do wrappers in rust: `struct Wrapper(pub MyType)` and then `let wrapper = Wrapper(mytype); let y = wrapper.0;`. Again, discussable, but perhaps not in this PR?..

ivc/src/ivc/interpreter.rs
volhovm: Will do!

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +920,101 @@ where
 {
 }
 
+/// Builds selectors for the IVC circuit.
+pub fn build_selectors<F, const N_COL_TOTAL: usize, const N_CHALS: usize>(
```
dannywillems, line 924: LGTM

ivc/src/ivc/columns.rs
volhovm: Yes, fixed.

ivc/src/ivc/mod.rs
volhovm: Will remove yes :)

ivc/src/ivc/columns.rs
```diff
@@ -212,6 +218,9 @@ pub type IVCPoseidonColumn = PoseidonColumn<IVC_POSEIDON_STATE_SIZE, IVC_POSEIDO
 // TODO: Can we pass just one coordinate and sign (x, sign) instead of (x,y) for hashing?
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum IVCColumn {
+    /// Selector for blocks. Inner usize is ∈ [0,#blocks).
```
dannywillems, line 221: I cannot find the up-to-date description of each different block. Is it the ones at the top of file?
My question is related to the function `build_selectors`, to verify the selectors are populated according to the spec.

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +920,101 @@ where
...
+    selectors
+}
+
+/// This function generates constraitns for the whole IVC circuit.
+pub fn constrain_selectors<F, Env>(env: &mut Env)
+where
+    F: PrimeField,
+    Env: ColAccessCap<F, IVCColumn>,
+{
+    for i in 0..N_BLOCKS {
```
dannywillems, line 968: If selectors are public values, i.e. part of the setup, I think we do not need to add constraints.

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +920,101 @@ where
...
+        // Each selector must have value either 0 or 1.
+        let sel = env.read_column(IVCColumn::BlockSel(i));
+        env.assert_zero(sel.clone() * (sel.clone() - Env::constant(F::one())));
+    }
+}
+
+/// This function generates constraitns for the whole IVC circuit.
```
dannywillems, line 975:
```suggestion
/// This function generates constraints for the whole IVC circuit.
```

ivc/src/ivc/interpreter.rs
volhovm: Yeah, in the following PR I introduce some converting methods to use multi-dimensional arrays more. Might be good to use more commonly indeed :)

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +920,101 @@ where
...
+    Ff: PrimeField,
+    Env: ColAccessCap<F, IVCColumn> + LookupCap<F, IVCColumn, IVCLookupTable<Ff>>,
+{
+    constrain_selectors(env);
+
+    let s0 = env.read_column(IVCColumn::BlockSel(0));
+    env.set_assert_mapper(Box::new(move |x| s0.clone() * x));
+    constrain_inputs(env);
+
+    // TODO FIXME add constraints for hashes
```
dannywillems, line 988: I suppose it is `BlockSel(1)`?

ivc/src/ivc/columns.rs
querolita: Oh I see then yes, like having the same Kimchi selector in consecutive rows.

ivc/src/ivc/columns.rs
querolita: Okay, let's keep an eye on this in any case

ivc/src/ivc/columns.rs
querolita: I am not sure I understand why you want redundant definitions of the same column. Like, it is not as if you could access multiple rows for each constraint no? Like, at most the current row and the next, but not 4 rows ahead...

msm/src/circuit_design/capabilities.rs
```diff
@@ -11,17 +11,25 @@ use ark_ff::PrimeField;
...
+    // our case. Necessary for `set_assert_mapper`
     type Variable: Clone
         + std::ops::Add<Self::Variable, Output = Self::Variable>
         + std::ops::Sub<Self::Variable, Output = Self::Variable>
         + std::ops::Mul<Self::Variable, Output = Self::Variable>
         + std::ops::Neg<Output = Self::Variable>
         + From<u64>
-        + std::fmt::Debug;
+        + std::fmt::Debug
+        + 'static;
```
dannywillems, line 24: :thinking: not sure I like this, but let's see. What will have `'static` lifetime?
A value with `'static` lifetime will be embedded in the read-only data segment of the binary.

msm/src/circuit_design/composition.rs
querolita: Sure sure, not a blocking comment

msm/src/circuit_design/composition.rs
```diff
@@ -70,6 +73,29 @@ pub trait MPrism {
     fn re_get(&self, target: Self::Target) -> Self::Source;
 }
 
+/// Identity `MPrism` from any type `T` to itself.
+#[derive(Clone, Copy, Debug)]
+pub struct IdMPrism<T>(pub PhantomData<T>);
```
dannywillems, line 78: Do I understand correctly that it will be used to represent compositions of circuits that "have the same shape"?

ivc/src/ivc/interpreter.rs
volhovm: Done

msm/src/circuit_design/constraints.rs
```diff
@@ -40,9 +40,11 @@ impl<F: PrimeField, CIx: ColumnIndexer, LT: LookupTableID> ColAccessCap<F, CIx>
     fn assert_zero(&mut self, cst: Self::Variable) {
         // FIXME: We should enforce that we add the same expression
```
dannywillems, line 41: This comment can also be removed. The comment was regarding the constrain_index.

msm/src/circuit_design/constraints.rs
```diff
@@ -17,17 +17,17 @@ pub struct ConstraintBuilderEnv<F: PrimeField, LT: LookupTableID> {
     /// The index can be used to differentiate the constraints used by different
```
dannywillems, line 17: This can be removed, now.

ivc/src/ivc/interpreter.rs
```diff
@@ -245,11 +246,11 @@ where
 #[allow(clippy::type_complexity)]
 pub fn process_inputs<F, Ff, Env, const N_COL_TOTAL: usize>(
     env: &mut Env,
-    comms: [[(Ff, Ff); N_COL_TOTAL]; 3],
+    comms: [Box<[(Ff, Ff); N_COL_TOTAL]>; 3],
 ) -> (
-    [[[F; 2 * N_LIMBS_SMALL]; N_COL_TOTAL]; 3],
-    [[[F; 2 * N_LIMBS_LARGE]; N_COL_TOTAL]; 3],
-    [[[F; 2 * N_LIMBS_XLARGE]; N_COL_TOTAL]; 3],
+    Box<[[[F; 2 * N_LIMBS_SMALL]; N_COL_TOTAL]; 3]>,
```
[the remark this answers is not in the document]
volhovm: Hmmm, right now I was focusing more on the stack space, and I think memory considerations sound like an early optimisation. But in the future, maybe.

msm/src/circuit_design/witness.rs
```diff
@@ -43,6 +43,9 @@ pub struct WitnessBuilderEnv<
     /// value for row #j of the selector #i.
     pub fixed_selectors: Vec<Vec<F>>,
 
+    /// Function used to map assertions.
+    pub assert_mapper: Box<dyn Fn(F) -> F>,
```
dannywillems, line 47: I am struggling understanding the real advantage of this, and I am not sure I fully understand it for now.

msm/src/circuit_design/witness.rs
```diff
@@ -67,7 +70,11 @@ impl<
     type Variable = F;
 
     fn assert_zero(&mut self, cst: Self::Variable) {
-        assert_eq!(cst, F::zero());
+        assert_eq!((self.assert_mapper)(cst), F::zero());
```
dannywillems, line 73: Ah, is `assert_mapper` only used for the `assert_zero`, so you can constrain multiple expressions at the same time for a given selector, like in constraint_ivc?

ivc/src/ivc/interpreter.rs
```diff
@@ -913,15 +920,101 @@ where
...
+    constrain_selectors(env);
+
+    let s0 = env.read_column(IVCColumn::BlockSel(0));
+    env.set_assert_mapper(Box::new(move |x| s0.clone() * x));
+    constrain_inputs(env);
+
+    // TODO FIXME add constraints for hashes
+
+    let s2 = env.read_column(IVCColumn::BlockSel(2));
+    env.set_assert_mapper(Box::new(move |x| s2.clone() * x));
```
dannywillems, line 991: IIUC, it will, for each constraint in constraint_ecadds, add a selector in front of the whole expression, increasing by one the degree?

msm/src/circuit_design/witness.rs
```diff
@@ -301,26 +308,29 @@ impl<
...
     }
 
-    /// Generates proof inputs, repacking/collecting internal witness builder state.
-    pub fn get_proof_inputs(
-        &self,
-        domain: EvaluationDomains<F>,
-        lookup_tables_data: BTreeMap<LT, Vec<F>>,
-    ) -> ProofInputs<N_COL, F, LT> {
+    /// Sets all fixed selectors directly. Each item in `selectors` is
+    /// a vector of `domain_size` length.
```
dannywillems, line 326: Nit: should we add an assertion on the size of each element of `selectors` to verify this?

msm/src/circuit_design/witness.rs
```diff
@@ -352,6 +362,15 @@ impl<
             witness.cols[N_REL + N_DSEL + i] = self.fixed_selectors[i].clone();
```
dannywillems, line 362: Maybe it is modified later in https://github.com/o1-labs/proof-systems/pull/2249/files, but we should be sure that `set_fixed_selectors` has been called before.

msm/src/circuit_design/witness.rs
```diff
@@ -301,26 +308,29 @@ impl<
...
+    pub fn set_fixed_selectors(&mut self, selectors: Vec<Vec<F>>) {
+        self.fixed_selectors = selectors
+    }
+
+    pub fn get_relation_witness(&self, domain: EvaluationDomains<F>) -> Witness<N_COL, Vec<F>> {
         let domain_size: usize = domain.d1.size as usize;
+
         // Boxing to avoid stack overflow
         let mut witness: Box<Witness<N_COL, Vec<F>>> = Box::new(Witness {
             cols: Box::new(std::array::from_fn(|_| Vec::with_capacity(domain_size))),
```
dannywillems, line 336: Nit: I would always add an assert at the end of the function checking that the size of the vector is *exactly* `domain_size`. `with_capacity` does not enforce that.

ivc/src/ivc/interpreter.rs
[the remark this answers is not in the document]
dannywillems: Opening https://github.com/o1-labs/proof-systems/issues/2252 for the future, not a priority.

ivc/src/ivc/interpreter.rs
dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/2253, just to keep track of it, not a priority.

ivc/src/ivc/columns.rs
dannywillems: I don't understand. We can talk about it in a call, easier.

ivc/src/ivc/columns.rs
volhovm: Yes, the general description of each block is in the top of `columns.rs`. I don't know how to create `const usize` definitions for block sizes, because they depend on `N_COL_TOTAL` and `N_CHALS` and generic consts in rust are not supported....

msm/src/circuit_design/witness.rs
dannywillems: Fixed, got it.

ivc/src/ivc/interpreter.rs
volhovm: Maybe, but I'm not entirely convinced -- it seems to me that when we're folding the circuit we'll have to constrain these values. For just regular proofs -- yes, it makes sense. I suggest we keep it just in case (it's cheap), and can remove it later. I added a comment above this function. OK with you?

dannywillems: Ok to merge like this, but not needed. I would merge the PR, and we can move from there.

dannywillems (approved):

ivc/src/ivc/columns.rs
volhovm: I fixed this by adding `block_height` function... but I see it's merged? heh

ivc/src/ivc/interpreter.rs
volhovm: The selector will be `BlockSel(1)` but right now the constraints and witness for hashes is generated by the same function, so I guess I need to pass it a zero input or something. Not immediately obvious...

msm/src/circuit_design/witness.rs
volhovm: Agreed. I want to rework the design later on to remove `set_fixed_selectors` altogether and take selectors from some trait for the column.

PR #2240, opened 2024-05-17
Folding: use () instead of trace structures

dannywillems:
Since https://github.com/o1-labs/proof-systems/pull/2239, no ProvableTrace is required.


querolita (approved):
Awesome cleanup

PR #2243, opened 2024-05-17
add doc about degree 3 folding

fabrizio-m:


volhovm (changes requested):
@fabrizio-m could you please add all the equations to the doc here? Like how we compute these `t_0` and `t_1` especially.

dannywillems:
@fabrizio-m: we can discuss about [this](https://github.com/o1-labs/proof-systems/pull/2243#pullrequestreview-2065856409) tomorrow at the offsite. I am happy to guide you on this. We should go through the codebase and the equations together.

folding/src/lib.rs
```diff
@@ -20,7 +20,37 @@
 //! expressions in orther to support multiple constraints.
 //! That results in 2 errors terms t_0 and t_1 and a slightly different computation
```
querolita, line 21: results in 2 error terms

dannywillems:
Most of the work in https://github.com/o1-labs/proof-systems/pull/2259 has been done by Fabrizio, I was guided by his explanation and he answered my question. I would suggest closing this one, and move on with 2259.

PR #2244, opened 2024-05-17
Folding: remove rows method in Witness trait

dannywillems:
It seems it was only used for checking.
Using a parameter domain is more scalable, and let the user pick any kind of domain they want. It doesn't require the user to implement a new Witness implementation.

Driving the road to use generically the `GenericWitness` structure.


volhovm:
(sorry, I thought the base branch was merged)

PR #2245, opened 2024-05-17
Implement FoldingWitness for MSM/Witness

dannywillems:


msm/src/witness.rs
```diff
@@ -158,3 +159,8 @@ impl<const N: usize, F: FftField> Foldable<F>
         a
     }
 }
+
+impl<const N: usize, G: CommitmentCurve> FoldingWitnessT<G>
```
volhovm, line 163: Is it /planned/ to be an empty trait, or is it going to get some methods in the future?..

volhovm (approved):

msm/src/witness.rs
dannywillems: I think the trait should not have a pre-defined implementation of `relax`. I would like to change this in the (near) future.

PR #2246, opened 2024-05-17
Folding: remove alphas from FoldingEnv as it is obsolete

dannywillems:
The alphas can be inferred while evaluating. It removes a method to implement within the folding environment. For the challenges, it will be done in a follow-up PR.

### DIFF
--- a/folding/src/error_term.rs
+++ b/folding/src/error_term.rs
@@ -4,7 +4,7 @@ use crate::{
     eval_leaf::EvalLeaf,
     expressions::{Degree, FoldingExp, IntegratedFoldingExpr, Sign},
     quadraticization::ExtendedWitnessGenerator,
-    FoldingConfig, FoldingEnv, RelaxedInstance, RelaxedWitness, ScalarField,
+    FoldingConfig, FoldingEnv, Instance, RelaxedInstance, RelaxedWitness, ScalarField,
 };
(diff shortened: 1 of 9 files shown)


querolita (approved):

folding/src/error_term.rs
```diff
@@ -187,13 +187,22 @@ pub(crate) fn compute_error<C: FoldingConfig>(
...
+        .get_alphas();
+
     let t_0 = {
         let t_0 = (zero(), zero());
         let (l, r) = exp.degree_0.iter().fold(t_0, |(l, r), (exp, sign, alpha)| {
             //could be left or right, doesn't matter for constant terms
             let exp = eval_exp_error(exp, env, Side::Left);
-            let alpha_l = env.inner().alpha(*alpha, Side::Left);
-            let alpha_r = env.inner().alpha(*alpha, Side::Right);
+            let alpha_l = alphas_l.get(*alpha).expect("alpha not present");
```
volhovm, line 204: So now alphas are part of a relaxed instances as opposed to the environment. Makes sense.

volhovm (approved):

folding/src/error_term.rs
dannywillems: They have always been part of a relaxed instance.

PR #2247, opened 2024-05-20
IVC: Add array allocators/helpers: reduce stack usage

volhovm:
Previous PR (https://github.com/o1-labs/proof-systems/pull/2233) has suffered from stack overflow because I used big arrays on stack. 

This PR: 
* Adds some helpers to deal with arrays without using stack (best of what I know! can be probably improved). 
* Rewrites some IVC code to rely on these, so decreases stack usage significantly.

### DIFF
--- a/ivc/src/ivc/mod.rs
+++ b/ivc/src/ivc/mod.rs
@@ -127,7 +127,7 @@ mod tests {
         let mut rng = o1_utils::tests::make_test_rng();
         build_ivc_circuit::<_, IVCLookupTable<Ff1>, _>(
             &mut rng,
-            1 << 8,
+            1 << 15,
             IdMPrism::<IVCLookupTable<Ff1>>::default(),
         );
     }
--- a/ivc/src/ivc/mod.rs
+++ b/ivc/src/ivc/mod.rs
@@ -134,8 +134,6 @@ mod tests {
     fn test_completeness_ivc() {
         let mut rng = o1_utils::tests::make_test_rng();
 
-        //let mut comms_left: Box<_> = box_array2![(Fp::zero(),Fp::zero()); TEST_N_COL_TOTAL; 3];
-
         let domain_size = 1 << 15;
         let domain = EvaluationDomains::<Fp>::create(domain_size).unwrap();
 
--- a/ivc/src/ivc/mod.rs
+++ b/ivc/src/ivc/mod.rs
@@ -29,6 +29,7 @@ mod tests {
         witness::Witness,
         BaseSponge, Ff1, Fp, OpeningProof, ScalarSponge, BN254,
     };
+    use o1_utils::box_array;
     use poly_commitment::pairing_proof::PairingSRS;
     use rand::{CryptoRng, RngCore};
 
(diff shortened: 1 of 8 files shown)


ivc/src/ivc/mod.rs
```diff
@@ -136,6 +134,8 @@ mod tests {
     fn test_completeness_ivc() {
         let mut rng = o1_utils::tests::make_test_rng();
 
+        //let mut comms_left: Box<_> = box_array2![(Fp::zero(),Fp::zero()); TEST_N_COL_TOTAL; 3];
```
dannywillems, line 137: Can be removed.

utils/src/array.rs
```diff
@@ -0,0 +1,122 @@
...
+///     let _: Box<[u8; 1024]> = box_array![0; 1024];
+/// ```
+///
+/// See
+/// <https://stackoverflow.com/questions/25805174/creating-a-fixed-size-array-on-heap-in-rust/68122278#68122278>
+#[macro_export]
+macro_rules! box_array {
+    ($val:expr ; $len:expr) => {{
+        // Use a generic function so that the pointer cast remains type-safe
+        fn vec_to_boxed_array<T>(vec: Vec<T>) -> Box<[T; $len]> {
```
dannywillems, line 52: As it does use unsafe, I would use `unsafe_` as a prefix in the function name, so the user is aware of this.

utils/src/array.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub type BaseField = <CurvePoint as AffineCurve>::BaseField;
+
+    #[test]
+    /// Tests whether initialising different arrays creates a stack
+    /// overflow. The usual default size of the stack is 128kB.
+    fn test_boxed_stack_overflow() {
+        // Each point is assumed to be 256 bits, so 512 points is
+        // 16MB. This often overflows the stack if created as an
+        // array.
+        let _boxed: Box<[[BaseField; 256]; 1]> =
```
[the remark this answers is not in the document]
querolita: Maybe he refers to the comment upfront, that it does not really take 16MB 🤔  Just guessing

dannywillems:
Merging, but not really happy with the tests as I said.

volhovm:
@dannywillems we can try to make a negative test, that is with smaller stack size this code fails. I can try, but I'm afraid it will instead crash the whole test suite, not just single test...

volhovm:
I wrote a negative test that creates stackoverflow, but indeed it just crashes the whole `cargo test` so I dunno how to test this tbh.

```
ї> RUST_MIN_STACK=1048576 cargo test --package o1-utils --lib -- array::tests::test_boxed_stack_overflow_fail --exact --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.12s
     Running unittests src/lib.rs (target/debug/deps/o1_utils-4f6b0c7216c06b3a)

running 1 test

thread 'array::tests::test_boxed_stack_overflow_fail' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `-p o1-utils --lib`

Caused by:
  process didn't exit successfully: `/home/volhovm/code/proof-systems/target/debug/deps/o1_utils-4f6b0c7216c06b3a 'array::tests::test_boxed_stack_overflow_fail' --exact --nocapture` (signal: 6, SIGABRT: process abort signal)
```

volhovm:
@dannywillems ok so yes these methods are not working really well. I think they cache some part of the process still :( Will have to improve.

PR #2248, opened 2024-05-20
WIP: get rid of challenge in folding environment

dannywillems:
Breaking things.
It is not a good time to improve this for now.
We should parametrize the expression framework defined in kimchi to use a generic type of challenge, and after that parametrize the folding expressions and this code.

The issue for now comes from the conversions.


dannywillems:
Closing. Old

PR #2249, opened 2024-05-20
Separate fixed selectors from the witness

volhovm:
With this PR, fixed selectors are not anymore part of the witness, and are passed separately. The intention is that they must be static for a given circuit.

### DIFF
--- a/msm/src/prover.rs
+++ b/msm/src/prover.rs
@@ -1,3 +1,6 @@
+#![allow(clippy::type_complexity)]
+#![allow(clippy::boxed_local)]
+
 use crate::{
     column_env::ColumnEnvironment,
     expr::E,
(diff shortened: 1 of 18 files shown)


msm/src/prover.rs
```diff
@@ -417,6 +482,16 @@ where
         .map(|poly| (coefficients_form(poly), hiding(1)))
         .collect();
 
+    // @volhovm: I'm not sure we need to prove opening of fixed
```
dannywillems, line 485: They should be treated as "common preprocessed input" from PlonK: https://eprint.iacr.org/2019/953.pdf, page 27.

dannywillems:
@volhovm I messed up with the merge, can you push force the old version, please?

dannywillems:
Oh, merging previous PR fixed the mess I made. Fixing conflicts again :)

volhovm:
@dannywillems merged the branch = fixed conflicts. Feel free to review. If you feel like there are more things to do I can do another PR round.

dannywillems (approved):
Accepting to move on. However, we should have an "index" or a setup phase as specified before.

PR #2259, opened 2024-05-29
Folding: more explanation of the addition of the u-term

dannywillems:


folding/src/expressions.rs
```diff
@@ -24,6 +24,186 @@
...
+//!   S(X_{1}, X_{2}, X_{3}, u, α_{1}, α_{2})
+//! = P'(X_{1}, X_{2}, X_{3}, u, α_{1}) + Q'(X_{1}, X_{2}, X_{3}, u, α_{2})`.
+//! = α_{1} X_{1} X_{2} +
+//!   α_{1} u^2 -
+//!   α_{1} u X_{3} +
+//!   α_{2} u X_{1} +
+//!   α_{2} u X_{2}
+//! ```
+//!
+//! Note that we end up with everything of the same degree, which is `3` in this
```
querolita, line 166: Maybe here it should be clarified that the powers of alpha are considered as degree 1 so the degree does not grow further than 3.

folding/src/expressions.rs
```diff
@@ -24,6 +24,186 @@
...
+//! ```text
+//!   (X_{1} + r X'_{1}) (X_{2} + r X'_{2}) \
+//! + (u + r u') [(u + r u') - (X_{3} + r X'_{3})]
+//! ```
+//! which can be simplified into:
+//! ```text
+//!   P_relaxed(X_{1}, X_{2}, X_{3}, u) + P_relaxed(r X_{1}', r X_{2}', r X_{3}', r u')
+//! + r [u (u' - X_{3}) + u' (u - X_{3})] + r [X_{1} X_{2}'   +   X_{2} X_{1}']
+//!   \---------------------------------/   \----------------------------------/
+//!  cross terms of monomials of degree 1   cross terms of monomials of degree 2
```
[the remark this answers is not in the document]
dannywillems: Yes, you are right.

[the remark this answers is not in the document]
dannywillems: What do you think of e8674cdc2c01451a7849bc7192f9e29c063c802e?

folding/src/expressions.rs
dannywillems: What do you think of e8674cdc2c01451a7849bc7192f9e29c063c802e?

querolita (approved):

folding/src/expressions.rs
```diff
@@ -24,6 +24,189 @@
...
+//! = α_{1} X_{1} X_{2} +
+//!   α_{1} u^2 -
+//!   α_{1} u X_{3} +
+//!   α_{2} u X_{1} +
+//!   α_{2} u X_{2}
+//! ```
+//!
+//! Note that we end up with everything of the same degree, which is `3` in this
+//! case. The variables `α_{1}` and `α_{2}` increase the degree of the
+//! homogeneous expressions by one.
```
volhovm, line 170: It'd be nice to add a remark later that for strict instances, during the process of their relaxation we're free to pick `alpha` challenges in whatever way as long as they are provably pseudorandom. Our subjective decision here is to pick `alpha_1 = Hash({C_{R,i}})`, and `alpha_i = alpha_1^i`.

volhovm:
Great stuff!

PR #2260, opened 2024-05-29
Fix allocation in o1utils array module

volhovm:
Fixes https://github.com/o1-labs/proof-systems/pull/2247. The functions before were still overflowing the stack. That was happening because e.g. `vec_of_slices2.iter().map(...)` was dereferencing inside the map (https://github.com/o1-labs/proof-systems/pull/2260/files#diff-892a56213c28a7749b5e25aa89c7067aac3c15d0d5d86293a00cb720ef88c6c4L20), thus bringing the value into the stack again. In practice, each element of that vector was put on stack and then added from stack to the vector again.

Main changes:
- Implement converters properly. Now new vectors are created from a slice, so from a memory reference directly.
- Tests changed to create much bigger arrays (that actually fail on `master`)

Introduces `Clone` but I haven't found another way. It's honestly a bit tough with rust that I can't just manipulate memory directly...

### DIFF
--- a/utils/src/array.rs
+++ b/utils/src/array.rs
@@ -7,33 +7,36 @@
 //! (non-fixed-sized) vectors.
 
 /// Converts a two-dimensional vector to a constant sized two-dimensional array.
-pub fn vec_to_boxed_array2<T, const N: usize, const M: usize>(
+pub fn vec_to_boxed_array2<T: Clone, const N: usize, const M: usize>(
     vec: Vec<Vec<T>>,
 ) -> Box<[[T; N]; M]> {
-    let vec_of_slices2: Vec<[T; N]> = vec
-        .into_iter()
-        .map(|x: Vec<T>| {
-            let y: Box<[T]> = x.into_boxed_slice();
-            let z: Box<[T; N]> = y
-                .try_into()
-                .unwrap_or_else(|_| panic!("vec_to_boxed_array2: length mismatch inner array"));
-            *z
-        })
-        .collect();
-    let array: Box<[[T; N]; M]> = vec_of_slices2
+    let mut vec_of_slices2: Vec<[T; N]> = vec![];
+    vec.into_iter().for_each(|x: Vec<T>| {
+        let y: Box<[T]> = x.into_boxed_slice();
+        let z: Box<[T; N]> = y
+            .try_into()
+            .unwrap_or_else(|_| panic!("vec_to_boxed_array2: length mismatch inner array"));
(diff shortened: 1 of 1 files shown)


dannywillems:
You can manipulate memory directly with `std::mem`

dannywillems (commented):
I don't understand the differences.

volhovm:
@dannywillems I clarified the description in the top post, how does this sound?

utils/src/array.rs
```diff
@@ -7,33 +7,36 @@
...
-        .collect();
-    let array: Box<[[T; N]; M]> = vec_of_slices2
+    let mut vec_of_slices2: Vec<[T; N]> = vec![];
+    vec.into_iter().for_each(|x: Vec<T>| {
+        let y: Box<[T]> = x.into_boxed_slice();
+        let z: Box<[T; N]> = y
+            .try_into()
+            .unwrap_or_else(|_| panic!("vec_to_boxed_array2: length mismatch inner array"));
+        let zz: &[[T; N]] = std::slice::from_ref(z.as_ref());
+        vec_of_slices2.extend_from_slice(zz);
```
dannywillems, line 20: From https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice
```
Iterates over the slice other, clones each element, and then appends it to this Vec. The other slice is traversed in-order.
```
Then the `Clone` requirement.

dannywillems:
I don't fully like it because of the clone. However, we must move one, and therefore I will accept.

The problem should be solvable easily by manipulating directly the memory with `std::mem` à la C (or higher rust-y interfaces of it), at the cost of being less generic over the type `T`. Note that we do not need the generic aspect of these functions. We should only need memory manipulation for field, commitments, and maybe other flat values, where `memcpy` and continuous flat arrays `Box<[T; N + M]` should be enough (and which can be done using macro?).

dannywillems (approved):

utils/src/array.rs
volhovm: Yep, exactly. Although I think "clones each element" here is misleading: the description seems to imply that it does put the element on stack, but it actually extends directly from memory. Which was the only way I found to add anything to a vector directly from memory.

volhovm:
Yep, doing it without `Clone` should be possible, I'm fairly sure, but I've spent a lot of time trying to search for a ready solution and haven't found any, and everything with `unsafe` I don't understand enough atm to confidently add to our system. So for the future.

PR #2267, opened 2024-05-30
Folding: output the relaxed extended instances

dannywillems:
It is useful information to pass to the IVC circuit. I would suggest to improve the interface later.

Adding also documentation.

Regarding the usage of `clone`, I am not happy with it. It will imply a possible non-negligeable memory overhead. We should avoid this when the first version is released.

### DIFF
--- a/folding/tests/test_decomposable_folding.rs
+++ b/folding/tests/test_decomposable_folding.rs
@@ -385,7 +385,11 @@ fn test_decomposable_folding() {
         let FoldingOutput {
             folded_instance,
             folded_witness,
-            ..
+            t_0: _,
+            t_1: _,
+            relaxed_extended_left_instance: _,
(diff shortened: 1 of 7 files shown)


folding/tests/test_decomposable_folding.rs
```diff
@@ -385,7 +385,11 @@ fn test_decomposable_folding() {
         let FoldingOutput {
             folded_instance,
             folded_witness,
-            ..
+            t_0: _,
```
dannywillems, line 388: It is a personal taste to explicitly unwrap it totally. It might be a pain when we add a field, but it gives information on the API. If there are many changes when unwrapping a structure, it means the API is not well implemented.

fabrizio-m (approved):

PR #2270, opened 2024-05-30
simpler folding

fabrizio-m:


dannywillems:
Can you have a look at the CI, please?

folding/src/columns.rs
```diff
@@ -26,3 +25,53 @@ pub enum ExtendedFoldingColumn<C: FoldingConfig> {
     /// while folding/accumulating multiple expressions.
     Selector(C::Selector),
 }
+
+impl<C: FoldingConfig> std::hash::Hash for ExtendedFoldingColumn<C> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        core::mem::discriminant(self).hash(state);
```
dannywillems, line 31: Only a comment: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations. Good reading.

dannywillems (changes requested):
We noticed that some tests are not passing, even though, logically, we think it should. See https://github.com/o1-labs/proof-systems/pull/2284.

I would be interested in having more granular testing for the hash, like maybe checking with a hashmap, to be sure that we have the expected behavior.

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+
+#[derive(Clone, Default)]
+/// Default type for when you don't need structure
+pub struct EmptyStructure<G: KimchiCurve>(PhantomData<G::ScalarField>);
+
+impl<G: KimchiCurve, Col> Index<Col> for EmptyStructure<G> {
+    type Output = Vec<G::ScalarField>;
+
+    fn index(&self, _index: Col) -> &Self::Output {
+        panic!("shouldn't reach this point, as this type only works with witness-only constraint systems");
```
dannywillems, line 20: By that, you mean that there is no some kind of selectors?

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+}
+
+/// An standard folding config that supports:
```
dannywillems, line 24:
```suggestion
/// A standard folding config that supports:
```

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+/// `G`: any curve
+/// `Col`: any column implementing [FoldingColumnTrait]
+/// `Chall`: any challenge
+/// `Sel`: any dynamic selector
+/// `Str`: structures that can be indexed by `Col`, thus implementing `Index<Col>`
+/// `I`: instances (implementing [Instance]) that can be indexed by `Chall`
+/// `W`: witnesses (implementing [Witness]) that can be indexed by `Col` and `Sel`
+/// ```ignore
```
dannywillems, line 32: Can we avoid ignoring it? So we build all the time with the doc and check it is up-to-date?

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+/// use ark_poly::{EvaluationDomain, Radix2EvaluationDomain};
+/// use mina_poseidon::FqSponge;
+/// use folding::{examples::{BaseSponge, Curve, Fp}, FoldingScheme};
+///
+/// // instanciating the config with our types and the defaults for selectors and structure
+/// type MyConfig = StandardConfig<Curve, MyCol, MyChallenge, MyInstance<Curve>, MyWitness<Curve>>;
+/// let constraints = vec![constraint()];
```
dannywillems, line 39: Ah, ok, we would need a lot of different value definitions like `constraint`

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+    I: Instance<G> + Index<Chall, Output = G::ScalarField> + Clone,
+    W: Witness<G> + Clone,
+    W: Index<Col, Output = Vec<G::ScalarField>> + Index<Sel, Output = Vec<G::ScalarField>>,
+    Col: Hash + Eq,
+{
+    instances: [I; 2],
+    witnesses: [W; 2],
+    next_evals: ColumnMemoizer<Col, G::ScalarField, 10>,
+    structure: Str,
+    _todo: PhantomData<(G, Col, Chall, Sel, Str)>,
```
dannywillems, line 133: todo?

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+                        let mut next = Vec::with_capacity(evals.len());
+                        next.extend(evals[1..].iter());
+                        next.push(evals[0]);
+                        next
+                    };
+                    self.next_evals.get_or_insert(col, f)
+                }
+            }
+        } else {
+            &self.structure[col]
```
dannywillems, line 204: ok nice, that's for the selectors the `is_witness`. Can we document that in the doc of `is_witness`, please?

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+                    let i = i - N;
+                    let new = || Box::new(Self::new());
+                    let next = self.next.get_or_init(new);
+                    next.get_or_insert(i, f)
+                }
+            }
+        }
+    }
+    /// a hashmap like data structure supporting get-or-insert with
+    /// an inmutable reference and returning an inmutable reference
```
dannywillems, line 265:
```suggestion
    /// an immutable reference and returning an inmutable reference
```

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+        instance[challenge]
+    }
+
+    fn col(&self, col: Col, curr_or_next: CurrOrNext, side: Side) -> &Vec<G::ScalarField> {
+        let witness = match side {
+            Side::Left => &self.witnesses[0],
+            Side::Right => &self.witnesses[1],
+        };
+        // this should hold as long the Index implementations are consistent with the
+        // FoldingColumnTrait implementation.
```
dannywillems, line 180: Feel free to use `FoldingColumnTrait` to link to the link, and detect old definitions while building the doc.

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+    W: Witness<G> + Clone,
+    W: Index<Col, Output = Vec<G::ScalarField>> + Index<Sel, Output = Vec<G::ScalarField>>,
+    Col: FoldingColumnTrait + Eq + Hash,
+    Sel: Copy,
+    Str: Clone + Index<Col, Output = Vec<G::ScalarField>>,
+{
+    type Structure = Str;
+
+    fn new(structure: &Self::Structure, instances: [&I; 2], witnesses: [&W; 2]) -> Self {
+        // cloning for now, ideally should work with references, but that requires deeper
```
dannywillems, line 151: +1

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+            // search the segments using the id
+            let mut ids = self.ids.borrow_mut();
+            let new_id = || self.next.fetch_add(1, Ordering::Relaxed);
+            let id = ids.entry(col).or_insert_with(new_id);
+            self.first_segment.get_or_insert(*id, f)
+        }
+    }
+}
+
+#[cfg(test)]
```
dannywillems, line 298: I start having a preference to move tests into the subdirectory `tests`. I would move it in `tests/test_standard_config.rs`. But not blocking ofc.

folding/src/standard_config.rs
```diff
@@ -0,0 +1,471 @@
...
+        let b = column(MyCol::B);
+        let c = column(MyCol::C);
+        let beta = chall(MyChallenge::Beta);
+        let gamma = chall(MyChallenge::Gamma);
+        (a + b - c) * (beta + gamma)
+    }
+
+    // here an example of how the config would be used, which is similar to any
+    // other custom config.
+    #[allow(dead_code)]
```
dannywillems, line 446: Maybe having this as a `test` and do a relatively “trivial” check?

folding/src/standard_config.rs
fabrizio-m: not in any case that makes use of this particular type, as it is only for cases where you don't need structure. I you have selectors you should use your own structure and implement `Index` accordingly.

folding/src/standard_config.rs
fabrizio-m: It is ignored because it isn't working code, it wouldn't even build as some of the types there are not defined.
I can later rewrite some of the examples we have using this config and make them into a doc, but I would prefer to do it in another PR.

folding/src/standard_config.rs
fabrizio-m: I don't think it should be in `is_witness`, as this property allowing to use `is_witness` to decide between witness and structure is more a particularity of this config, and not a general requirement for any witness/columns.

folding/src/standard_config.rs
fabrizio-m: right, I actually wanted just `#[cfg(feature = "bn254")]`, as I don't have any proper test, just a function that shows the code builds, but I don't run it as that would creating the pairs and some prover/checker, which is already covered in the tests we have.

folding/src/standard_config.rs
fabrizio-m: it doesn't work as a test, as I don't instance witness pairs to use as inputs, I made it just to show how it would be used and check there is no issue building it. But for a proper test I think it would be better to rewrite some of the ones we already have with this generic config.

folding/src/standard_config.rs
fabrizio-m: I'll remove `#[cfg(test)]` for nwo

folding/src/standard_config.rs
dannywillems: Sure, not urgent at all. I noticed it in [this comment[(https://github.com/o1-labs/proof-systems/pull/2270/files#r1626589975).

folding/src/standard_config.rs
dannywillems: Yes, you're right. I didn't know that rustdoc wasn't checking `//` :(.

folding/src/standard_config.rs
dannywillems: Agreed.

dannywillems (approved):

PR #2271, opened 2024-05-30
Folding: simplify the interface and rename fields

dannywillems:
The folding interface works as follow:
An original instance is extended using quadriticization. When the instance has been extended, it is relaxed.

A relaxed instance will therefore contain the extended instance, and the extended instance will contain the extra columns added by the quadriticization process.

Before the changes, the interface was using the concept of "inner", and was defining methods on structures to get reference to different fields. It does add unecessary code, and also using shared terms between structure complexify the understanding.

This commit aims to simplify the reading of the interface, by using explicit field name. Also, it makes the fields of the structure accessible publicly, and allows to remove some methods that adds "noise" to the codebase. References are automatically infered, and therefore functions returning references are not required.


folding/src/checker.rs
```diff
@@ -69,8 +69,8 @@ impl<C: FoldingConfig> ExtendedProvider<C> {
         witness: RelaxedWitness<C::Curve, C::Witness>,
     ) -> Self {
         let inner_provider = {
-            let instance = instance.inner_instance().inner.clone();
-            let witness = witness.inner().inner.clone();
+            let instance = instance.extended_instance.instance.clone();
```
querolita, line 72: Will create a follow-up PR to get rid of `extended` in between.

querolita (approved):

folding/src/checker.rs
dannywillems: I agree it would be nice to have a way to "hide" the extension from the "relaxed instance", but it might (not sure, yet) to know which and how many columns have been added by quadri.

PR #2273, opened 2024-05-31
Folding: add additional doc on how polynomial IOPs are handled

dannywillems:


folding/src/expressions.rs
```diff
@@ -207,6 +207,55 @@
...
+//! online
+//! documentation](https://o1-labs.github.io/proof-systems/rustdoc/kimchi_msm/logup/index.html).
+//! We will suppose we have only one table `T` and Alice wants to prove to Bob
+//! that she knows that all evaluations of `f(X)` is in `t(X)`. The additive
+//! lookup argument is described by the polynomial equation:
+//! ```text
+//! β + f(x) = m(x) (β + t(x))
+//! ```
+//! where β is the challenge, `f(x)` is the polynomial whose evaluations describe
+//! the value Alice wants to prove to Bob that they are in the table, `m(x)` is
```
querolita, line 237:
```suggestion
//! the value Alice wants to prove to Bob that is in the table, `m(x)` is
```

querolita (approved):

PR #2274, opened 2024-05-31
Fix constraints of `SyscallReadPreimage`

querolita:
This PR solves the (several) failing constraints that were happening in the `SyscallReadPreimage` instruction. In particular, this PR mainly changes parts of the `request_preimage_write()` function, both in the witness and constraints side. It also refactors the purpose of some witness columns of the scratch state, which were not being used properly.

In particular, these are the changes that have been addressed:

- Removal of `is_syscall` and `reading_preimage` flags:
  - These were meant to differentiate between rows performing oracle syscalls (vs. any other MIPS instruction) and whether or not the current row was reading any preimage byte or not (if not, it meant it was reading only the bytelength). 
  - `is_syscall` is no longer necessary now that we are going to be creating one separate circuit for each separate instruction. Meaning, there is no need to include the "selector" inside the actual constraints, but it will be added later by the tracer for folding.
  - `reading_preimage` is not used anymore: it has been replaced with an equivalent combination of other witness cells (it is reading the preimage as long as it reads at least 1 preimage byte: reading 0 bytes means it is reading the length).
 
- Fix confusion between number of total bytes read from the oracle (including bytelength), total bytes read from the preimage, bytes read in the current instruction (could include length bytes), and bytes read in the current instruction from the preimage.
  - The reason some constraints were failing is that I had misunderstood what each variable was representing. In the constraints side, we mainly care about the preimage bytes processed in the current row, but other values were being used instead.
  - Change the returned value of the constraints side to match that of the witness side. Meaning, it returns in `pos` the number of bytes read in the current instruction (could include both length bytes and preimage bytes). This is then used in the interpreter to update the total number of bytes read for the offset of the following call.

- Fixed the expression returning whether the exact number of preimage bytes were read or not. 
  - This is a high-degree variable that is nonzero when N bytes are read (and zero otherwise). Nonetheless, after removing the `reading_preimage` flag,  one needs to include the case that 0 bytes are read (because it was only reading the length). If not, then the chunk constraints would not be satisfied in those cases.

- In the witness side, fix the byte index and compute the chunk of preimage being read.
  - Here again, there was a confusion with what each variable represented, so bytes read were not being stored in the right place of the scratch state, and the 4-byte chunk was not being computed anywhere. 

- Remove the number of bytes left that remain. 
  - This was not being used to constrain anything (like, it was not computed as the subtraction of the previous amount and the current number of bytes read or anything close).
  - Instead, this was just being used as a flag to indicate when the preimage was finished. But now there's the `end_of_preimage` flag. 

- Fit to 80-char lines.

- Added a boolean helper (unused right now) that could be useful at some point if we wanted to optimize any high degree constraints to check if a variable is zero. 

- Moved the unit tests of MIPS instructions to their own module inside the tests, foreseeing that there will be more in there shortly.

TODO: I left a bunch of comments with ideas to improve the code. 
- And in particular, potential extra constraints that we might have to include in order to link the "total number of bytes" with the "preimage number of bytes". 
- Also, comments about whether we want to constrain anything about the length bytes read or not (like, checking they are 8-bits in length). Preimage bytes are checked to be so on the Keccak side, but that is not done for the length bytes right now. Doing so would imply having to add some columns to store length bytes and add lookups for them.
- Another one about how we want to check that the end of the preimage has actually arrived or if setting `end_of_preimage=1` is sufficient.


o1vm/src/mips/constraints.rs
```diff
@@ -44,14 +44,13 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
-    // of columns is defined upfront by crate::mips::witness::SCRATCH_SIZE.
+    // Use one of the available columns. It won't create a new column every time
+    // this function is called. The number of columns is defined upfront by
+    // crate::mips::witness::SCRATCH_SIZE.
     fn alloc_scratch(&mut self) -> Self::Position {
         // All columns are implemented using a simple index, and a name is given
-        // to the index.
-        // See crate::SCRATCH_SIZE for the maximum number of columns the circuit
-        // can use.
+        // to the index. See crate::SCRATCH_SIZE for the maximum number of
```
dannywillems, line 52:
```suggestion
        // to the index. See `crate::SCRATCH_SIZE` for the maximum number of
```

dannywillems: Only to check that the doc is up-to-date when building it.

o1vm/src/mips/constraints.rs
```diff
@@ -44,14 +44,13 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
     /// variable/input of our circuit.
     type Position = MIPSColumn;
 
-    // Use one of the available columns. It won't
-    // create a new column every time this function is called. The number
-    // of columns is defined upfront by crate::mips::witness::SCRATCH_SIZE.
+    // Use one of the available columns. It won't create a new column every time
+    // this function is called. The number of columns is defined upfront by
+    // crate::mips::witness::SCRATCH_SIZE.
```
dannywillems, line 49:
```suggestion
    // `crate::mips::witness::SCRATCH_SIZE`
```

o1vm/src/mips/constraints.rs
```diff
@@ -374,75 +373,95 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         _len: &Self::Variable,
```
dannywillems, line 373: Would be nice in the near future to add documentation on what `request_preimage_write` is supposed to do, on the constraint side.

o1vm/src/mips/constraints.rs
```diff
@@ -374,75 +373,95 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
-                * (row_bytes.clone() - Expr::from(3));
-
-            // Note there is no need to multiply by the Syscall flag because the constraints are zero when the witnesses are zero
+            for var in has_n_bytes.iter() {
+                self.assert_boolean(var.clone());
+            }
+            self.assert_boolean(end_of_preimage.clone());
+        }
+
+        {
```
dannywillems, line 418: +1

o1vm/src/mips/constraints.rs
```diff
@@ -374,75 +373,95 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
-                // Constrain the byte decomposition of the preimage chunk
-                // TODO: smaller degree?
-                // When only 1 byte is read, the chunk is equal to the byte[0]
-                self.constraints.push(
-                    reading_preimage.clone()
-                        * read_1.clone()
-                        * (this_chunk.clone() - bytes[0].clone()),
-                );
-                // When 2 bytes are read, the chunk is equal to the byte[0] * 2^8 + byte[1]
+                // Constrain the byte decomposition of the preimage chunk When
```
dannywillems, line 441: Looks like the only change is `reading_preimage` that disappears (expected).

o1vm/src/mips/constraints.rs
```diff
@@ -451,58 +470,45 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
-                }
-            }
-
-            // Constrain the bytes flags depending on the number of bytes read in this row
+            // Constrain the bytes flags depending on the number of bytes read
+            // in this row
             {
-                // When at least has_1_byte, then any number of bytes can be read
-                // <=> Check that you can only read 1, 2, 3 or 4 bytes
+                // When at least has_1_byte, then any number of bytes can be
```
dannywillems, line 476: Same here. Only important change is `read_preimage` has been removed (expected).

o1vm/src/mips/constraints.rs
```diff
@@ -511,28 +517,30 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
                 ],
             ));
         }
-        // COMMUNICATION CHANNEL: Read hash output
+        // COMMUNICATION CHANNEL: Read hash output FIXME: is it a problem that
```
dannywillems, line 520: Nit: I suppose the `FIXME` was supposed to be on the next line?

dannywillems: To answer to the FIXME: we do only care about the 248 first bits. The last byte is reserved by OP.

o1vm/src/mips/tests.rs
```diff
@@ -72,9 +73,10 @@ fn test_mips_number_constraints() {
                 | MultiplyUnsigned | Div | DivUnsigned => assert_num_constraints(&instr, 6),
                 SyscallOther => assert_num_constraints(&instr, 10),
                 SyscallMmap => assert_num_constraints(&instr, 11),
-                SyscallReadPreimage => assert_num_constraints(&instr, 21),
-                SyscallFcntl => assert_num_constraints(&instr, 22),
+                SyscallFcntl | SyscallReadPreimage => assert_num_constraints(&instr, 22),
                 SyscallWritePreimage => assert_num_constraints(&instr, 30),
+                // FIXME: for some reason it does not matter if we comment out
```
dannywillems, line 78: I'll need a call tomorrow to understand this.

o1vm/src/mips/tests.rs
```diff
@@ -234,8 +225,7 @@ mod folding {
         // same register
         let reg_src = 1;
         let reg_dest = 2;
-        // Instruction: 0b00100100001000010110110011101000
-        // addiu $at, $at, 27880
+        // Instruction: 0b00100100001000010110110011101000 addiu $at, $at, 27880
```
dannywillems, line 228: I would leave it as before.

o1vm/src/mips/tests.rs
```diff
@@ -261,11 +251,8 @@ mod folding {
         let mut rng = o1_utils::tests::make_test_rng();
         // lw instruction
         let mut dummy_env = dummy_env(&mut rng);
-        // Instruction: 0b10001111101001000000000000000000
-        // lw $a0, 0(29)
-        // a0 = 4
-        // Random address in SP
-        // Address has only one index
+        // Instruction: 0b10001111101001000000000000000000 lw $a0, 0(29) a0 = 4
```
dannywillems, line 254: I would leave it as before.

o1vm/src/mips/tests.rs
```diff
@@ -297,8 +284,7 @@ mod folding {
         let mut rng = o1_utils::tests::make_test_rng();
         // We only care about instruction parts and instruction pointer
         let mut dummy_env = dummy_env(&mut rng);
-        // Instruction: 0b10001111101001000000000000000000
-        // addi	a1,sp,4
+        // Instruction: 0b10001111101001000000000000000000 addi a1,sp,4
```
dannywillems, line 287: I would leave it as before.

o1vm/src/mips/tests.rs
```diff
@@ -322,8 +308,7 @@ mod folding {
         let mut rng = o1_utils::tests::make_test_rng();
         // We only care about instruction parts and instruction pointer
         let mut dummy_env = dummy_env(&mut rng);
-        // Instruction: 0b00111100000000010000000000001010
-        // lui at, 0xa
+        // Instruction: 0b00111100000000010000000000001010 lui at, 0xa
```
dannywillems, line 311: I would leave it as before.

o1vm/src/mips/witness.rs
```diff
@@ -143,9 +144,9 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
     }
 
     fn add_constraint(&mut self, _assert_equals_zero: Self::Variable) {
-        // No-op for witness
-        // Do not assert that _assert_equals_zero is zero here! Some variables may have
-        // placeholders that do not faithfully represent the underlying values.
+        // No-op for witness Do not assert that _assert_equals_zero is zero
```
dannywillems, line 147: I would leave it as before.

o1vm/src/mips/witness.rs
```diff
@@ -641,59 +633,87 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
                 - preimage_offset;
         // We read at most 4 bytes, ensuring that we respect word alignment.
         let actual_read_len = std::cmp::min(max_read_len, 4 - (addr & 3));
-        // This variable will contain the amount of bytes read which belong to the actual preimage
+        // This variable will contain the amount of bytes read which belong to
```
dannywillems, line 636: I'm curious if we can write some unit tests. We can fake the preimage_oracle by reading files on disk.

o1vm/src/mips/witness.rs
```diff
@@ -641,59 +633,87 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
 
+        // Number of preimage bytes processed in this instruction
+        self.write_column(
+            Column::ScratchState(MIPS_NUM_BYTES_READ_OFF),
+            preimage_read_len,
+        );
+
         // Update the flags to count how many bytes are contained at least
-        for i in 0..MIPS_CHUNK_BYTES_LENGTH {
+        for i in 0..MIPS_CHUNK_BYTES_LEN {
```
dannywillems, line 698: Note: seems not to be an important change.

o1vm/src/mips/witness.rs
```diff
@@ -641,59 +633,87 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+                let byte_i = (idx - LENGTH_SIZE) % MIPS_CHUNK_BYTES_LEN;
+
                 // This should really be handled by the keccak oracle.
                 let preimage_byte = self.preimage.as_ref().unwrap()[idx - LENGTH_SIZE];
-                // Write the individual byte to the witness
+
+                // Write the individual byte of the preimage to the witness
                 self.write_column(
-                    Column::ScratchState(MIPS_PREIMAGE_BYTES_OFFSET + i as usize),
+                    Column::ScratchState(MIPS_PREIMAGE_BYTES_OFF + byte_i),
```
dannywillems, line 663: Is the column therefore more `MIPS_PREIMAGE_CURRENT_CHUNK_BYTE_OFFSET`? It should always be between 0 and 4, right? It does change the semantic of the column compared to before. Before it was the index on the whole byte sequence. Here it is only 0, 1, 2, 3.

o1vm/src/mips/witness.rs
```diff
@@ -641,59 +633,87 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        //        anywhere else where the bytes are stored in the scratch state?
+        self.write_column(Column::ScratchState(MIPS_PREIMAGE_CHUNK_OFF), chunk);
+
+        // Update the number of bytes read from the oracle in this step (can
+        // include bytelength and preimage bytes)
         self.write_column(pos, actual_read_len);
 
+        // Number of preimage bytes processed in this instruction
+        self.write_column(
+            Column::ScratchState(MIPS_NUM_BYTES_READ_OFF),
```
dannywillems, line 693: This looks to be the previous version https://github.com/o1-labs/proof-systems/pull/2274/files#r1626634133.

dannywillems (commented):
I'll give another pass tomorrow + I would like to suggest to have some tests with files on disk. We can discuss it tomorrow.

o1vm/src/mips/constraints.rs
dannywillems: Ah. Just noticed that it is only in documentation starting with /// and it has to be between square brackets.

o1vm/src/mips/constraints.rs
dannywillems: See https://github.com/o1-labs/proof-systems/pull/2274/files#r1627503637

dannywillems:
Opened https://github.com/o1-labs/proof-systems/pull/2288 to provide a test environment for the syscalls.

dannywillems:
@querolita new conflicts :(

o1vm/src/mips/constraints.rs
querolita: Yes, but the interpreter is storing the hash in 8 chunks of 4 bytes each. I'll work on a test with your framework to understand if the leading chunk includes the whole 4 bytes or just 3, leaving the 32nd byte empty (for the 0x02). If that was the case, then there is no need to merge https://github.com/o1-labs/proof-systems/pull/2278 (only valid if the real 32 digest is stored). If that is not the case, then we should think of a different way to store the hash bytes on the MIPS side so we can access bytes independently and leave the 32nd outside of the channel (or have extra columns to constrain the decomposition of the most significant chunk into 1 and 3 bytes).

o1vm/src/mips/tests.rs
querolita: stupid command I found for vscode to limit 80 chars...

o1vm/src/mips/witness.rs
querolita: Need to write one yes, I will likely look into it today

o1vm/src/mips/witness.rs
querolita: Oh yes, it is indeed referring to the chunk (the max bytes you can read per instruction). But I didn't want to make the constant ultra verbose... I can rename it though if it makes it clearer from outside 😄

querolita: According to the doc comment of the constant:
```
/// The at most 4-bytes of the preimage that are currently being processed
```

querolita: Actually, after observing the behavior of the oracle, I noticed that it only reads a number of bytes that is different from 4 at the very end. Meaning, it reads 4 bytes by default, until reaching the preimage end (and in that call it can be 1, 2, 3, or 4 bytes read). Then, that means that the first 2 calls will entirely represent the bytelength. Meaning, that the chunk that is read at each step never contains both bytes from the length and the preimage, what simplifies some things in this PR https://github.com/o1-labs/proof-systems/pull/2277

querolita: Also, `MIPS_PREIMAGE_BYTES_OFF` is renamed with `MIPS_ORACLE_BYTES_OFF` in https://github.com/o1-labs/proof-systems/pull/2277 because it looked more meaningful.

dannywillems:
I'm running e2e tests

querolita:
Making it a draft because despite constraints are satisfied after this PR, the demo halts 🤦🏻 Investigating

dannywillems:
Back on this now

o1vm/src/keccak/helpers.rs
```diff
@@ -83,6 +83,24 @@ where
         Self::is_one(x * x_inv)
```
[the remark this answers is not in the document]
dannywillems: At the moment, I'm adding some debug! in the code. I can push on another branch.
Note: there is already one in witness:
```
debug!("Preimage has been read entirely, triggering Keccak process");
```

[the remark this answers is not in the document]
dannywillems: Can be activated using `RUST_LOG=debug bash run-code.sh` - for info.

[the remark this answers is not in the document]
dannywillems: Oh, we added some already:
```
[2024-06-17T13:34:44Z DEBUG o1vm::mips::witness] Preimage has been read entirely, triggering Keccak process
[2024-06-17T13:34:45Z DEBUG o1vm::preimage_oracle] Reading response
[2024-06-17T13:34:45Z DEBUG o1vm::preimage_oracle] Extracting contents
[2024-06-17T13:34:45Z DEBUG o1vm::preimage_oracle] Got preimage of length 616
     f90265a067b9860843a15812dba28180f4053e75bfdb849eb2abe26d03216c9707f80ae3a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d4934794c4bfccb1668d6e464f33a76badd8c8d7d341e04aa04fb68c89de3da2624bac23278619d56adb9be5eb2b4aff84c61fd2ac448bb64aa024bd76698ee6ba4e030e2dd10fc4bc37d9336abeed624e78e745edaee5146343a0301f8c1ac833a55fbcd030bd305fc8497bb8dfabd5e3ffef924512f63de825e8b90100b69330eb0a1fd75603fe789acc124c2ce06513f0bc1c17d411f225fe44a1892ab0cb2c16510e240d1cf7b213d63469e6dc5a479f8aa7053610f593394ffc804bb158b6eaed34961e68d810ef73f53c6cffed50a0f7d4ed68217f79aed81afa3d18b40ac45acd3bcb7712950d7f81cf45828cf5e65e550c6ee5ef37f1b6fe7a74ef2d655dc488ba896d9e029ce3785f23a572400e5e041b9f39ed5a6337cde28eea09352af52229369950872e862825ca1470f8c29f168f14bb82efe7a2c202feff49db531ffc1fcc9f3f56178e021d936dd3de949d5ad026cb05b6731d2de9a56a760cb9a07d0e83b26d5cc8fd3e3f56a68a71b04db7cb1eba3342bedc004e3b80835cd9208401c9c38084013f0318846668221899d883010e00846765746888676f312e32322e32856c696e7578a09b5313a28445f8dfa266725d847fa197ec0578ea94d01ba694cb03581e736c8a88000000000000000085060cca7404a039c24023a36d6e5cb9e696fbd6d563ffa3ab163e846fe1d66b1986c6b0db0c7683020000830a0000a0aeb14d7977e47d8b36bef74b9dcd3356fda2d8367495512d27af7e4c62fad866
```

o1vm/src/mips/column.rs
```diff
@@ -14,14 +14,22 @@ use strum::EnumCount;
...
+/// The number of hashes performed so far in the block
+pub(crate) const MIPS_HASH_COUNTER_OFF: usize = 80;
+/// The number of bytes of the preimage that have been read so far in this hash
+pub(crate) const MIPS_BYTE_COUNTER_OFF: usize = 81;
+/// A flag indicating whether the preimage has been read fully or not
+pub(crate) const MIPS_END_OF_PREIMAGE_OFF: usize = 82;
+/// The number of preimage bytes processed in this step
+pub(crate) const MIPS_NUM_BYTES_READ_OFF: usize = 83;
+/// The at most 4-byte chunk of the preimage that has been read in this step
+pub(crate) const MIPS_PREIMAGE_CHUNK_OFF: usize = 84;
```
dannywillems, line 26: Nit: (might be only for me) I think it might be worth to add some small examples on what the registers at these offsets are supposed to embed? Like integers, bytes, etc. Just to know, when we read this, where we are supposed to find them in the code.

o1vm/src/mips/column.rs
```diff
@@ -14,14 +14,22 @@ use strum::EnumCount;
...
+/// The at most 4-bytes of the preimage that are currently being processed
+pub(crate) const MIPS_PREIMAGE_BYTES_OFF: usize = 85;
+/// Flags indicating whether at least N bytes have been processed in this step
+pub(crate) const MIPS_HAS_N_BYTES_OFF: usize = 89;
+/// The maximum size of a chunk (4 bytes)
```
dannywillems, line 31: Nit (sorry, very "stupid"): only maybe adding a small notes saying that the values above are offsets for the registers, and this one is simply a constant.

dannywillems: Also, i do not understand why this variable is in `columns.rs`. Why not in libs.rs or somewhere else?

o1vm/ethereum-optimism
dannywillems, line 1: Why are you updating optimism?

dannywillems: It seems to be the old version of optimism. Please revert.

o1vm/src/mips/witness.rs
```diff
@@ -43,7 +43,7 @@ pub const NUM_INSTRUCTION_LOOKUP_TERMS: usize = 5;
 pub const NUM_LOOKUP_TERMS: usize =
     NUM_GLOBAL_LOOKUP_TERMS + NUM_DECODING_LOOKUP_TERMS + NUM_INSTRUCTION_LOOKUP_TERMS;
 // TODO: Delete and use a vector instead
-pub const SCRATCH_SIZE: usize = 93; // MIPS + hash_counter + is_syscall + bytes_read + bytes_left + bytes + has_n_bytes + reading_preimage
+pub const SCRATCH_SIZE: usize = 93; // MIPS + hash_counter + chunk_read + bytes_read + bytes_left + bytes + has_n_bytes
```
dannywillems, line 46: Nit: seems that we can decrease the number? reading_preimage was a +1. trying quickly with a smaller value.

dannywillems: Ignore. 93 is required.

o1vm/src/mips/constraints.rs
```diff
@@ -376,81 +375,104 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
 
     fn report_exit(&mut self, _exit_code: &Self::Variable) {}
 
+    /// This function checks that the preimage is read correctly.
+    /// It adds 13 constraints, and 5 lookups for the communication channel.
```
dannywillems, line 379: The test mentions 22 constraints.

o1vm/src/mips/constraints.rs
```diff
@@ -376,81 +375,104 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+    /// In particular, at every step it writes the bytes of the preimage into
+    /// the channel (excluding the length bytes) and it reads the hash digest
+    /// from the channel when the preimage is fully read.
```
dannywillems, line 382:
```suggestion
    /// from the channel when the preimage is fully read.
    /// The output is the actual number of bytes that have been read.
```

dannywillems: Nit (not for this PR): a good documentation, IMO, is also a documentation that describes what the parameters of the function does, and in particular, explains why there are some unused parameters when it is in the implementation of a trait.
For instance, here, I would have explained why `_addr` and `_len` are unused in the constraints side, when it is used in the witness.

o1vm/src/mips/constraints.rs
```diff
@@ -376,81 +375,104 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
     fn request_preimage_write(
         &mut self,
         _addr: &Self::Variable,
         _len: &Self::Variable,
```
dannywillems, line 386: I'm surprised the length column is not constrained. For a follow-up.

o1vm/src/mips/witness.rs
```diff
@@ -109,13 +109,14 @@ fn memory_size(total: usize) -> String {
...
-        // Famous last words: 1023 exabytes ought to be enough for anybody        //
-        //                                                                        //
-        // Corollary: unwrap() below shouldn't fail                               //
-        //                                                                        //
-        // The maximum representation for usize corresponds to 16 exabytes anyway //
-        ////////////////////////////////////////////////////////////////////////////
+        ////////////////////////////////////////////////////////////////////////
+        // Famous last words: 1023 exabytes ought to be enough for anybody    //
+        //                                                                    //
+        // Corollary:                                                         //
```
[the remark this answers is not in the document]
querolita: I just realized that the 80-char script had messed up the format.

o1vm/src/mips/witness.rs
```diff
@@ -667,57 +660,85 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
...
 
-        // Update how many bytes are left to be read
-        self.write_column(
-            Column::ScratchState(MIPS_PREIMAGE_LEFT_OFFSET),
-            (preimage_len as u64) - self.preimage_bytes_read,
-        );
-
         // If we've read the entire preimage, trigger Keccak workflow
         if self.preimage_bytes_read == preimage_len as u64 {
+            self.write_column(Column::ScratchState(MIPS_END_OF_PREIMAGE_OFF), 1);
```
dannywillems, line 740: Even though the default value is 0 for the scratch state, I would also write 0 then in the other case.

dannywillems (approved):
Ok for now. I would keep in mind that soundness might need another pass, and in particular, we need to test that the registers are written with the appropriate values. The VM runs as expected, therefore I'm fine with merging this one + the test.

o1vm/src/mips/constraints.rs
querolita: Either way I agree that this instruction needs another pass for soundness.

o1vm/src/mips/constraints.rs
```diff
@@ -481,6 +482,31 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
                 );
             }
 
+            // Constrain that at most you read `len` bytes
```
dannywillems, line 485: Don't we simply want to check that `actual_read_len <= len`, i.e `0 <= len - actual_read_len <= 4`. Can be done using a lookup.

o1vm/src/mips/constraints.rs
dannywillems: I'm opening a follow-up for the soundness of the syscall.
A first step would be to list all the constraints that the function is supposed to perform, as multi-variate polynomials and inequalities. From there, we can see what's the best way to achieve the soundness property.

dannywillems:
I'm merging to move on as it is blocking merging other PRs. I would like to discuss the constraints you added recently @querolita. The VM runs fine for me.

o1vm/src/mips/constraints.rs
dannywillems: Let's move the conversation in https://github.com/o1-labs/proof-systems/issues/2335.

querolita: That's a good thing to check, apart from distinguishing between bytelength bytes and preimage bytes. Because only the preimage bytes must be part of the absorb lookup. So the reason for having chunks is to prove decomposition of preimage bytes as 4byte chunks. 

But yeah, I like your approach if we don't mind adding a small table for numbers 0-4.

PR #2277, opened 2024-05-31
Optimize syscall lookups by sharing the bytes vector

querolita:
This PR addresses the concern I raised in https://github.com/o1-labs/proof-systems/pull/2276 about reducing the number of witness columns and lookups required to constrain that the bytes read from the oracle are indeed 8-bits long.

After having tested the behavior of the oracle using our demo, I realized that it always reads the 8 leading bytes corresponding to the preimage length in two instructions reading 4 bytes in each row. This means, it never shares both length bytes with preimage bytes in the same row. 

This simplifies the approach in the previous PR, because now we checked that the vector of bytes can contain length bytes (always 4) or preimage bytes (at most 4). The rest remains unchanged, meaning `has_n_bytes` still refers to preimage bytes. Keeping that meaning results in no changes required for the constraints. And more importantly, this allows us to bring down the number of extra lookups from 8 to 4 per `SyscallReadPreimage` instruction.

### DIFF
--- a/o1vm/src/mips/column.rs
+++ b/o1vm/src/mips/column.rs
@@ -24,12 +24,10 @@ pub(crate) const MIPS_END_OF_PREIMAGE_OFF: usize = 82;
 pub(crate) const MIPS_NUM_BYTES_READ_OFF: usize = 83;
 /// The at most 4-byte chunk of the preimage that has been read in this step
 pub(crate) const MIPS_PREIMAGE_CHUNK_OFF: usize = 84;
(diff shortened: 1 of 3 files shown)


querolita:
As part of locally running the syscall instruction as a unit test I found out that this is not necessarily true. Specially, if the `addr` passed to `request_preimage_write` is not aligned. In that case, one can start by reading a non-4 amount of bytes of the length, and then the third (?) call could contain both length bytes and real preimage bytes.

querolita:
https://github.com/o1-labs/proof-systems/issues/2312

PR #2281, opened 2024-06-04
Use random values for MIPS instructions unit tests

querolita:
This PR initializes the RW memory and the general purpose registers of the dummy environment used in unit tests of MIPS instructions to random values. This gives us better guarantees.

### DIFF
--- a/o1vm/src/mips/tests.rs
+++ b/o1vm/src/mips/tests.rs
@@ -157,7 +157,8 @@ mod unit {
             // Only 8kb of memory (two PAGE_ADDRESS_SIZE)
             memory: vec![
                 // Read/write memory
-                (0, vec![0; PAGE_SIZE as usize]),
+                // Initializing with random data
+                (0, vec![rng.gen_range(0u8..=255); PAGE_SIZE as usize]),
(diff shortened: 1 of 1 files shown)


o1vm/src/mips/tests.rs
```diff
@@ -148,8 +148,11 @@ mod unit {
             // Only 8kb of memory (two PAGE_ADDRESS_SIZE)
             memory: vec![
                 // Read/write memory
-                (0, vec![0; PAGE_SIZE as usize]),
-                // Executable memory. Allocating 4 * 4kB
+                // Initializing with random data
+                (
+                    0,
+                    (0..PAGE_SIZE).map(|_| rng.gen_range(0u8..=255)).collect(),
+                ), // Executable memory. Allocating 4 * 4kB
```
dannywillems, line 155: Your plugin tries to align after the `,`. I think it should not be the case.

dannywillems (approved):
modulo comment

o1vm/src/mips/tests.rs
querolita: Oh, fun because I didnt run the 80-char align command on that file this time... will fix it manually

PR #2288, opened 2024-06-05
o1VM: make the preimage oracle abstract

dannywillems:
For test environment, we can implement a fake preimage oracle that will read/write values on disk/memory.

This PR implements the fake on-disk preimage oracle.
It consists of reading files from the directory `o1vm/resources/test/`, formatted as `0xPREIMAGE_KEY.txt`, as given by the op-program. The content is the preimage data.

The interpreter uses the preimage oracle through `request_preimage_write` which calls the `preimage_oracle.get_preimage` (on-disk for the test).

Testing the syscall write can consist now in writing correctly the preimage key in the correct registers (+ preimage key offset, etc), call the interpreter on the syscall, and check if the memory has been set correctly.


dannywillems:
Note that I would be careful to not push too many files to avoid growing the git repository space.

querolita (approved):

Issue #2287, opened 2024-06-05
Bug: SRS BN254 G2 h is not in subgroup


volhovm:
This fails, due to the fact how `verifier_srs` is created.
```
    let actual: BN254G2Affine = srs.verifier_srs.h;
    assert!(actual.is_in_correct_subgroup_assuming_on_curve());
```

The issue is in `point_of_random_bytes()` most likely.

dannywillems:
Use `mul_by_cofactor`

PR #2289, opened 2024-06-05
Improve SRS caching for tests

volhovm:
What this PR does:
- Adds code for creating/reading BN254 "cached" SRS.
- Makes the "lagrange" field "optional on read". For future/past compatibility: it's not logically "essential", but very good to have.
- Generates the cached BN254 SRS with Lagrange bases (all up to the power), now used by tests.
- Makes serialization for these "test" SRS unchecked -- that helps to save time (5 sec for SRS, about as much for each Lagrange basis).
- Regenerates Pallas/Vesta SRS (with Lagrange bases), now applies to kimchi tests too. Renames them from `pallas.srs` to `test_pallas.srs` to emphasize that these are /not/ used in production and can be regenerated at any point if necessary.

Now many tests will be using the pre-computed SRS, which leads to approximately this kind of improvements for 2^15 SRS (release mode):
- Generate SRS + Lagrange:  1m13s user, 24sec real
  - Out of which SRS takes:  4.4sec user / real. Sequential.
- All cached with Lagrange: 0.5sec, sequential.

(as in unix `time`, "user" is total core-time, while "real" is how much physical time it actually took on all the cores simultaneously)

While doing it, I discovered this bug: https://github.com/o1-labs/proof-systems/issues/2287, now fixed.

### DIFF
--- a/kimchi/src/precomputed_srs.rs
+++ b/kimchi/src/precomputed_srs.rs
@@ -5,21 +5,95 @@
 //! We generate the SRS within the test in this module.
 //! If you modify the SRS, you will need to regenerate the SRS by passing the `SRS_OVERWRITE` env var.
 
-use std::{fs::File, io::BufReader, path::PathBuf};
-
-use poly_commitment::srs::SRS;
-
 use crate::curve::KimchiCurve;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use poly_commitment::{srs::SRS, PolyComm};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use std::{collections::HashMap, fs::File, io::BufReader, path::PathBuf};
+
+/// We store several different types of SRS objects. This enum parameterizes them.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum StoredSRSType {
+    Test,
+    Prod,
+}
+
+/// A clone of the SRS that is serialized in a test-optimised way.
+///
+/// NB: Serialization of these fields is unchecked (and fast). If you
+/// want to make sure the data is checked on deserialization, this code
+/// must be changed; or you can check it externally.
+#[serde_as]
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(bound = "G: CanonicalDeserialize + CanonicalSerialize")]
+pub struct TestSRS<G> {
+    /// The vector of group elements for committing to polynomials in
+    /// coefficient form.
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAsUnchecked>")]
+    pub g: Vec<G>,
+
+    /// A group element used for blinding commitments
+    #[serde_as(as = "o1_utils::serialization::SerdeAsUnchecked")]
+    pub h: G,
+
+    /// Commitments to Lagrange bases, per domain size
+    #[serde_as(as = "HashMap<_,Vec<PolyComm<o1_utils::serialization::SerdeAsUnchecked>>>")]
+    pub lagrange_bases: HashMap<usize, Vec<PolyComm<G>>>,
+}
+
+impl<G> From<SRS<G>> for TestSRS<G> {
+    fn from(value: SRS<G>) -> Self {
+        TestSRS {
+            g: value.g,
+            h: value.h,
+            lagrange_bases: value.lagrange_bases,
+        }
+    }
+}
+
+impl<G> From<TestSRS<G>> for SRS<G> {
+    fn from(value: TestSRS<G>) -> Self {
+        SRS {
+            g: value.g,
+            h: value.h,
+            lagrange_bases: value.lagrange_bases,
+        }
+    }
+}
 
 /// The size of the SRS that we serialize.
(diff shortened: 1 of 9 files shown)


poly-commitment/src/commitment.rs
```diff
@@ -39,7 +41,7 @@ use super::evaluation_proof::*;
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(bound = "C: CanonicalDeserialize + CanonicalSerialize")]
 pub struct PolyComm<C> {
-    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAs>")]
+    #[serde_as(as = "Vec<o1_utils::serialization::SerdeAsUnchecked>")]
```
dannywillems, line 44: Not sure if this changes breaks or no other code in develop/berkeley.

kimchi/src/precomputed_srs.rs
```diff
@@ -62,17 +64,15 @@ mod tests {
...
         test_regression_serialization_srs_with_generators::<Vesta>(exp_output.to_string())
     }
 
     #[test]
     fn test_regression_serialization_srs_with_generators_pallas() {
-        // This is the same as Vesta as we encode the coordinate x only.
-        // Generated with commit 4c69a4defdb109b94f1124fe93283e728f1d8758
-        let exp_output = "9291c421010000000000000000000000000000000000000000000000000000000000000000c421010000000000000000000000000000000000000000000000000000000000000000";
+        // Generated with commit 309e1f1d4bc76fd37bc5403f7c728238f6a39aee
+        let exp_output = "9391c4410100000000000000000000000000000000000000000000000000000000000000bb2aedca237acf1971473d33d45b658f54ee7863f0a9df537c93120aa3b5741b00c4410100000000000000000000000000000000000000000000000000000000000000bb2aedca237acf1971473d33d45b658f54ee7863f0a9df537c93120aa3b5741b0080";
```
dannywillems, line 75: I don't really like we change the encoding. Can we not keep the same?

poly-commitment/src/commitment.rs
volhovm: Good catch, the change is not necessary. Will revert now.

kimchi/src/precomputed_srs.rs
volhovm: The difference is that now we serialize the Lagrange bases as well, so this extra tail is that.

msm/src/precomputed_srs.rs
```diff
@@ -1,18 +1,116 @@
...
+use ark_serialize::Write;
 use kimchi::circuits::domains::EvaluationDomains;
 use poly_commitment::pairing_proof::PairingSRS;
-
-use crate::{Fp, BN254};
+use rand::{prelude::StdRng, SeedableRng};
+use std::{fs::File, io::BufReader, path::PathBuf};
 
 /// Obtains an SRS for a specific curve from disk, or generates it if absent.
 pub fn get_bn254_srs(domain: EvaluationDomains<Fp>) -> PairingSRS<BN254> {
```
dannywillems, line 12: As discussed in our call some minutes ago, I am in favor of having the deserializer as you implemented, but not change the actual serializer used currently. I am sure there should be a way to use a custom serializer with serde.

volhovm:
@dannywillems now reimplemented in a fully backwards-compatible way!

kimchi/src/prover_index.rs
```diff
@@ -213,7 +213,7 @@ pub mod testing {
                 let log2_size = size.ilog2();
                 let mut srs = if log2_size <= precomputed_srs::SERIALIZED_SRS_SIZE {
                     // TODO: we should trim it if it's smaller
-                    precomputed_srs::get_srs()
+                    precomputed_srs::get_srs_test()
```
dannywillems, line 216: :grimacing:

poly-commitment/src/commitment.rs
```diff
@@ -58,6 +60,65 @@ impl<T> PolyComm<T> {
     }
 }
 
+impl<T, U> SerializeAs<PolyComm<T>> for PolyComm<U>
```
dannywillems, line 63: Nit (follow-up): a comment would be welcome for our future us.

msm/src/precomputed_srs.rs
```diff
@@ -1,18 +1,145 @@
...
+    pub fn create_or_check_srs() {
+        let domain_size = DOMAIN_SIZE;
+        create_and_store_srs_with_path(false, domain_size, get_bn254_srs_path());
+    }
+
+    #[test]
+    /// Fast test for a small-sized SRS. Always writes & reads.
+    pub fn check_bn256_srs_serialization() {
+        let domain_size = 1 << 8;
+        let test_srs_path = PathBuf::from("/tmp/test_bn254.srs");
```
dannywillems, line 133: I hope it works in the CI (and other platforms - permissions for instance). I would use https://doc.rust-lang.org/std/env/fn.temp_dir.html

kimchi/src/prover_index.rs
dannywillems: oh

PR #2299, opened 2024-06-06
Unit test for `LoadWordLeft`

querolita:
<details>

<summary>
This PR provides a unit test for the `LoadWordLeft` instruction. It matches closely the Cannon definition of the instruction given by 

```
		case 0x22: // lwl
			val := mem << ((rs & 3) * 8)
			mask := uint32(0xFFFFFFFF) << ((rs & 3) * 8)
			return (rt & ^mask) | val
```

I used this test to realize that our current implementation of the interpreter for that function was not working properly. In particular, bytes were being mapped to different positions. I changed the interpreter code to avoid the confusing (and error prone) use of `overwrite_n` in favor of a more direct set of variables called `mod_n`. I included comments to help understand what the above function does and what it expects in each byte position of the destination register.

</summary> 

Old description

This unit test matches what the interpreter of `LoadWordLeft` is doing. Nonetheless, the constraints still do not pass when trying to make a proof for them. In particular:

```bash
"Unsatisfied expression: ((((Curr(x[19]) - 0x0100000000000000000000000000000000000000000000000000000000000000) * Curr(x[23])) + Curr(x[22])) - 0x0100000000000000000000000000000000000000000000000000000000000000)"
```

Still figuring out why. The obvious answer would be dummy rows initialized to zeros so the `-1` will not hold. But we already solved that issue when creating the trace, by padding with a dummy row that is known to be a "real" row.~

</details>

### DIFF
--- a/o1vm/src/mips/interpreter.rs
+++ b/o1vm/src/mips/interpreter.rs
@@ -2086,6 +2086,8 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
                 unsafe { env.bitmask(&addr, 2, 0, pos) }
             };
 
+            // From docs: "EffAddr is the address of the most-significant of
+            //             four consecutive bytes word in memory"
             // Big-endian: smallest address is the most significant byte
             let m0 = env.read_memory(&addr);
             let m1 = env.read_memory(&(addr.clone() + Env::constant(1)));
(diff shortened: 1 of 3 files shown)


querolita:
After applying https://github.com/o1-labs/proof-systems/pull/2311 to this PR, I will run the nodes again to see if the constraint holds now.

querolita:
Demo halts, making the PR a draft until solved

querolita:
Closing due to https://github.com/o1-labs/proof-systems/pull/2729/

PR #2302, opened 2024-06-06
IVC: add top level documentation

dannywillems:
We must keep somewhere our discussion.

### DIFF
--- a/ivc/src/lib.rs
+++ b/ivc/src/lib.rs
@@ -1,3 +1,116 @@
+//! This crate provides a circuit to achieve Incremental Verifiable
+//! Computation (IVC) based on a variant of the folding scheme described in the
+//! paper [Nova](https://eprint.iacr.org/2021/370.pdf). For the rest of the
+//! document, we do suppose that the curve is BN254.
+//!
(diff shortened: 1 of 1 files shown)


ivc/src/lib.rs
```diff
@@ -1,3 +1,116 @@
...
+//! see below (TODO).
+//! Note that the columns used by the poseidon hash must also be aggregated
+//! while running the IVC.
+//!
+//! The Poseidon hash is used to absorb all the instance commitments, the
+//! challenges and the scaled commitments.
+//! In our circuit, we first start by absorbing all the elliptic curve points. A
+//! particularity is that we will use different Poseidon instances for each
+//! "side" of the addition. For each point, we do assume (*at the moment*) that
+//! each point is encoded as 2 field elements in the field of the circuit
```
volhovm, line 92: In this paragraph (and a bit in the previous one, and in the following one) we are being overly specific with parameters and layouts that we might change next week. Having these numbers in the column file and only general structure of the module in this comment would minimise maintenance overhead.

volhovm (approved):

PR #2304, opened 2024-06-10
IVC: use poseidon degree 2

dannywillems:
Move to an instance of Poseidon suitable for IVC


ivc/src/ivc/columns.rs
```diff
@@ -245,7 +252,13 @@ pub enum IVCColumn {
     Block1InputRepacked150(usize),
 
     /// 1 hash per row
-    Block2Hash(PoseidonColumn<IVC_POSEIDON_STATE_SIZE, IVC_POSEIDON_NB_FULL_ROUND>),
+    Block2Hash(
```
volhovm, line 255: Do I understand correctly that the fixed selector changes are not yet in master?

volhovm (approved):
Looks good!

ivc/src/ivc/columns.rs
dannywillems: It is in master. Why makes you think it is not?

PR #2306, opened 2024-06-10
IVC: Poseidon with full rounds and partial rounds

dannywillems:
Ready. First version to get the IVC working. Can be improved later.

Implement a version of poseidon with partial rounds and full rounds.
It is implemented based on a copy of the existing poseidon implementations.
It is not elegant in terms of design because:
- it duplicates code
- it doesn't share code between the "only full rounds" version and the "full + partial rounds" version.

In a future PR, we should:
- [ ] make the code more elegant to merge the different interpreters
- [ ] add the tests in poseidon crate -> seems to be in poseidon, with half_full_rounds. Seems only to change the the first element of the state instead of the last element.
- [x] move the round constants into "fixed selectors", see https://github.com/o1-labs/proof-systems/pull/2314


ivc/src/poseidon_8_56_5_3_2/interpreter.rs
```diff
@@ -0,0 +1,356 @@
...
+        .map(|(i, res)| env.hcopy(res, PoseidonColumn::FullRound(round, nb_red * i + 3)))
+        .collect();
+
+    res_state
+        .try_into()
+        .expect("Resulting state must be of state size (={STATE_SIZE}) length")
+}
+
+/// Compute one partial round of the Poseidon permutation
+fn compute_one_partial_round<
```
dannywillems, line 273: Mmh, it seems to be RC + SBOX + MDS, not SBOX + MDS + RC.

dannywillems: Fixing it quickly

ivc/src/lib.rs
```diff
@@ -1,7 +1,19 @@
 pub mod ivc;
-pub mod poseidon;
-pub mod poseidon_params;
-pub mod poseidon_quadri;
+/// Poseidon hash function with 55 full rounds, 0 partial rounds, sbox 7, a
+/// state of 3 elements and constraints of degree 2
+pub mod poseidon_55_0_7_3_2;
```
volhovm, line 4: More Poseidons to the god of Poseidons (Zeus?).

ivc/src/poseidon_8_56_5_3_2/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        TestPoseidonColumn,
+        { <TestPoseidonColumn as ColumnIndexer>::N_COL },
+        { <TestPoseidonColumn as ColumnIndexer>::N_COL },
+        0,
+        0,
+        DummyLookupTable,
+    >;
+
+    #[test]
+    #[ignore = "There is no support for partial rounds in the crate poseidon"]
```
volhovm, line 56: Do I understand correctly that this test is disabled because you have no specification-level native implementation for this type of poseidon? 

Maybe the test is worth deleting though?

volhovm (approved):
I'm a tiny bit worried we don't have a native implementation for this Poseidon to compare with (probably an important todo for later), but otherwise looks all good structurally!

ivc/src/poseidon_8_56_5_3_2/mod.rs
dannywillems: See https://github.com/o1-labs/proof-systems/pull/2329 and https://github.com/o1-labs/proof-systems/pull/2328

PR #2310, opened 2024-06-11
Fix `equal` for witness interpreter

querolita:
This PR proposes a different definition of the `equal` function of the MIPS interpreter so that it does not generate underflows when trying to subtract `x-y` and `y` is larger than `x` (for `u32` types). This is not a problem in fields, but it is when running unit tests built with unsigned types.

This is not the only place where subtraction underflows happen in unit tests. An upcoming PR will address the rest, which corresponds to the `FIXME` to allow multiple register accesses per instruction call.

### DIFF
--- a/o1vm/src/mips/interpreter.rs
+++ b/o1vm/src/mips/interpreter.rs
@@ -648,9 +648,8 @@ pub trait InterpreterEnv {
         res
     }
 
-    fn equal(&mut self, x: &Self::Variable, y: &Self::Variable) -> Self::Variable {
-        self.is_zero(&(x.clone() - y.clone()))
-    }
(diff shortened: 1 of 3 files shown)


o1vm/src/mips/interpreter.rs
```diff
@@ -159,7 +159,7 @@ impl IntoIterator for Instruction {
 
 pub trait InterpreterEnv {
     /// A position can be seen as an indexed variable
-    type Position;
+    type Position: std::fmt::Debug;
```
dannywillems, line 162: I would remove this from the PR.

querolita: oh okay I will undo it

dannywillems (approved):

PR #2305, opened 2024-06-10
Unit test for `SyscallReadPreimage`

querolita:
This uses @dannywillems's framework to emulate the preimage oracle in unit tests for the Keccak syscall. 

This PR does not only include a unit test using the sample file, but it also proposes two main fixes:

1. The `get_preimage()` function of the oracle trait was not reading the right bytes from the file, but in ASCII. Thus, I decode them as hexadecimal bytes to store them correctly.
2. The interpreter for the `SyscallReadPreimage` was not updating the address with the offset correctly. Thus, the addresses where the preimage bytes were being stored in memory were mis-aligned. The unit test checks that now the bytes are stored correctly where desired.

PR #2293, opened 2024-06-05
Unit test for `LoadWordRight`

querolita:
Fixes interpreter to behave like Cannon.

PR #2311, opened 2024-06-11
Fix instruction counter to allow multiple register/memory accesses per instruction call

querolita:
This PR is meant to fix the subtraction underflows that appeared in unit tests when the same instruction was accessing the same register multiple times. This PR solves that issue by artificially increasing the instruction counter every time that register access takes place (so that the elapsed time is always a positive value). 

It compliments https://github.com/o1-labs/proof-systems/pull/2310. 

I tested it in a local branch with `LoadWordLeft` which used to fail for this reason (https://github.com/o1-labs/proof-systems/actions/runs/9447808562/job/26020430671?pr=2299), and now with the same seed the test `RUSTFLAGS="-Coverflow-checks=y -Cdebug-assertions=y" cargo test --package o1vm --lib --all-features --release -- mips::tests::unit::itype::test_unit_lwl_instruction --exact --show-output` works.

I will be updating the unit tests accordingly now, meaning
https://github.com/o1-labs/proof-systems/pull/2299
https://github.com/o1-labs/proof-systems/pull/2293
https://github.com/o1-labs/proof-systems/pull/2305

to check that this solves the issue.

Nonetheless, we need to run the whole demo to make sure that this fix does not break other parts of the system.

### DIFF
--- a/o1vm/src/mips/witness.rs
+++ b/o1vm/src/mips/witness.rs
@@ -169,6 +169,10 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
         self.instruction_counter
     }
 
+    fn increase_instruction_counter(&mut self) {
+        self.instruction_counter += 1;
+    }
+
     unsafe fn fetch_register(
         &mut self,
         idx: &Self::Variable,
--- a/o1vm/src/mips/witness.rs
+++ b/o1vm/src/mips/witness.rs
@@ -1207,7 +1207,8 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> Env<Fp, PreImageOracle> {
     fn pp_info(&mut self, at: &StepFrequency, meta: &Meta, start: &Start) {
         if self.should_trigger_at(at) {
             let elapsed = start.time.elapsed();
-            let step = self.instruction_counter;
+            // Compute the step number removing the MAX_ACC factor
+            let step = self.instruction_counter / MAX_ACC;
             let pc = self.registers.current_instruction_pointer;
 
             // Get the 32-bits opcode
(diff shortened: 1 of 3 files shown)


dannywillems:
As discussed some min ago, and suggested by @mrmr1993, we should get the number of times we read/write/access a memory cell/register, and has the instruction counter being a multiple of it. It eases the debugging when we want to stop at instruction N for instance. See https://github.com/o1-labs/proof-systems/pull/2311#issuecomment-2160979048 for a branch printing the number of times we acess a register/memory cell. My testing shows it is 12 atm.

dannywillems (changes requested):
See  comment + discussions.

dannywillems (changes requested):
The log is wrong:
```
step=380000000 pc=00003391f8 insn=008c870080 ips=1759259.26
```

ips and step

querolita:
> The log is wrong:

@dannywillems will fix it, forgot to adapt the log

o1vm/src/mips/witness.rs
```diff
@@ -1050,6 +1064,24 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> Env<Fp, PreImageOracle> {
...
+    /// NOTE: actually, in practice it will be less than that, as there is no
+    ///       single instruction that performs all of them.
+    ///
+    /// This means that the actual number of instructions executed will result
+    /// from dividing the instruction counter by MAX_ACC (floor).
+    ///
+    /// Then, in order to update the instruction counter, we need to add 1 to
+    /// the real instruction counter and multiply it by MAX_ACC to have a unique
+    /// representation of each step (which is helpful for debugging).
+    pub fn update_instruction_counter(&mut self) {
```
dannywillems, line 1081: I think we might need to add another constraint. We might be able to add intermediate constraints. Will postpone the discussion to the soundness proof step, and consider this a good thing to move on.

dannywillems (approved):

o1vm/src/mips/witness.rs
querolita: https://github.com/o1-labs/proof-systems/issues/2322

PR #2314, opened 2024-06-12
IVC/Poseidon: make the round constants public values

dannywillems:
Built on top of !https://github.com/o1-labs/proof-systems/pull/2306
in 2306, the round constants are considered as witness values. Simply making them public to not fold them later.

### DIFF
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -25,6 +25,9 @@ pub fn block_height<const N_COL_TOTAL: usize, const N_CHALS: usize>(block_num: u
 pub const IVC_POSEIDON_STATE_SIZE: usize = 3;
 pub const IVC_POSEIDON_NB_FULL_ROUND: usize = 55;
 
+pub const IVC_NB_TOTAL_FIXED_SELECTORS: usize =
+    IVC_POSEIDON_NB_FULL_ROUND * IVC_POSEIDON_STATE_SIZE + N_BLOCKS;
+
 pub type IVCPoseidonColumn = PoseidonColumn<IVC_POSEIDON_STATE_SIZE, IVC_POSEIDON_NB_FULL_ROUND>;
 
 /// The IVC circuit is tiled vertically. We assume we have as many
--- a/ivc/src/ivc/interpreter.rs
+++ b/ivc/src/ivc/interpreter.rs
@@ -33,6 +33,8 @@ use kimchi_msm::{
 use num_bigint::BigUint;
 use std::marker::PhantomData;
 
+use super::columns::IVC_NB_TOTAL_FIXED_SELECTORS;
+
 /// The biggest packing variant for foreign field. Used for hashing. 150-bit limbs.
 pub const LIMB_BITSIZE_XLARGE: usize = 150;
 /// The biggest packing format, 2 limbs.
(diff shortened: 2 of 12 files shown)


dannywillems:
Investigating CI, seems to be related to the IVC circuit. https://github.com/o1-labs/proof-systems/actions/runs/9505793514/job/26201502174?pr=2314

dannywillems:
Related to the fixed selectors. Must be mapped correctly in the IVC circuit.

ivc/src/ivc/columns.rs
```diff
@@ -25,6 +25,9 @@ pub fn block_height<const N_COL_TOTAL: usize, const N_CHALS: usize>(block_num: u
 pub const IVC_POSEIDON_STATE_SIZE: usize = 3;
 pub const IVC_POSEIDON_NB_FULL_ROUND: usize = 55;
 
+pub const IVC_NB_TOTAL_FIXED_SELECTORS: usize =
+    IVC_POSEIDON_NB_FULL_ROUND * IVC_POSEIDON_STATE_SIZE + N_BLOCKS;
```
volhovm, line 29: Cosmetic: maybe create  `N_FSEL_POSEIDON = IVC_POSEIDON_NB_FULL_ROUND * IVC_POSEIDON_STATE_SIZE` and refer to that?

volhovm: Generally where possible it's good to keep to the same notation `prover` is using: `N_WIT, N_FSEL, N_DSEL, N_REL` etc

ivc/src/ivc/interpreter.rs
```diff
@@ -887,16 +889,23 @@ where
 }
 
 /// Builds selectors for the IVC circuit.
+/// The round constants for Poseidon are not added in this function, and must be
+/// done separately.
+/// The size of the array is the total number of public values required for the
+/// IVC. Therefore, it includes the potential round constants required by
+/// the hash function.
+// FIXME: rc should be handled here or in the lens
```
volhovm, line 897: I'm working on this (slowly, there's a PR...) -- my current idea is to put it into `ColumnIndexer`, and thus the `fixed_selectors` for `IVCColumn` would call `fixed_selectors` for `PoseidonColumn`... For now this is fine, but I'd also prefer to just add `build_selectors` function for poseidon, and call it from this IVC function. Anyway, if it works it works :)

ivc/src/ivc/mod.rs
```diff
@@ -91,8 +94,20 @@ mod tests {
...
 
         println!("Building fixed selectors");
-        let fixed_selectors: Vec<Vec<Fp>> =
-            build_selectors::<_, TEST_N_COL_TOTAL, TEST_N_CHALS>(domain_size).to_vec();
+        let mut fixed_selectors: Vec<Vec<Fp>> =
+            build_selectors::<Fp, TEST_N_COL_TOTAL, TEST_N_CHALS>(domain_size).to_vec();
+
+        // Write constants
+        {
+            let rc = PoseidonBN254Parameters.constants();
```
volhovm, line 102: Any reason this block that's repeated twice cannot be a separate function `build_selectors() -> Vec<Vec<Fp>>`?

volhovm (approved):
Cool! One cosmetic comment: the "building fixed selectors" block of 15 lines is repeated exactly 8 times, so unless it's problematic for some reason it'd be nice to have it in a separate function. Or two. I realise different Poseidons have different constants...

ivc/src/ivc/mod.rs
dannywillems: Interpreter depends on the field.

PR #2316, opened 2024-06-13
Add completeness test for boxed arrays

volhovm:
Subj.


dannywillems (approved):
Curious you want to test that. But fine for me.

PR #2324, opened 2024-06-13
MSM: fixed selectors with fixed values, 42, instead of random

dannywillems:
To handle the case the polynomial evaluations are always the same


querolita (approved):

dannywillems:
Force pushing a modified commit message to run the CI. It was stuck

PR #2325, opened 2024-06-13
o1VM: set correct values of step in snapshot

dannywillems:


dannywillems:
Closing in favor of https://github.com/o1-labs/proof-systems/pull/2326

PR #2328, opened 2024-06-14
Poseidon: apply linear layer

dannywillems:


poseidon/src/permutation.rs
```diff
@@ -62,7 +62,10 @@ pub fn half_rounds<F: Field, SC: SpongeConstants>(
         for state_i in state.iter_mut() {
             *state_i = sbox::<F, SC>(*state_i);
         }
-        apply_mds_matrix::<F, SC>(params, state);
+        let res = apply_mds_matrix::<F, SC>(params, state);
+        for (i, state_i) in state.iter_mut().enumerate() {
```
querolita, line 66: Does this correspond to this part in the [Mina book](https://o1-labs.github.io/proof-systems/specs/poseidon.html)?
```
def apply_mds(state):
    n = [0, 0, 0]
    n[0] = state[0] * mds[0][0] + state[1] * mds[0][1] + state[2] * mds[0][2]
    n[1] = state[0] * mds[1][0] + state[1] * mds[1][1] + state[2] * mds[1][2]
    n[2] = state[0] * mds[2][0] + state[1] * mds[2][1] + state[2] * mds[2][2]
    return n
```

dannywillems: Seems ok (modulo a transpose, depends on how you represent your matrix). However, the Mina book does describe an instance of Poseidon with only full rounds (the one actually used in Mina at the moment).

dannywillems: The one changed here is the "traditional" one, where we have full and partial rounds. See Figure 2 of the paper https://eprint.iacr.org/2019/458

dannywillems: Note that this PR is about the fact that the output of the function `apply_mds_matrix` was just simply never kept. It was doing the computation, but the state wasn't updated.

volhovm (approved):
My main question was "this seems like a core poseidon spec implementation, why changing it does not affect any code? prod code?"

Approving on assumption that it's actually not used anywhere else.

dannywillems:
> My main question was "this seems like a core poseidon spec implementation, why changing it does not affect any code? prod code?"

It seems not used. The only Poseidon implementation we use is with full rounds. `half_rounds` is activated when the parameter [HALF_FULL_ROUNDS](https://github.com/o1-labs/proof-systems/blob/master/poseidon/src/constants.rs#L7) is non-zero.

PR #2329, opened 2024-06-14
IVC/Poseidon: activate test

dannywillems:


ivc/src/poseidon_params_8_56_5_3.rs
```diff
@@ -979,9 +979,9 @@ impl SpongeConstants for PlonkSpongeConstantsIVC {
     const SPONGE_CAPACITY: usize = 1;
     const SPONGE_WIDTH: usize = 3;
     const SPONGE_RATE: usize = 2;
-    const PERM_ROUNDS_FULL: usize = 8;
+    const PERM_ROUNDS_FULL: usize = 0;
```
volhovm, line 982: I want to understand this better: this poseidon is `8_56`, so it's 4 full rounds, 56 partial, and 4 again. Why is `PERM_ROUNDS_FULL=0` then? And what are "HALF_ROUNDS_FULL?"...

ivc/src/poseidon_params_8_56_5_3.rs
```diff
@@ -1008,9 +1008,9 @@ mod tests {
             Fp::from_str("1").unwrap(),
         ];
         let exp_output_str = [
-            "1f1a9ee0bafa6811f1f9bd1e9a7a1ffe6bed28608a11a07833058622f2089b0e",
```
volhovm, line 1011: Do we have any other regression tests that make sure that other parameterizations were not changed in the previous PR that changed the poseidon spec in `poseidon/` crate?

ivc/src/poseidon_params_8_56_5_3.rs
volhovm: `SpongeConstants` is undocumented :(

volhovm: Ah OK I see now:
* If `SC::PERM_HALF_ROUNDS_FULL > 0` we're in half-round mode, then Poseidon will do `PERM_HALF_ROUNDS_FULL` full rounds, then `PERM_ROUNDS_PARTIAL` partial rounds, then again `PERM_HALF_ROUNDS_FULL` ones
* Otherwise we're in full-round mode and Poseidon will do `PERM_ROUNDS_FULL` full rounds.

volhovm (approved):

volhovm (approved):

PR #2336, opened 2024-06-17
Move byte-checks of Keccak to Syscall side and include lookups for length bytes

dannywillems:
Cherry-picked from https://github.com/o1-labs/proof-systems/pull/2276/


dannywillems:
Closing to keep https://github.com/o1-labs/proof-systems/pull/2276. Was simpler to use git merge than cherry-picking.

PR #2269, opened 2024-05-30
WIP - Integrate folding with IVC

dannywillems:

PR #2338, opened 2024-06-17
IVC/columns: add MISC comments

dannywillems:
Cherry-picked from https://github.com/o1-labs/proof-systems/pull/2269/

### DIFF
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -48,6 +48,11 @@ pub type IVCPoseidonColumn = PoseidonColumn<
 ///
 ///```text
 ///
+///         Inputs:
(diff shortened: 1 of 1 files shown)


ivc/src/ivc/columns.rs
```diff
@@ -67,7 +72,7 @@ pub type IVCPoseidonColumn = PoseidonColumn<
 ///      0       ...     34*2    76    80
 ///
 ///
-///                      Hashes
+///                      Hashes (temporarily DISABLED)
```
volhovm, line 75: It'd make more sense to also add disabling of hashes into this PR perhaps. I think cherry-picking the IVC bits is harder than folding...

volhovm (approved):

PR #2340, opened 2024-06-17
Poseidon: expose publicly bn254 parameters for 8-56-5-3-2 128 bits

dannywillems:
This will make it reusable outside. Also, it is important that the user knows that it is only for bn254, and no other curves. The parameters must be different for other curves.

The dependency on bn254 should be later considered as a feature. The IVC circuit is generic enough over any elliptic curve without cycle.

In https://github.com/o1-labs/proof-systems/pull/2269/files, it redefines the same structure. Removing here to avoid duplication.


ivc/src/poseidon_8_56_5_3_2/bn254/mod.rs
```diff
@@ -1,10 +1,18 @@
...
+use ark_bn254::Fr;
 use mina_poseidon::{constants::SpongeConstants, poseidon::ArithmeticSpongeParams};
 use once_cell::sync::Lazy;
 
+pub const STATE_SIZE: usize = 3;
+pub const NB_FULL_ROUND: usize = 8;
+pub const NB_PARTIAL_ROUND: usize = 56;
+pub const NB_TOTAL_ROUND: usize = NB_FULL_ROUND + NB_PARTIAL_ROUND;
+
+pub type Column = PoseidonColumn<STATE_SIZE, NB_FULL_ROUND, NB_PARTIAL_ROUND>;
```
volhovm, line 14: This could be redefined `PoseidonColumn` maybe... I find generic local-scope names always very confusing.

volhovm (approved):

PR #2341, opened 2024-06-17
IVC: reuse definitions of Poseidon parameters for bn254

dannywillems:
In https://github.com/o1-labs/proof-systems/pull/2269/files, it redefines the same structure. Removing here to avoid duplication.

### DIFF
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -1,4 +1,14 @@
-use crate::{ivc::interpreter::N_LIMBS_XLARGE, poseidon_8_56_5_3_2::columns::PoseidonColumn};
+use crate::{
+    ivc::interpreter::N_LIMBS_XLARGE,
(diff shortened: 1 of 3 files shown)


ivc/src/ivc/columns.rs
```diff
@@ -1,4 +1,14 @@
-use crate::{ivc::interpreter::N_LIMBS_XLARGE, poseidon_8_56_5_3_2::columns::PoseidonColumn};
+use crate::{
+    ivc::interpreter::N_LIMBS_XLARGE,
+    poseidon_8_56_5_3_2::{
+        bn254::{
+            Column as IVCPoseidonColumn, NB_FULL_ROUND as IVC_POSEIDON_NB_FULL_ROUND,
```
volhovm, line 5: This approach with renaming I'm somewhat happy about... at least there are no 5 types of `Column` in the same file :)

volhovm (approved):

PR #2342, opened 2024-06-17
IVC/Column: add FoldIteration

dannywillems:
All columns must be off by one. A better method should be used to avoid the duplicated code `add_rel_offset`, but it is left for another commit.

Also, this doesn't implement how the FoldIteration column is filled in the interpreter. It will come in another commit.


ivc/src/ivc/columns.rs
```diff
@@ -336,80 +341,86 @@ impl ColumnIndexer for IVCColumn {
...
+            IVCColumn::FoldIteration => Column::Relation(0),
             IVCColumn::BlockSel(i) => {
                 assert!(i < N_BLOCKS);
                 Column::FixedSelector(i)
             }
 
             IVCColumn::Block1Input(i) => {
                 assert!(i < 2 * N_LIMBS_SMALL);
-                Column::Relation(i)
+                Column::Relation(i).add_rel_offset(1)
```
volhovm, line 352: In the IVC PR, a lot of this copy-paste is avoided by applying `add_rel_offset` to the whole block.

volhovm (approved):

PR #2343, opened 2024-06-17
WIP: define a trait that defines IVC properties of an env

dannywillems:
We should follow the same design as in `o1vm/src/mips`. The witness environment carries some additional values that are updated over the rows. The state of the hash(es) is (are) saved in the environment.
The current approach of the interpreter is different: the state is passed as parameters. We should update that.

### DIFF
--- a/ivc/src/ivc/interpreter.rs
+++ b/ivc/src/ivc/interpreter.rs
@@ -36,6 +36,37 @@ use super::{
     LIMB_BITSIZE_XLARGE, N_LIMBS_XLARGE,
 };
 
+pub enum HashSide {
+    Left,
+    Right,
+    RightScaled,
+    Challenges,
+}
+
+/// A trait that defines methods an environment must have to be able to
+/// incrementally verifying the computation.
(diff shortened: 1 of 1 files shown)


ivc/src/ivc/interpreter.rs
```diff
@@ -36,6 +36,37 @@ use super::{
...
+pub enum HashSide {
+    Left,
+    Right,
+    RightScaled,
+    Challenges,
+}
+
+/// A trait that defines methods an environment must have to be able to
+/// incrementally verifying the computation.
+pub trait IVCEnv<F: PrimeField, Ff: PrimeField>:
```
volhovm, line 48: This trait makes the whole intention of a granular capability access pointless: the code that runs under this environment is fixed forever and cannot be reused for anything else. If everyone's on the same page here than we can proceed. I don't find the design approach convenient; to me it looks like `public static` in java: global state, global effects.

(on top of that we lose composability; but I assume we anyway cannot imagine two `IVCBuilderEnv`s with e.g. different lookup tables, since it's a top-level wrapper)

PR #2347, opened 2024-06-18
Add test for correctness of FEC ADD

volhovm:
Provides a test that checks that FEC ADD circuit actually computes `p + q`. Now in the test the point is actually valid, and we compare the output with the native implementation on Pallas.

Closes #2230.


dannywillems (approved):

dannywillems (approved):
LGTB. Tested locally with
```
cargo nextest run "test_fec_addition_circuit" --release --all-features --nocapture
```

PR #2351, opened 2024-06-18
Fix `ColumnIndexer` in MIPS (wrong order of selector columns)

querolita:
As part of the debugging process I realized that the witness columns storing the values for the dynamic selectors (instructions) in MIPS were not being mapped to the right indices. In particular, they were appearing _at the beginning_ of the witness, instead of _at the end_, which could be causing potential overwrites of relation witness data (for example, the selector column for `SyscallReadPreimage` was 11 instead of 106, which is 95 (relation columns) + 11, but the scratch state 11 is meant to store other type of data).

I thought that could have been the problem behind the failing constraints containing `self.equal()` (these have a `-1` which can only hold when real data is used, meaning not all zeros). But those are still failing after this fix, so it must be something else. I am printing out the witness for some instructions to check by hand why those constraints are not satisfied.

### DIFF
--- a/o1vm/src/lib.rs
+++ b/o1vm/src/lib.rs
@@ -49,7 +49,7 @@ pub use ramlookup::{LookupMode as RAMLookupMode, RAMLookup};
 /// trace.
 /// As a reminder, a constraint can be formally defined as a multi-variate
 /// polynomial over a finite field. The variables of the polynomial are defined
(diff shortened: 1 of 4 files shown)


dannywillems (approved):

dannywillems:
Ran the tests locally. LGTM.

Issue #2335, opened 2024-06-17
Soundness of syscall read preimage


dannywillems:
Nit (not for this PR): a good documentation, IMO, is also a documentation that describes what the parameters of the function does, and in particular, explains why there are some unused parameters when it is in the implementation of a trait.
For instance, here, I would have explained why `_addr` and `_len` are unused in the constraints side, when it is used in the witness. And also explain what `pos` is expected to be.

_Originally posted by @dannywillems in https://github.com/o1-labs/proof-systems/pull/2274#discussion_r1642929659_

dannywillems:
Left some comments on the PR !2355 we should have a look at later.

PR #2355, opened 2024-06-18
Soundness: use lookup table to check length of `SyscallReadPreimage`

querolita:
Partially addresses https://github.com/o1-labs/proof-systems/issues/2335

As suggested by Danny, I am using here a simple lookup to check that the values of `len`, `actual_read_len` and `num_preimage_bytes_read` are in the right range: `0 <= num_preimage_bytes_read <= actual_read_len <= len <= 4`.

TODO: in the future, as soon as we manage to satisfy the instructions containing `self.equal()` constraints, substitute the product definition of`num_preimage_bytes_read` with the `self.equal()` expressions that are commented in text. This will increase the number of constraints in a first approach, but they will be smaller degree (so after quadri, it is likely comparable).


o1vm/src/mips/constraints.rs
```diff
@@ -415,65 +416,75 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
-            let read_4 = (num_read_bytes.clone())
-                * (num_read_bytes.clone() - Expr::from(1))
-                * (num_read_bytes.clone() - Expr::from(2))
-                * (num_read_bytes.clone() - Expr::from(3));
-
-            // Note these constraints also hold when 0 preimage bytes are read
+            // of preimage bytes are read (case 0 bytes used when bytelength is read)
+            // TODO: embed any more complex logic to know how many bytes are read
+            //       depending on the address and length as in the witness?
+            // FIXME: use the lines below when the issue with `equal` is solved
```
dannywillems, line 438: I do not understand what you mean with "when the issue with 'equal' is solved". The commented lines below will never be satisfied all at the same time, as only one will be true.

o1vm/src/mips/constraints.rs
```diff
@@ -482,67 +493,66 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+                // When has_4_byte, then only can read 4 preimage bytes
+                self.add_constraint(
+                    has_n_bytes[3].clone() * (num_preimage_bytes_read.clone() - Expr::from(4)),
+                );
             }
         }
 
+        // FIXED LOOKUPS
+        // Check that 0 <= preimage read <= actual read <= len <= 4
+        self.add_lookup(Lookup::write_one(
```
dannywillems, line 535: Note (reminder for me): good that we check. We must check that `len`, a value coming from a register, is smaller than 4 because we do not do it in any other place.

o1vm/src/mips/constraints.rs
```diff
@@ -482,67 +493,66 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+            LookupTableIDs::AtMost4Lookup,
+            vec![actual_read_bytes.clone()],
+        ));
+        self.add_lookup(Lookup::write_one(
+            LookupTableIDs::AtMost4Lookup,
+            vec![num_preimage_bytes_read.clone()],
+        ));
+        self.add_lookup(Lookup::write_one(
+            LookupTableIDs::AtMost4Lookup,
+            vec![len.clone() - actual_read_bytes.clone()],
```
dannywillems, line 549: Note (for me): checking that `actual_read_bytes <= len`, i.e. we check that we didn't read more bytes that we have been asked for.

o1vm/src/mips/constraints.rs
```diff
@@ -482,67 +493,66 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        ));
+        self.add_lookup(Lookup::write_one(
+            LookupTableIDs::AtMost4Lookup,
+            vec![actual_read_bytes.clone() - num_preimage_bytes_read.clone()],
```
dannywillems, line 553: Note (for me): I always forget the difference between these two variables. The variable `actual_read_bytes` is the number of bytes that it reads from the oracle, which can be either the first 8 bytes that contain the length or the actual bytes of the preimage. `num_preimage_bytes_read` is only non-zero when we are actually getting values of the preimage, and it will be maximum the number of bytes we actually read from the preimage. I think we still need to verify later the whole semantics. I am quite sure we can simplify the logic.

dannywillems (approved):

o1vm/src/mips/constraints.rs
querolita: I meant that the lines using `equal` are equivalent but cleaner than the lines using the explicit values with multiplications. Nonetheless, they were failing: I know only one will be true and the rest won't, same as with the multiplication values. What I meant is that the constraints using `preimage_n` below fail when `preimage_n` is defined with `equal` instead of the multiplications approach. And I believe we should express them with `equal` in the future and fix that issuel.

PR #2357, opened 2024-06-18
Soundness: range checks for instruction parts

querolita:


o1vm/src/mips/interpreter.rs
```diff
@@ -501,6 +501,44 @@ pub trait InterpreterEnv {
...
+    fn lookup_16bits(&mut self, value: &Self::Variable) {
+        self.add_lookup(Lookup::read_one(
+            LookupTableIDs::RangeCheck16Lookup,
+            vec![value.clone()],
+        ));
+    }
+
+    /// Range checks with 2 lookups to the RangeCheck16Lookup table that a value
+    /// is at most 2^`bits`-1  (bits <= 16).
+    fn range_check16(&mut self, value: &Self::Variable, bits: u32) {
```
dannywillems, line 514: We do have only a SRS of size 2^15. Is it not problematic to perform a lookup this way?

o1vm/src/mips/interpreter.rs
```diff
@@ -501,6 +501,44 @@ pub trait InterpreterEnv {
...
+
+    /// Range checks with 2 lookups to the ByteLookup table that a value
+    /// is at most 2^`bits`-1  (bits <= 8).
+    fn range_check8(&mut self, value: &Self::Variable, bits: u32) {
+        assert!(bits <= 8);
+        // 0 <= value < 2^bits
+        // First, check lowerbound: 0 <= value < 2^8
+        self.lookup_8bits(value);
+        // Second, check upperbound: value + 2^8 - 2^bits < 2^8
+        self.lookup_8bits(&(value.clone() + Self::constant(1 << 8) - Self::constant(1 << bits)));
```
dannywillems, line 539: Interesting approach. It does save an additional table and an additional multiplicity polynomial IIUC.

o1vm/src/mips/interpreter.rs
```diff
@@ -1164,18 +1220,18 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
             // with the rest until there is a success). This also simplifies the implementation
             // here, so we will follow suit.
             let bytes_to_preserve_in_register = {
-                // FIXME: Requires a range check
                 let pos = env.alloc_scratch();
                 unsafe { env.bitmask(&write_length, 2, 0, pos) }
             };
+            env.range_check8(&bytes_to_preserve_in_register, 2);
```
dannywillems, line 1226: Nit: there is a table you recently added that checks if a value is in `[0, 4(`.

o1vm/src/mips/interpreter.rs
```diff
@@ -1236,6 +1292,7 @@ pub fn interpret_rtype<Env: InterpreterEnv>(env: &mut Env, instr: RTypeInstructi
                         let pos = env.alloc_scratch();
                         unsafe { env.bitmask(&addr, 2, 0, pos) }
                     };
+                    env.range_check8(&byte_subaddr, 2);
```
dannywillems, line 1295: Nit: there is a table you recently added that checks if a value is in [0, 4(.

o1vm/src/mips/interpreter.rs
```diff
@@ -2152,10 +2228,10 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
             };
 
             let byte_subaddr = {
-                // FIXME: Requires a range check
                 let pos = env.alloc_scratch();
                 unsafe { env.bitmask(&addr, 2, 0, pos) }
             };
+            env.range_check8(&byte_subaddr, 2);
```
dannywillems, line 2234: Nit: there is a table you recently added that checks if a value is in [0, 4(.

o1vm/src/mips/interpreter.rs
```diff
@@ -2377,6 +2457,7 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
...
```
dannywillems, line 2460: Nit: there is a table you recently added that checks if a value is in [0, 4(.

o1vm/src/mips/interpreter.rs
```diff
@@ -2463,10 +2544,10 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: ITypeInstructi
             };
 
             let byte_subaddr = {
-                // FIXME: Requires a range check
                 let pos = env.alloc_scratch();
                 unsafe { env.bitmask(&addr, 2, 0, pos) }
             };
+            env.range_check8(&byte_subaddr, 2);
```
dannywillems, line 2550: Nit: there is a table you recently added that checks if a value is in [0, 4(.

o1vm/src/mips/tests.rs
```diff
@@ -42,7 +42,7 @@ fn test_mips_number_constraints() {
     for instr in Instruction::iter().flat_map(|x| x.into_iter()) {
         match instr {
             RType(rtype) => match rtype {
-                JumpRegister | SyscallExitGroup | Sync => assert_num_constraints(&instr, 0),
+                JumpRegister | SyscallExitGroup | Sync => assert_num_constraints(&instr, 1),
```
dannywillems, line 45: Note: the additional constraint seems to be the one for the instruction decomposition. :+1:

o1vm/src/mips/witness.rs
```diff
@@ -177,6 +177,7 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
 
     fn add_lookup(&mut self, _lookup: Lookup<Self::Variable>) {
         // No-op, constraints only
+        // TODO: keep track of multiplicities of fixed tables here as in Keccak?
```
dannywillems, line 180: Yes, we should

dannywillems (commented):
LGTM.
Comments:
- see notes when checking values in [0, 4(, we can save 1 lookup.
- the fact we do 2 lookups when checking we are in `[0, 2^bits(` sounds interesting at a first sight as it does save one additional table.
- the lookup into 16 bits table value seems wrong, no? We do have only a SRS of size 2^15.

dannywillems (approved):

volhovm:
@dannywillems should we merge this before it gets unsynchronized from master too much?

querolita:
> the lookup into 16 bits table value seems wrong, no? We do have only a SRS of size 2^15.

@dannywillems 

I was expecting that the lookup argument could split the 2^16 entry table into two of 2^15, so that it becomes transparent at the caller function level.

PR #2365, opened 2024-06-18
IVC: activate hashes

dannywillems:
1.  made less generic the IVC interpreter. The Poseidon parameters are implemented only for the scalar field BN254 at the moment. We should make the Poseidon parameters more generic. I would prefer to use this method for now as the IVC interpreter is specialised for bn254 for 1.0.0
2. Update regression tests
3. The lens is not obvious. It took me a long time to understand how to use it :(.
4. Comment and add some FIXME in `process_hashes` on the witness side


ivc/tests/add.rs
```diff
@@ -336,7 +336,7 @@ pub fn test_simple_add() {
 
     let ivc_constraints: Vec<E<Fp>> = {
         let mut ivc_constraint_env = ConstraintBuilderEnv::<Fp, IVCLookupTable<Fq>>::create();
-        constrain_ivc::<Fp, Fq, _>(&mut ivc_constraint_env);
+        constrain_ivc::<Fq, _>(&mut ivc_constraint_env);
```
volhovm, line 339: Unfortunate, but minor, let's fix it later.

volhovm (approved):

PR #2366, opened 2024-06-19
[EASY] IVC: add the challenges required by the block selection and expose the IVC challenges count

dannywillems:
It is required to have a correctly shaped/sound IVC circuit.

### DIFF
--- a/ivc/src/ivc/mod.rs
+++ b/ivc/src/ivc/mod.rs
@@ -4,6 +4,9 @@ pub mod helpers;
 pub mod interpreter;
 pub mod lookups;
 
+use self::columns::N_BLOCKS;
+use crate::poseidon_8_56_5_3_2::bn254::NB_CONSTRAINTS as IVC_POSEIDON_NB_CONSTRAINTS;
+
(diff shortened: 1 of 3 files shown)


ivc/src/ivc/mod.rs
```diff
@@ -16,6 +19,14 @@ pub const N_LIMBS_XLARGE: usize = 2;
 // [folding::tests::test_quadraticization]
 pub const N_ADDITIONAL_WIT_COL_QUAD: usize = 48;
 
+/// Number of challenges in the IVC circuit.
+/// It is the maximum number of constraints per row.
+/// We do suppose the Poseidon circuit has the highest number of constraints,
+/// for now.
+/// We also add the number of "blocks" in the IVC circuit as there will be
+/// [N_BLOCKS] alphas required to aggregate all blocks on each row.
+pub const IVC_NB_CHALLENGES: usize = IVC_POSEIDON_NB_CONSTRAINTS + N_BLOCKS;
```
volhovm, line 28: Right now this will not be enough: folding needs the /sum/ of constraints, not the max(block_constraints) + common_constraints. To be optimised in the future.

volhovm (approved):

PR #2367, opened 2024-06-20
[EASY] IVC/Columns: simply move circuit layout to top-level + doc block height

dannywillems:
This is useful to generate the doc, and verifying the code matches the specification while reviewing/implementing.
I noticed some inconsistencies between the code and the specification.
Also, changing 6 to 12 for the number of hashes required per commitment. See the comment in the code. I added https://github.com/o1-labs/proof-systems/pull/2368 to verify we can still handle the MIPS interpreter.

### DIFF
--- a/ivc/src/ivc/columns.rs
+++ b/ivc/src/ivc/columns.rs
@@ -1,4 +1,216 @@
+//! IVC circuit layout - Top level documentation outdated
+//!
+//! The IVC circuit is tiled vertically. We assume we have as many
+//! rows as we need: if we don't, we wrap around and continue.
+//!
+//! The biggest blocks are hashes and ECAdds, so other blocks may be wider.
+//!
+//! `N := N_IVC + N_APP` is the total number of columns in the circuit.
+//!
+//! Vertically stacked blocks are as follows:
+//!
+//!```text
+//!
+//!         Inputs:
+//!      Each point is 2 base field coordinates in 17 15-bit limbs
+//!       recomposed as 8 75-bit limbs
+//!       recomposed as 4 150-bit limbs.
+//!
+//!              34            8      4
+//!            Input1         R75   R150
+//!  1   |-----------------|-------|----|
+//!      |      C_L1       |       |    |
+//!      |      C_L2       |       |    |
+//!      |                 |       |    |
+//!      |      ...        |       |    |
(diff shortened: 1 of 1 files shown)


ivc/src/ivc/columns.rs
```diff
@@ -1,3 +1,214 @@
+//! IVC circuit layout - Top level documentation outdated
```
dannywillems, line 1: Simply moving to the top + add one ```text`

ivc/src/ivc/columns.rs
```diff
@@ -1,3 +1,214 @@
...
+//!    |--|--------|--------------------|
+//!
+//!
+//! 2^15 |---- --------------------------------------|
+//!```
+//!
+//!
+//! Assume that IVC circuit takess CELL cells, e.g. CELL = 10000*N.
+//! Then we can calculate N_IVC as dependency of N_APP in this way:
+//! ```text
```
dannywillems, line 187: Text has been added here.

ivc/src/ivc/columns.rs
```diff
@@ -19,230 +231,45 @@ use kimchi_msm::{
...
+        // chunk, therefore we need (2 * 2 * 3) (= 12) hashes per commitment.
+        // FIXME: why an additional 2?
+        // FIXME: use the encoding of the bit of y instead. It would reduce to 6
+        // hashes instead of 12.
+        //   For a EC point (x, y), we can only rely on x and the bit sign of y.
+        //   We decompose x into `x_{0..149}` and `x_{150..255}` (only 255 as
+        //   the foreign field is 255 bits), and we add into the value
+        //   `x_{150..255}` the bit sign of `y`. From there, we can hash the two
+        //   values, which fits in our state of size 3.
+        1 => 12 * N_COL_TOTAL + 2,
```
dannywillems, line 255: See comment for this change. Was 6 before. Added https://github.com/o1-labs/proof-systems/pull/2368 to verify it is still working for our first real usecase.

dannywillems: My comment was wrong. I deleted the commit https://github.com/o1-labs/proof-systems/commit/17f4d5dec8658299a42d18642b2eb9023c0e415f

volhovm (approved):

PR #2369, opened 2024-06-21
test: add the unit test for computing the `domain_size`

### LINKED ISSUE #1335: Add a test checking the domain size when building a circuit with gates using same table.
Approving, but would have been nice to have a test, even it looks fine regarding the code. Opening a follow-up issue for a 10 min task.

_Originally posted by @dannywillems in https://github.com/o1-labs/proof-systems/pull/1334#pullrequestreview-1734526395_
           
 Before this PR, it was duplicating the domain if two gates were using the same table.

duguorong009:
## Description
Add the unit test for checking the `domain_size` of circuit, when the gates use the same table

## Related issues
- Resolve #1335 

## Changes
- add the `test_lookup_domain_size_computation` in `kimchi/src/circuits/constraints.rs`

### DIFF
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -970,9 +970,9 @@ pub mod tests {
         let (_, xor_gates_1) = CircuitGate::<Fp>::create_xor_gadget(next_start, 3); /* 1 xor gate */
         let circuit_gates: Vec<CircuitGate<Fp>> = range_check_gates_0
             .into_iter()
-            .chain(range_check_gates_1.into_iter())
-            .chain(xor_gates_0.into_iter())
-            .chain(xor_gates_1.into_iter())
+            .chain(range_check_gates_1)
+            .chain(xor_gates_0)
+            .chain(xor_gates_1)
             .collect(); /* 2 range check gates + 2 xor gates */
 
         // inputs + expected output
--- a/kimchi/src/circuits/constraints.rs
+++ b/kimchi/src/circuits/constraints.rs
@@ -914,7 +914,8 @@ impl<F: PrimeField + SquareRootField> Builder<F> {
 #[cfg(test)]
 pub mod tests {
     use super::*;
-    use mina_curves::pasta::Fp;
+    use mina_curves::pasta::{Fp, Fq};
+    use o1_utils::FieldHelpers;
 
     impl<F: PrimeField + SquareRootField> ConstraintSystem<F> {
         pub fn for_testing(gates: Vec<CircuitGate<F>>) -> Self {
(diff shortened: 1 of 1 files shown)


dannywillems (commented):
Thanks for the contribution! Really appreciated!
The code looks good and clean, good job!

However, I do not think it does fulfil all the requirements to close the issue. The idea of the issue was to verify that if a circuit uses the gate [Foreign Field Multiplication](https://o1-labs.github.io/proof-systems/kimchi/foreign_field_mul.html) alongside the gate "rangecheck", then, the table "rangecheck" will be added only once in the domain size computation. You can see that the foreign field multiplication gate does require performing a range check.

The tests you are suggested are still correct. I would leave this review as a comment, for now, in case you would like to add additional tests based on my comment above. Happy to give more guidance, if needed.

duguorong009:
> Thanks for the contribution! Really appreciated! The code looks good and clean, good job!
> 
> However, I do not think it does fulfil all the requirements to close the issue. The idea of the issue was to verify that if a circuit uses the gate [Foreign Field Multiplication](https://o1-labs.github.io/proof-systems/kimchi/foreign_field_mul.html) alongside the gate "rangecheck", then, the table "rangecheck" will be added only once in the domain size computation. You can see that the foreign field multiplication gate does require performing a range check.
> 
> The tests you are suggested are still correct. I would leave this review as a comment, for now, in case you would like to add additional tests based on my comment above. Happy to give more guidance, if needed.

Thanks for review! @dannywillems 
How about this simple addition? https://github.com/o1-labs/proof-systems/pull/2369/commits/7c22692fa36f87acfa611d6a4b9c64129efd9a46

kimchi/src/circuits/constraints.rs
```diff
@@ -961,4 +962,54 @@ pub mod tests {
...
+            (
+                (0, 0),
+                8192, /* 8192 > 0 * 0 + 1 * 4096 + 1 * 256 + 1 + zk_row */
+            ),
+            (
+                (5, 100),
+                8192, /* 8192 > 5 * 100 + 1 * 4096 + 1 * 256 + 1 + zk_row */
+            ),
+        ];
+        for ((number_of_table_ids, size), expected_domain_size) in data.into_iter() {
```
dannywillems, line 997: Nit: personal preference to use `data.into_iter().for_each()` for a functional approach.

dannywillems (approved):
LGTM, thank you! I approve.

kimchi/src/circuits/constraints.rs
duguorong009: Done https://github.com/o1-labs/proof-systems/pull/2369/commits/32637674c1975382ae82179db22b9324a11f8c43

PR #2371, opened 2024-06-24
MSM: add regression tests for FEC

dannywillems:


msm/src/fec/mod.rs
```diff
@@ -83,6 +83,56 @@ mod tests {
...
+                *constraints_degrees.entry(degree).or_insert(0) += 1;
+            });
+
+            assert_eq!(constraints_degrees.get(&1), Some(&1));
+            assert_eq!(constraints_degrees.get(&2), Some(&5));
+            assert_eq!(constraints_degrees.get(&3), Some(&21));
+            assert_eq!(constraints_degrees.get(&4), None);
+            assert_eq!(constraints_degrees.get(&5), Some(&2));
+            assert_eq!(constraints_degrees.get(&6), None);
+            assert_eq!(constraints_degrees.get(&7), Some(&34));
```
volhovm, line 132: Degree 7 here is because of the lookups I assume.

volhovm (approved):

msm/src/fec/mod.rs
dannywillems: Yes.

PR #2372, opened 2024-06-24
IVC: base case - temporary solution

dannywillems:
TODO:
- [x] update regression tests for the degree of the constraints
- [x] add a test with fold_iteration sets to 0
- [x] add a test with fold_iteration sets to 1


volhovm (approved):

dannywillems:
I deactivate the completeness tests as they were failing. See #2373 for the reactivation of them.

PR #2373, opened 2024-06-25
Nextest: reactivate completeness

dannywillems:


.config/nextest.toml
```diff
@@ -122,4 +130,10 @@ failure-output = "immediate-final"
 
 # The number of threads to run tests with. Supported values are either an integer or
 # the string "num-cpus". Can be overridden through the `--test-threads` option.
-test-threads = "num-cpus"
\ No newline at end of file
+test-threads = "num-cpus"
+
+[[profile.ci.overrides]]
+filter = "test(completeness)"
```
dannywillems, line 136: https://nexte.st/docs/filtersets/

volhovm: Nextest docs have a section on heavy tests: https://nexte.st/docs/configuration/threads-required/

.config/nextest.toml
```diff
@@ -122,4 +130,10 @@ failure-output = "immediate-final"
...
+threads-required = 2
```
volhovm, line 137: My prediction would be that it's more optimal to give all the heavy tests *all* the threads (or at least half), because they're quite parallelisable anyway. But we can do it in the future when we hit the ceiling next time.

volhovm (approved):

PR #2375, opened 2024-06-25
IVC: regression tests for the number of additional columns added by the reduction to degree 2

dannywillems:
The tests pass, but it is only because the additional columns are not considered in the completeness tests for now. It will fail when we do use folding with it.


volhovm (approved):
Ultimately though I believe this belongs to `folding`, since anything quadraticization is not a property of a gadget, but a property of the quadraticization algorithm (that can be implemented in many ways).

dannywillems:
> What you're testing is that quadriticization works "as expected" on a particular example (here, IVC)

No, It is a regression test to keep track of the number of columns added by quadri for the IVC circuit.

Issue #2313, opened 2024-06-12
Folding + IVC for Add Circuit


volhovm:
We have a circuit for Add, we have IVC, but these two need to be concatenated properly. 

This task can be closed when:
- IVC + Add circuit are concatenated and folded properly
- The combined witness satisfies the combined expression
- We can pass the combined instance into (a variant) of plonkish prover (to be implemented within this task)
- This prover succeeds and veriifes (completeness test)

PR #2378, opened 2024-06-26
[#2313] Porting IVC+ADD 3: Pass on the usize constants

volhovm:


mrmr1993 (approved):

ivc/src/ivc/mod.rs
```diff
@@ -19,21 +19,28 @@ pub const N_LIMBS_XLARGE: usize = 2;
...
 
-/// Number of challenges in the IVC circuit.
-/// It is the maximum number of constraints per row.
-/// We do suppose the Poseidon circuit has the highest number of constraints,
-/// for now.
-/// We also add the number of "blocks" in the IVC circuit as there will be
-/// [N_BLOCKS] alphas required to aggregate all blocks on each row.
-pub const IVC_NB_CHALLENGES: usize = IVC_POSEIDON_NB_CONSTRAINTS + N_BLOCKS;
+/// Number of constraints used by the IVC circuit.
+pub const N_CONSTRAINTS: usize = N_CONSTRAINTS_POSEIDON + N_BLOCKS + 61;
```
fabrizio-m, line 23: where does this `+ 61` come from?

volhovm: It's ... the rest of the IVC :) Like EC additions, decomposing inputs, and so on. We could split it down into a sum of 6 other constants, but I see no need for now. Maybe later?

dannywillems: Would be good to mention where it comes. Unclear for now.

PR #2380, opened 2024-06-26
[#2313] Porting IVC+ADD 5: Add inductive case IVC & second folding call

volhovm:


dannywillems (approved):
Still a lot of FIXME, and not correct code, but fine to move on. I opened some PR to follow-up on some suggestions.

PR #2381, opened 2024-06-26
[#2313] Porting IVC+ADD 6: Add folding expression evaluation

volhovm:


ivc/tests/add.rs
```diff
@@ -341,7 +346,7 @@ pub fn heavy_test_simple_add() {
         }
     }
 
-    type Config = StandardConfig<
+    type Config<const N_COL: usize, const N_FSEL: usize> = StandardConfig<
```
dannywillems, line 349: Why this change?

dannywillems: It is adding noise, but ok... I would simply remove this.

ivc/tests/add.rs
```diff
@@ -416,13 +562,13 @@ pub fn heavy_test_simple_add() {
     );
 
     // Make the constraints folding compatible
-    let app_compat_constraints: Vec<FoldingCompatibleExpr<Config>> = app_constraints
+    let app_compat_constraints: Vec<FoldingCompatibleExpr<MainTestConfig>> = app_constraints
```
dannywillems, line 565: Noise from the change above. The PR should be only the changes above.

ivc/tests/add.rs
```diff
@@ -350,6 +355,147 @@ pub fn heavy_test_simple_add() {
         (),
         GenericVecStructure<Curve>,
     >;
+    type MainTestConfig = Config<N_COL_TOTAL, N_FSEL_TOTAL>;
+
+    ////////////////////////////////////////////////////////////////////////////
+    // Evaluations
+    ////////////////////////////////////////////////////////////////////////////
+
+    #[allow(dead_code)]
```
dannywillems, line 364: I would guess it is used later.

ivc/tests/add.rs
volhovm: It will be generalized and moved out of this file (into the library) 5 PRs later, to be used with different circuits.

Here it was originally introduced because I was experimenting with several setups at the same time.

dannywillems (approved):
Approving as other PRs are built on top of this, but I would still remove the noise added by the type alias.

ivc/tests/add.rs
volhovm: Is removed eventually as far as I remember.

PR #2388, opened 2024-06-28
FEC ADD: Decrease degree to 2

volhovm:
FEC ADD had degree 3 because of `q1_sign`, and about `229` columns. Now it has degree 2 and about `245` columns because I did quadraticisation in place. I added "big" q1/q2/q3 limbs, and I multiply them by sign separately instead of multiplying all together.


fabrizio-m (approved):

volhovm:
I'm forever fixing an infinite amount of regression tests that changed because number of columns and expressions changed. Really doubting if it was worth it.

volhovm:
At last.

PR #2389, opened 2024-06-28
Adding multirow lookup into MSM serialization

volhovm:
Adds:
* Capability for vector lookup (no actual support, just interface)
* Some adjustments to support runtime lookup tables
* The "multiplication bus" runtime lookup table and support columns in the MSM serialization circuit
* Code invoking this lookup capability.


volhovm:
Closed as split into several sub-PRs.

PR #2392, opened 2024-07-01
Folding Prover 2: Move IVC folding expresion evaluation into separate module

volhovm:
Moves code into a separate module. This time it's the code that evaluates folding expressions.


dannywillems (approved):
I suppose evaluate.rs can be removed later.

volhovm:
Ah yes, indeed.... :facepalm:

PR #2396, opened 2024-07-08
Folding Prover 4: Add actual prover and verifier

volhovm:
This PR:
* Adds prover and verifier together with some prover types. This is 95%+ copy paste from the MSM.
* Generalizes evaluator for folding expressions to support vectors. This is needed because in the prover we evaluate over columns (polynomials `P(X)`), while in the verifier we evaluate expressions over zeta-evals (evaluations `P(zeta)`).
* Fixer a minor bug where `N_COL` (total column number) for Poseidon was only counting `N_WIT` (total non-fixed column number).

### DIFF
--- a/ivc/src/verifier.rs
+++ b/ivc/src/verifier.rs
@@ -1,0 +1,259 @@
+use crate::{
+    expr_eval::GenericEvalEnv,
+    plonkish_lang::{PlonkishChallenge, PlonkishWitnessGeneric},
+    prover::Proof,
+};
+use ark_ff::{Field, One};
+use ark_poly::{univariate::DensePolynomial, Evaluations, Radix2EvaluationDomain as R2D};
+use folding::{
+    eval_leaf::EvalLeaf, instance_witness::ExtendedWitness, FoldingCompatibleExpr, FoldingConfig,
+};
+use kimchi::{
+    self, circuits::domains::EvaluationDomains, curve::KimchiCurve, groupmap::GroupMap,
+    plonk_sponge::FrSponge, proof::PointEvaluations,
+};
+use kimchi_msm::columns::Column as GenericColumn;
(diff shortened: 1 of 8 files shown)


ivc/src/verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+        .chain(&proof_comms.witness_comms)
+        .for_each(|comm| absorb_commitment(&mut fq_sponge, comm));
+
+    ////////////////////////////////////////////////////////////////////////////
+    // Sampling scalars
+    ////////////////////////////////////////////////////////////////////////////
+
+    //let (joint_combiner, beta) = (None, Fp::zero());
+
+    // Sample α with the Fq-Sponge.
```
volhovm, line 108: I could remove these comments. I think `alpha` has to go for sure (as in we don't need to sample it). The joint combiner and beta we don't need it seems? I guess I wanted to double check.

ivc/tests/add.rs
```diff
@@ -866,16 +881,17 @@ pub fn heavy_test_simple_add() {
                 extended: folded_witness
                     .extended_witness
                     .extended
+                    .clone()
```
dannywillems, line 884: You don't need to clone. I will remove it in a next PR.

ivc/src/verifier.rs
```diff
@@ -0,0 +1,259 @@
...
+    };
+
+    let ft_eval0 = {
+        // We evaluate only at zeta
+        let point_eval_to_vec = |x: PointEvaluations<_>| vec![x.zeta];
+
+        let witness_evals_vecs = proof_evals
+            .witness_evals
+            .cols
+            .clone()
```
dannywillems, line 177: Again, we don't have to clone. I'll change this.

dannywillems (approved):
Modulo some changes that will come in a future PR

PR #2397, opened 2024-07-08
cache lagrange basis

### LINKED ISSUE #2414: Cache Lagrange Basis
We want to be able to cache the lagrange basis to speed up the mina CI tests. Full SRS caching was done in this repo in #2286 so  this doesn't need to be integrated into CI for this repo. Still there should be some basic tests that it's working (e.g. tests in the `poly-commitment` package)

- [x] add an `add_lagrange_basis_with_cache` method to the SRS type @martyall
- [x] allow for using local disk as cache backend @martyall 
- [x] verify disk cache works locally (e.g. using tests in `commitment.rs`) @martyall 
- [ ] allow using remote cache backend (either google cloud / aws bucket) @svv232 
- [ ] verify remote cache works  @svv232

martyall:
Fixes #2414


poly-commitment/src/srs.rs
```diff
@@ -104,6 +108,21 @@ impl<G: CommitmentCurve> SRS<G> {
...
+        //self.g.hash(&mut g_hasher);
+
+        //let mut h_hasher = DefaultHasher::new();
+        //self.h.hash(&mut h_hasher);
+
+        //println!("g: {:?}, h: {:?}, domain: {:?}", g_hasher.finish(), h_hasher.finish(), n);
+
         // Let V be a vector space over the field F.
         //
         // Given
```
svv232, line 128: we should remove this eventually

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+pub struct GoogleCloudCache<G> {
+    url: String,
+    point_type: PhantomData<G>,
+}
+
+impl<G> GoogleCloudCache<G> {
+    pub fn new(url: PathBuf) -> Self {
+        todo!("GoogleCloudCache::new")
+    }
+}
```
svv232, line 176: In theory we can use any remote file store not a google cloud bucket. I think something like `RemoteCache` is a slightly better naming scheme.

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn default() -> Self {
+        LagrangeCache::Local(FileCache::new(PathBuf::from("/tmp/lagrange_cache")))
+    }
+}
+
+// Keep this enum private for type safety
+pub enum LagrangeCacheKey {
+    Local(PathBuf),
+    Remote(()),
+}
```
svv232, line 28: imo a name like `LagrangeCacheFile` is a clearer name as Key implies this is for some map type. We should comment why we've called it `LagrangeCache[Key|File]` by explaining the storage scheme.

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl<G: CommitmentCurve> LagrangeCacheTrait<G> for LagrangeCache<G> {
+    type CacheKey = LagrangeCacheKey;
+
+    fn lagrange_basis_cache_key(&self, g: &Vec<G>, domain: &D<G::ScalarField>) -> Self::CacheKey {
```
svv232, line 33: did we also need h here?

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl<G: CommitmentCurve> LagrangeCacheTrait<G> for FileCache<G> {
+    type CacheKey = PathBuf;
+
+    fn lagrange_basis_cache_key(&self, g: &Vec<G>, domain: &D<G::ScalarField>) -> Self::CacheKey {
+        let mut hasher = DefaultHasher::new();
+        g.hash(&mut hasher);
+        domain.size.hash(&mut hasher);
```
svv232, line 111: same question as before i think we need h here

poly-commitment/src/lagrange_cache.rs
svv232: Now that I think about it, I see why this makes sense. Every struct should have a different fetching implementation, so you would implement this separately per service.

poly-commitment/src/srs.rs
```diff
@@ -29,6 +30,9 @@ pub struct SRS<G> {
     /// Commitments to Lagrange bases, per domain size
     #[serde(skip)]
     pub lagrange_bases: HashMap<usize, Vec<PolyComm<G>>>,
+
+    #[serde(skip)]
+    lagrange_bases_cache: Option<LagrangeCache<G>>,
```
volhovm, line 35: I'm not sure this is the best place to keep it. The only reason I see to keep the cache type as part of the SRS is that we're assuming that this cache might be updated when the SRS is updated.

Our current model is not like that: the SRS can be only updated (with more Lagrange bases) in memory, but this never affects the disk. To update the disk, you need to run a test (see `srs_serialization.rs`). This is optimal in the sense that SRS is only updated when we explicitly want it (and run the test), and not otherwise (which can be very confusing).


I'd probably suggest keeping `LagrangeCache` as a parameter to this test. Like, create a different test that is called something like `populate_lagrange_bases` and if it is run (with a `LagrangeCache` parameter), the test adds/removes bases and updates the cache. 

Am I missing some advantages of the suggested approach?

poly-commitment/src/lagrange_cache.rs
mrmr1993: Suggest: `source`

poly-commitment/src/lagrange_cache.rs
mrmr1993: You shouldn't do, `h` isn't included in the lagrange basis

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+}
+
+impl<G: CommitmentCurve> LagrangeCacheTrait<G> for FileCache<G> {
+    type CacheKey = PathBuf;
+
+    fn lagrange_basis_cache_key(&self, g: &Vec<G>, domain: &D<G::ScalarField>) -> Self::CacheKey {
+        let mut hasher = DefaultHasher::new();
+        g.hash(&mut hasher);
```
mrmr1993, line 110: Isn't this going to be rather expensive? The lagrange basis has been unchanged for several years, I think it's okay to assume it's constant.

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    fn cache_lagrange_basis(
+        &self,
+        g: &Vec<G>,
+        domain: &D<G::ScalarField>,
+        basis: &Vec<PolyComm<G>>,
+    ) {
+        let cache_key = self.lagrange_basis_cache_key(g, domain);
+        if Path::exists(&cache_key) {
+            println!("Lagrange basis cache file {:?} already exists", cache_key);
```
mrmr1993, line 150: TODO: Delete

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+            return;
+        } else {
+            println!("Caching lagrange basis to file {:?}", cache_key);
```
mrmr1993, line 153: TODO: Delete

poly-commitment/src/lagrange_cache.rs
```diff
@@ -0,0 +1,201 @@
...
+        basis: &Vec<PolyComm<G>>,
+    ) {
+        match self {
+            LagrangeCache::Local(cache) => cache.cache_lagrange_basis(g, domain, basis),
+            LagrangeCache::Remote(cache) => cache.cache_lagrange_basis(g, domain, basis),
+        }
+    }
+}
+
+pub trait LagrangeCacheTrait<G: CommitmentCurve> {
```
mrmr1993, line 68: It's generally considered bad practice/redundant to call a trait `Trait`

poly-commitment/src/srs.rs
mrmr1993: Agree: the cache is something you use, not something that we should carry around in the SRS. Consider a different approach.

poly-commitment/src/srs.rs
```diff
@@ -104,6 +108,21 @@ impl<G: CommitmentCurve> SRS<G> {
             return;
         }
 
+        self.lagrange_bases_cache.as_ref().map(|cache| {
```
mrmr1993, line 111: Similarly here: it's surprising if `add_lagrange_basis` actually does network IO. The cache shouldn't be integrated here

poly-commitment/src/lagrange_cache.rs
martyall: @mrmr1993 I noticed from debug printing during tests that the same domain size can appear for multiple different `G`-bases (the `g` field in the `SRS` struct) . Was I hallucinating or is this something that can happen?

Maybe the length of `g` is the only variable, in which case the cache key could be `(g.len(), domain.size())` ?

mrmr1993: The real variable is `domain.size()` and the curve, we use the same `g`s for all SRSs over the curve (and truncate if the domain doesn't require the full thing).

martyall: got it :ok_hand:

poly-commitment/src/lagrange_cache.rs
martyall: This type is no longer needed, I just needed to get my rust legs back and remember how to use the languaeg

poly-commitment/src/srs.rs
martyall: agreed, is now kept separately

poly-commitment/src/lagrange_cache.rs
martyall: As we discussed, it turns out that for reasons related to chunking we also need to keep the srs size as part of the cache key. Might not even be necessary within the `mina` repo, but it is within `proof-systems`. Still better than hashing the `g` field

poly-commitment/src/lagrange_cache.rs
martyall: will do, these are only here for sanity checking while we're getting set up

poly-commitment/src/lagrange_cache.rs
martyall: Yes it's lame, thankfully don't need this trait anymore

martyall:
closing this in favor of https://github.com/MinaProtocol/mina/pull/15843 as it is only used in the mina test suite.

PR #2404, opened 2024-07-09
Suggestion for 2380

dannywillems:


dannywillems:
This is ready now.

volhovm (approved):
Nice. I'm looking forward in going over the whole codebase and trying to use references more. I wanted to make things work ASAP, but 32GB+ memory consumption from cloning is not fun... and some big clones require changing structure of folding witness/instance, adding a lifetime parameter, etc...

PR #2405, opened 2024-07-09
IVC/folding: remove unessary clone derive

dannywillems:


volhovm:
Same as https://github.com/o1-labs/proof-systems/pull/2404#issuecomment-2217444789: this behaviour will be changed in the head of the PR train. We'll have to clone the folded witness and instance because there are a lot of pieces of code that don't work with references, and changing them is probably 500~+ lines of code. I ignore cloning in this train and suggest to handle it once-and-for-all after.

dannywillems:
Closing for now.

PR #2411, opened 2024-07-09
Move code and remove duplicated check

dannywillems:


ivc/src/expr_eval.rs
```diff
@@ -17,6 +19,21 @@ use kimchi::{
 use kimchi_msm::columns::Column as GenericColumn;
 use strum::EnumCount;
 
+#[derive(Clone)]
+/// Generic structure containing column vectors.
+pub struct GenericVecStructure<G: KimchiCurve>(pub Vec<Vec<G::ScalarField>>);
+
+impl<G: KimchiCurve> Index<GenericColumn> for GenericVecStructure<G> {
```
volhovm, line 26: I guess this instance I wrote implies that it should be called `FixSelectorsStructure` and not `GenericVecStructure`, but ... we can leave it for now.

volhovm (approved):

PR #2415, opened 2024-07-10
WIP: BN254 powers of tau parsing

dannywillems:
This is only to keep track of the research.
The format is not totally clear. It would be nice to have a separate library, and maybe publish it somewhere (and maintain it) for the community, and for visibility.
In the library, we should clearly specify the format, and make it easy to anyone to reproduce it in another programming language.

The SRS should be easily loadable to be used by tests in IVC/o1VM.
The user should be allowed to use any SRS size.
The code should not download the SRS, but a path should be given in a variable environment.

The goal is to use it for committing to polynomials using KZG.


dannywillems:
Closing as old, and we won't use it.

PR #2416, opened 2024-07-10
IVC: make some completeness tests heavy

dannywillems:


volhovm (approved):
I don't mind, but I think it's not terribly heavy -- 40 seconds on my machine and just a few gigs of RAM. I guess it can be...

dannywillems:
> I don't mind, but I think it's not terribly heavy -- 40 seconds on my machine and just a few gigs of RAM. I guess it can be seen as such...

It was failing on a 16GB RAM machine, cc @marcbeunardeau88

volhovm:
Ah OK my bad. I looked at it at a wrong time. The max memory consumption I observe is 7GB, which should still work on a 16GB machine. It's more than I thought (2-3) but still not 16+ :) can be probably optimised.

volhovm:
@marcbeunardeau88 note that the tests are supposed to be run with `--release`, otherwise we don't guarantee it won't be terribly slow:
```
 cargo test --release --package ivc --lib -- ivc::tests::heavy_test_completeness_ivc_general_case
```

marcbeunardeau88:
I ran only the `heavy_test_completeness_ivc_general_case` and indeed I'm around 7GB and it passes
Isn't the test command running several tests in parrallel and that would explain the higher peak I got ?
I believe @dannywillems told me that.
anyway the good news is that now my tests are passing when skipping heavy ones : )

PR #2418, opened 2024-07-10
Remove unnecessary fixed_selector double-building

volhovm:
`build_selectors` already contains building selectors for the Poseidon circuit, so these calls were unnecessary.


dannywillems (approved):
Nice

Issue #2089, opened 2024-04-18
MSM Serialization: Use lookups for multi-row communication


volhovm:
Serialization circuit row #i needs to "query" the data from some row #i/2 (roughly, see docs for precise numbers). This "query" needs to be done via a lookup.

PR #2421, opened 2024-07-12
[#2089] Introduce vector lookup interface

volhovm:
This makes it possible to write circuits that `lookup` vectors not single values. 

Functionally, it's a stub. The actual implementation for proving vector lookup is TBD.


volhovm:
Closed in favour of https://github.com/o1-labs/proof-systems/pull/2424

PR #2424, opened 2024-07-12
Re-add vector lookups

mrmr1993:


volhovm (approved):
I'd prefer to have `lookup_single` since it's a common use case, but we can do it later.

PR #2426, opened 2024-07-13
Super IVC: based on a variant of SuperNova

dannywillems:
Using the old "interpreter" model with a specific witness environment, not the one defined with capabilities.
Based on SuperNova, with possible small modifications.
It is compatible with the folding scheme defined in the folding library.
The decoding function `phi` from the paper will use the 5 columns decoding the instruction (or maybe only the one describing the encoded 32 bits instruction value). We must precompute the polynomial that evaluates to 1, 2, ..., l for the values encoding the instructions, and reduce it to degree 2. As we do have 435 columns at the least in the IVC for Poseidon, we can allow a non-negligeable number of different instructions. 
Each `F_i` from the paper will be the constraints related to the specific instruction.
Each accumulator U_i will be encapsulated in a `FoldingOutput` value.
A state for each accumulator/instruction will be kept.


dannywillems:
Closing as it will never be used.

PR #2432, opened 2024-07-16
Add support for runtime tables

volhovm:
Work in progress branch. 

I'm extending the existing `logup` module to support non-compile-time-fixed tables.

Addresses #2089

### DIFF
--- a/msm/src/circuit_design/witness.rs
+++ b/msm/src/circuit_design/witness.rs
@@ -33,10 +33,20 @@ pub struct WitnessBuilderEnv<
     /// `i` was looked up.
     pub lookup_multiplicities: BTreeMap<LT, Vec<F>>,
 
+    // @volhovm: It seems much more ineffective to store
+    // Vec<BTreeMap<LT, Vec<Logup>>> than BTreeMap<LT, Vec<Vec<Logup>>>.
     /// Lookup requests. Each vector element represents one row, and
     /// each row is a map from lookup type to a vector of concrete
     /// lookups requested.
     pub lookups: Vec<BTreeMap<LT, Vec<Logup<F, LT>>>>,
+
+    /// Values of the runtime tables. Each element (value) in the map
+    /// is a on-the-fly built column.
+    pub runtime_tables: BTreeMap<LT, Vec<Vec<F>>>,
+
+    /// A runtime table resolver; the inverse of `runtime_tables`:
+    /// maps runtime lookup table to its `usize` position in `runtime_tables`.
+    pub runtime_tables_resolver: BTreeMap<LT, BTreeMap<Vec<F>, usize>>,
 
     /// Fixed values for selector columns. `fixed_selectors[i][j]` is the
     /// value for row #j of the selector #i.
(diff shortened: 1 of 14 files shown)


msm/src/circuit_design/witness.rs
```diff
@@ -184,7 +193,37 @@ impl<
     > LookupCap<F, CIx, LT> for WitnessBuilderEnv<F, CIx, N_WIT, N_REL, N_DSEL, N_FSEL, LT>
 {
     fn lookup(&mut self, table_id: LT, value: Vec<<Self as ColAccessCap<F, CIx>>::Variable>) {
-        let value_ix = table_id.ix_by_value(&value);
+        let value_ix = if table_id.is_fixed() {
+            table_id
+                .ix_by_value(&value)
+                .expect("Could not resolve lookup for a fixed table")
+        } else {
+            // For runtime tables, we check if this value was already
```
mrmr1993, line 201: Is this necessarily the desired behaviour? I would have thought that the index would be 'fixed' for all our uses, given that we treat the tables effectively as arrays. That is to say, usually we'll be doing a lookup of the form `table_id + value_index * joint_combiner + value * joint_combiner^2`, and `value_index` determines exactly which value we're hoping to access.

msm/src/verifier.rs
```diff
@@ -151,6 +151,12 @@ where
                 .values()
                 .for_each(|comm| absorb_commitment(&mut fq_sponge, comm));
 
+            // FIXME @volhovm it seems that the verifier does not
```
dannywillems, line 154: It does not because fixed lookup tables should be treated as public columns, in particular when implementing vector lookup. Columns used by lookups should be considered as part of the "witness" (which includes private + public inputs), and therefore, the check would be performed automatically, if public inputs are checked.

volhovm: Hm... I guess that's quite similar to what I was trying to say. I do understand that a lot of logup columns are just "witness" (like the accumulator and so on), but the actual fixed table columns should be treated similarly to the public input. Or rather as fixed selectors tbh. "Public input" is something that changes almost every time you use the circuit, but fixed table content is mostly fixed across many public inputs. So TBD.

msm/src/circuit_design/witness.rs
volhovm: Hm. I'll think about it a bit later, but my current take is this:
1. Fixed tables have a fixed mapping `index -> value`, and we look up `table_id + value[0] * joint_combiner + value[1] * joint_combiner^2 + ...`
    * So `index` is not actually present in the lookup. It's only used in the code that constructs the lookup for counting multiplicities etc, since we request lookup by value, not by index. It's just an interface thing, we could do otherwise. But for now, if there are four value is the table, `555, 666, 777, 888`, we call it `lookup(column_777)` instead of `lookup_ix(column_777, 3)`. And so we record that lookup to 777 was done, and add `+1` to its multiplicity.
2. My current plan with runtime tables is to just mimic this behaviour. The only difference is that we don't know `index -> value` mapping in advance, so the code you're commenting on looks it up from a dynamically on-the-fly constructed map (resolver),  which is also maintained on-the-fly. Essentially we push elements to the current runtime table column, and when value from it is requested, we go to that value, and add `+1` to its multiplicity based on this index. It's very similar.

volhovm:
@dannywillems @mrmr1993 This is ready for a pass. This is obviously only part of the work, something self-contained to add runtime lookups, with limitations. More will be coming.

msm/src/circuit_design/witness.rs
mrmr1993: I trust that you've thought about this, but I find it unlikely that this is actually the behaviour we want. I won't block though.

msm/src/circuit_design/witness.rs
```diff
@@ -184,8 +194,40 @@ impl<
...
+            let multiplicities = self.lookup_multiplicities.get_mut(&table_id).unwrap();
+            // Since we allow multiple lookups per row, runtime tables
+            // can in theory grow bigger than the domain size. We
+            // still collect multiplicities as if runtime table vector
+            // is not height-bounded, but we will split it into chunks
+            // later.
+            if !table_id.is_fixed() && value_ix > multiplicities.len() {
+                multiplicities.resize(value_ix, F::zero());
+            }
+            multiplicities[value_ix] += F::one();
```
mrmr1993, line 229: It's unfortunate that we're using the field here. Again, not a blocker.

mrmr1993 (approved):
Some concerns, but otherwise LGTM

msm/src/circuit_design/witness.rs
volhovm: @mrmr1993 Do I understand correctly that your suggestion wrt runtime tables specifically is to lookup `index || value_vec` instead of just `value_vec`? If yes, I probably understand the idea, but not the trade-offs: we seem to still need to maintain the reverse map `value -> index` while constructing the witness, so I'm not exactly seeing what's the advantage of this approach... what am I missing?

msm/src/circuit_design/witness.rs
volhovm: Fixed: https://github.com/o1-labs/proof-systems/pull/2432/commits/36a554ff6d396fc659df0a88b30c01d24af2fb91

I quickly checked and it seems that it only makes sense to use `u64` for multiplicities for witness building; when the multiplicities are all known and circuit is built, we have to convert them into the field, so I did not change that part since the performance would be the same it seems. Correct me if I misunderstood you.

volhovm:
@dannywillems just remembered: if you have any suggestions regarding docs, feel free to suggest even post-merge!

PR #2433, opened 2024-07-16
Arrabiata - first big PR

dannywillems:
This library provides an implementation of a generic recursive zero-knowledge argument based on folding schemes (initially defined in [Nova](https://eprint.iacr.org/2021/370)), over the [pasta]() curves and using the IPA polynomial commitment.


dannywillems:
Rebasing on top of master. Commits history and time will be different :(. Before this rebase, commit was [56999c8](https://github.com/o1-labs/proof-systems/pull/2433/commits/56999c8374b588eb484caf9b4c07a6e74cb9916b).

dannywillems:
cc @marcbeunardeau88.
Future code will be on fork of this branch.
The branch is clean in terms of history, but not clean in terms of final code. I will clean step by step. We are good now on the path to use.

dannywillems:
The code has also no impact on the rest of the codebase. It is totally isolated.

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! ## Fiat-Shamir challenges
+//!
+//! The challenges sent by the verifier must also be simulated by the IVC
+//! circuit.
+//!
+//! For a step `i + 1`, the challenges of the step `i` must be computed by the
+//! verifier, and check that it corresponds to the ones received as a public
+//! input.
+//!
+//! TBD/FIXME: specify. Might require foreign field arithmetic.
```
dannywillems, line 283: Note that we will verify at step "i + 2". No foreign field arithmetic. I'll update later.

arrabiata/README.md
```diff
@@ -0,0 +1,93 @@
...
+### Implementation details
+
+We provide a binary to run arbitrarily large computation.
+The implementation of the circuits will follow the one used by the o1vm
+interpreter. An interpreter will be implemented over an abstract environment.
+
+The environment used for the witness will contain all the information required
+to make an IVC proof, i.e. the current witness in addition to the oracle, the
+folding accumulator, etc. And the implementation attempts to have the smallest
+memory footprint it can have by optimising the use of the CPU cache and using
```
marcbeunardeau88, line 40: Cpu cache does not reduce memory footprint

arrabiata/README.md
```diff
@@ -0,0 +1,93 @@
...
+references as much as it can. The implementation attempts to prioritize
+allocations on the stack when possible, and attempts to keep in the CPU cache as
+many data as required as long as possible to avoid indirection.
+
+The witness interpreter attempts to perform most of the algebraic operations
+using integers, and not field element. The results are reduced into the field
```
marcbeunardeau88, line 46: I dont get how that's possible since you don't have any bounds on the witness elements at this point

arrabiata/README.md
```diff
@@ -0,0 +1,93 @@
...
+when mandatory.
+
+While building the witness, the cross terms are also computed on the fly, to be
+used in the next iteration. This way, the prover only pays the price of the
```
marcbeunardeau88, line 50: I don't get how the on the fly computation makes the prover only pay for the activated gate

arrabiata/src/columns.rs
```diff
@@ -0,0 +1,82 @@
+use kimchi::circuits::expr::{CacheId, ConstantExpr, Expr, FormattedOutput};
+use std::collections::HashMap;
+use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
+
+/// This enum represents the different gadgets that can be used in the circuit.
+/// The selectors are defined at setup time, can take only the values `0` or
```
marcbeunardeau88, line 6: Don't we want selectors to be other values than 0/1, eg. poseidon related constants

arrabiata/src/constraints.rs
```diff
@@ -0,0 +1,359 @@
...
+use log::debug;
+use num_bigint::BigInt;
+use o1_utils::FieldHelpers;
+
+pub struct Env<Fp: Field> {
+    pub poseidon_mds: Vec<Vec<Fp>>,
+    /// The parameter a is the coefficients of the elliptic curve in affine
+    /// coordinates.
+    // FIXME: this is ugly. Let use the curve as a parameter. Only lazy for now.
+    pub a: BigInt,
```
marcbeunardeau88, line 21: I don't see that using a big integer is better than a field element

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! At the moment, all constraints must be of maximum degree
+//! [crate::MAX_DEGREE], but it might change in the future.
+//!
+//! The implementation relies on a representation of the circuit as a 2D array
+//! of "data points" the interpreter can use.
+//!
+//! An interpreter defines what a "position" is in the circuit and allow to
+//! perform operations using these positions.
+//! Some of these positions will be considered as public inputs and might be
+//! fixed at setup time while making a proof, when other will be considered as
```
marcbeunardeau88, line 13: public inputs are not fixed at setup time
Note that in some use cases some PI are also computed by the prover
eg. crafting a proof that a transition from an intial state to a resulting state requires the prover to compute the resulting state, although it is marked as a PI

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//!
+//! For given inputs (x1, y1) and (x2, y2), the layout will be as follow:
+//!
+//! ```text
+//! | C1 | C2 | C3 | C4 | C5 | C6 | C7 | C8 | C9 | C10 | C11 | C12 | C13 | C14 | C15 | C16 | C17 |
+//! | -- | -- | -- | -- | -- | -- | -- | -- | -- | --- | --- | --- | --- | --- | --- | --- | --- |
+//! | x1 | y1 | x2 | y2 | b0 | λ  | x3 | y3 |    |     |     |     |     |     |     |     |     |
+//! ```
+//!
+//! where `b0` is equal two `1` if the points are the same, and `0` otherwise.
```
marcbeunardeau88, line 76: I would separate the double and the add, it will give a much better scalar mul
And most of time we have an overwhelming chance of not doubling 'by chance', eg. when adding two commitments to different polynomial.

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//!
+//! The permutation argument described in the PlonK paper is a kind of "inverse
+//! lookup" protocol, like Plookup. The polynomials are defined as follows:
+//!
+//! ```text
+//!            Can be seen as T[f(X)] = Χ
+//!          --------
+//!          |      |
+//! f'(X) = f(X) + β X + γ
+//!                      |--- Can be seen as the evaluation point.
```
marcbeunardeau88, line 226: gamma cannot be seen as the evaluation point

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! Communication between rows must be done using a permutation argument. The
+//! argument we use will be a generalisation of the one used in the [PlonK
+//! paper](https://eprint.iacr.org/2019/953).
+//!
+//! The construction of the permutations will be done using the methods prefixed
+//! `save` and `load`. The index of the current row and the index of the
+//! time the value has been written will be used to generate the permutation on
+//! the fly.
+//!
+//! The permutation argument described in the PlonK paper is a kind of "inverse
```
marcbeunardeau88, line 218: it's not an 'inverse lookup'
Also I would not try to give an intuition of plonk's perm argument by mentionning plookup, which itself uses the techniques of the perm argument with more complex equations

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! lookup" protocol, like Plookup. The polynomials are defined as follows:
+//!
+//! ```text
+//!            Can be seen as T[f(X)] = Χ
```
marcbeunardeau88, line 222: I don't understand this line

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! The verifier at the next step will verify that the challenges have been
+//! correctly computed.
+//! In the implementation, the accumulator will be computed after the challenges
+//! and the commitments. Note that the accumulator must also be aggregated, and
+//! the aggregation must be performed by the verifier at the next step.
+//!
+//! The methods `save` and `load` will accept as arguments only a column that is
+//! included in the permutation argument. For instance, `save_poseidon_state`
+//! will only accept columns with index 3, 4 and 5, where the
+//! `load_poseidon_state` will only accepts columns with index 0, 1 and 2.
```
marcbeunardeau88, line 247: why are the save and load in different columns ?
if a perm argument  is applied to a set of columns, one can load and save from/to any columns of the set

arrabiata/src/columns.rs
dannywillems: Selectors and public values are handled differently.

arrabiata/src/constraints.rs
```diff
@@ -0,0 +1,359 @@
...
+use kimchi::circuits::{
+    expr::{ConstantTerm::Literal, Expr, ExprInner, Operations, Variable},
+    gate::CurrOrNext,
+};
+use log::debug;
+use num_bigint::BigInt;
+use o1_utils::FieldHelpers;
+
+pub struct Env<Fp: Field> {
+    pub poseidon_mds: Vec<Vec<Fp>>,
```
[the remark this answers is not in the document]
dannywillems: Yes, I want to remove it. I think there is a comment somewhere saying "it is ugly but ok for now".

arrabiata/src/interpreter.rs
dannywillems: The sentence is not phrased correctly. The "and might be fixed at setup time" are not the public inputs, but for "positions".

arrabiata/src/interpreter.rs
dannywillems: We can discuss offline. I can remove this (it is mostly a note from the draft PR that shouldn't come into this top-level doc).

arrabiata/src/interpreter.rs
dannywillems: We can discuss offline. I can remove this (it is mostly a note from the draft PR that shouldn't come into this top-level doc).

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//!
+//! The permutations values will be saved in public values, and will contain the
```
marcbeunardeau88, line 249: Not sure what you mean by permutation values

arrabiata/src/constraints.rs
dannywillems: We should bench later.

arrabiata/README.md
dannywillems: Because not all columns are used by all gates.

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! index of the row. The permutation values will be encoded with a 32 bits
+//! value (u32) as we can suppose a negligible probability that a user will use
+//! more than 2^32 rows.
+//!
+//! The permutation argument also generates constraints that will be
+//! homogeneized with the gadget constraints.
+//!
+//! Note all rows might require to use the permutation argument. Therefore, a
```
marcbeunardeau88, line 257: I understand that you want the save some computation if a written circuit does not copy constrains some columns ?
You do not need selectors to do that

arrabiata/README.md
dannywillems: It is not phrased correctly. "decreasing memory access and footprint" would be better.

arrabiata/src/interpreter.rs
dannywillems: We can discuss offline. I can remove this (it is mostly a note from the draft PR that shouldn't come into this top-level doc).

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! the aggregation must be performed by the verifier at the next step.
+//!
+//! The methods `save` and `load` will accept as arguments only a column that is
+//! included in the permutation argument. For instance, `save_poseidon_state`
+//! will only accept columns with index 3, 4 and 5, where the
+//! `load_poseidon_state` will only accepts columns with index 0, 1 and 2.
+//!
+//! The permutations values will be saved in public values, and will contain the
+//! index of the row. The permutation values will be encoded with a 32 bits
+//! value (u32) as we can suppose a negligible probability that a user will use
```
marcbeunardeau88, line 251: it's not really a probability, it's not like the user flips random coins to decide on the number of row :)

dannywillems: It wasn't "negligible probability" as in cryptography :smile:. True, I should rephrase it.

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//!
+//! TBD:
+//! - number of columns
+//! - accumulator column
+//! - folding of the permutation argument
+//!
+//! TBD/FIXME: do we use a additive permutation argument to increase the number
+//! of columns we can perform the permutation on?
+//!
+//! TBD/FIXME: We can have more than one permutation argument. For instance, we
```
marcbeunardeau88, line 270: doing this prevents communication from column 1 to column 5
to reduce the degree we can use chunking : P = P_lo + X^k P_hi

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+//! forming the constraint.
+//! Next to this, additional information, like the cross-terms and the error
+//! terms must be computed.
+//!
+//! This computation depends on the constraints, and in particular on the
+//! monomials describing the constraints.
+//! The computation of the cross-terms and the error terms happen after the
+//! witness has been built and the different arguments like the permutation or
+//! lookup have been done. Therefore, the interpreter must provide a method to
+//! compute it, and the constraints should be passed as an argument.
```
marcbeunardeau88, line 301: I would give the symbolic cross term (computed at setup time) as an argument, rather than the contraints

arrabiata/README.md
dannywillems: I think it is a leftover of the initial design, and I realised after that I had to use BigInt in any case.

arrabiata/src/interpreter.rs
```diff
@@ -0,0 +1,962 @@
...
+/// next iteration.
+/// A row must be created to generate a challenge to combine the constraints
+/// later. The challenge will be also accumulated over time.
+///
+/// FIXME: homogeneize
+/// FIXME: compute error terms
+/// FIXME: the resulting constraints do not include the selectors. We decided to
+/// use this design to only keep in the constraints the columns that are not
+/// fixed by the relation.
+pub fn run_ivc<E: InterpreterEnv>(env: &mut E, instr: Instruction) {
```
marcbeunardeau88, line 637: I am confused here, I thought the instructions were building blocks to craft the IVC circuit, but the signature of this function make me understand that you create an IVC of one of those instruction

arrabiata/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+pub mod logup;
+pub mod poseidon_3_60_0_5_5_fp;
+pub mod poseidon_3_60_0_5_5_fq;
+pub mod proof;
+pub mod prover;
+pub mod verifier;
+pub mod witness;
+
+/// The maximum degree of the polynomial that can be represented by the
+/// polynomial-time function the library supports.
```
marcbeunardeau88, line 16: I think we need a better name than polynomial time function
There is no context reagarding assymptotic complexity here
I believe the usual name for IVC is application

arrabiata/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+pub const MAX_DEGREE: u64 = 5;
+
+/// The minimum SRS size required to use Nova, in base 2.
+/// Requiring at least 2^16 to perform 16bits range checks.
```
marcbeunardeau88, line 20: Is this assuming we use plookup ?

arrabiata/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+pub const MIN_SRS_LOG2_SIZE: usize = 16;
+
+/// The number of rows the IVC circuit requires.
+// FIXME: that might change. We use a vertical layout for now.
+pub const IVC_CIRCUIT_SIZE: usize = 1 << 13;
+
+/// The maximum number of columns that can be used in the circuit.
+pub const NUMBER_OF_COLUMNS: usize = 17;
```
marcbeunardeau88, line 28: it is also the minimum, since you plan to use 17 columns for the recursive part of the computation, right ?

arrabiata/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// The maximum number of public inputs the circuit can use per row
```
marcbeunardeau88, line 30: Is this the public input of the application, or of the IVC part ?
Why do you say 'per row' ?
I don't understand the justification of the computation

arrabiata/src/lib.rs
dannywillems: No. The recursive part takes less than 17 columns, now. We can decrease it later. I think it is 15 at the moment (Poseidon takes 15).

arrabiata/src/lib.rs
dannywillems: Plookup or logup. But no lookup argument is implemented for now. There was an initial plan to use one, but I would revisit this later, after the PoC.

arrabiata/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+/// In addition to the 12 public inputs required for the rounds, we add 2 more
+/// for the values to absorb.
+// FIXME: we can do 5 rounds per row by using "the next row"
+pub const NUMBER_OF_PUBLIC_INPUTS: usize = 12 + 2;
+
+/// The low-exponentiation value used by the Poseidon hash function for the
+/// substitution box.
+///
+/// The value is used to perform initialisation checks with the fields.
+pub const POSEIDON_ALPHA: u64 = 5;
```
marcbeunardeau88, line 41: all the poseidon related constant should not be there
Maybe they can be given by the curves, since they come from there

arrabiata/src/interpreter.rs
dannywillems: We can change the interface later. `run_ivc` should execute each instruction, in order, to build the IVC circuit.

arrabiata/src/lib.rs
dannywillems: They do not come necessarily from the curves. They come from the design. 7 could also be used for instance.

arrabiata/src/logup.rs
```diff
@@ -0,0 +1 @@
+//! This file will implement a logup argument to allow users performing lookup in their circuits.
```
marcbeunardeau88, line 1: I would name that lookup

arrabiata/src/lib.rs
dannywillems: It is for both.
15 is required for the Poseidon gadget (3 round constants, 5 permutation per row). 2 is for the sponge (absorb).

arrabiata/src/poseidon_3_60_0_5_5_fp.rs
```diff
@@ -0,0 +1,903 @@
+use mina_curves::pasta::Fp;
```
marcbeunardeau88, line 1: dont we have this poseidon already ?
Also at some point we should move all the poseidon together

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// The index of the latest allocated public inputs in the circuit.
+    /// It is used to allocate new public inputs without having to keep track of
+    /// the position.
+    pub idx_var_pi: usize,
+
+    /// Current processing row. Used to build the witness.
+    pub current_row: usize,
+
+    /// State of the current row in the execution trace
+    pub state: [BigInt; NUMBER_OF_COLUMNS],
```
marcbeunardeau88, line 83: We will need two vectors like this when we will have access to the next row ?

arrabiata/src/poseidon_3_60_0_5_5_fp.rs
dannywillems: No, we do have a poseidon with alpha = 7 and 55 rounds.
Yes, we should move all poseidon instances together one day.

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// also be used to verify the consistency of the computation by hashing the
+    /// public IO.
+    // IMPROVEME: use a list of BigInt? It might be faster as the CPU will
+    // already have in its cache the values, and we can use a flat array
+    pub sponge_e1: [BigInt; POSEIDON_STATE_SIZE],
+    pub sponge_e2: [BigInt; POSEIDON_STATE_SIZE],
+
+    /// List of public inputs, used first to verify the consistency of the
+    /// previous iteration.
+    pub current_iteration: u64,
```
marcbeunardeau88, line 120: that is not a list

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+
+    /// Selectors to activate the gadgets.
+    /// The size of the outer vector must be equal to the number of gadgets in
+    /// the circuit.
+    /// The size of the inner vector must be equal to the number of rows in
+    /// the circuit.
+    ///
+    /// The layout columns/rows is used to avoid rebuilding the arrays per
+    /// column when committing to the witness.
+    pub selectors: Vec<Vec<bool>>,
```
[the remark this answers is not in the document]
dannywillems: I think it is described somewhere, but I think about having a different "interpreter" for the setup, and, indeed, the selectors won't be in the witness env anymore.

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    // already have in its cache the values, and we can use a flat array
+    pub sponge_e1: [BigInt; POSEIDON_STATE_SIZE],
+    pub sponge_e2: [BigInt; POSEIDON_STATE_SIZE],
+
+    /// List of public inputs, used first to verify the consistency of the
+    /// previous iteration.
+    pub current_iteration: u64,
+
+    /// A previous hash, encoded in 2 chunks of 128 bits.
+    pub previous_hash: [u128; 2],
```
marcbeunardeau88, line 123: why not a field element

dannywillems: Because it has to be instantiated for both fields. We might want to parametrize later. We should bench if it is faster or not than bigint.

arrabiata/src/witness.rs
dannywillems: Right. I think it is an old comment.

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    ///
+    /// The registers can, and must, be cleaned after the gadget is computed.
+    ///
+    /// The values are considered as BigInt, even though we should add some
+    /// type. As we want to apply the KISS method, we tend to avoid adding
+    /// types. We leave this for future work.
+    ///
+    /// Two registers are provided, represented by a tuple for the coordinates
+    /// (x, y).
+    pub temporary_accumulators: ((BigInt, BigInt), (BigInt, BigInt)),
```
marcbeunardeau88, line 143: I would have something more abstract here, the detail of the EC computation should not appear here
Also we should have a name suggesting it is related to EC

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    // ----------------
+    /// The witness of the current instance of the circuit.
+    /// The size of the outer vector must be equal to the number of columns in the
+    /// circuit.
+    /// The size of the inner vector must be equal to the number of rows in
+    /// the circuit.
+    ///
+    /// The layout columns/rows is used to avoid rebuilding the witness per
+    /// column when committing to the witness.
+    pub witness: Vec<Vec<BigInt>>,
```
marcbeunardeau88, line 156: Should't we use the columns type here ?

arrabiata/src/witness.rs
dannywillems: Yes, I do agree. It will be abstracted by the perm argument later. The temporary accumulators are removed step by step in following PR.

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+
+    // --------------
+    // Inputs
+    /// Initial input
+    pub z0: BigInt,
```
marcbeunardeau88, line 161: we will need to generalize this at some point, the application can have more than one arg

dannywillems: Agreed

arrabiata/src/witness.rs
dannywillems: Why?

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// The values are considered as BigInt, even though we should add some
+    /// type. As we want to apply the KISS method, we tend to avoid adding
+    /// types. We leave this for future work.
+    ///
+    /// Two registers are provided, represented by a tuple for the coordinates
+    /// (x, y).
+    pub temporary_accumulators: ((BigInt, BigInt), (BigInt, BigInt)),
+
+    /// Index of the values to absorb in the sponge
+    pub idx_values_to_absorb: usize,
```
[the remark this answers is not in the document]
dannywillems: Agreed. It is temporary.

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    // Data only used by the interpreter while building the witness over time
+    /// The index of the latest allocated variable in the circuit.
+    /// It is used to allocate new variables without having to keep track of the
+    /// position.
+    pub idx_var: usize,
+
+    /// The index of the latest allocated public inputs in the circuit.
+    /// It is used to allocate new public inputs without having to keep track of
+    /// the position.
+    pub idx_var_pi: usize,
```
marcbeunardeau88, line 77: we do not expect to allocate public input in the middle of constructing the circuit right ?

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// Initial input
+    pub z0: BigInt,
+
+    /// Current input
+    pub zi: BigInt,
+    // ---------------
+
+    // ---------------
+    // Only used to have type safety and think about the design at the
+    // type-level
```
marcbeunardeau88, line 169: you don't need i, the number of the iteration ?

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+        } else {
+            Fq::modulus_biguint().into()
+        };
+        let v = v.mod_floor(&modulus);
+        self.public_state[idx] = v.clone();
+        v
+    }
+
+    /// Activate the gadget for the current row
+    fn activate_gadget(&mut self, gadget: Gadget) {
```
marcbeunardeau88, line 242: this should not be in the witness computation right ?

arrabiata/src/witness.rs
dannywillems: Yes. As it was based on Nova initially, `i` appears. We might want to change it.

arrabiata/src/witness.rs
dannywillems: See my comment above about "setup environment".

arrabiata/src/witness.rs
dannywillems: See my comment above about "setup environment".

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+            let modulus: BigInt = Fp::modulus_biguint().into();
+            self.sponge_e1[i] = x.mod_floor(&modulus)
+        } else {
+            let modulus: BigInt = Fq::modulus_biguint().into();
+            self.sponge_e2[i] = x.mod_floor(&modulus)
+        }
+    }
+
+    // The following values are expected to be absorbed in order:
+    // - vk
```
marcbeunardeau88, line 430: Not sure we need to abosrb the vk, let's talk about this offline

arrabiata/src/witness.rs
```diff
@@ -0,0 +1,1098 @@
...
+    ///
+    /// For a step i + 1, the IVC circuit receives as public input the following
+    /// values:
+    ///
+    /// - The commitments to the previous witnesses.
+    /// - The previous challenges (α_{i}, β_{i}, γ_{i}) - the challenges β and γ
+    /// are used by the permutation argument where α is used by the quotient
+    /// polynomial, generated after also absorbing the accumulator of the
+    /// permutation argument.
+    /// - The previous accumulators (acc_1, ..., acc_17).
```
marcbeunardeau88, line 988: Not sure what those acc are

arrabiata/src/witness.rs
dannywillems: You're right. We do not need. Leftover of Nova reading.

arrabiata/src/witness.rs
dannywillems: Commitments to the columns.

marcbeunardeau88 (approved):

arrabiata/src/witness.rs
dannywillems: Addressed in 9452aa6be30ea23751930b575d69743ba6fb5f1a

arrabiata/src/witness.rs
dannywillems: Agreed on the idea in a call.

arrabiata/src/witness.rs
dannywillems: Discussed in a call, see below.

arrabiata/src/witness.rs
dannywillems: Should we rename it in trace?

arrabiata/src/witness.rs
dannywillems: Yes, it is the case, see new PR.

arrabiata/README.md
dannywillems: My comment isn't correct. The witness is the evaluation, but the cross-terms require the coefficients of the polynomial.

arrabiata/src/interpreter.rs
dannywillems: Discussed in a call: we talk about rows, and not columns.

arrabiata/src/columns.rs
dannywillems: Note: should we consider Poseidon constants as "selectors"?

PR #2434, opened 2024-07-18
SRS: digest into the given sponge

dannywillems:
Implemented for the IPA PCS

### DIFF
--- a/poly-commitment/src/srs.rs
+++ b/poly-commitment/src/srs.rs
@@ -7,6 +7,7 @@ use ark_poly::{EvaluationDomain, Radix2EvaluationDomain as D};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use blake2::{Blake2b512, Digest};
 use groupmap::GroupMap;
+use mina_poseidon::FqSponge;
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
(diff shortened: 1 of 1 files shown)


poly-commitment/src/srs.rs
```diff
@@ -265,6 +266,16 @@ impl<G: CommitmentCurve> SRS<G> {
...
         }
     }
+
+    /// Absorb all the information of the verification key into the given sponge.
+    ///
+    /// It is useful in SNARK protocols where the verification key must be
+    /// absorbed at the beginning of the protocol.
+    /// It does absorb first the basis G and after that the blinder H.
+    pub fn digest<Sponge: FqSponge<G::BaseField, G, G::ScalarField>>(&self, sponge: &mut Sponge) {
+        self.g.iter().for_each(|g| sponge.absorb_g(&[*g]));
```
volhovm, line 276: Remind me, how are `g` generated? They're not independent, right? If they're not (e.g. they're powers of each other, or progressive hashes), maybe we can hash just first/last value...

volhovm (approved):

poly-commitment/src/srs.rs
dannywillems: They can be arbitrary values for the IPA.

mrmr1993 (changes requested):
Please no

PR #2435, opened 2024-07-24
utils/field_helpers: introduce from_biguint_err alias

dannywillems:
Simple alias to avoid the `unwrap()`. I use a lot of from_biguint in arrabiata, and it is only to make it shorter. I would prefer going with `exn` for the suffix. I leave the reviewer decides.


dannywillems:
Can be ignored. I moved to BigInt.

Issue #2427, opened 2024-07-15
Runtime logup tables support


volhovm:
In the MSM project we only support fixed tables at the moment. Supporting runtime tables requires changing some table interfaces.

PR #2436, opened 2024-07-24
Adding runtime tables with arbitrary writes

volhovm:
This PR adds support for runtime tables which write explicitly into witness columns, possible multiple times.

In practice, this means we can do multi-row communication, which is shown in the `serialization/interpreter.rs` circuit.

A lot of restructuring along the way, including some tests for existing circuits, support for multiple multiplicities vectors, etc.

Addresses #2427


volhovm:
@fabrizio-m feel free to highlight places that require more docs.

fabrizio-m (approved):

PR #2437, opened 2024-07-24
Bifolding IVC: Add Poseidon gadget for hashing commitments

fabrizio-m:


ivc/src/lib.rs
```diff
@@ -112,6 +112,8 @@
 //! executed, `0` otherwise.
 
 pub mod ivc;
+/// poseidon implementation that hashes commitments to the columns containing values
+pub mod poseidon;
```
volhovm, line 116: This needs a more distinctive name. `poseidon_hashed_com`?..

msm/src/circuit_design/capabilities.rs
```diff
@@ -36,6 +36,10 @@ pub trait ColAccessCap<F: PrimeField, CIx: ColumnIndexer> {
     /// Turns a constant value into a variable.
     fn constant(value: F) -> Self::Variable;
 }
+/// Environment capability for accessing the next row, constraint only.
+pub trait NextCap<F: PrimeField, CIx: ColumnIndexer>: ColAccessCap<F, CIx> {
+    fn read_next(&self, col: CIx) -> Self::Variable;
```
volhovm, line 41: I'd strongly suggest to put this into `ColAccessCap` as `read_column_next`: it's literally the same semantics (reading a column). Any reasons why you didn't go for this design?

msm/src/circuit_design/capabilities.rs
```diff
@@ -36,6 +36,10 @@ pub trait ColAccessCap<F: PrimeField, CIx: ColumnIndexer> {
     /// Turns a constant value into a variable.
     fn constant(value: F) -> Self::Variable;
 }
+/// Environment capability for accessing the next row, constraint only.
+pub trait NextCap<F: PrimeField, CIx: ColumnIndexer>: ColAccessCap<F, CIx> {
```
volhovm, line 40: Another question on this framework: don't you also need `write_column_next`?

msm/src/circuit_design/constraints.rs
```diff
@@ -56,6 +55,16 @@ impl<F: PrimeField, CIx: ColumnIndexer, LT: LookupTableID> ColAccessCap<F, CIx>
         Expr::Atom(ExprInner::Constant(cst_expr_inner))
     }
 }
+impl<F: PrimeField, CIx: ColumnIndexer, LT: LookupTableID> NextCap<F, CIx>
```
volhovm, line 58: You have to implement lenses for your new capability too (missing in this PR), otherwise it won't compose well in the future.

ivc/src/poseidon/interpreter.rs
```diff
@@ -0,0 +1,156 @@
...
+    let absorb_selector = env.read_column(PoseidonColumn::Mode(Selector::Absorb));
+    let [a0, a1] = absorb;
+    let s0 = s0 + a0 * absorb_selector.clone();
+    let s1 = s1 + a1 * absorb_selector;
+
+    let state = [s0, s1, s2];
+    state
+}
+
+///constrains a round
```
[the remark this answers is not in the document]
volhovm: (this comment applies to a lot of places in this file, and to other PRs generally too)

ivc/src/poseidon/interpreter.rs
```diff
@@ -0,0 +1,156 @@
...
+    let s1 = env.read_column(PoseidonColumn::State(1));
+    let s2 = env.read_column(PoseidonColumn::State(2));
+    let state = [s0, s1, s2];
+    let a0 = env.read_column(PoseidonColumn::Absorb(0));
+    let a1 = env.read_column(PoseidonColumn::Absorb(1));
+    let absorb = [a0, a1];
+    let check_out = env.read_column(PoseidonColumn::Check);
+    let res = round(env, state, absorb, check_out);
+    for i in 0..=2 {
+        let next = env.read_next(PoseidonColumn::State(i));
```
volhovm, line 127: I don't understand the semantics of this. The idea behind `constrain_X` is that it can be used as a real assertion in the witness-building code. 
* When you execute `constraint_round` in the `ConstraintBuilder` mode, you'll get the constraint, alright.
* But when you execute it in the `WitnessBuilder` mode, this will fail, because ... current `WitnessBuilder` does not allow you to write into the future cells. 

I suggest this mechanic is added within this PR. You'll just need to:
* Store `cur_row` explicitly in the `WitnessBuilder`, so that `cur_row = witness.len() - 2` (e.g. witness is two rows, but current row is 0, and row 1 is "next").
* Extend capabilities of read/write column to read/write next column. 
* Add `write_column_next` to `compute_round`.

PR #2439, opened 2024-07-25
Generate URS using parallelisation

mrmr1993:


poly-commitment/src/srs.rs
```diff
@@ -245,6 +245,7 @@ where
         let m = G::Map::setup();
 
         let g: Vec<_> = (0..depth)
+            .into_par_iter()
```
dannywillems, line 248: There is the same below, `create_parallel`. Do we want to keep only one?

Fizzixnerd:
@mrmr1993 is this still open?

dannywillems:
I am closing it as it is already implemented, and it seems to be stalled. Feel free to reopen it, if needed.

PR #2441, opened 2024-07-29
IVC: Adding SNARK on the final fold

volhovm:
Previously, `simple.rs` was SNARKing the pre-last folded instance; in this PR we SNARK the last one.

### DIFF
--- a/ivc/src/prover.rs
+++ b/ivc/src/prover.rs
@@ -261,7 +261,7 @@ where
     let (_, endo_r) = G::endos();
 
     let quotient_poly = {
-        let evaluation_domain = domain.d2;
+        let evaluation_domain = domain.d4;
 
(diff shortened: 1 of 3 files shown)


ivc/src/prover.rs
```diff
@@ -329,8 +329,8 @@ where
         }
     };
 
-    // we assume our folding degree is always 2, so number of chunks should be 1
-    let num_chunks: usize = 1;
+    // we interpolate over d4, so number of chunks should be 3
```
fabrizio-m, line 332: what is `num_chunk` exactly? if this is about splitting the final constraint polynomials into sections that fit in d1, shouldn't it be 2/4? rather than 1/3.
Or is 1 chunk assumed to be zero?

fabrizio-m (approved):

PR #2446, opened 2024-07-29
Poly-commitment: additional top-level documentation

dannywillems:
We should have more documentation in this file, and we should maybe rename it.


poly-commitment/src/evaluation_proof.rs
```diff
@@ -1,3 +1,8 @@
+//! This module contains the implementation of the polynomial commitment scheme
+//! called the Inner Product Argument (IPA) as described in [Efficient
```
volhovm, line 2: I never understand why this module is called `evaluation_proof`... I guess it's something archaic before KZG was written?.. :thinking:

volhovm (approved):

poly-commitment/src/evaluation_proof.rs
dannywillems: Yes. We must rename it! It would be very nice as a first task!

PR #2447, opened 2024-07-29
Poly-commitment/Tests: move IPA related tests into ipa_commitment

dannywillems:
Moving into tests subdirectory helps checking the API and give a context for users to use the library.


volhovm (approved):
I assume all these big test PRs are just moving code (with no actual code changes)

PR #2462, opened 2024-07-29
Mina_curves/tests: move into subdirectory

dannywillems:


curves/src/pasta/fields/mod.rs
```diff
@@ -3,6 +3,3 @@ pub use self::fp::*;
 
 pub mod fq;
 pub use self::fq::*;
-
-#[cfg(test)]
-mod tests;
```
volhovm, line 8: How did the previous PR compile? :thinking:

volhovm (approved):

PR #2473, opened 2024-08-15
Bump toolchain version and dependencies upgrade.

shimkiv:


dannywillems:
Closing as we're not actively on this, and it seems to be a single bump up of rust-toolchain. I'm still in favor of updating the toolchain!

PR #2475, opened 2024-08-18
MVPoly: multivariate polynomial library, handling homogeneous forms

dannywillems:


sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+
+use ark_ff::PrimeField;
+use num_integer::binomial;
+use o1_utils::FieldHelpers;
+
+use crate::utils::{
+    compute_all_two_factors_decomposition, naive_prime_factors, PrimeNumberGenerator,
+};
+
+/// Represents a multivariate polynomial of degree `D` in `N` variables.
```
marcbeunardeau88, line 18: of degree less than

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// - 2 -> 3
+    /// - 3 -> 4
+    /// - 4 -> 6
+    /// - 5 -> 9
+    /// ```
+    pub fn compute_normalized_indices() -> Vec<usize> {
+        let length = Self::dimension();
+        let mut normalized_indices = vec![1; length];
+        let mut prime_gen = PrimeNumberGenerator::new();
+        let primes = prime_gen.get_first_nth_primes(N);
```
marcbeunardeau88, line 69: why do you need this function, since you store the prime numbers in constant.rs

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// - 0 -> 1
+    /// - 1 -> 2
+    /// - 2 -> 3
+    /// - 3 -> 4
+    /// - 4 -> 6
+    /// - 5 -> 9
+    /// ```
+    pub fn compute_normalized_indices() -> Vec<usize> {
+        let length = Self::dimension();
+        let mut normalized_indices = vec![1; length];
```
marcbeunardeau88, line 67:
```suggestion
        let mut normalized_indices = vec![1; Self::dimension()];
```

sympolyc/src/expr.rs
marcbeunardeau88, line 18:
```suggestion
/// Represents a multivariate polynomial of degree less than `D` in `N` variables.
```

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+        let mut prime_gen = PrimeNumberGenerator::new();
+        let primes = prime_gen.get_first_nth_primes(N);
+        let max_index = primes[N - 1].checked_pow(D as u32);
+        assert!(
+            max_index.is_some(),
+            "Overflow in computing the maximum index"
+        );
+        let max_index = max_index.unwrap();
```
marcbeunardeau88, line 75: can't you merge the unwrap and assert some ?

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+    compute_all_two_factors_decomposition, naive_prime_factors, PrimeNumberGenerator,
+};
+
+/// Represents a multivariate polynomial of degree `D` in `N` variables.
+/// The natural representation is the coefficients given in the monomial basis.
+pub struct MVPoly<F: PrimeField, const N: usize, const D: usize> {
+    coeff: Vec<F>,
+    // keeping track of the indices of the monomials that are normalized
+    // to avoid recomputing them
+    normalized_indices: Vec<usize>,
```
marcbeunardeau88, line 24: Is this guy supposed to be with the polynomial ?
He only depends on N and D, we shouldn't carry him with every polynomial, nor compute him every time we create a polynomial

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn is_homogeneous(&self) -> bool {
+        let normalized_indices = self.normalized_indices.clone();
+        let mut prime_gen = PrimeNumberGenerator::new();
+        let is_homogeneous = normalized_indices
+            .iter()
+            .zip(self.coeff.clone())
+            .all(|(idx, c)| {
+                let decomposition_of_i = naive_prime_factors(*idx, &mut prime_gen);
+                let monomial_degree = decomposition_of_i.iter().fold(0, |acc, (_, d)| acc + d);
+                !(monomial_degree != D && c != F::zero())
```
marcbeunardeau88, line 104:
```suggestion
                (monomial_degree = D || c = F::zero())
```

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+                    .position(|&x| x == *b)
+                    .unwrap();
+                let a_coeff = self.coeff[inv_a];
+                let b_coeff = other.coeff[inv_b];
+                let product = a_coeff * b_coeff;
+                sum += product;
+            });
+            result.push(sum);
+        });
+        Self::from_coeffs(result)
```
marcbeunardeau88, line 197: Here the normalized indices will be computed in function of the degree bound D, which is not invariant by multiplication.
Can't that be an issue ?

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    fn mul(self, other: Self) -> Self {
+        let mut acc = HashMap::new();
+        let mut prime_gen = PrimeNumberGenerator::new();
+        let mut result = vec![];
+        (0..self.coeff.len()).for_each(|i| {
+            let mut sum = F::zero();
+            let normalized_index = self.normalized_indices[i];
+            let two_factors_decomposition =
+                compute_all_two_factors_decomposition(normalized_index, &mut acc, &mut prime_gen);
```
marcbeunardeau88, line 177: This is a weird way to iterate, or is it just me ?
You are finding all the possible product of monomials that can contribute to a coeff of the result.
I would do the schoolbook way, which iterates on both polynomials, and compute which contribution goes where.

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+        result
+    }
+}
+
+// Multiplication
+impl<F: PrimeField, const N: usize, const D: usize> Mul<MVPoly<F, N, D>> for MVPoly<F, N, D> {
+    type Output = Self;
+
+    fn mul(self, other: Self) -> Self {
+        let mut acc = HashMap::new();
```
marcbeunardeau88, line 170: is this acc ever usefull ?

sympolyc/src/utils.rs
marcbeunardeau88, line 1: aren't a bunch of function of this file already implemented ?

sympolyc/src/expr.rs
```diff
@@ -0,0 +1,241 @@
...
+        let mut prime_gen = PrimeNumberGenerator::new();
+        self.coeff.iter().enumerate().for_each(|(i, c)| {
+            let normalized_idx = self.normalized_indices[i];
+            if normalized_idx == 1 {
+                write!(f, "{}", c.to_biguint()).unwrap();
+            } else {
+                let prime_decomposition = naive_prime_factors(normalized_idx, &mut prime_gen);
+                write!(f, "{}", c.to_biguint()).unwrap();
+                prime_decomposition.iter().for_each(|(p, d)| {
+                    // FIXME: not correct
```
Fizzixnerd, line 220: Should this be committed?

sympolyc/src/expr.rs
fabrizio-m: you can also use Option::expect()

sympolyc/src/expr.rs
dannywillems: The prime numbers in constant.rs are only used for testing for now. I will move it in the testing library.

sympolyc/src/expr.rs
dannywillems: It is a debug method. I will fix it in a follow-up PR, if it is fine for you.

sympolyc/src/expr.rs
dannywillems: Yes, it is a cache to avoid recomputing decomposition. Example: 20 (2 * 2 * 5). If 10 has been computed before (which is the case as X1 * X3 is handled before x1^2 * x3, you don't want to recompute the two factors decomposition.

dannywillems: I renamed acc in cache in c05812ef828c1c2884ff7b326e55d950d15ff053

sympolyc/src/expr.rs
dannywillems: It is the whole point of this approach using prime numbers.

sympolyc/src/expr.rs
dannywillems: I forgot to mention in the doc: it is expected for now. This first PR is a PoC regarding multiplication using prime numbers. I will change later or add some warnings/raise exception in future PR.

sympolyc/src/expr.rs
dannywillems: Done in 275f8517c32cb8d7ee34ca8577beb7471546b146

sympolyc/src/expr.rs
dannywillems: > Is this guy supposed to be with the polynomial ?
> He only depends on N and D, we shouldn't carry him with every polynomial, nor compute him every time we create a polynomial

No, I want to move it out. Only for now.

Regarding the degree bound, answered below.

sympolyc/src/expr.rs
dannywillems: Done in 4e7b0f2c96dd2c5ac1606d36a61ff08f1c90d76d

sympolyc/src/expr.rs
dannywillems: Done in aa960633d2e6c8dcaaa74ef358237256444ae524

sympolyc/src/expr.rs
dannywillems: removed in d4bed9e9ac75a95d54f8ec4be5ffca4a78a59dd5

sympolyc/src/expr.rs
dannywillems: Note that it is like this for now as it is initially supposed to be used from the expression framework, and with constraints where we know the number of variables and the degree.

marcbeunardeau88 (approved):

PR #2476, opened 2024-08-21
Add serde_as regression test for pasta

volhovm:


utils/src/serialization.rs
```diff
@@ -87,3 +87,61 @@ where
...
+        // reference serialized value
+        let buf_expected: Vec<u8> = vec![
+            146, 196, 33, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 196, 33, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        ];
+
+        let mut buf_written: Vec<u8> = vec![0; buf_expected.len()];
+
+        let srs_bytes = rmp_serde::to_vec(&data_expected).unwrap();
```
dkijania, line 126: Nit pick: don't use unwrap()  but expects. Generally it's easier to find what is wrong if you have more than one unwrap() in  the test

kimchi/src/tests/framework.rs
```diff
@@ -122,7 +122,6 @@ where
     }
 
     /// creates the indexes
-    #[must_use]
```
volhovm, line 125: Offtop but: I removed this since it was giving a warning: `setup_with_custom_srs` is indeed used, but only in tests. `must_use` does not understand this and produces a warning.

volhovm: This did not work.

PR #2477, opened 2024-08-22
Fix setup_with_custom_srs warning

volhovm:
The function is actually used (in tests), but was producing an annoying warning. So `allow(dead_code)` it is.

### DIFF
--- a/kimchi/src/tests/framework.rs
+++ b/kimchi/src/tests/framework.rs
@@ -121,7 +121,10 @@ where
         self
     }
 
+    // Re allow(dead_code): this method is used in tests; without the annotation it warns unnecessarily.
     /// creates the indexes
+    #[must_use]
+    #[allow(dead_code)]
     pub(crate) fn setup_with_custom_srs<F: FnMut(D<G::ScalarField>, usize) -> OpeningProof::SRS>(
         mut self,
         get_srs: F,


kimchi/src/tests/framework.rs
```diff
@@ -121,7 +121,10 @@ where
         self
     }
 
+    // Re allow(dead_code): this method is used in tests; without the annotation it warns unnecessarily.
     /// creates the indexes
+    #[must_use]
+    #[allow(dead_code)]
```
Fizzixnerd, line 127: Are these two things not conflicting?  I'm still learning rust, so please forgive, but naively I'd expect `must_use` ensures a value is used, while allowing it to be dead reverses that.

volhovm: `must_use` has a really weird semantics: it forces the caller of the function to consider its output explicitly (and forbids to ignore it):
* https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute
* https://std-dev-guide.rust-lang.org/policy/must-use.html

So I guess it does not contradict `dead_code`. The combination above means "it's fine to not use the function, but if you use it, you have to examine its output".

Fizzixnerd:
LGTM!

kimchi/src/tests/framework.rs
```diff
@@ -121,7 +121,10 @@ where
         self
     }
 
+    // Re allow(dead_code): this method is used in tests; without the annotation it warns unnecessarily.
     /// creates the indexes
+    #[must_use]
```
dannywillems, line 126: I would be in favor of moving tests in subdirectory later.

dannywillems (approved):

PR #2480, opened 2024-08-26
Revisit 2278: Unify hash digest format on both sides (MIPS + Keccak)

querolita:
This is based upon https://github.com/o1-labs/proof-systems/pull/2278/.

The only difference is that the above PR assumed that bytes corresponding to the length of the preimage were always read in a different instruction than actual preimage bytes. But I am no longer sure that I had checked that thoroughly back then, so I am creating this PR to solely include the fix for the hash digest format: that is, making sure that both in MIPS and Keccak the data sent to the communication channel corresponds to the 31 bytes of the preimage (no MSB in either of them). I left the check of the length bytes as a FIXME for the future.


o1vm/src/mips/witness.rs
```diff
@@ -44,7 +45,9 @@ pub const NUM_INSTRUCTION_LOOKUP_TERMS: usize = 5;
 pub const NUM_LOOKUP_TERMS: usize =
     NUM_GLOBAL_LOOKUP_TERMS + NUM_DECODING_LOOKUP_TERMS + NUM_INSTRUCTION_LOOKUP_TERMS;
 // TODO: Delete and use a vector instead
-pub const SCRATCH_SIZE: usize = 97; // MIPS + hash_counter + byte_counter + eof + num_bytes_read + chunk + bytes + length + has_n_bytes
+// MIPS + hash_counter + byte_counter + eof + num_bytes_read + chunk + bytes
+// + length + has_n_bytes + chunk_bytes + preimage
+pub const SCRATCH_SIZE: usize = 98;
```
marcbeunardeau88, line 50: the comment says you're adding two things, chunk_bytes + preimage, and yet the size only grows by one

marcbeunardeau88 (approved):

o1vm/src/mips/witness.rs
querolita: It's just the preimage key that is being added wrt what's in master. The problem is that in the old comment the chunk_bytes were not written down, but they were in any case part of the state https://github.com/o1-labs/proof-systems/blob/7d51847c00ff790bc22af3c83ddedc1deb070e38/o1vm/src/mips/column.rs#L26C1-L27C1

PR #2482, opened 2024-08-26
MVPoly: implement random

dannywillems:


Fizzixnerd (commented):
I feel like having `random` be part of the public interface is a little dangerous, as it may provide guarantees about randomness that you didn't mean to.

dannywillems:
> about randomness that you didn't mean to.

What do you mean?

We often provide a rand/random methods to generate values for testing, in particular. I do not plan to use it, but for testing. I can move (in a follow-up PR) the method into the tests directory, if you don't want to have it in the public interface. I'm fine with this.

Fizzixnerd:
> > about randomness that you didn't mean to.
> 
> What do you mean?
> 
> We often provide a rand/random methods to generate values for testing, in particular. I do not plan to use it, but for testing. I can move (in a follow-up PR) the method into the tests directory, if you don't want to have it in the public interface. I'm fine with this.

I was worried that someone might use these in a probabilistic protocol as a source of randomness, which is presumably not what you wanted.  If people aren't going to be writing code like that, I'm happy to approve.

Fizzixnerd (approved):

dannywillems:
> > > > about randomness that you didn't mean to.
> > > 
> > > 
> > > What do you mean?
> > > We often provide a rand/random methods to generate values for testing, in particular. I do not plan to use it, but for testing. I can move (in a follow-up PR) the method into the tests directory, if you don't want to have it in the public interface. I'm fine with this.
> > 
> > 
> > I was worried that someone might use these in a probabilistic protocol as a source of randomness, which is presumably not what you wanted. If people aren't going to be writing code like that, I'm happy to approve.
> 
> I see. Worth using `unsafe` then. Doing in a follow-up.

See https://github.com/o1-labs/proof-systems/pull/2498

PR #2484, opened 2024-08-26
MVPoly: add double

dannywillems:
To match expr interface


dannywillems:
> Does rust not have method `map` for `coeff`?

See https://github.com/o1-labs/proof-systems/pull/2496

Fizzixnerd (approved):

PR #2485, opened 2024-08-26
MVPoly: add tests for double

dannywillems:


dannywillems:
>  For example, I doubt the RNG is printed in that case.

See https://github.com/o1-labs/proof-systems/blob/master/utils/src/lib.rs#L33

Fizzixnerd (approved):
Even if I purpose this line of reasoning, it's better to be in a separate PR.  Approving this.

PR #2488, opened 2024-08-27
MVPoly: implement Index/IndexMut for prime::Dense

dannywillems:


Fizzixnerd:
Should you add a `map` method too?

dannywillems:
> Should you add a `map` method too?

Can be useful in the future. Keeping in mind if I see the needs.

Fizzixnerd (approved):

PR #2489, opened 2024-08-27
MVPoly: implement mul_by_const for prime::Dense

dannywillems:


Fizzixnerd (approved):
NitPick (NP): Possibly a better name would be `scalar_mul` or `mul_by_scalar`, as you are using the word `const` here to mean "constant polynomial", but then you pass in a field element, i.e. not a polynomial (but obviously could represent one).  Up to you, I leave to your judgement.

dannywillems:
Agree with mul_by_scalar. Changing.

dannywillems:
> NitPick (NP): Possibly a better name would be `scalar_mul` or `mul_by_scalar`, as you are using the word `const` here to mean "constant polynomial", but then you pass in a field element, i.e. not a polynomial (but obviously could represent one). Up to you, I leave to your judgement.

See https://github.com/o1-labs/proof-systems/pull/2497

Fizzixnerd (approved):
approving as part of train.

PR #2492, opened 2024-08-27
MVPoly: implement from Expr

dannywillems:


mvpoly/src/prime.rs
```diff
@@ -542,4 +549,138 @@ impl<F: PrimeField, const N: usize, const D: usize> From<F> for Dense<F, N, D> {
...
+impl<F: PrimeField, const N: usize, const D: usize> From<Operations<ConstantExprInner<F>>>
+    for Dense<F, N, D>
+{
+    fn from(op: Operations<ConstantExprInner<F>>) -> Self {
+        use kimchi::circuits::expr::Operations::*;
+        match op {
+            Atom(op_const) => Self::from(op_const),
+            Add(c1, c2) => Self::from(*c1) + Self::from(*c2),
+            Sub(c1, c2) => Self::from(*c1) - Self::from(*c2),
+            Mul(c1, c2) => Self::from(*c1) + Self::from(*c2),
```
Fizzixnerd, line 598: This should use mul, I would imagine (not +).

Fizzixnerd (changes requested):

mvpoly/src/prime.rs
Fizzixnerd: I'm a little worried about the effectiveness of the tests for not catching this, as well, but that's another issue.

Fizzixnerd: I feel like I don't have the necessary background knowledge to properly review this commit.

cc @marcbeunardeau88

dannywillems: huhu, nice catch :see_no_evil:

dannywillems: image

Fizzixnerd (approved):

mvpoly/src/prime.rs
```diff
@@ -585,4 +592,138 @@ impl<F: PrimeField, const N: usize, const D: usize> From<F> for Dense<F, N, D> {
     }
 }
 
+impl<F: PrimeField, const N: usize, const D: usize> From<ConstantExprInner<F>> for Dense<F, N, D> {
+    fn from(expr: ConstantExprInner<F>) -> Self {
+        match expr {
+            // The unimplemented methods might be implemented in the future if
+            // we move to the challenge into the type of the constant
+            // terms/expressions
+            // Unrolling for visibility
```
marcbeunardeau88, line 601: I would not unroll, I don't think it helps reading the code

mvpoly/src/prime.rs
```diff
@@ -585,4 +592,138 @@ impl<F: PrimeField, const N: usize, const D: usize> From<F> for Dense<F, N, D> {
...
+                        unimplemented!("Not used in this context")
+                    }
+                    ExprInner::Constant(c) => Self::from(c),
+                    ExprInner::Cell(Variable { col, row }) => {
+                        assert_eq!(row, CurrOrNext::Curr, "Only current row is supported for now. You cannot reference the next row");
+                        Self::from_variable(col)
+                    }
+                }
+            }
+            Add(e1, e2) => {
```
marcbeunardeau88, line 687: there is no way to not duplicate eg. this add, as it is already there for inner expression ?

mvpoly/src/prime.rs
```diff
@@ -585,4 +592,138 @@ impl<F: PrimeField, const N: usize, const D: usize> From<F> for Dense<F, N, D> {
...
+            Pow(c, e) => {
+                // FIXME: dummy implementation
+                let p = Dense::from(*c);
+                let mut result = p.clone();
+                for _ in 0..e {
+                    result = result.clone() * p.clone();
+                }
+                result
+            }
+            Cache(_c, _) => {
```
marcbeunardeau88, line 719: Is this true that we do not need to convert from expressions using cache ?
same for iffeatures

mvpoly/tests/prime.rs
```diff
@@ -400,3 +400,56 @@ fn test_evaluation_predefined_polynomial() {
...
+fn test_eval_pbt_neg() {
+    let mut rng = o1_utils::tests::make_test_rng(None);
+
+    let random_evaluation: [Fp; 6] = std::array::from_fn(|_| Fp::rand(&mut rng));
+    let p1 = Dense::<Fp, 6, 4>::random(&mut rng);
+    let p2 = -p1.clone();
+    let eval_p1 = p1.eval(&random_evaluation);
+    let eval_p2 = p2.eval(&random_evaluation);
+    assert_eq!(eval_p2, -eval_p1);
+}
```
marcbeunardeau88, line 455: you don't check the multiplication?

marcbeunardeau88 (approved):

mvpoly/src/prime.rs
dannywillems: For the current use case (Arrabiata), we do not need. We might need in the future.

mvpoly/tests/prime.rs
dannywillems: Addressed in https://github.com/o1-labs/proof-systems/pull/2511/

PR #2494, opened 2024-08-27
MVPoly: implement from_variable in MVPoly

dannywillems:


mvpoly/src/prime.rs
```diff
@@ -241,6 +241,21 @@ impl<F: PrimeField, const N: usize, const D: usize> Dense<F, N, D> {
         }
     }
 
+    pub fn from_variable<C: Into<usize>>(var: C) -> Self {
+        let mut res = Self::zero();
+        let mut prime_gen = PrimeNumberGenerator::new();
+        let primes = prime_gen.get_first_nth_primes(N);
+        let var_usize: usize = var.into();
+        assert!(primes.contains(&var_usize), "The usize representation of the variable must be a prime number, and unique for each variable");
```
dannywillems, line 249: Maybe we want the opposite? Maybe we want to have the index of the prime number?

Fizzixnerd: I think this way is more clear, personally.  All converting to indices will do is force the programmer to do this calculation outside of their rust code, and possibly mess it up.

Fizzixnerd (approved):

PR #2499, opened 2024-08-27
MVPoly: implement eval

dannywillems:
Dummy implementation. It can be improved, see comment.


mvpoly/src/prime.rs
```diff
@@ -321,6 +321,37 @@ impl<F: PrimeField, const N: usize, const D: usize> Dense<F, N, D> {
...
+                    // vector space
+                    let prime_decomposition = naive_prime_factors(normalized_index, &mut prime_gen);
+                    let mut monomial = F::one();
+                    prime_decomposition.iter().for_each(|(p, d)| {
+                        // IMPROVEME: we should keep the inverse indices
+                        let inv_p = primes.iter().position(|&x| x == *p).unwrap();
+                        let x_p = x[inv_p].pow([*d as u64]);
+                        monomial *= x_p;
+                    });
+                    acc + *c * monomial
```
Fizzixnerd, line 351: little confused here: why do you deref c in one case and not the other?

dannywillems: `iter` returns a reference: https://stackoverflow.com/questions/34733811/what-is-the-difference-between-iter-and-into-iter. Therefore we need to dereference.
A deref can also be automatically performed, if the operation is implemented for references. I would guess that the multiplication is not implemented between reference (`c`) and direct value (`monomial`), where it is for addition (`acc + c`).

I would guess we have the following implementations:
```
impl Add for Fp {
  type Output = Self;

  fn add(self, other: &Self);
}

impl Add for Fp {
  type Output = Self;

  fn add(self, other: Self);
}
```
but we don't have
```
//           |--- direct value
impl Mul for Fp {
  type Output = Self;
  //                  |-- Ref here
  fn mul(self, other: &Self);
}
```

dannywillems:
Seems it was already on master, and rebasing closed the PR. Oopsie.

PR #2502, opened 2024-08-28
MVPoly/prime: implement compute_cross_terms

dannywillems:


marcbeunardeau88 (approved):

mvpoly/src/prime.rs
```diff
@@ -439,6 +440,126 @@ impl<F: PrimeField, const N: usize, const D: usize> Dense<F, N, D> {
...
+    // IMPROVEME: Dummy implementation, a cache can be used to save the
+    // previously computed multiplications and powers.
+    // Maybe using a symbolic form could speed up the computation.
+    pub fn compute_cross_terms(
+        &self,
+        eval1: &[F; N],
+        eval2: &[F; N],
+        u1: F,
+        u2: F,
+    ) -> HashMap<usize, F> {
```
Fizzixnerd, line 466: Maybe this is a dumb question, but I don't understand why we use a HashMap here instead of just a Vector.

Fizzixnerd: Just noticed this is still draft.  Sorry.  Feel free to resolve.

dannywillems:
Leaving like this for now, as focusing on monomials form and Arrabiata for now.

PR #2504, opened 2024-08-28
CI refactoring, test coverage data gathering and reports generation.

shimkiv:
- This PR attempts to generalize the complex scripts usage by wrapping them with the `Makefile`.
  - This will make it impossible to run `make` command on Windows natively (not sure though if we have users who build stuff on Windows without the WSL).
  - Should we create the Windows CMD/Powershell scripts to emulate some `Make` targets?
- CI is updated accordingly to use the Makefile.
  - We need to decide how we want to run tests with code coverage data gathering and reporting. Please refer to [internal](https://o1-labs.slack.com/archives/C07F4MR66G2/p1724784387708559) discussion.
- Formatting applied against the root `.md` files.
  - More information and usage examples will be added as part of the followup tasks.
- [Codecov](https://app.codecov.io/gh/o1-labs/proof-systems) and current repository are configured for coverage reports uploading.

---

Tasks left:

- [x] Run it for one specific Rust version only (no reason to do it for whole the `matrix`).
  - [x] Conditionally run `make nextest-with-coverage` for one of the `rust_toolchain_version: ["1.71", "1.72", "1.73", "1.74"]`.
  - [x] Extract the `make nextest-heavy-with-coverage` into Nightly builds with on-demand execution option.
- [x] Document changes and usage options in `CONTRIBUTING.md`.
  - `CARGO_EXTRA_ARGS="-p poly-commitment" make test-all-with-coverage`
  - `BIN_EXTRA_ARGS="-p poly-commitment" make nextest-all-with-coverage`


dannywillems:
> Should we create the Windows CMD/Powershell scripts to emulate some Make targets?

No, we should not. It is fine for now, IMO.

Makefile
```diff
@@ -0,0 +1,106 @@
...
+		# https://nexte.st/book/pre-built-binaries.html#using-nextest-in-github-actions
+		# FIXME: update to 0.9.68 when we get rid of 1.71 and 1.72.
+		cargo install cargo-nextest@=0.9.67 --locked
+		# FIXME: latest 0.8.19+ requires rustc 1.74+
+		cargo install grcov@=0.8.13 --locked
+		@echo ""
+		@echo "Test dependencies installed."
+		@echo ""
+
+# Default target
```
dannywillems, line 21: Should we make `release` the default target?
I don't have a strong opinion, but running `make` will always clean the cargo workspace, which might not be the default behavior a user expects? Up to you.

dannywillems (commented):
LGTM so far. Feel free to continue with the tasks left.
I tested a few `make` targets.

.github/workflows/rust.yml
```diff
@@ -170,30 +165,35 @@ jobs:
...
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
 
       - name: Run heavy tests
         if: ${{ matrix.rust_toolchain_version == '1.74' }}
         run: |
           eval $(opam env)
-          cargo nextest run --all-features --release --profile ci -E "test(heavy)"
-
+          make nextest-heavy-with-coverage
```
dannywillems, line 191: I would run the heavy tests with coverage only on demand.

shimkiv: I'll extract it into the Nightly/on-demand.
And yes it is possible it might killed the runner.
image

Makefile
shimkiv: `release` target will be the default one (without the cleaning).

.github/workflows/rust.yml
shimkiv: Added this to the PR's tasks list.

dannywillems (approved):
Already approving as it seems it is only related to the linter now.

dannywillems (approved):

shimkiv:
Hey folks, 
let's iterate over this changes later to see if we can improve PR checks timing.

dannywillems:
> Hey folks, let's iterate over this changes later to see if we can improve PR checks timing.

Agree.

dannywillems (approved):

volhovm:
@shimkiv is it possible to get this into `compatible` too?

shimkiv:
@volhovm everything is possible, I'm just thinking which changes you'd like to have in `compatible`? If we are talking about "quality-of-live" changes only - then no problem. But if you were also thinking to have code coverage uploading to the Codecov service - here we might start to have issues. 
Codecov is configured to consider `master` as default/main branch it will compare coverage against for diff reports. And if `compatible` is different to `main` (`master`) by design, then you might end up with constant "failed" jobs on CI side (unless such comparison will be intentionally disabled).

volhovm:
^ (discussed yesterday in PM, but copying): I'd ideally like to see the CI files/Makefile structure synchronized between this repo's `master` and `develop`, because rn `develop` is outdated.

PR #2506, opened 2024-08-28
MVPoly: add max_degree optional parameter into random

dannywillems:


mvpoly/src/prime.rs
```diff
@@ -234,12 +237,29 @@ impl<F: PrimeField, const N: usize, const D: usize> Dense<F, N, D> {
...
+                        .fold(0, |acc, (_, d)| acc + d);
+                    if degree > max_degree {
+                        F::zero()
+                    } else {
+                        F::rand(rng)
+                    }
+                })
+                .collect::<Vec<F>>()
+        } else {
+            normalized_indices.iter().map(|_| F::rand(rng)).collect()
```
marcbeunardeau88, line 261: why isn't it collect()::<Vec<F>>() here ?
It was the case before changing and is the case when max degree is provided

dannywillems: The typer can infer the type.

PR #2508, opened 2024-08-28
MVPoly: add tests for is_constant

dannywillems:


mvpoly/tests/prime.rs
```diff
@@ -663,3 +663,24 @@ fn test_degree_random_degree() {
     let degree = unsafe { p.degree() };
     assert!(degree <= max_degree);
 }
+
+#[test]
+fn test_is_constant() {
+    let mut rng = o1_utils::tests::make_test_rng(None);
+    let c = Fp::rand(&mut rng);
```
marcbeunardeau88, line 670: although it's not an issue, I'd rather we reserved PBT when then make sense.
I don't see eg. here how PBT makes it safer

marcbeunardeau88 (approved):

PR #2509, opened 2024-08-28
Add initial flake for installing build dependencies for `proof-systems`

Fizzixnerd:
This is the start of a very simple flake.

In scope for this version of the flake:

1. `nix develop .` will open a dev shell with everything you need to develop on various parts of `proof-systems` in scope (caveat about rust and ocaml toolchain version in next point).
2. You must still use `rustup` to manage your rust toolchains.
3. You must still use `opam` to manage your ocaml toolchains.

Out of scope for this version of the flake:

1. Distributed caching -- only if the others in this list come.
2. Actually building things using `nix` -- likely to come.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,3 +81,7 @@ The CI will build different targets.
   When CI passes on master, the documentation built from the rust code will be
   available [here](https://o1-labs.github.io/proof-systems/rustdoc) and the book
   will be available [here](https://o1-labs.github.io/proof-systems).
+
+## Nix for Dependencies (WIP)
+
+If you have `nix` installed and in particular, `flakes` enabled, you can install the dependencies for these projects using nix.  Simply `nix develop .` inside this directory to bring into scope `rustup`, `opam`, and `go` (along with a few other tools).  You will have to manage the toolchains yourself using `rustup` and `opam`, in the current iteration.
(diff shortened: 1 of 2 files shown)


dannywillems (commented):
Can we have a continuous check in the CI that it works, and the code can be built, please?

Fizzixnerd:
> Can we have a continuous check in the CI that it works, and the code can be built, please?

Absolutely.  I have tested

`cargo build` and that succeeds.  What other checks would you like?  My understanding is there is go and ocaml code somewhere in this repo.

dannywillems (approved):
Approving as it is pretty simple. As a non-nix user, I simply tried the commands in the README.
Feel free to merge.

Fizzixnerd:
> > > Can we have a continuous check in the CI that it works, and the code can be built, please?
> > 
> > 
> > Absolutely. I have tested
> > `cargo build` and that succeeds. What other checks would you like? My understanding is there is go and ocaml code somewhere in this repo.
> 
> Would it be worth, even if the nix config is simple, to check in the CI that it works? By running, for instance, `cargo build` in the nix environment? IIRC, we do that in the Mina repository.

I will make an issue so I don't forget to.

PR #2514, opened 2024-08-30
MVPoly: util to compute nested loops indices

dannywillems:
It will be used later in the cross-term computation to perform the multi-nomial sum

PR #2516, opened 2024-09-03
MVPoly: implement monomials form

dannywillems:
It aims to share the same interface as `prime::Dense`, but it is a WIP. I would suggest to add a trait later.

Built on top of #2514

### DIFF
--- a/mvpoly/src/monomials.rs
+++ b/mvpoly/src/monomials.rs
@@ -1,0 +1,480 @@
+use ark_ff::{One, PrimeField, Zero};
+use rand::RngCore;
+use std::{
+    collections::HashMap,
+    fmt::Debug,
+    ops::{Add, Mul, Neg, Sub},
+};
+
+use crate::{
+    prime,
+    utils::{naive_prime_factors, PrimeNumberGenerator},
+};
+
+/// Represents a multivariate polynomial in `N` variables with coefficients in
+/// `F`. The polynomial is represented as a sparse polynomial, where each
+/// monomial is represented by a vector of `N` exponents.
+// We could use u8 instead of usize for the exponents
+// FIXME: the maximum degree D is encoded in the type to match the type
+// prime::Dense
+#[derive(Clone)]
+pub struct Sparse<F: PrimeField, const N: usize, const D: usize> {
+    pub monomials: HashMap<[usize; N], F>,
+}
+
+impl<const N: usize, const D: usize, F: PrimeField> Add for Sparse<F, N, D> {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        let mut monomials = self.monomials.clone();
+        for (exponents, coeff) in other.monomials {
+            monomials
+                .entry(exponents)
+                .and_modify(|c| *c += coeff)
+                .or_insert(coeff);
+        }
+        // Remove monomials with zero coefficients
+        let monomials: HashMap<[usize; N], F> = monomials
+            .into_iter()
+            .filter(|(_, coeff)| !coeff.is_zero())
+            .collect();
+        // Handle the case where the result is zero because we want a unique
+        // representation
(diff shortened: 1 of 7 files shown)


mvpoly/src/monomials.rs
```diff
@@ -0,0 +1,480 @@
...
+        } else {
+            Sparse::<F, N, D> { monomials }
+        }
+    }
+}
+
+impl<const N: usize, const D: usize, F: PrimeField> Add<&Sparse<F, N, D>> for Sparse<F, N, D> {
+    type Output = Sparse<F, N, D>;
+
+    fn add(self, other: &Sparse<F, N, D>) -> Self::Output {
```
marcbeunardeau88, line 54: isn't the same function implemented l28 ?

mvpoly/src/monomials.rs
```diff
@@ -0,0 +1,480 @@
...
+        } else {
+            Sparse::<F, N, D> { monomials }
+        }
+    }
+}
+
+impl<const N: usize, const D: usize, F: PrimeField> Add<Sparse<F, N, D>> for &Sparse<F, N, D> {
+    type Output = Sparse<F, N, D>;
+
+    fn add(self, other: Sparse<F, N, D>) -> Self::Output {
```
marcbeunardeau88, line 80: same comment

mvpoly/src/monomials.rs
```diff
@@ -0,0 +1,480 @@
...
+        } else {
+            Sparse::<F, N, D> { monomials }
+        }
+    }
+}
+
+impl<const N: usize, const D: usize, F: PrimeField> Add<&Sparse<F, N, D>> for &Sparse<F, N, D> {
+    type Output = Sparse<F, N, D>;
+
+    fn add(self, other: &Sparse<F, N, D>) -> Self::Output {
```
marcbeunardeau88, line 106: same comment

mvpoly/src/monomials.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+/// Equality is defined as equality of the monomials.
+impl<const N: usize, const D: usize, F: PrimeField> PartialEq for Sparse<F, N, D> {
+    fn eq(&self, other: &Self) -> bool {
+        self.monomials == other.monomials
+    }
+}
+
+impl<const N: usize, const D: usize, F: PrimeField> Eq for Sparse<F, N, D> {}
```
marcbeunardeau88, line 322:
```suggestion

```

mvpoly/src/monomials.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+impl<const N: usize, const D: usize, F: PrimeField> Zero for Sparse<F, N, D> {
+    fn is_zero(&self) -> bool {
+        self.monomials.len() == 1
+            && self.monomials.contains_key(&[0; N])
+            && self.monomials[&[0; N]].is_zero()
+    }
+
+    fn zero() -> Self {
```
marcbeunardeau88, line 339: why is zero not the empty map ?

marcbeunardeau88 (approved):

mvpoly/src/monomials.rs
dannywillems: :sweat_smile: 
We can use macros, but it is not really readable.
There might also be a better way that I don't know.

mvpoly/src/monomials.rs
dannywillems: No, it is necessary to implement the trait `Eq`.

mvpoly/src/monomials.rs
dannywillems: I wanted to have a single representation for the zero polynomial, and I picked the constant polynomial with `0`. No strong opinion. Feel free to change it later.

PR #2520, opened 2024-09-04
MVPoly/monomials: compute cross-terms

dannywillems:


mvpoly/src/monomials.rs
```diff
@@ -447,6 +448,106 @@ impl<const N: usize, const D: usize, F: PrimeField> MVPoly<F, N, D> for Sparse<F
     fn from_expr<Column: Into<usize>>(expr: Expr<ConstantExpr<F>, Column>) -> Self {
         prime::Dense::from_expr(expr).into()
     }
+
```
marcbeunardeau88, line 451:
```suggestion
/// see https://hackmd.io/@dannywillems/Syo5MBq90
```

dannywillems: The documentation is already added from the trait.

marcbeunardeau88 (approved):
I'd like to change that to a functions that computes an MVPoly rather than directly evaluating, but let's merge that for now

PR #2521, opened 2024-09-04
Improved comments for chunked Lagrange evaluations

querolita:
This PR is part of https://github.com/o1-labs/o1js/issues/1806 .

It is a no-op, simply adding more verbosity to the specs and comments related to the chunked Lagrange evaluations.

### DIFF
--- a/kimchi/src/lagrange_basis_evaluations.rs
+++ b/kimchi/src/lagrange_basis_evaluations.rs
@@ -100,8 +100,7 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
         let stride = p.evals.len() / self.evals[0].len();
         self.evals
             .iter()
-            .enumerate()
-            .map(|(j, evals)| {
+            .map(|evals| {
                 let mut result = F::zero();
                 for (i, e) in evals.iter().enumerate() {
                     if !p.evals[stride * i].is_zero() {
--- a/kimchi/src/lagrange_basis_evaluations.rs
+++ b/kimchi/src/lagrange_basis_evaluations.rs
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
 use ark_poly::{EvaluationDomain, Evaluations, Radix2EvaluationDomain as D};
 use rayon::prelude::*;
 
-/// The evaluations of all normalized lagrange basis polynomials at a given
-/// point. Can be used to evaluate an `Evaluations` form polynomial at that point.
+/// Evaluations of all normalized lagrange basis polynomials at a given point.
+/// Can be used to evaluate an `Evaluations` form polynomial at that point.
+///
+/// The Lagrange basis for polynomials of degree <= d over a domain
+/// {ω_0,...,ω_d} is the set of (d+1) polynomials ${l_0,...,l_d}$ of degree d
+/// that equal 1 at ω_i and 0 in the rest of the domain terms. They can be used
+/// to evaluate polynomials in evaluation form efficiently in O(d) time.
+///
+/// When chunking is in place, the domain size is larger than the maximum
+/// polynomial degree allowed. In particular, the SRS is not long enough to
+/// support higher degrees. Thus, we cannot obtain a polynomial f with degree
+/// c·n (with domain size n and number of chunks c) as usual with the equation:
+///
+/// f(X) = x_0 · l_0(X) + ... + x_n · l_n(X) + ... + x_{c*n} · l_{c*n}(X)
+///
+/// Instead, this struct will contain the (c*n+1) coefficients of the polynomial
+/// that is equal to the powers of the point x in the positions corresponding to
+/// the chunk, and 0 elsewhere in the domain. This is useful to evaluate the
+/// chunks of polynomials of degree c·n given in evaluation form at the point.
+///
 pub struct LagrangeBasisEvaluations<F> {
+    /// If no chunking:
+    /// - evals is a vector of length 1 containing a vector of size n
+    ///   corresponding to the evaluations of the Lagrange polynomials, which
+    ///   are the polynomials that equal 1 at ω_i and 0 elsewhere in the domain.
+    /// If chunking (a vector of size c * N )
+    /// - the first index refers to the chunks
+    /// - the second index refers j-th coefficient of the i-th chunk of the
+    ///   polynomial that equals the powers of the point and 0 elsewhere.
+    ///
     evals: Vec<Vec<F>>,
 }
 
 impl<F: FftField> LagrangeBasisEvaluations<F> {
-    /// Given the evaluations form of a polynomial, directly evaluate that polynomial at a point.
+    /// Given the evaluations form of a polynomial, directly evaluate that
+    /// polynomial at a point.
+    ///
+    /// The Lagrange basis evaluations can be used to evaluate a polynomial of
+    /// given in evaluation form efficiently in O(n) time, of degree d <= n
+    /// where n is the domain size, without the need of interpolating it first.
+    ///
+    /// Recall that a polynomial can be represented as the sum of the scaled
+    /// Lagrange polynomials using its evaluations on the domain:
+    /// $ f(x) = x_0 · l_0(x) + ... + x_n · l_n(x) $
+    ///
+    /// Recall that when chunking is in place, we want to evaluate a
+    /// polynomial f of degree c · m at point z, expressed as
+    ///
+    /// f(z) = a_0·z^0 + ... + a_{n-1}·z^{n-1} + ... + a_{c*n-1}·z^{c*n-1}
+    ///      = z^0 · f_0(z) + z^n · f_1(z) + ... + z^{(c-1)n} · f_{c-1}(z)
+    ///
+    /// where f_i(X) is the i-th chunked polynomial of degree m of f:
+    /// f_i(x) = a_{i*n} · x^0 + ... + a_{(i+1)n-1} · x^{n-1}
+    ///
+    /// Returns the evaluation of each chunk of the polynomial at the point
+    /// (when there is no chunking, the result is a vector of length 1). They
+    /// correspond to the f_i(z) in the equation above.
+    ///
     pub fn evaluate<D: EvaluationDomain<F>>(&self, p: &Evaluations<F, D>) -> Vec<F> {
+        // The domain size must be a multiple of the number of evaluations so
+        // that the degree of the polynomial can be split into chunks of equal size.
         assert_eq!(p.evals.len() % self.evals[0].len(), 0);
+        // The number of chunks c
         let stride = p.evals.len() / self.evals[0].len();
         let p_evals = &p.evals;
+
+        // Performs the operation
+        //                         n-1
+        // j ∈ [0, c) : eval_{j} =  Σ   p_{i · c} · l_{j,i}
+        //                         i=0
+        //
+        // Note that in the chunking case, the Lagrange basis contains the
+        // coefficient form of the polynomial that evaluates to the powers of z
+        // in the chunk positions and zero elsewhere.
+        //
+        // Then, the evaluation of f on z can be computed as the sum of the
+        // products of the evaluations of f in the domain and the Lagrange
+        // evaluations.
+
         (&self.evals)
             .into_par_iter()
             .map(|evals| {
@@ -33,10 +100,11 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
         let stride = p.evals.len() / self.evals[0].len();
         self.evals
             .iter()
-            .map(|evals| {
+            .enumerate()
+            .map(|(j, evals)| {
                 let mut result = F::zero();
                 for (i, e) in evals.iter().enumerate() {
-                    if !p.evals[stride * i].is_zero() {
+                    if !p.evals[stride * j + i].is_zero() {
                         result += e;
                     }
                 }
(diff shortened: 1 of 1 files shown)


kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -33,10 +100,11 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
         let stride = p.evals.len() / self.evals[0].len();
         self.evals
             .iter()
-            .map(|evals| {
+            .enumerate()
+            .map(|(j, evals)| {
                 let mut result = F::zero();
                 for (i, e) in evals.iter().enumerate() {
-                    if !p.evals[stride * i].is_zero() {
+                    if !p.evals[stride * j + i].is_zero() {
```
dannywillems, line 107: :eyes:

querolita: Oh thanks for pointing out! I left that line after doing some prints, will delete it asap

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -33,7 +100,8 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
         let stride = p.evals.len() / self.evals[0].len();
         self.evals
             .iter()
-            .map(|evals| {
+            .enumerate()
```
dannywillems, line 103: I suppose this can also be removed as `j` is not used.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
 use ark_poly::{EvaluationDomain, Evaluations, Radix2EvaluationDomain as D};
 use rayon::prelude::*;
 
-/// The evaluations of all normalized lagrange basis polynomials at a given
-/// point. Can be used to evaluate an `Evaluations` form polynomial at that point.
+/// Evaluations of all normalized lagrange basis polynomials at a given point.
+/// Can be used to evaluate an `Evaluations` form polynomial at that point.
+///
+/// The Lagrange basis for polynomials of degree <= d over a domain
+/// {ω_0,...,ω_d} is the set of (d+1) polynomials ${l_0,...,l_d}$ of degree d
```
dannywillems, line 9: Note that LaTeX notations like `$` do not work in the doc. \` has to be used, like `{l_0, ..., l_d}`. It is pure Markdown that is accepted. See the generated doc by
```
RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps
```
and open
```
firefox target/doc/kimchi/index.html
```

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+    ///
     evals: Vec<Vec<F>>,
 }
 
 impl<F: FftField> LagrangeBasisEvaluations<F> {
-    /// Given the evaluations form of a polynomial, directly evaluate that polynomial at a point.
+    /// Given the evaluations form of a polynomial, directly evaluate that
+    /// polynomial at a point.
+    ///
+    /// The Lagrange basis evaluations can be used to evaluate a polynomial of
```
dannywillems, line 42: polynomial of ...?

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+    /// given in evaluation form efficiently in O(n) time, of degree d <= n
+    /// where n is the domain size, without the need of interpolating it first.
+    ///
+    /// Recall that a polynomial can be represented as the sum of the scaled
+    /// Lagrange polynomials using its evaluations on the domain:
+    /// $ f(x) = x_0 · l_0(x) + ... + x_n · l_n(x) $
+    ///
+    /// Recall that when chunking is in place, we want to evaluate a
+    /// polynomial f of degree c · m at point z, expressed as
```
dannywillems, line 51:
```suggestion
    /// polynomial f of degree c · n at point z, expressed as
```

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+    ///
+    /// f(z) = a_0·z^0 + ... + a_{n-1}·z^{n-1} + ... + a_{c*n-1}·z^{c*n-1}
+    ///      = z^0 · f_0(z) + z^n · f_1(z) + ... + z^{(c-1)n} · f_{c-1}(z)
+    ///
+    /// where f_i(X) is the i-th chunked polynomial of degree m of f:
```
dannywillems, line 56:
```suggestion
    /// where f_i(X) is the i-th chunked polynomial of degree n of f:
```

dannywillems: And it is actually `n - 1` here.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -106,19 +173,35 @@ impl<F: FftField> LagrangeBasisEvaluations<F> {
...
+        // Example:
+        //                                  i-th chunk
+        //                          -----------------------
+        //   chunked: [ 0, ..., 0,  1, x, x^2, ..., x^{m-1}, 0, ..., 0 ]
+        //   indices:   0    i·m-1  i·m            (i+1)m-1  (i+1)m  cm-1=n-1
+        //
+        // A total of n coefficients are returned. These will be helpful to
+        // evaluate the chunks of polynomials of degree c·n at the point x.
+        //
         let n = domain.size();
```
dannywillems, line 205: Note that everywhere in the code, there is no check about if n < max_poly_size (therefore chunks must be equal to 1, and not 0 as it is), and also there is no check when they do not divide each other (might a be a bug you're looking for). I would always add a check when we divide.

dannywillems: And of course, add a test to check if it handled correctly.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+    /// f_i(x) = a_{i*n} · x^0 + ... + a_{(i+1)n-1} · x^{n-1}
+    ///
+    /// Returns the evaluation of each chunk of the polynomial at the point
+    /// (when there is no chunking, the result is a vector of length 1). They
+    /// correspond to the f_i(z) in the equation above.
+    ///
     pub fn evaluate<D: EvaluationDomain<F>>(&self, p: &Evaluations<F, D>) -> Vec<F> {
+        // The domain size must be a multiple of the number of evaluations so
+        // that the degree of the polynomial can be split into chunks of equal size.
         assert_eq!(p.evals.len() % self.evals[0].len(), 0);
```
dannywillems, line 66: It would be nice to have an alias for `self.evals[0].len()` as `self.domain_size()`.
We should also have tests enforcing that each way of constructing a structure of type `LagrangeBasisEvaluations` respect some invariants, like every chunks is of the same size (and therefore, using `self.evals[0].len()` or `self.evals[1].len()` is `self.evals.len() > 1` return the same output).

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+/// support higher degrees. Thus, we cannot obtain a polynomial f with degree
+/// c·n (with domain size n and number of chunks c) as usual with the equation:
+///
+/// f(X) = x_0 · l_0(X) + ... + x_n · l_n(X) + ... + x_{c*n} · l_{c*n}(X)
+///
+/// Instead, this struct will contain the (c*n+1) coefficients of the polynomial
+/// that is equal to the powers of the point x in the positions corresponding to
+/// the chunk, and 0 elsewhere in the domain. This is useful to evaluate the
+/// chunks of polynomials of degree c·n given in evaluation form at the point.
+///
```
dannywillems, line 24:
```suggestion

```

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
 pub struct LagrangeBasisEvaluations<F> {
+    /// If no chunking:
+    /// - evals is a vector of length 1 containing a vector of size n
+    ///   corresponding to the evaluations of the Lagrange polynomials, which
+    ///   are the polynomials that equal 1 at ω_i and 0 elsewhere in the domain.
+    /// If chunking (a vector of size c * N )
+    /// - the first index refers to the chunks
+    /// - the second index refers j-th coefficient of the i-th chunk of the
+    ///   polynomial that equals the powers of the point and 0 elsewhere.
+    ///
```
dannywillems, line 34:
```suggestion

```

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+    /// $ f(x) = x_0 · l_0(x) + ... + x_n · l_n(x) $
+    ///
+    /// Recall that when chunking is in place, we want to evaluate a
+    /// polynomial f of degree c · m at point z, expressed as
+    ///
+    /// f(z) = a_0·z^0 + ... + a_{n-1}·z^{n-1} + ... + a_{c*n-1}·z^{c*n-1}
+    ///      = z^0 · f_0(z) + z^n · f_1(z) + ... + z^{(c-1)n} · f_{c-1}(z)
+    ///
+    /// where f_i(X) is the i-th chunked polynomial of degree m of f:
+    /// f_i(x) = a_{i*n} · x^0 + ... + a_{(i+1)n-1} · x^{n-1}
```
dannywillems, line 57:
```suggestion
    /// ```
    /// f_i(x) = a_{i*n} · x^0 + ... + a_{(i+1)n-1} · x^{n-1}
    /// ```
```

I would use something like this for comments.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+    /// Given the evaluations form of a polynomial, directly evaluate that
+    /// polynomial at a point.
+    ///
+    /// The Lagrange basis evaluations can be used to evaluate a polynomial of
+    /// given in evaluation form efficiently in O(n) time, of degree d <= n
+    /// where n is the domain size, without the need of interpolating it first.
+    ///
+    /// Recall that a polynomial can be represented as the sum of the scaled
+    /// Lagrange polynomials using its evaluations on the domain:
+    /// $ f(x) = x_0 · l_0(x) + ... + x_n · l_n(x) $
```
dannywillems, line 48:
```suggestion
    /// `f(x) = x_0 · l_0(x) + ... + x_n · l_n(x)`
```

It is markdown, not LaTeX.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+/// Evaluations of all normalized lagrange basis polynomials at a given point.
+/// Can be used to evaluate an `Evaluations` form polynomial at that point.
+///
+/// The Lagrange basis for polynomials of degree <= d over a domain
+/// {ω_0,...,ω_d} is the set of (d+1) polynomials ${l_0,...,l_d}$ of degree d
+/// that equal 1 at ω_i and 0 in the rest of the domain terms. They can be used
+/// to evaluate polynomials in evaluation form efficiently in O(d) time.
+///
+/// When chunking is in place, the domain size is larger than the maximum
+/// polynomial degree allowed. In particular, the SRS is not long enough to
```
dannywillems, line 14:
```suggestion
/// polynomial degree allowed. In particular, the SRS (used by the polynomial commitment to commit to) is not long enough to
```

or something like this, to just remind that the SRS (nit - we should stop using the term SRS as it is always "structured") is used by the PCS, and the domain size is related to the field we use to evaluate the polynomials at - the evaluation domain size is unrelated to the SRS size.

kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -2,18 +2,85 @@ use ark_ff::{batch_inversion_and_mul, FftField};
...
+/// that is equal to the powers of the point x in the positions corresponding to
+/// the chunk, and 0 elsewhere in the domain. This is useful to evaluate the
+/// chunks of polynomials of degree c·n given in evaluation form at the point.
+///
 pub struct LagrangeBasisEvaluations<F> {
+    /// If no chunking:
+    /// - evals is a vector of length 1 containing a vector of size n
+    ///   corresponding to the evaluations of the Lagrange polynomials, which
+    ///   are the polynomials that equal 1 at ω_i and 0 elsewhere in the domain.
+    /// If chunking (a vector of size c * N )
```
dannywillems, line 30:
```suggestion
    /// If chunking (a vector of size c * N)
```

dannywillems (approved):
It is blocking PRs built on top of this one as it is not built on top of latest master with the latest CI configuration. Accepting it and merging as it is a no-op. Comments can be addressed in subsequent PRs.

kimchi/src/lagrange_basis_evaluations.rs
querolita: used m standing for maximum degree, given that in the function `new_with_chunked_segments` they use `n` to refer to the domain size

kimchi/src/lagrange_basis_evaluations.rs
querolita: Actually the check that n < max_poly_size is performed here -> https://github.com/o1-labs/proof-systems/blob/fbf0572a456b7a6ced34dd493f62562d220d1274/kimchi/src/lagrange_basis_evaluations.rs#L243

PR #2527, opened 2024-09-04
LagrangeBasisEvaluations: make tests more random regarding the domain size

dannywillems:


kimchi/src/lagrange_basis_evaluations.rs
```diff
@@ -240,7 +245,6 @@ mod tests {
                     Fp::zero()
                 });
             }
-            e = vec![Fp::zero(), Fp::one()];
```
querolita, line 243: oh, I wonder why this was here... Basically this was reverting the lines above 😞

querolita (approved):

PR #2531, opened 2024-09-05
MVPoly: add_monomial in trait

dannywillems:


marcbeunardeau88 (approved):
That one feels weird in the trait, but no strong opinion against that, maybe we'll remove later

PR #2536, opened 2024-09-05
MVPoly: method to generate random multilinear polynomial

dannywillems:


mvpoly/src/lib.rs
```diff
@@ -59,4 +61,60 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
     /// This is a dummy implementation. A cache can be used for the monomials to
     /// speed up the computation.
     fn eval(&self, x: &[F; N]) -> F;
+
+    /// Returns true if the polynomial is homogeneous (of degree `D`).
+    /// As a reminder, a polynomial is homogeneous if all its monomials have the
+    /// same degree.
```
Fizzixnerd, line 67: TIL I was confusing homogeneous ODEs with homogeneous polynomials.  I thought homogeneous just mean didn't have a constant term!

dannywillems:
Leaving like this for now, not useful. But leaving open in case we move to multilinear PCS

PR #2540, opened 2024-09-06
[DO NOT MERGE] MVPoly: make tests generic over the MVPoly trait

dannywillems:

PR #2541, opened 2024-09-08
MVPoly tests generalization based on example.

shimkiv:
#### Examples provided:

- https://github.com/o1-labs/proof-systems/pull/2540

#### Notes

- It was required for generic tests to extend the `MVPoly` trait with the:

  ```rust
  fn modify_monomial_with_scalar(&mut self, scalar: F);
  ```

  And implement it for `Sparse` and `Dense` structs.
  Not sure though if it makes sense to do for `MVPoly` or maybe we should do it with some helper trait instead.

#### Tasks left:
- [ ] Generalize the `test_homogeneous_eval` tests.
- [ ] Generalize the `test_add_monomial` tests.

### DIFF
--- a/mvpoly/src/lib.rs
+++ b/mvpoly/src/lib.rs
@@ -5,6 +5,7 @@ use kimchi::circuits::expr::{ConstantExpr, Expr};
 use rand::RngCore;
 
 pub mod monomials;
+pub mod pbt;
 pub mod prime;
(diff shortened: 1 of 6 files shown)


dannywillems (approved):
Pre-approving, in case you want to add the change suggested in the Slack conversation later, and not in this PR.

PR #2394, opened 2024-07-02
Speed up poseidon for kimchi by hand-unrolling

mrmr1993:
Running an experiment using the Mina code, I see a ~50% speed-up in poseidon by hand-unrolling the `x^7` sbox vs using the built-in `pow` function.

PR #2558, opened 2024-09-16
Manually unroll MDS calculation

georgeee:
Modify `full_round` to perform round inline when `params.mds.len() == 3`.

This optimization gave a ~3% speedup (measured by `mina ledger test apply` on top of a PR stack #2394, https://github.com/MinaProtocol/mina/pull/15980 and _TODO open a PR_.

Although 3% isn't much, I expect speed-up to be better when a batch version of hash (https://github.com/MinaProtocol/mina/issues/16053) is delivered.

P.S. change's diff is bloated by `^M` characters present at the end of each line for the previous version of the `poseidon/src/permutation.rs` file.

### DIFF
--- a/poseidon/src/permutation.rs
+++ b/poseidon/src/permutation.rs
@@ -33,12 +33,34 @@ pub fn full_round<F: Field, SC: SpongeConstants>(
     state: &mut Vec<F>,
     r: usize,
 ) {
-    for state_i in state.iter_mut() {
-        *state_i = sbox::<F, SC>(*state_i);
-    }
-    *state = apply_mds_matrix::<F, SC>(params, state);
-    for (i, x) in params.round_constants[r].iter().enumerate() {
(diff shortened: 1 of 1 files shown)


marcbeunardeau88 (approved):

dannywillems:
As discussed with @georgeee online, we can close this for now.

georgeee:
> let's close this one for now. Initially I thought it would be useful. But now I'm not that sure. For the work I was doing I developed a quite different version of this computation (relying on parallelism). So the usage against which I was seeing 3% speed-up (i.e. ocaml bindings) will end up not involving the optimized call. And I have no idea whether this optimization would still make sense in other contexts (like WASM) or is likely to worsen things.
> 3% is also not that convincing. I thought there will be a follow-up with more increases, but it turned out to be a futile endeavor.

PR #2559, opened 2024-09-17
Arrabiata: interface for "next row" and keep "next state" in the env

dannywillems:


arrabiata/src/witness.rs
```diff
@@ -203,6 +208,13 @@ where
         pos
     }
 
+    fn access_next_row(&self, pos: Self::Position) -> Self::Variable {
+        let Column::X(idx) = pos else {
+            unimplemented!("Only works for private inputs")
+        };
+        self.witness[idx][self.current_row + 1].clone()
```
dannywillems, line 215: This is changed in a future PR in `self.next_state[idx].clone()`.

arrabiata/src/constraints.rs
```diff
@@ -55,6 +55,13 @@ impl<Fp: PrimeField> InterpreterEnv for Env<Fp> {
         pos
     }
 
+    fn access_next_row(&self, pos: Self::Position) -> Self::Variable {
```
dannywillems, line 58: In a future PR, I will add the "row" (i.e. Curr or Next) in the type `Self::Position`. It will simplify this code.

marcbeunardeau88 (approved):

PR #2562, opened 2024-09-17
Arrabiata: implement PoseidonNextRow - a variant of the gadget Poseidon using "next row"

dannywillems:
Note: we will certainly deprecate other versions later. Only leaving it for now.
Constraints checks come right after.

Fix https://github.com/o1-labs/proof-systems/issues/2565

### DIFF
--- a/arrabiata/src/interpreter.rs
+++ b/arrabiata/src/interpreter.rs
@@ -401,6 +401,10 @@ pub trait InterpreterEnv {
     /// Return the corresponding variable at the given position, for the next row.
     fn access_next_row(&self, pos: Self::Position) -> Self::Variable;
 
+    /// Return the corresponding variable at the given position, for the current
+    /// row.
+    fn access_current_row(&self, pos: Self::Position) -> Self::Variable;
+
     fn allocate_public_input(&mut self) -> Self::Position;
 
     /// Set the value of the variable at the given position for the current row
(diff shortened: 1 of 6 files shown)


arrabiata/src/interpreter.rs
```diff
@@ -401,6 +408,10 @@ pub trait InterpreterEnv {
     /// Return the corresponding variable at the given position, for the next row.
     fn access_next_row(&self, pos: Self::Position) -> Self::Variable;
 
+    /// Return the corresponding variable at the given position, for the current
+    /// row.
+    fn access_current_row(&self, pos: Self::Position) -> Self::Variable;
```
dannywillems, line 413: Not fully happy with the naming. Might change in a follow-up PR.

arrabiata/src/interpreter.rs
```diff
@@ -973,6 +1026,108 @@ pub fn run_ivc<E: InterpreterEnv>(env: &mut E, instr: Instruction) {
...
         }
+        Instruction::PoseidonNextRow(curr_round) => {
+            env.activate_gadget(Gadget::PoseidonNextRow);
+            debug!("Executing instruction PoseidonNextRow({curr_round})");
+            if curr_round < POSEIDON_ROUNDS_FULL {
+                // Values to be absorbed are 0 when when the round is not zero,
+                // i.e. when we are processing the rounds.
+                let values_to_absorb: Vec<E::Variable> = (0..POSEIDON_STATE_SIZE - 1)
+                    .map(|_i| {
+                        let pos = env.allocate_public_input();
```
marcbeunardeau88, line 1037: I see that in this use case we absorb PI, but this is not always the case (eg. the internal node of a merkle tree)
Can we not harcode the fact that we absorb PI ?

marcbeunardeau88 (approved):

PR #2566, opened 2024-09-17
Expr.rs can be parametrized by challenge

marcbeunardeau88:
The expression Framework harcoded the plonk + plookup challenges.
This MR allows an expression to be parametrised by different challenges.
The old harcoded challenges are renamed BerkeleyChallenges/BerkeleyChallengterm, and are used in pre-existing code.
This MR is usefull for Arriabatta, to add the folding challenge in expressions.


kimchi/src/circuits/expr.rs
```diff
@@ -51,15 +51,36 @@ pub enum ExprError<Column> {
     MissingRuntime,
 }
 
-pub struct Challenges<F> {
+pub struct BerkeleyChallenges<F> {
```
dannywillems, line 54: Note: for a follow-up, moving this in `berkeley_columns.rs`

dannywillems:
And https://github.com/o1-labs/proof-systems/issues/2556

kimchi/src/circuits/expr.rs
```diff
@@ -154,13 +181,32 @@ pub struct Variable<Column> {
...
+        use BerkeleyChallengeTerm::*;
+        let str = match self {
+            Alpha => "alpha".to_string(),
+            Beta => "beta".to_string(),
+            Gamma => "gamma".to_string(),
+            JointCombiner => "joint_combiner".to_string(),
+        };
+        write!(f, "{}", str)
+
+        // Debug::fmt(&self, f)
```
dannywillems, line 202: Can be removed.

kimchi/src/circuits/expr.rs
```diff
@@ -3152,13 +3258,13 @@ pub mod constraints {
     /// This trait defines a common arithmetic operations interface
     /// that can be used by constraints.  It allows us to reuse
     /// constraint code for witness computation.
-    pub trait ExprOps<F>:
```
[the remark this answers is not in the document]
dannywillems: Oh, I see. It is used for Keccak and FF arithmetic in Kimchi. I would definitely specialized it only for Berkeley for now.

kimchi/src/circuits/expr.rs
```diff
@@ -3206,35 +3312,40 @@ pub mod constraints {
...
 
         /// Cache item
         fn cache(&self, cache: &mut Cache) -> Self;
     }
-
-    impl<F> ExprOps<F> for Expr<ConstantExpr<F>, berkeley_columns::Column>
+    // TODO generalize with generic Column/challengeterm
+    // We need to create a trait for berkeley_columns::Environment
+    impl<F> ExprOps<F, BerkeleyChallengeTerm>
+        for Expr<ConstantExpr<F, BerkeleyChallengeTerm>, berkeley_columns::Column>
```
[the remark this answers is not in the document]
dannywillems: If we want to keep it, we can simply implement it generically above, around line 1838.

kimchi/src/circuits/expr.rs
```diff
@@ -1761,15 +1827,22 @@ impl<F: FftField, Column: Copy> Expr<ConstantExpr<F>, Column> {
             }
         }
     }
-
+}
+impl<F: FftField, Column: Copy> Expr<ConstantExpr<F, BerkeleyChallengeTerm>, Column> {
     /// The expression `beta`.
     pub fn beta() -> Self {
```
dannywillems, line 1833: Even though it is out of the scope of this PR first, I would get rid of this alias. It is only used in `lookups/constraints.rs`, where `berkeley_columns::E` is used, and `E` is already an alias for the expressions used by Berkeley. It would clean the code.

kimchi/src/verifier.rs
```diff
@@ -886,11 +889,15 @@ where
...
+            let challenges = BerkeleyChallenges {
                 alpha: oracles.alpha,
                 beta: oracles.beta,
                 gamma: oracles.gamma,
-                joint_combiner: oracles.joint_combiner.as_ref().map(|j| j.1),
+                joint_combiner: oracles
+                    .joint_combiner
+                    .as_ref()
+                    .map(|j| j.1)
+                    .unwrap_or(G::ScalarField::zero()),
```
dannywillems, line 900: Even though it seems ok at a first sight, I do not remember if we rely on the fact the joint combiner is None somewhere.

kimchi/src/circuits/expr.rs
```diff
@@ -51,15 +51,36 @@ pub enum ExprError<Column> {
...
-pub struct Challenges<F> {
+pub struct BerkeleyChallenges<F> {
     /// The challenge alpha from the PLONK IOP.
     pub alpha: F,
     /// The challenge beta from the PLONK IOP.
     pub beta: F,
     /// The challenge gamma from the PLONK IOP.
     pub gamma: F,
     /// The challenge joint_combiner which is used to combine joint lookup tables.
-    pub joint_combiner: Option<F>,
```
dannywillems, line 62: Sometimes, the joint combiner is non-existent, for instance when we do have a single column.

dannywillems (approved):

kimchi/src/circuits/expr.rs
marcbeunardeau88: see previous comment, it's ok to make that change
It is also necessarry to index the challenge term

kimchi/src/circuits/expr.rs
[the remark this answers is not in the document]
marcbeunardeau88: Don't we want to have a general interface which exposes the basic constraints ?

Issue #2545, opened 2024-09-09
use generator style


marcbeunardeau88:
We need to rewrite the code in ‘generator’ style, roughly `generate_next_proof : state -> (state, proof)`.
Then we can decide how many steps can be executed in proof based on the domain size.

Issue #2543, opened 2024-09-09
Add Selectors


marcbeunardeau88:
The current desgin of the o1vm is one circuit per MIPS instructions.
We want to change that and get one big circuit which does all the instruction depending on which selector is activated.

We need to add ‘selectors’ for (see [kimchi](https://github.com/o1-labs/proof-systems/blob/997847b767ecc0e9a48d1a9ae521d59e8bc7671c/kimchi/src/prover.rs#L654) and/or [the original zkVM prototype](https://github.com/o1-labs/mips-demo), the plonk paper for the concept), and multiplex the instructions using those. The [original zkVM](https://github.com/o1-labs/mips-demo/blob/70c93219a826028abe519f20e6698c07d0cafd24/kimchi/src/mips/witness.rs#L576) is probably the best guide for this, this is how it originally worked.

Fizzixnerd:
CC @dannywillems @marcbeunardeau88 

I was wondering: do we want one giant selector `q_MIPS` which serves as essentially the opcode for MIPS, or several smaller selectors `q_leftshift` and friends that are in the range `{0, 1}`.

Or, is the selector something more clever than what I'm initially thinking of?  For example, it might be possible to construct `Z` such that

```Z(g^i, x) = MIPS_instruction(i)(x)```

By using the same trick in Section 5 of PlonK.

Fizzixnerd:
Closed in #2671

PR #2567, opened 2024-09-18
Initial selectorified API

Fizzixnerd:
I have here a skeleton API for a selectorified API.  This begins to address #2543 and #2545 (Perhaps?  Comments?) by switching from the API used in [`o1vm/src/mips/interpreter.rs`](./o1vm/src/mips/interpreter.rs) to that detailed in [`o1vmsrc/mips/singlestep.rs`](./o1vm/src/mips/singlestep.rs).

The main changes are:

* No use of mutability, instead using the "generator style" of `a -> State -> (b, State)`
* Significantly pared down API, will likely add to it as I realize it was needed, but for now start small!
* Instructions now "know" their arguments, by packing them in the `enum` fields with them.  This allows the generator style to uniformly look like `Instructions -> State -> (Witness, State)` and for witness generation, without having to worry about different arities and register arguments. [Of course, the old instructions with their arguments can be relatively easily converted to this form -- to come].
* A small example of the API in use is below the API with a couple of test curves and an example instruction-set based on a significantly simplified MIPS instruction set, using just one of each of a small class of instructions, for parallel debugging and sanity-checking purposes.
* Simplified register set, again for debugging and sanity-checking
* Simplified `ExampleS` previously an implementor of `InterpreterEnv` (`Env` in `witness.rs` and the like) which will almost surely grow as the number of instructions returns to parity.

The point of these changes is to provide a basic test-bed for the selector changes to come.

### DIFF
--- a/o1vm/src/mips/singlestep.rs
+++ b/o1vm/src/mips/singlestep.rs
@@ -1,0 +1,116 @@
+// Amalgamation of the different MIPS circuits into a "single" circuit.
+
+use super::column;
+
+pub trait SingleStepper {
+    type S<I, P, V>;
+
+    type Position;
+
+    type Instruction;
+
+    type Variable: Clone
+        + std::ops::Add<Self::Variable, Output = Self::Variable>
+        + std::ops::Sub<Self::Variable, Output = Self::Variable>
+        + std::ops::Mul<Self::Variable, Output = Self::Variable>
+        + std::fmt::Debug
+        + ark_ff::Zero
+        + ark_ff::One;
+
+    type State = S<Self::Instruction, Self::Position, Self::Variable>;
+
+    /// Helpers
+
+    fn alloc_scratch(state: Self::State) -> (Self::Position, Self::State);
+
+    fn variable(state: &Self::State, column: Self::Position) -> Self::Variable;
+
+    fn add_constraint(state: Self::State, assert_equals_zero: Self::Variable) -> Self::State;
+
+    // ABORTS
+    fn check_is_zero(assert_equals_zero: &Self::Variable);
+
+    fn assert_is_zero(state: Self::State, assert_equals_zero: Self::Variable) -> Self::State {
+        Self::check_is_zero(&assert_equals_zero);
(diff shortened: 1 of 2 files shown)


marcbeunardeau88 (commented):
I understand that you are re-creating from scratch
Shouldn't we rather build on top of https://github.com/o1-labs/proof-systems/blob/master/o1vm/src/mips/interpreter.rs ?
I would have expected a new function in interpreter similar to https://github.com/o1-labs/proof-systems/blob/master/o1vm/src/mips/interpreter.rs#L987C8-L987C23
But something done differently here : https://github.com/o1-labs/proof-systems/blob/master/o1vm/src/mips/interpreter.rs#L1047

o1vm/src/mips/singlestep.rs
```diff
@@ -0,0 +1,116 @@
...
+    D,
+    Lo,
+    Hi,
+    Ip,
+    NextIp,
+}
+
+pub type ExamplePosition = column::ColumnAlias;
+
+pub type ExampleVariable = ark_bn254::Bn254;
```
marcbeunardeau88, line 84: I understand that variables are a way to index to positions in the plonk matrix.
Therfore, I would use something like unsigned integer, but not a crypto related object

dannywillems:
As discussed in PM, I would simply use an "iter" on the instructions with the constraint environment, multiplying by a selector, adding constraints on them on the way.
The selectors are added to a witness structure/proof structure. I'm not sure that we need a new trait.

dannywillems:
C.f Arrabiata

o1vm/src/mips/singlestep.rs
Fizzixnerd: Yes, I changed to unsigned integer in the latest version.

Regarding the from-scratch approach shown here, this is mostly a playground where I can test out ideas which I will then port back to the regular interpreter, I think, without the full complexity of the regular interpreter.  It didn't start out as that, mind you!

But I think you and Danny are right.

PR #2588, opened 2024-09-20
Update arkworks to 0.4.2 (master)

volhovm:
Main part of update (towards `develop` = mina's `compatible`) has been done here: https://github.com/MinaProtocol/mina/pull/15940

This upgrades `master` too. It's non-trivial since `master` has a lot of stuff `develop` doesn't.


o1vm/src/legacy/tests.rs
```diff
@@ -145,33 +148,33 @@ pub mod mips {
                 RType(rtype) => match rtype {
                     JumpRegister | SyscallExitGroup | Sync => assert_num_constraints(&instr, 1),
                     ShiftLeftLogical
-                        | ShiftRightLogical
```
dannywillems, line 148: Ah? Why this change? It is syntactically only, but weird to have this here.

dannywillems:
Regarding commit messages, I am against having something like:
> [Fix serialization? (what? why?)](https://github.com/o1-labs/proof-systems/pull/2588/commits/e0d92f468ff4c485c76c0dd31203bddf0d6a18ae)

Could you change to something more descriptive (by answering the questions, for instance, in the commit message)?

msm/src/fec/mod.rs
```diff
@@ -35,14 +38,17 @@ mod tests {
         rng: &mut RNG,
         domain_size: usize,
     ) -> FECWitnessBuilderEnv {
-        use mina_curves::pasta::{Fp, Pallas};
+        // NOTE: this uses Pallas-in-Pallas emulation.
+        use mina_curves::pasta::Pallas;
+        type Fp = <Pallas as AffineRepr>::ScalarField;
```
dannywillems, line 43: The scalar field of Pallas is `Fq`: https://github.com/o1-labs/proof-systems/blob/master/curves/src/pasta/curves/pallas.rs#L13

msm/src/serialization/interpreter.rs
```diff
@@ -802,10 +802,9 @@ mod tests {
             lookups::LookupTable,
             N_INTERMEDIATE_LIMBS,
         },
-        Ff1, LIMB_BITSIZE, N_LIMBS,
+        Ff1, Fp, LIMB_BITSIZE, N_LIMBS,
```
dannywillems, line 805: Not that this is not the same field. `Fp` points to `ark_bn254::Fp` in `msm/src/lib.rs`. Is it expected to change the field?

o1vm/src/legacy/tests.rs
```diff
@@ -475,6 +479,8 @@ pub mod keccak {
     // (Step, Left, Right)
     type KeccakFoldingPair = (Steps, KeccakFoldingSide, KeccakFoldingSide);
 
+    type KeccakDefaultFqSponge = DefaultFqSponge<ark_bn254::g1::Config, PlonkSpongeConstantsKimchi>;
```
dannywillems, line 482: Just a note: seems an additional alias added. LGTM.

poly-commitment/src/commitment.rs
```diff
@@ -232,7 +230,7 @@ impl<'a, 'b, C: AffineCurve> Add<&'a PolyComm<C>> for &'b PolyComm<C> {
     }
 }
 
-impl<'a, 'b, C: AffineCurve> Sub<&'a PolyComm<C>> for &'b PolyComm<C> {
+impl<'a, 'b, C: AffineRepr + Sub<Output = C::Group>> Sub<&'a PolyComm<C>> for &'b PolyComm<C> {
```
dannywillems, line 233: LGTM, even though I am surprised you do have to add this additional trait. Can we not get rid of it?

poly-commitment/src/kzg.rs
```diff
@@ -401,25 +402,35 @@ impl<
...
         ];
         // the result here is numerator_commitment * 1 - quotient * divisor_commitment
         // Note that the unwrap cannot fail as the output of a miller loop is non zero
-        let res = Pair::final_exponentiation(&(Pair::miller_loop(&to_loop))).unwrap();
-
-        res == Pair::Fqk::one()
+        let res = Pair::final_exponentiation(Pair::multi_miller_loop(to_loop_left, to_loop_right))
+            .unwrap();
+
+        res.0 == Pair::TargetField::one()
```
dannywillems, line 423: Checking this.

poly-commitment/src/kzg.rs
```diff
@@ -401,25 +402,35 @@ impl<
...
+
+        // @VOLHOVM remove this if not necessary
```
dannywillems, line 425: Is it a leftover from your tests?

poseidon/export_test_vectors/Cargo.toml
```diff
@@ -10,11 +10,11 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-ark-ff = { version = "0.3.0", features = [ "parallel", "asm" ] }
```
dannywillems, line 13: Could we also use workspace?

poseidon/src/sponge.rs
```diff
@@ -108,10 +106,10 @@ pub struct DefaultFrSponge<Fr: Field, SC: SpongeConstants> {
 }
 
 fn pack<B: BigInteger>(limbs_lsb: &[u64]) -> B {
-    let mut res: B = 0.into();
+    let mut res: B = 0u64.into();
     for &x in limbs_lsb.iter().rev() {
         res.muln(64);
-        res.add_nocarry(&x.into());
```
dannywillems, line 114: Is it not contradictory?

poly-commitment/src/kzg.rs
dannywillems: You can use `res.is_zero()`, see
```
impl<P: Pairing> Zero for PairingOutput<P> {
    /// The identity element, or "zero", of the group is the identity element of the multiplicative group of the underlying field, i.e., `P::TargetField::one()`.
    fn zero() -> Self {
        Self(P::TargetField::one())
    }

    fn is_zero(&self) -> bool {
        self.0.is_one()
    }
}
```

It is better to rely on the interface provided by the library than the inner representation (c.f. affine vs projective coordinates where we can reference `x` and `y` in both structures).

poly-commitment/src/kzg.rs
```diff
@@ -401,25 +402,35 @@ impl<
...
+            // Note that we do a neagtion here, to put everything on the same side
+            ark_ec::pairing::prepare_g1::<Pair>(self.quotient.into_group().neg()),
+        ];
+        let to_loop_right = [
+            ark_ec::pairing::prepare_g2::<Pair>(Pair::G2Affine::generator()),
+            ark_ec::pairing::prepare_g2::<Pair>(divisor_commitment),
         ];
         // the result here is numerator_commitment * 1 - quotient * divisor_commitment
         // Note that the unwrap cannot fail as the output of a miller loop is non zero
-        let res = Pair::final_exponentiation(&(Pair::miller_loop(&to_loop))).unwrap();
```
dannywillems, line 421: Note that there is now the pairing `multi_pairing`, for exactly the same code.

poly-commitment/tests/kzg.rs
```diff
@@ -112,9 +112,10 @@ fn check_srs_g2_valid_and_serializes() {
 
         // Check it serializes well
         let actual_y: BN254G2BaseField = actual.y;
-        let res = actual_y.serialize(vec.as_mut_slice());
+        let res = actual_y.serialize_compressed(vec.as_mut_slice());
```
dannywillems, line 115: I think it would be nice to have regression tests for these changes.

poseidon/export_test_vectors/src/vectors.rs
```diff
@@ -89,8 +90,8 @@ pub fn generate(mode: Mode, param_type: ParamType) -> TestVectors {
             .collect();
         let mut output_bytes = vec![];
         output
-            .into_repr()
-            .serialize(&mut output_bytes)
+            .0
```
dannywillems, line 93: I would try to rely on a method on `output` instead of getting the inner element of the structure.
Also, simply adding a single test checking the serialization is still the same would be nice - it is used for further testing, not production code, but always good.

poseidon/src/sponge.rs
dannywillems: LGTM after checking the doc... Weird.

utils/src/field_helpers.rs
```diff
@@ -122,26 +122,27 @@ pub trait FieldHelpers<F> {
...
     {
-        BigUint::from_bytes_le(&F::Params::MODULUS.to_bytes_le())
+        BigUint::from_bytes_le(&F::MODULUS.to_bytes_le())
     }
 }
 
 impl<F: Field> FieldHelpers<F> for F {
     fn from_bytes(bytes: &[u8]) -> Result<F> {
-        F::deserialize(&mut &*bytes).map_err(|_| FieldHelpersError::DeserializeBytes)
+        F::deserialize_uncompressed(&mut &*bytes).map_err(|_| FieldHelpersError::DeserializeBytes)
```
dannywillems, line 139: In arkworks.0.3.0, we have:
```rust
pub trait CanonicalDeserialize: Sized {
    /// Reads `Self` from `reader`.
    fn deserialize<R: Read>(reader: R) -> Result<Self, SerializationError>;

    /// Reads `Self` from `reader` without compression.
    #[inline]
    fn deserialize_uncompressed<R: Read>(reader: R) -> Result<Self, SerializationError> {
        Self::deserialize(reader)
    }

    /// Reads `self` from `reader` without compression, and without performing
    /// validity checks. Should be used *only* when the input is trusted.
    #[inline]
    fn deserialize_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
        Self::deserialize_uncompressed(reader)
    }
}
```
Then, `deserialize` is exactly `deserialize_uncompressed`. it seems to perform the check at the same time, based on the documentation.
In arkworks 0.4.2, we have
```rust
pub trait CanonicalDeserialize: Valid {
    /// The general deserialize method that takes in customization flags.
    fn deserialize_with_mode<R: Read>(
        reader: R,
        compress: Compress,
        validate: Validate,
    ) -> Result<Self, SerializationError>;

    fn deserialize_compressed<R: Read>(reader: R) -> Result<Self, SerializationError> {
        Self::deserialize_with_mode(reader, Compress::Yes, Validate::Yes)
    }

    fn deserialize_compressed_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
        Self::deserialize_with_mode(reader, Compress::Yes, Validate::No)
    }

    fn deserialize_uncompressed<R: Read>(reader: R) -> Result<Self, SerializationError> {
        Self::deserialize_with_mode(reader, Compress::No, Validate::Yes)
    }

    fn deserialize_uncompressed_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
        Self::deserialize_with_mode(reader, Compress::No, Validate::No)
    }
}
```
Which seems to be the expected behavior, as we would have called `deserialize_uncompressed_unchecked` if we did not want to validate.

utils/src/serialization.rs
```diff
@@ -23,7 +23,7 @@ pub mod ser {
         S: serde::Serializer,
     {
         let mut bytes = vec![];
-        val.serialize(&mut bytes)
+        val.serialize_compressed(&mut bytes)
```
dannywillems, line 26: This doesn't seem right.
In arkworks 0.3.0:
```rust
pub trait CanonicalSerialize {
    /// Serializes `self` into `writer`.
    /// It is left up to a particular type for how it strikes the
    /// serialization efficiency vs compression tradeoff.
    /// For standard types (e.g. `bool`, lengths, etc.) typically an uncompressed
    /// form is used, whereas for algebraic types compressed forms are used.
    ///
    /// Particular examples of interest:
    /// `bool` - 1 byte encoding
    /// uints - Direct encoding
    /// Length prefixing (for any container implemented by default) - 8 byte encoding
    /// Elliptic curves - compressed point encoding
    fn serialize<W: Write>(&self, writer: W) -> Result<(), SerializationError>;

    fn serialized_size(&self) -> usize;

    /// Serializes `self` into `writer` without compression.
    #[inline]
    fn serialize_uncompressed<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
        self.serialize(writer)
    }

    /// Serializes `self` into `writer` without compression, and without
    /// performing validity checks. Should be used *only* when there is no
    /// danger of adversarial manipulation of the output.
    #[inline]
    fn serialize_unchecked<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
        self.serialize_uncompressed(writer)
    }

    #[inline]
    fn uncompressed_size(&self) -> usize {
        self.serialized_size()
    }
}
```
And in arkworks 0.4.2:
```rust
pub trait CanonicalSerialize {
    /// The general serialize method that takes in customization flags.
    fn serialize_with_mode<W: Write>(
        &self,
        writer: W,
        compress: Compress,
    ) -> Result<(), SerializationError>;

    fn serialized_size(&self, compress: Compress) -> usize;

    fn serialize_compressed<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
        self.serialize_with_mode(writer, Compress::Yes)
    }

    fn compressed_size(&self) -> usize {
        self.serialized_size(Compress::Yes)
    }

    fn serialize_uncompressed<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
        self.serialize_with_mode(writer, Compress::No)
    }

    fn uncompressed_size(&self) -> usize {
        self.serialized_size(Compress::No)
    }
}
```

dannywillems: Ah well. As the doc says:
```
    /// It is left up to a particular type for how it strikes the
    /// serialization efficiency vs compression tradeoff.
```
I would guess the implementation was left to the user, and we did have the compressed version by default?

utils/src/serialization.rs
```diff
@@ -100,7 +100,7 @@ where
         S: serde::Serializer,
     {
         let mut bytes = vec![];
-        val.serialize_unchecked(&mut bytes)
+        val.serialize_uncompressed(&mut bytes)
```
dannywillems, line 103: Are we sure it doesn't check? It would imply a performance regression otherwise. I would be happy to see some regression tests for this.

utils/src/serialization.rs
```diff
@@ -124,6 +124,6 @@ where
         } else {
             Bytes::deserialize_as(deserializer)?
         };
-        T::deserialize_unchecked(&mut &bytes[..]).map_err(serde::de::Error::custom)
+        T::deserialize_uncompressed_unchecked(&mut &bytes[..]).map_err(serde::de::Error::custom)
```
dannywillems, line 127: Same than above.

utils/tests/field_helpers.rs
```diff
@@ -125,7 +128,7 @@ fn field_big() {
     let field_zero = BaseField::from(0u32);
 
     assert_eq!(
-        BigUint::from_bytes_be(&field_zero.into_repr().to_bytes_be()),
+        BigUint::from_bytes_be(&field_zero.0.to_bytes_be()),
```
dannywillems, line 131: Can we use a method on it instead of getting the field?

utils/src/field_helpers.rs
```diff
@@ -122,26 +122,27 @@ pub trait FieldHelpers<F> {
...
     }
 
     fn from_hex(hex: &str) -> Result<F> {
         let bytes: Vec<u8> = hex::decode(hex).map_err(|_| FieldHelpersError::DecodeHex)?;
-        F::deserialize(&mut &bytes[..]).map_err(|_| FieldHelpersError::DeserializeBytes)
+        F::deserialize_uncompressed(&mut &bytes[..])
```
dannywillems, line 144: As always, I would be happy to have some regression tests. One test could be checking that deserialization/serialization on Fp::MODULUS + 1 works/doesn't work on both version. Same for the serialization. And also have some PBT (serialization followed by deserialization gives the same value).

utils/src/serialization.rs
dannywillems: I would be happy if we could have some PBT tests regarding the serialization/deserialization, like my comment mentions above. I would be also in favor of having quick unit tests checking if there is any regression tests between both, including quick benchmark.

dannywillems (changes requested):
See comment.

circuit-construction/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
```
dannywillems, line 1: As said for develop, I do not know if this is used.
Let's keep it, but we should see if we can remove it.
Also, in a follow-up PR, I would use the workspace everywhere in this Cargo.toml file.

groupmap/src/lib.rs
```diff
@@ -38,12 +38,13 @@ pub struct BWParameters<G: SWModelParameters> {
...
 
 /// returns the right-hand side of the Short Weierstrass curve equation for a given x
-fn curve_eqn<G: SWModelParameters>(x: G::BaseField) -> G::BaseField {
+fn curve_eqn<G: SWCurveConfig>(x: G::BaseField) -> G::BaseField {
     let mut res = x;
     res *= &x; // x^2
     res += &G::COEFF_A; // x^2 + A
     res *= &x; // x^3 + A x
     res += &G::COEFF_B; // x^3 + A x + B
+
```
dannywillems, line 47: Fine, but use less new line :).

utils/tests/field_helpers.rs
volhovm: Not only we can but we should, since now that I'm looking at it (it was some code written before me) the serialization is not correct. It should be `PrimeField::into_bigint()` instead of `.0`, and these two calls give different bigints (surprisingly!). I'll add a fix here + hotfix on `develop` + regression test.

circuit-construction/Cargo.toml
volhovm: Will fix the workspace. This workspace was not unified in `master`, and `master` does not yet have the changes of `develop` where /all/ the workspaces are unified.

volhovm (commented):
Thanks for the review! 
* I'll fix the workspace issues and remove the unnecessary comment
* Will add a hotfix for `export_test_vectors` on `develop` and cherry-pick it here too.
* All the other comments especially w.r.t. serialization and regression tests I suggest to address within a new PR on `develop`. They're reasonable, but arguably regression tests are not part of moving arkworks 0.4.2 to master in particular.

msm/src/serialization/interpreter.rs
volhovm: OK indeed you're right. I have no idea why this module uses `native = foreign` in tests instead, but I returned it back/made it clear now.

volhovm: quite confusing actually, I spent some time thinking about the same issue in the other file :)

msm/src/fec/mod.rs
volhovm: Renamed :+1:

o1vm/src/legacy/tests.rs
dannywillems: Ah no, it was mostly a comment.

arrabiata/tests/witness.rs
```diff
@@ -291,8 +292,9 @@ where
     let res_y: BigInt = env.state[1].clone();
 
     let p1_proj: ProjectivePallas = p1.into();
-    let p1_r: Pallas = p1_proj.mul(r.clone().to_u64_digits().1).into();
-    let exp_res: Pallas = p1_r + env.srs_e2.h;
+    // @volhovm TODO check if mul_bigint is what was intended
```
dannywillems, line 295: Yes it is fine. And it is a test.
I'll remove in a follow-up

circuit-construction/Cargo.toml
dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/2607

poly-commitment/src/commitment.rs
dannywillems: It is better to leave it like than going into projective coordinates and come back.

poly-commitment/src/commitment.rs
```diff
@@ -241,7 +239,7 @@ impl<'a, 'b, C: AffineCurve> Sub<&'a PolyComm<C>> for &'b PolyComm<C> {
         let n2 = other.elems.len();
         for i in 0..std::cmp::max(n1, n2) {
             let pt = if i < n1 && i < n2 {
-                self.elems[i] + (-other.elems[i])
+                (self.elems[i] - other.elems[i]).into_affine()
```
dannywillems, line 242: Is it not less efficient? IIUC, it goes into projective coordinates and come back into affine, performing an inverse for each element?

poly-commitment/src/commitment.rs
dannywillems: It seems that before, there were simply doing in affine coordinates: https://github.com/arkworks-rs/algebra/blob/v0.3.0/ec/src/models/short_weierstrass_jacobian.rs#L246.

poly-commitment/src/commitment.rs
dannywillems: See

poly-commitment/src/commitment.rs
```diff
@@ -220,7 +218,7 @@ impl<'a, 'b, C: AffineCurve> Add<&'a PolyComm<C>> for &'b PolyComm<C> {
         let n2 = other.elems.len();
         for i in 0..std::cmp::max(n1, n2) {
             let pt = if i < n1 && i < n2 {
-                self.elems[i] + other.elems[i]
+                (self.elems[i] + other.elems[i]).into_affine()
```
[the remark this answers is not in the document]
dannywillems: See dw/arkworks-042-bench and https://github.com/o1-labs/proof-systems/pull/2608.
Result: this branch is even faster, around 6%. LGTM.

dannywillems (approved):

poly-commitment/src/commitment.rs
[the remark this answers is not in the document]
volhovm: Cool! Somewhat unexpected tbh hehe.

PR #2589, opened 2024-09-22
[o1vm] Dynamic Selectors for MIPS

Fizzixnerd:


Fizzixnerd:
Should be ready for review;

I will add my own review covering stuff I'm not sure of.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -82,6 +83,7 @@ pub struct Env<Fp, PreImageOracle: PreImageOracleT> {
     pub scratch_state: [Fp; SCRATCH_SIZE],
     pub halt: bool,
     pub syscall_env: SyscallEnv,
+    pub selectors: Vec<Fp>,
```
dannywillems, line 86: I suppose you can use a boolean instead of a field element.

dannywillems: You can also keep simply the integer value of the current selector.

o1vm/src/interpreters/mips/interpreter.rs
```diff
@@ -977,7 +982,27 @@ pub trait InterpreterEnv {
...
+    Env: InterpreterEnv<Position = crate::interpreters::mips::column::ColumnAlias>,
+>(
+    env: &mut Env,
+    instr: Instruction,
+) {
+    // We need to subtract the relations from the instruction usize.
+    env.activate_selector(usize::from(instr) - N_MIPS_REL_COLS);
+
+    // Assert that exactly one selector is active at a time.
+    env.add_constraint(
```
dannywillems, line 995: I would not add this in the interpreter, but instead outside, when getting the constraints.

o1vm/src/interpreters/mips/interpreter.rs
```diff
@@ -977,7 +982,27 @@ pub trait InterpreterEnv {
...
+        Instruction::iter().fold(Env::constant(0), |acc, x| {
+            acc + env.variable(x.to_column_alias())
+        }) - Env::constant(1),
+    );
+
+    // Assert all selectors are either zero or one.
```
dannywillems, line 1001: I would not add this in the interpreter, but instead outside, when getting the constraints.

o1vm/src/interpreters/mips/tests_helpers.rs
```diff
@@ -89,6 +91,9 @@ where
         halt: false,
         // Keccak related
         syscall_env: SyscallEnv::default(),
+        selectors: Vec::from(std::array::from_fn::<Fp, N_MIPS_SEL_COLS, _>(|_| {
```
dannywillems, line 94: It might be ok for a first version, but I would simply keep track of the actual integer representing the instruction, and in the `main.rs` file, I would build the actual vectors of evaluation points.

dannywillems (commented):
Look at how it was done initially: https://github.com/o1-labs/proof-systems/blob/5b687813895fef22a5fc199103b70c3fb0d2360e/optimism/src/main.rs#L84.

I would go with this kind of version.

o1vm/src/interpreters/mips/constraints.rs
```diff
@@ -68,6 +68,10 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         }))
     }
 
+    // TODO: Make sure this is correct: only the witness needs to
+    // worry about this method.
+    fn activate_selector(&mut self, _selector_idx: usize) {}
```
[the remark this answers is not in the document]
marcbeunardeau88: It's not correct since it's empty
Or you mean you want to do this in a follow up ?

Fizzixnerd:
@dannywillems I deleted the code I added to the witness, as it's not clear it's needed now.  Will add back later if it is, but for now this just adds the selectors to the proof inputs.

o1vm/src/interpreters/mips/constraints.rs
```diff
@@ -72,6 +73,16 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         self.constraints.push(assert_equals_zero)
     }
 
+    fn activate_selector(&mut self, selector_idx: usize) {
+        let mut acc : Self::Variable = -Expr::Atom(ExprInner::Constant(ConstantExpr::Atom(ConstantExprInner::Constant(kimchi::circuits::expr::ConstantTerm::<Fp>::Literal(Fp::one())))));
```
Fizzixnerd, line 77: Is there a less verbose way of writing this?  it's just `-1`.

o1vm/src/interpreters/mips/constraints.rs
```diff
@@ -72,6 +73,16 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        for i in 0..N_MIPS_SEL_COLS {
+            let var = self.variable(MIPSColumn::Selector(i));
+            acc += var.clone();
+            self.add_constraint(var.clone() * var.clone() - var);
```
Fizzixnerd, line 81: Add that they are all either zero or one

o1vm/src/interpreters/mips/constraints.rs
```diff
@@ -72,6 +73,16 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        }
+        self.add_constraint(acc);
```
Fizzixnerd, line 83: Because `acc` starts at -1, this is zero if only one is one.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -141,23 +143,8 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
...
-                Instruction::IType(i) => self.selectors[i as usize + RTypeInstruction::COUNT],
-                Instruction::JType(j) => {
-                    self.selectors[j as usize + RTypeInstruction::COUNT + ITypeInstruction::COUNT]
-                }
-            },
-            Column::InstructionCounter => self.instructions_counter,
-        }
-        */
+    fn variable(&self, column: Self::Position) -> Self::Variable {
+        u64::try_from(usize::from(column)).unwrap()
```
dannywillems, line 147: This is not correct:
`variable` must return the actual value saved at the position `column`.
If the column is a `SCRATCH_STATE`, we get the value from `self.scratch_state`. If the value is a `Selector(idx)`, you should return 1 if it is the index of the selector for the current instruction, otherwise 0.

o1vm/src/pickles/main.rs
```diff
@@ -84,6 +85,14 @@ pub fn main() -> ExitCode {
             .push(Fp::from(mips_wit_env.instruction_counter));
         // FIXME: Might be another value
         curr_proof_inputs.evaluations.error.push(Fp::rand(&mut rng));
+
+        curr_proof_inputs
+            .evaluations
+            .selector
+            .push(Fp::from(ark_ff::BigInt::from(
```
dannywillems, line 92: Note: you can simply use `Fp::from(instr as u64)`

o1vm/src/interpreters/mips/column.rs
```diff
@@ -159,3 +171,9 @@ impl ColumnIndexer for Instruction {
         Column::DynamicSelector(usize::from(self) - N_MIPS_REL_COLS)
     }
 }
+
+impl Instruction {
+    pub fn to_selector_column_idx(self) -> usize {
+        usize::from(self) - N_MIPS_REL_COLS
```
dannywillems, line 177: This is not correct. You get an underflow.

o1vm/src/interpreters/mips/constraints.rs
```diff
@@ -72,6 +74,20 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         self.constraints.push(assert_equals_zero)
     }
 
+    fn activate_selector(&mut self, _selector_idx: usize) {
+        let mut acc: Self::Variable = -Expr::Atom(ExprInner::Constant(ConstantExpr::Atom(
```
dannywillems, line 78: You can use `Self::Variable::one()` and import `ark_ff::One`.

dannywillems (changes requested):
Running with the nodes give errors:
```
[2024-10-02T17:29:43Z INFO  o1vm::interpreters::mips::witness] processing step=0 pc=00000ae88c insn=000802ba1f ips=0.00 pages=8598 mem=33.6 MiB name=_rt0_mips_linux
[2024-10-02T17:29:46Z INFO  o1vm::interpreters::mips::witness] Snapshot state in snapshot-state-0.json, step 0
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TryFromIntError(())', o1vm/src/pickles/main.rs:93:81
```

o1vm/src/interpreters/mips/column.rs
dannywillems: This is correct. `usize::from(self)` is not the conversion from `EnumIter` into a `usize`, it is actually an implementation from Instruction to `usize` defined above. Sorry.

dannywillems:
See https://github.com/o1-labs/proof-systems/pull/2649

dannywillems (approved):
As https://github.com/o1-labs/proof-systems/pull/2649/ has been approved, and has been built on top of this PR, I would approve this one, and merge right after 2649.

dannywillems:
Taking care of fixing conflicts

o1vm/src/interpreters/mips/constraints.rs
marcbeunardeau88: There is a From<u64> for expr

PR #2590, opened 2024-09-23
[easy] Fix lower bound of zk rows for chunking

querolita:
When chunking is in place, the number of rows reserved for zero knowledge must be augmented accordingly. In particular, 

```
zk_rows > (2 * permuts * chunks - 2) / 7
```

that means that `zk_rows` must at least be ` (2 * permuts * chunks - 2) / 7 + 1` or what's equivalent,  (2 * permuts * chunks - 5) / 7. Examples of this are [here](https://github.com/MinaProtocol/mina/blob/b6bdcb1a9c8f04534974f3736eb0d20c7c4552b8/src/lib/pickles/step_branch_data.ml#L200).

Nonetheless, on the Kimchi side, the strict lower bound `(2 * permuts * chunks - 2) / 7` is created [here](https://github.com/o1-labs/proof-systems/blob/97950b5ed4735a56ffab6324d88b2247e5a089d8/kimchi/src/circuits/constraints.rs#L622) and is used incorrectly when triggering the error message `ProverError::NotZeroKnowledge`. Meaning, it is raised whenever the number of rows is smaller than that value, when it should be raised anytime it is smaller or equal (because by definition, the number of rows must be **larger** than that value). 

This PR fixes that mismatch and adapts the error message accordingly. 

Example: 

With 2 chunks, the message used to read `there are not enough random rows to achieve zero-knowledge (expected: 4, got: 3)` whereas with 2 chunks you need at least 5 rows for zero knowledge.

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -208,9 +208,9 @@ where
             .ok_or(ProverError::NoRoomForZkInWitness)?;
 
         let zero_knowledge_limit = zk_rows_strict_lower_bound(num_chunks);
-        if (index.cs.zk_rows as usize) < zero_knowledge_limit {
+        if (index.cs.zk_rows as usize) <= zero_knowledge_limit {
             return Err(ProverError::NotZeroKnowledge(
-                zero_knowledge_limit,
+                zero_knowledge_limit + 1,
                 index.cs.zk_rows as usize,
             ));
         }


fabrizio-m (approved):

kimchi/src/circuits/constraints.rs
```diff
@@ -618,6 +618,10 @@ impl<F: PrimeField + SquareRootField> ConstraintSystem<F> {
     }
 }
 
+//
+pub const NUM_CHUNKS_BY_DEFAULT: usize = 1;
+pub const ZK_ROWS_BY_DEFAULT: u64 = 3;
+
 pub fn zk_rows_strict_lower_bound(num_chunks: usize) -> usize {
```
dannywillems, line 625: Note: this is *strict*, i.e. it will give 4 when `num_chunks = 2`, and therefore, we should have *at least* 5 in index.cs.zk_rows.
(I am annoying, but could we simply add a small comment please?)

kimchi/src/prover.rs
```diff
@@ -208,9 +208,9 @@ where
             .ok_or(ProverError::NoRoomForZkInWitness)?;
 
         let zero_knowledge_limit = zk_rows_strict_lower_bound(num_chunks);
-        if (index.cs.zk_rows as usize) < zero_knowledge_limit {
+        if (index.cs.zk_rows as usize) <= zero_knowledge_limit {
```
dannywillems, line 211: And therefore, here, before if `zk_rows` was equals to 4, it would not have raised the error, which was not the expected behavior. Because 5 is the minimum number of `zk_rows`.

dannywillems (approved):

PR #2594, opened 2024-09-23
Arkworks 0.4.2: fix test vectors serialization issue

volhovm:
Problem: in `export_test_vectors` instead of (`f_elem.into_repr() -> f_elem.into_bigint()`) the renaming during the arkworks upgrade was mistakenly (`f_elem.into_repr() -> f_elem.0`), which also returns a bigint, but it's a different bigint!

I also scanned for all `into_repr()` renamings in the original arkworks PR diff and only found another similar issue in `field_helpers`, but there it's immaterial (fixed anyway).

Regression test confirmed to work on `1494cf973d40fb276465929eb7db1952c5de7bdc` (last commit on `develop` before arkworks 0.4.2).

### DIFF
--- a/poseidon/export_test_vectors/src/vectors.rs
+++ b/poseidon/export_test_vectors/src/vectors.rs
@@ -1,5 +1,5 @@
 use super::{Mode, ParamType};
-use ark_ff::UniformRand as _;
+use ark_ff::{PrimeField, UniformRand as _};
 use ark_serialize::CanonicalSerialize as _;
 use mina_curves::pasta::Fp;
 use mina_poseidon::{
@@ -78,7 +78,7 @@ pub fn generate(mode: Mode, param_type: ParamType) -> TestVectors {
             .into_iter()
             .map(|elem| {
                 let mut input_bytes = vec![];
-                elem.0
+                elem.into_bigint()
                     .serialize_uncompressed(&mut input_bytes)
                     .expect("canonical serialiation should work");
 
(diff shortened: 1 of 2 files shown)


poseidon/export_test_vectors/src/vectors.rs
```diff
@@ -112,3 +112,112 @@ pub fn generate(mode: Mode, param_type: ParamType) -> TestVectors {
...
+            [
+                42, 33, 199, 187, 104, 139, 231, 56, 52, 166, 8, 70, 141, 53, 158, 96, 175, 246,
+                75, 186, 160, 9, 17, 203, 83, 113, 240, 208, 235, 33, 111, 41,
+            ],
+            [
+                133, 233, 196, 82, 62, 17, 13, 12, 173, 230, 192, 216, 56, 126, 197, 152, 164, 155,
+                205, 238, 73, 116, 220, 196, 21, 134, 120, 39, 171, 177, 119, 25,
+            ],
+        ];
+
```
dannywillems, line 179: Can you add the commit the values have been generated with? It is a good practice when adding regression tests. In particular, the commit should be one *before* the test has been written.

dannywillems (approved):
Modulo the comment. I approved based on the hypothesis the regression test has been generated *before* the change in the codebase.

poseidon/export_test_vectors/src/vectors.rs
dannywillems: Feel free to do it in a follow-up.

volhovm: Done!

volhovm:
@dannywillems thanks! Yes, I did `checkout` the older commit and generated values at that point.

PR #2596, opened 2024-09-23
Enable rust-cache GH action

Fizzixnerd:
See https://github.com/marketplace/actions/rust-cache

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
           rust_toolchain_version: ${{ matrix.rust_toolchain_version }}
 
       - name: Rust Cache
-        uses: Swatinem/rust-cache@v2.7.3
+        uses: Swatinem/rust-cache@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -84,6 +84,9 @@ jobs:
         with:
           rust_toolchain_version: ${{ matrix.rust_toolchain_version }}
 
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2.7.3
```
shimkiv, line 88:
```suggestion
        uses: Swatinem/rust-cache@v2
```

shimkiv: Let's use major versions so that it will pick up whatever is the latest.

shimkiv (changes requested):
LGTM, requested simple change.
@Fizzixnerd it also will be good to port the changes to `develop`!

shimkiv (approved):

PR #2598, opened 2024-09-24
Serializaiton regression tests for proof-systems and parts of kimchi

volhovm:
This PR adds some serialization regression tests, in particular for `SRS`, `PolyComm`, `OpeningProof`, `ProverProof`.


kimchi/src/prover.rs
```diff
@@ -165,13 +169,15 @@ where
     pub fn create_recursive<
```
dannywillems, line 169: That seems to introduce a change also in the whole Mina repository. We might want to be more careful regarding this. Can we avoid this in this PR?

dannywillems (changes requested):
I don't think we should introduce the rng parameter in the prover in this PR.

kimchi/src/prover.rs
volhovm: We can't really, since then the prover is non-deterministic and it serializes to a different value every time.

volhovm:
@dannywillems Do you want it in a different PR or...? I think it's a fairly small cosmetic change, and it's necessary to be able to run the prover in a reproducible way, so why not?

dannywillems:
> @dannywillems Do you want it in a different PR or...? I think it's a fairly small cosmetic change, and it's necessary to be able to run the prover in a reproducible way, so why not?
> 
> Edit: you're right that it introduces a change in the mina repo, but only in like ... two places in bindings. So it's still quite lightweight.

Ok, giving a second thought...
Can you make the change in Mina accordingly so we can merge all of them? It will imply changing the o1js-bindings also.

volhovm:
Sounds good, will do tomorrow!

volhovm:
Done for mina in https://github.com/MinaProtocol/mina/pull/16139. WIll see regarding o1js.

kimchi/src/tests/and.rs
```diff
@@ -143,6 +143,36 @@ where
...
+/// Generic test for checking serialization & regression of AND circuit.
+fn prove_and_check_serialization_regression<G: KimchiCurve, EFqSponge, EFrSponge>(
+    bytes: usize,
+    buf_expected: Vec<u8>,
+) where
+    G::BaseField: PrimeField,
+    EFqSponge: Clone + FqSponge<G::BaseField, G, G::ScalarField>,
+    EFrSponge: FrSponge<G::ScalarField>,
+{
+    let rng = &mut StdRng::from_seed(RNG_SEED);
```
dannywillems, line 155: For a follow up: use `o1_utils::tests:make_test_rng`. More random tests this way.

dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/2610

poly-commitment/src/tests/commitment.rs
```diff
@@ -216,6 +215,19 @@ fn test_randomised<RNG: Rng + CryptoRng>(mut rng: &mut RNG) {
...
 
+    (proofs, time_commit, time_open)
+}
+
+fn test_randomised<RNG: Rng + CryptoRng>(mut rng: &mut RNG) {
+    let group_map = <Vesta as CommitmentCurve>::Map::setup();
+    // create an SRS optimized for polynomials of degree 2^7 - 1
+    let srs = SRS::<Vesta>::create(1 << 7);
+
+    // TODO: move to bench
```
dannywillems, line 226: Opening follow-up

dannywillems: See https://github.com/o1-labs/proof-systems/issues/2611

utils/Cargo.toml
```diff
@@ -12,24 +12,25 @@ license = "Apache-2.0"
...
-num-integer.workspace = true 
+num-integer.workspace = true
 num-traits.workspace = true
-rmp-serde.workspace = true 
+rmp-serde.workspace = true
 sha2.workspace = true
 thiserror.workspace = true
 rand.workspace = true
 rand_core.workspace = true
+
```
dannywillems, line 30: image

kimchi/src/tests/and.rs
```diff
@@ -328,3 +358,309 @@ fn test_bad_and() {
         ))
     );
 }
+
+#[test]
+fn test_serialization_regression() {
+    // Generated with commit 1494cf973d40fb276465929eb7db1952c5de7bdc
+    let buf_expected = vec![
+        149, 148, 159, 145, 145, 196, 33, 36, 165, 245, 213, 186, 207, 201, 96, 141, 145, 71, 154,
```
dannywillems, line 366: Nit: could be nice to load the value from a file.

poly-commitment/src/tests/serialization.rs
```diff
@@ -0,0 +1,160 @@
...
+pub fn ser_regression_canonical_srs() {
+    use mina_curves::pasta::{Fp, Fq, Pallas, Vesta};
+
+    let rng = &mut StdRng::from_seed([0u8; 32]);
+
+    let td1 = Fp::rand(rng);
+    let data_expected = SRS::<Vesta>::create_trusted_setup(td1, 1 << 3);
+    // Generated with commit 1494cf973d40fb276465929eb7db1952c5de7bdc
+    let buf_expected: Vec<u8> = vec![
+        146, 152, 196, 33, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
dannywillems, line 18: Nit: could be nice to load the value from a file

poly-commitment/src/tests/serialization.rs
```diff
@@ -0,0 +1,160 @@
...
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    ];
+
+    test_generic_serialization_regression_serde(data_expected, buf_expected);
+
+    let td2 = Fq::rand(rng);
+    let data_expected = SRS::<Pallas>::create_trusted_setup(td2, 1 << 3);
+    // Generated with commit 1494cf973d40fb276465929eb7db1952c5de7bdc
+    let buf_expected: Vec<u8> = vec![
+        146, 152, 196, 33, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
dannywillems, line 43: Nit: could be nice to load the value from a file

poly-commitment/src/tests/serialization.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    let BlindedCommitment {
+        commitment: chunked_commitment,
+        blinders: _blinders,
+    } = srs.commit(&poly, num_chunks, rng);
+
+    let data_expected = chunked_commitment;
+    // Generated with commit 1494cf973d40fb276465929eb7db1952c5de7bdc
+    let buf_expected: Vec<u8> = vec![
+        145, 150, 196, 33, 36, 158, 70, 161, 147, 233, 138, 19, 54, 52, 87, 58, 158, 154, 255, 197,
```
dannywillems, line 87: Nit: could be nice to load the value from a file

poly-commitment/src/tests/serialization.rs
```diff
@@ -0,0 +1,160 @@
...
+    let srs = SRS::<Vesta>::create(1 << 7);
+
+    let data_expected: OpeningProof<Vesta> =
+        crate::tests::commitment::generate_random_opening_proof(rng, &group_map, &srs).0[0]
+            .proof
+            .clone();
+
+    // Generated with commit 1494cf973d40fb276465929eb7db1952c5de7bdc
+    let buf_expected: Vec<u8> = vec![
+        149, 151, 146, 196, 33, 166, 183, 76, 209, 121, 62, 56, 194, 134, 133, 238, 52, 190, 163,
```
dannywillems, line 123: Nit: could be nice to load the value from a file

dannywillems (approved):

kimchi/src/tests/and.rs
volhovm: @dannywillems the whole point of this test is to use a fixed RNG because we want this to be static and reproducible. I'm not sure what you mean?

utils/Cargo.toml
volhovm: This was intentional actually, the idea is to keep in-project dependencies separately from the outer-project ones. Note that we do this quite a lot both in other cargo files, and in rust imports within the codebase too.

poly-commitment/src/tests/serialization.rs
volhovm: I'd rather keep it here tbh. I know it's big and ugly, but file management is more annoying imo, you rely on extra kernel calls, it's more files, and if we change serialization in a certain way git would probably just think the whole binary file was replaced, even if it's just a bit. The last case we had before actually when I was replacing the SRS serialization, you could clearly see that the new serialization is just one bit off, which helps with debugging.

poly-commitment/src/tests/commitment.rs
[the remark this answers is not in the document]
volhovm: Agree, not my comment though, I just moved it :wink:

kimchi/src/tests/and.rs
dannywillems: Right. Sorry, it is a regression test. Not a valid comment.

dannywillems:
Feel free to merge when the PR in o1js/o1js-bindings are good to be merged too.

dannywillems (approved):

PR #2601, opened 2024-09-24
Rustfmt-unify imports in develop

volhovm:
This was done half a year ago in `master` but never reached `develop` which is a big source of annoying diffs. 

https://github.com/o1-labs/proof-systems/pull/2044

### DIFF
--- a/poseidon/src/permutation.rs
+++ b/poseidon/src/permutation.rs
@@ -1,7 +1,9 @@
 //! The permutation module contains the function implementing the permutation used in Poseidon
 
-use crate::constants::SpongeConstants;
-use crate::poseidon::{sbox, ArithmeticSpongeParams};
+use crate::{
+    constants::SpongeConstants,
+    poseidon::{sbox, ArithmeticSpongeParams},
+};
 use ark_ff::Field;
 
 fn apply_mds_matrix<F: Field, SC: SpongeConstants>(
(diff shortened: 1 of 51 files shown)


poseidon/src/permutation.rs
```diff
@@ -1,105 +1,107 @@
-//! The permutation module contains the function implementing the permutation used in Poseidon
-
-use crate::constants::SpongeConstants;
-use crate::poseidon::{sbox, ArithmeticSpongeParams};
-use ark_ff::Field;
-
-fn apply_mds_matrix<F: Field, SC: SpongeConstants>(
```
volhovm, line 7: Regarding this diff: this is caused by whitespace format most likely. If you click on "hide whitespace" in github panel above, the whole diff disappears.

(see the sreenshot)
screen_2024-09-24_003

volhovm: If someone can run `cargo fmt` on their machine and check if they get the same diff that'd be great. In case it's reproducible I'm pro leaving it, since our policy seems to be relying on automatic tools regarding styling.

querolita: I actually get a lot of `o1vm` op-program files with hash digests that I don't think should be there. 
Captura de pantalla 2024-09-24 a las 20 00 24
Captura de pantalla 2024-09-24 a las 20 00 30

dannywillems: LGTM.
Process:
- checkout on this branch.
- take `develop` version of the file.
- copy in this branch
- check there are actual changes (there are)
- run `cargo +nightly fmt`.
- check `git diff` is empty
- enjoy and approve PR.

dannywillems (approved):

PR #2602, opened 2024-09-24
o1vm/pickles: importing proof and building it in main

dannywillems:
See https://github.com/o1-labs/proof-systems/blob/5b687813895fef22a5fc199103b70c3fb0d2360e/optimism/src/main.rs#L84

### DIFF
--- a/o1vm/src/pickles/proof.rs
+++ b/o1vm/src/pickles/proof.rs
@@ -1,0 +1,27 @@
+use ark_ff::Zero;
+use kimchi::curve::KimchiCurve;
+
+pub struct WitnessColumns<G> {
+    // FIXME: add selectors
+    pub scratch: [G; crate::interpreters::mips::witness::SCRATCH_SIZE],
+    pub instruction_counter: G,
+    pub error: G,
+}
+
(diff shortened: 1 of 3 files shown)


o1vm/src/pickles/proof.rs
```diff
@@ -0,0 +1,27 @@
...
+use kimchi::curve::KimchiCurve;
+
+pub struct WitnessColumns<G> {
+    // FIXME: add selectors
+    pub scratch: [G; crate::interpreters::mips::witness::SCRATCH_SIZE],
+    pub instruction_counter: G,
+    pub error: G,
+}
+
+pub struct ProofInputs<G: KimchiCurve> {
```
marcbeunardeau88, line 11: I'd call that extended witness
There often is an ambiguity between the witness as the necessarry info to craft a proof (eg. a pre image, or the witness in a more complexity theory sense) and the witness columns which include all the intermediary variables of the constraint system.
I'd like to avoid that ambiguity, and proof inputs is ambigious, or makes me think of the necessarry info

marcbeunardeau88 (approved):

PR #2603, opened 2024-09-24
o1vm/pickles: prover

dannywillems:
It is only a sketch, and only consists of committing to the polynomials at the moment, and evaluates them at \zeta and \zeta \omega.
We still need to:
- add the quotient polynomial -> will be in other PR, after !2604
- eval on a bigger domain. At the moment, we only interpolate on d1 and commit. We should evaluate on a bigger domain depending on the degree of the constraints.

Built on top of https://github.com/o1-labs/proof-systems/pull/2649

### DIFF
--- a/o1vm/src/pickles/proof.rs
+++ b/o1vm/src/pickles/proof.rs
@@ -1,4 +1,5 @@
 use kimchi::curve::KimchiCurve;
+use poly_commitment::{ipa::OpeningProof, PolyComm};
 
 pub struct WitnessColumns<G> {
     pub scratch: [G; crate::interpreters::mips::witness::SCRATCH_SIZE],
@@ -22,4 +23,13 @@ impl<G: KimchiCurve> ProofInputs<G> {
             },
         }
     }
+}
+
+// FIXME: should we blind the commitment?
+pub struct Proof<G: KimchiCurve> {
+    pub commitments: WitnessColumns<PolyComm<G>>,
+    pub zeta_evaluations: WitnessColumns<G::ScalarField>,
+    pub zeta_omega_evaluations: WitnessColumns<G::ScalarField>,
+    /// IPA opening proof
+    pub opening_proof: OpeningProof<G>,
 }
--- a/o1vm/src/pickles/prover.rs
+++ b/o1vm/src/pickles/prover.rs
@@ -1,0 +1,221 @@
+use ark_ec::AffineRepr;
+use ark_ff::{PrimeField, Zero};
+use ark_poly::{univariate::DensePolynomial, Evaluations, Polynomial, Radix2EvaluationDomain as D};
+use kimchi::{
+    circuits::domains::EvaluationDomains, curve::KimchiCurve, groupmap::GroupMap,
+    plonk_sponge::FrSponge,
+};
+use mina_poseidon::{sponge::ScalarChallenge, FqSponge};
+use poly_commitment::{
+    commitment::{absorb_commitment, PolyComm},
+    ipa::{DensePolynomialOrEvaluations, OpeningProof, SRS},
+    OpenProof as _, SRS as _,
+};
+use rand::{CryptoRng, RngCore};
+use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
+
+use super::proof::{Proof, ProofInputs, WitnessColumns};
+
+/// Make a PlonKish proof for the given circuit. As inputs, we get the execution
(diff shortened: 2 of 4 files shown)


dannywillems:
Rebasing on top of https://github.com/o1-labs/proof-systems/pull/2649 as I need the fix in @Fizzixnerd branch to continue and e2e tests.

dannywillems:
I did rebase on top of the latest version of https://github.com/o1-labs/proof-systems/pull/2649.

o1vm/src/pickles/proof.rs
```diff
@@ -23,3 +24,12 @@ impl<G: KimchiCurve> ProofInputs<G> {
         }
     }
 }
+
+// FIXME: should we blind the commitment?
+pub struct Proof<G: KimchiCurve> {
+    pub commitments: WitnessColumns<PolyComm<G>>,
```
marcbeunardeau88, line 30: we'll need more explicit names when we'll have more polynomials than the witness columns, but ok for now

o1vm/src/pickles/prover.rs
```diff
@@ -0,0 +1,221 @@
...
+    };
+    let commitments = {
+        let WitnessColumns {
+            scratch,
+            instruction_counter,
+            error,
+            // FIXME
+            selector: _,
+        } = &polys;
+        // Note: we do not blind. We might want in the near future in case we
```
marcbeunardeau88, line 90: AFAIK it's not planned for this proof to be zk

o1vm/src/pickles/prover.rs
```diff
@@ -0,0 +1,221 @@
...
+            instruction_counter: eval(instruction_counter),
+            error: eval(error),
+            // FIXME
+            selector: eval(error),
+        }
+    };
+    // All evaluations at ζ
+    let zeta_evaluations = evals(&zeta);
+    // All evaluations at ζω
+    let zeta_omega_evaluations = evals(&zeta_omega);
```
marcbeunardeau88, line 150: it's unclear to me which poly should be evaluated at zeta_omega
Also this info can (should?) be derived from the constraints rather than hardcoded

o1vm/src/pickles/prover.rs
```diff
@@ -0,0 +1,221 @@
...
+    let mut polynomials: Vec<_> = polys.scratch.into_iter().collect();
+    polynomials.push(polys.instruction_counter);
+    polynomials.push(polys.error);
+    // FIXME: add selectors
+    let polynomials: Vec<_> = polynomials
+        .iter()
+        .map(|poly| {
+            (
+                DensePolynomialOrEvaluations::DensePolynomial(poly),
+                // We do not have any blinder, therefore we set to 0.
```
marcbeunardeau88, line 186: I thought the field PolyComm contains the commitment
This comment makes me think it contains the blinder

o1vm/src/pickles/prover.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    // Computing the opening proof for the IPA PCS
+    let opening_proof = OpeningProof::open::<_, _, D<G::ScalarField>>(
+        srs,
+        &group_map,
+        polynomials.as_slice(),
+        &[zeta, zeta_omega],
+        v,
+        u,
+        fq_sponge_before_evaluations,
```
marcbeunardeau88, line 211: why is that sponge state given ?

marcbeunardeau88 (approved):

o1vm/src/pickles/prover.rs
```diff
@@ -0,0 +1,221 @@
...
+    ////////////////////////////////////////////////////////////////////////////
+
+    // FIXME: add selectors
+    // FIXME: evaluate on a domain higher than d1 for the quotient polynomial.
+    let ProofInputs { evaluations } = inputs;
+    let polys = {
+        let WitnessColumns {
+            scratch,
+            instruction_counter,
+            error,
```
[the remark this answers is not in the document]
dannywillems: No, it is the error code of the process, updated after each instruction.

o1vm/src/pickles/prover.rs
dannywillems: > Also this info can (should?) be derived from the constraints rather than hardcoded

Yes, we will probably use it with the lookup argument soon.

o1vm/src/pickles/prover.rs
dannywillems: As discussed earlier this week, PolyComm is abstracted over any type `T`, and sometimes a scalar is used.

PR #2606, opened 2024-09-25
Auto-align kimchi md file (tabs/spaces?) [cosmetics/easy]

volhovm:
Different from https://github.com/o1-labs/proof-systems/pull/2599!

Can someone please double check? Whenever I'm on `develop` and run `cargo clean && cargo build --all-targets --all-features --tests` I get the changes in `kimchi.md` which are committed in this PR. If it's what `cargo` does then let's commit it I guess. If not, happy to close/investigate.

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -611,16 +611,16 @@ every input (thinking of a two-bit input as a number in $\{0, 1, 2, 3\}$), one f
 are given by
 
 * `c_func(x)`, defined by
-     * `c_func(0) = 0`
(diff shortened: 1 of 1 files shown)


querolita:
I just executed this command on my machine and I also get that diff. But instead of adding more spaces, it reduces one level of indentation in my machine (example attached) 
Captura de pantalla 2024-09-25 a las 15 30 14

querolita (approved):
Oh, I just saw this PR also reduces one level. It's just the way Git was showing the diff that made me think it was the other way round. Accepting now.

PR #2609, opened 2024-09-25
Poly_comm: benchmark add and sub over Pallas/Vesta

dannywillems:


poly-commitment/benches/poly_comm.rs
```diff
@@ -0,0 +1,112 @@
...
+use poly_commitment::PolyComm;
+use rand_core::{CryptoRng, RngCore};
+
+fn helper_generate_random_elliptic_curve_point<RNG, P>(rng: &mut RNG) -> Affine<P>
+where
+    P::BaseField: PrimeField,
+    RNG: RngCore + CryptoRng,
+    P: SWCurveConfig,
+{
+    let p1_x = P::BaseField::rand(rng);
```
volhovm, line 17: You could have just generated a scalar and scaled by it probably? I'm also thinking there's /likely/ a way to generate random points right away, no?

volhovm (approved):

poly-commitment/benches/poly_comm.rs
dannywillems: Huhu. YES!

dannywillems: "Pourquoi faire simple quand on peut faire compliqué ?"

dannywillems: See https://github.com/o1-labs/proof-systems/pull/2612

PR #2617, opened 2024-09-25
Mina_curves: activating ADX instructions by default

dannywillems:
Spotted by @sebastiencs.

ADX instructions are not activated otherwise. However, this change activates it all the time, and compiling on machines not supporting ADX will fail. Is it what we want?

### DIFF
--- a/curves/Cargo.toml
+++ b/curves/Cargo.toml
@@ -19,3 +19,7 @@ rand.workspace = true
 ark-algebra-test-templates.workspace = true
 ark-std.workspace = true
 ark-serialize.workspace = true
+
+[features]
+asm = []
+default = ["asm"]


tizoc:
https://docs.minaprotocol.com/berkeley-upgrade/requirements says AVX here but probably meant to say ADX

dannywillems:
Reg Mina: it seems we build by default with the flags: https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_bindings/stubs/dune#L59 (and https://github.com/MinaProtocol/mina/blob/compatible/src/lib/crypto/kimchi_bindings/stubs/dune#L13), therefore we should not expect any change in Mina?

curves/Cargo.toml
```diff
@@ -19,3 +19,7 @@ rand.workspace = true
 ark-algebra-test-templates.workspace = true
 ark-std.workspace = true
```
sebastiencs, line 20: `ark-std` need to be moved from `dev-dependencies` to `dependencies`

sebastiencs, line 20: `ark-std` need to be moved from `dev-dependencies` to `dependencies`

sebastiencs (changes requested):

PR #2626, opened 2024-09-25
Poly_commitment: move verify method on SRS from commitment into srs

dannywillems:


marcbeunardeau88 (changes requested):
I don't see how moving an opening verification into srs makes sense (even though the method is on srs)

dannywillems:
It is already in SRS. We can move later outside of the structure SRS. It is only to move between files and rename after.

marcbeunardeau88 (approved):

PR #2627, opened 2024-09-25
Poly_commitment: move open SRS into srs.rs

dannywillems:


marcbeunardeau88 (changes requested):
I would not make that change (even though the metod is implemented for the srs type)

dannywillems:
See https://github.com/o1-labs/proof-systems/pull/2626#issuecomment-2376226442

marcbeunardeau88 (approved):

PR #2632, opened 2024-09-26
poly-commitment: move SRS structure into `evaluation_proof.rs`

dannywillems:
It is error-prone to use simply `SRS`. Reorganizing to match KZG and IPA.
It is also to integrate more easily other polynomial commitment scheme we might want to use in the future.

Next step is to rename `evaluation_proof` into `ipa`.

While doing the changes, I realised we were not using the right PC (i.e. KZG with bn254) in the ivc repository, including the tests in the o1vm.


ivc/src/plonkish_lang.rs
```diff
@@ -130,8 +129,8 @@ impl<G: KimchiCurve, const N_COL: usize, const N_CHALS: usize, const N_ALPHAS: u
     }
 }
 
-impl<G: KimchiCurve, const N_COL: usize, const N_CHALS: usize, const N_ALPHAS: usize> Instance<G>
-    for PlonkishInstance<G, N_COL, N_CHALS, N_ALPHAS>
+impl<G: CommitmentCurve, const N_COL: usize, const N_CHALS: usize, const N_ALPHAS: usize>
```
dannywillems, line 132: Also changing this to `CommitmentCurve` as `bn254` does not implement the trait `KimchiCurve`. And nothing from the trait `KimchiCurve` is required.

ivc/src/plonkish_lang.rs
```diff
@@ -153,13 +152,16 @@ impl<G: KimchiCurve, const N_COL: usize, const N_CHALS: usize, const N_ALPHAS: u
...
 
 // Implementation for 3 challenges; only for now.
-impl<G: KimchiCurve, const N_COL: usize, const N_ALPHAS: usize>
+impl<G: CommitmentCurve, const N_COL: usize, const N_ALPHAS: usize>
     PlonkishInstance<G, N_COL, 3, N_ALPHAS>
 {
-    pub fn from_witness<EFqSponge: FqSponge<G::BaseField, G, G::ScalarField>>(
+    pub fn from_witness<
+        EFqSponge: FqSponge<G::BaseField, G, G::ScalarField>,
+        Srs: SRS<G> + std::marker::Sync,
```
dannywillems, line 160: Also this type is added to support KZG and IPA, even though we only need KZG.

kimchi/src/bench.rs
```diff
@@ -34,7 +34,7 @@ pub struct BenchmarkCtx {
 
 impl BenchmarkCtx {
     pub fn srs_size(&self) -> usize {
-        math::ceil_log2(self.index.srs.max_degree())
+        math::ceil_log2(self.index.srs.max_poly_size())
```
dannywillems, line 37: This is another change. I got rid of the method `max_degree` which was exactly the same than `max_poly_size`.

o1vm/src/legacy/folding.rs
```diff
@@ -377,7 +377,7 @@ pub mod keccak {
         type Selector = Steps;
         type Challenge = Challenge;
         type Curve = Curve;
-        type Srs = SRS<Curve>;
+        type Srs = PairingSRS<Pairing>;
```
dannywillems, line 380: We are in the legacy flavor, therefore only need to deal with KZG.

o1vm/src/legacy/main.rs
```diff
@@ -102,7 +103,7 @@ pub fn main() -> ExitCode {
                 BaseSponge,
             >>::folding_constraints(&mips_trace),
             vec![],
-            &srs.full_srs,
+            &srs,
```
dannywillems, line 106: Note: we should change the KZG SRS structure later. It is ugly.

poly-commitment/src/evaluation_proof.rs
```diff
@@ -3,17 +3,33 @@
 //! Zero-Knowledge Arguments for Arithmetic Circuits in the Discrete Log
 //! Setting](https://eprint.iacr.org/2016/263)
 
-use crate::{commitment::*, srs::SRS, PolynomialsToCombine};
-use ark_ec::AffineRepr;
-use ark_ff::{FftField, Field, One, PrimeField, Zero};
-use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial, EvaluationDomain, Evaluations};
+use crate::{
```
dannywillems, line 6: No changes here. Mostly moving from `srs`. I did remove one method (`max_degree`) and remove some duplicated methods.

poly-commitment/src/lib.rs
```diff
@@ -21,7 +20,7 @@ use ark_poly::{
 use mina_poseidon::FqSponge;
 use rand_core::{CryptoRng, RngCore};
 
-pub trait SRS<G: CommitmentCurve>: Clone {
+pub trait SRS<G: CommitmentCurve>: Clone + Sized {
```
dannywillems, line 23: Because `Sized` is required for the folding library to pass it in the type, IIRC.

marcbeunardeau88 (approved):

PR #2639, opened 2024-10-01
MVPoly/tests: fix maximum degree for multiplication

dannywillems:
The multiplication doesn't check that the resulting degree is higher than the one encoded in the type. We do have to change the type signature in the near future.


Fizzixnerd (approved):
This is just dividing by two so that it never exceeds the static maximum, yes?

PR #2644, opened 2024-10-02
Update book description with auto-generated stuff [1min cosmetic]

volhovm:
After this PR was merged (https://github.com/o1-labs/proof-systems/pull/2598), the book was updated, however this was not committed in the PR itself.

I think there was a CI test before that was checking whether auto-generated book matches the commited one. We should probably have this again.

### DIFF
--- a/book/src/specs/kimchi.md
+++ b/book/src/specs/kimchi.md
@@ -1940,7 +1940,7 @@ A proof consists of the following data structures:
 ```rs
 /// Evaluations of a polynomial at 2 points
 #[serde_as]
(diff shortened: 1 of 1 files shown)


volhovm:
OK so apparently the job `Check that up-to-date specification is checked in` exists and it fails (https://github.com/o1-labs/proof-systems/actions/runs/11142333403/job/30965055783?pr=2643). Wondering why it didn't fail on that PR.

querolita (approved):

dannywillems (approved):

PR #2649, opened 2024-10-02
Suggestions for dynamic selectors

dannywillems:
Running the o1vm with this code seems to work.

cc @Fizzixnerd

### DIFF
--- a/o1vm/src/pickles/main.rs
+++ b/o1vm/src/pickles/main.rs
@@ -1,19 +1,23 @@
 use ark_ff::UniformRand;
 use kimchi::circuits::domains::EvaluationDomains;
+use kimchi_msm::expr::E;
 use log::debug;
 use o1vm::{
     cannon::{self, Meta, Start, State},
     cannon_cli,
     interpreters::mips::{
         column::N_MIPS_REL_COLS,
-        constraints as mips_constraints,
+        constraints as mips_constraints, interpreter,
+        interpreter::InterpreterEnv,
         witness::{self as mips_witness},
+        Instruction,
     },
     pickles::proof::ProofInputs,
     preimage_oracle::PreImageOracle,
 };
 use poly_commitment::{ipa::SRS, SRS as _};
 use std::{fs::File, io::BufReader, process::ExitCode};
+use strum::IntoEnumIterator;
 
 use mina_curves::pasta::{Fp, Vesta};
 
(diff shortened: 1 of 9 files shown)


Fizzixnerd (approved):

dannywillems:
Rebasing on top of https://github.com/o1-labs/proof-systems/pull/2589 and amending the commit messages to a better description.

dannywillems:
I'm adding regression tests, in particular for the degree of the constraints. This PR doesn't include any.

o1vm/src/pickles/main.rs
```diff
@@ -65,11 +69,29 @@ pub fn main() -> ExitCode {
     // Initialize the environments
     let mut mips_wit_env =
         mips_witness::Env::<Fp, PreImageOracle>::create(cannon::PAGE_SIZE as usize, state, po);
-    let mut _mips_con_env = mips_constraints::Env::<Fp>::default();
+
+    // TODO: give this to the prover + verifier
+    let _constraints = {
+        let mut mips_con_env = mips_constraints::Env::<Fp>::default();
+        let mut constraints = Instruction::iter().fold(vec![], |mut acc, instr| {
```
dannywillems, line 76: `iter` is not correctly used here. Detected while writing regression tests to check the number of constraitns and their degree. `IntoIter` must be reimplemented. Doing it.

dannywillems: Or `flat_map` can be used. I think it is a better solution for now.

o1vm/src/pickles/main.rs
```diff
@@ -65,11 +69,31 @@ pub fn main() -> ExitCode {
...
+        let mut constraints = Instruction::iter()
+            .flat_map(|instr_typ| instr_typ.into_iter())
+            .fold(vec![], |mut acc, instr| {
+                interpreter::interpret_instruction(&mut mips_con_env, instr);
+                let selector = mips_con_env.get_selector();
+                let constraints_with_selector: Vec<E<Fp>> = mips_con_env
+                    .get_constraints()
+                    .into_iter()
+                    .map(|c| selector.clone() * c)
+                    .collect();
```
Fizzixnerd, line 85: I don't think this is correct, since this would iterate over each instruction once.  We want to iterate over the instructions that were actually run in the MIPS program, not every instruction once, in order.

Fizzixnerd (changes requested):

o1vm/src/pickles/main.rs
Fizzixnerd: I understand now: this is just for the constraints.

Fizzixnerd (approved):

PR #2652, opened 2024-10-03
o1vm/pickles: introduce the computation of the quotient polynomial

dannywillems:
TODO:
- absorb the commitments
- add in the opening proof

Note that this will break the e2e tests because selectors polynomials are not defined correctly at the moment.

### DIFF
--- a/o1vm/src/pickles/prover.rs
+++ b/o1vm/src/pickles/prover.rs
@@ -29,12 +29,6 @@ use thiserror::Error;
 /// Errors that can arise when creating a proof
 #[derive(Error, Debug, Clone)]
 pub enum ProverError {
-    #[error("the proof could not be constructed: {0}")]
-    Generic(&'static str),
-
-    #[error("the provided constraint was not satisfied: {0}")]
-    ConstraintNotSatisfied(String),
-
     #[error("the provided constraint has degree {0} > allowed {1}; expr: {2}")]
     ConstraintDegreeTooHigh(u64, u64, String),
 }
--- a/o1vm/src/pickles/prover.rs
+++ b/o1vm/src/pickles/prover.rs
@@ -191,30 +191,6 @@ where
     };
 
     let quotient_poly: DensePolynomial<G::ScalarField> = {
-        let mut last_constraint_failed = None;
-        // Only for debugging purposes
-        for expr in constraints.iter() {
-            let fail_q_division =
-                ProverError::ConstraintNotSatisfied(format!("Unsatisfied expression: {:}", expr));
-            // Check this expression are witness satisfied
-            let (_, res) = expr
-                .evaluations(&column_env)
-                .interpolate_by_ref()
-                .divide_by_vanishing_poly(domain.d1)
-                .ok_or(fail_q_division.clone())?;
-            if !res.is_zero() {
-                eprintln!("Unsatisfied expression: {}", expr);
-                //return Err(fail_q_division);
-                last_constraint_failed = Some(expr.clone());
-            }
-        }
-        if let Some(expr) = last_constraint_failed {
-            return Err(ProverError::ConstraintNotSatisfied(format!(
-                "Unsatisfied expression: {:}",
-                expr
-            )));
-        }
-
         // Compute ∑ α^i constraint_i as an expression
         let combined_expr =
             E::combine_constraints(0..(constraints.len() as u32), (constraints).to_vec());
--- a/o1vm/src/pickles/prover.rs
+++ b/o1vm/src/pickles/prover.rs
@@ -24,6 +24,20 @@ use super::{
     proof::{Proof, ProofInputs, WitnessColumns},
 };
 use crate::E;
+use thiserror::Error;
+
+/// Errors that can arise when creating a proof
+#[derive(Error, Debug, Clone)]
+pub enum ProverError {
+    #[error("the proof could not be constructed: {0}")]
+    Generic(&'static str),
+
+    #[error("the provided (witness) constraints was not satisfied: {0}")]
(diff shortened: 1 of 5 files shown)


o1vm/src/pickles/prover.rs
```diff
@@ -24,6 +24,20 @@ use super::{
...
 use crate::E;
+use thiserror::Error;
+
+/// Errors that can arise when creating a proof
+#[derive(Error, Debug, Clone)]
+pub enum ProverError {
+    #[error("the proof could not be constructed: {0}")]
+    Generic(&'static str),
+
+    #[error("the provided (witness) constraints was not satisfied: {0}")]
```
marcbeunardeau88, line 35: either the constraints were not satisfied, or the witness satisfying

o1vm/src/pickles/prover.rs
```diff
@@ -24,6 +24,20 @@ use super::{
...
+    ConstraintNotSatisfied(String),
+
+    #[error("the provided (witness) constraint has degree {0} > allowed {1}; expr: {2}")]
```
marcbeunardeau88, line 38:
```suggestion
    #[error("the provided constraint has degree {0} > allowed {1}; expr: {2}")]
```

o1vm/src/pickles/prover.rs
marcbeunardeau88, line 35:
```suggestion
    #[error("the provided constraint was not satisfied: {0}")]
```

o1vm/src/pickles/prover.rs
```diff
@@ -176,7 +190,72 @@ where
...
+        let mut last_constraint_failed = None;
+        // Only for debugging purposes
+        for expr in constraints.iter() {
+            let fail_q_division =
+                ProverError::ConstraintNotSatisfied(format!("Unsatisfied expression: {:}", expr));
+            // Check this expression are witness satisfied
+            let (_, res) = expr
+                .evaluations(&column_env)
+                .interpolate_by_ref()
+                .divide_by_vanishing_poly(domain.d1)
```
marcbeunardeau88, line 203: are we paying for a division for each constraints just to check, and re doing that division line 246 ?

o1vm/src/pickles/prover.rs
```diff
@@ -176,7 +190,72 @@ where
...
+        // must be equal to 0 as the constraints polynomial and Z_H(X) are both
+        // equal on H.
+        if !res.is_zero() {
+            fail_final_q_division();
+        }
+
+        quotient
+    };
+
+    let _t_comm = srs.commit_non_hiding(&quotient_poly, 7);
```
marcbeunardeau88, line 258: we should absorb the t before squeezing the next FS challenge !

dannywillems: It is in the future (draft) PR, see https://github.com/o1-labs/proof-systems/pull/2655. It is also written in the description of this PR that it is a TODO.

o1vm/src/pickles/prover.rs
dannywillems: Yes, I am adding an option soon to activate only this in a "debug" mode. PR coming.

marcbeunardeau88: I'd remove that, doesn't it pollute the code ?
Note that we can check the reminder when dividing by the vanishing poly if we really want a check

dannywillems: I did remove https://github.com/o1-labs/proof-systems/pull/2652/commits/8e87244a98d55b90d175ad389611a3c98bd16b34 and https://github.com/o1-labs/proof-systems/pull/2652/commits/96d42c127ef025869c91e5fed680e0e383108706 as you seem to have a strong opinion on this. I would probably keep the commits somewhere on my side for debugging purposes when finishing the prover.

o1vm/src/pickles/prover.rs
```diff
@@ -176,7 +184,48 @@ where
         }
     };
 
-    // FIXME: add quotient polynomial
+    let quotient_poly: DensePolynomial<G::ScalarField> = {
+        // Compute ∑ α^i constraint_i as an expression
+        let combined_expr =
+            E::combine_constraints(0..(constraints.len() as u32), (constraints).to_vec());
```
marcbeunardeau88, line 190: not a big deal, but I'd rather have this computed elsewhere
The expressions should not be computed at proving time, for separation of concept

o1vm/src/pickles/prover.rs
```diff
@@ -176,7 +184,48 @@ where
...
+
+        // And we interpolate using the evaluations
+        let expr_evaluation_interpolated = expr_evaluation.interpolate();
+
+        let fail_final_q_division = || {
+            panic!("Division by vanishing poly must not fail at this point, we checked it before")
+        };
+        // We compute the polynomial t(X) by dividing the constraints polynomial
+        // by the vanishing polynomial, i.e. Z_H(X).
+        let (quotient, res) = expr_evaluation_interpolated
```
marcbeunardeau88, line 215: shouldn't res be called rem for remainder ?

o1vm/src/pickles/prover.rs
```diff
@@ -176,7 +184,48 @@ where
...
+            .divide_by_vanishing_poly(domain.d1)
+            .unwrap_or_else(fail_final_q_division);
```
marcbeunardeau88, line 217: Why do we have twice the same error for two different fails ?
Also (not a  part of this PR) I don't see the fail happening in divide by vanishing poly, mb it should not returns an option

marcbeunardeau88 (approved):

PR #2657, opened 2024-10-04
Poly-commitment: remove prover_plms_to_verif_evals as unused

dannywillems:
Even in Mina it is unused.


marcbeunardeau88 (changes requested):
since its a debugging function, I'd keep it commented

marcbeunardeau88 (approved):

PR #2658, opened 2024-10-04
Poly-commitment: rename elems field into chunks

dannywillems:
It is more meaningful because it is actually chunks. In following pull requests, the abstraction over PolyComm will be changed, to avoid erroneous usage of PolyComm.

It is part of the implementation of the prover into pickles, as I noticed I made a small error while splitting the quotient polynomial in chunks.

I am breaking the usage of PolyComm in develop, and in Mina, cc @volhovm as you will work on merging develop into master for proof-systems.

### DIFF
--- a/arrabiata/src/witness.rs
+++ b/arrabiata/src/witness.rs
@@ -496,7 +496,7 @@ where
                 let idx_col = idx / 2;
                 debug!("Absorbing the accumulator for the column index {idx_col}. After this, there will still be {} elements to absorb", NUMBER_OF_VALUES_TO_ABSORB_PUBLIC_IO - idx - 1);
                 if self.current_iteration % 2 == 0 {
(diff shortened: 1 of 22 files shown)


querolita (approved):
Approving, but let's make sure the mina side is not broken if trying to access `elems`

dannywillems:
> Approving, but let's make sure the mina side is not broken if trying to access `elems`

It will be broken.

dannywillems:
> > Approving, but let's make sure the mina side is not broken if trying to access `elems`
> 
> It will be broken.

I've checked to backport in develop by cherry-picking, and update Mina accordingly, but proof-systems develop/master are unsynced. I think @volhovm is working on this.

dannywillems:
I'm taking the responsibility to introduce this breaking change. If you have difficulties when merging develop into master @volhovm, I think it is fair for me to help you to propagate this change in mina + o1js.

PR #2659, opened 2024-10-04
PolyComm: use method len directly instead of chunks.len

dannywillems:
`len()` on PolyComm is exactly implemented as `chunks.len()` It is part of a work abstracting the structure `PolyComm`, and remove the access to the field `chunks`.

### DIFF
--- a/kimchi/src/verifier.rs
+++ b/kimchi/src/verifier.rs
@@ -239,11 +239,11 @@ where
         let alpha = alpha_chal.to_field(endo_r);
 
         //~ 1. Enforce that the length of the $t$ commitment is of size 7.
-        if self.commitments.t_comm.chunks.len() > chunk_size * 7 {
+        if self.commitments.t_comm.len() > chunk_size * 7 {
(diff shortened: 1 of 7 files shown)


kimchi/src/verifier.rs
```diff
@@ -239,11 +239,11 @@ where
         let alpha = alpha_chal.to_field(endo_r);
 
         //~ 1. Enforce that the length of the $t$ commitment is of size 7.
-        if self.commitments.t_comm.chunks.len() > chunk_size * 7 {
+        if self.commitments.t_comm.len() > chunk_size * 7 {
             return Err(VerifyError::IncorrectCommitmentLength(
                 "t",
                 chunk_size * 7,
```
querolita, line 245: Unrelated, but this 7 should be a `PERMUTS` or similar

querolita (approved):

kimchi/src/verifier.rs
querolita: https://github.com/o1-labs/proof-systems/issues/2666

PR #2664, opened 2024-10-04
PolyComm: use new instead of constructing with fields

dannywillems:


querolita (commented):
Meh... no strong feeling about this one 😞

dannywillems:
> Meh... no strong feeling about this one 😞

It is mostly to make the field `chunks` abstract later, instead of public.

PR #2665, opened 2024-10-04
PolyComm/Kimchi: use into_iter instead of accessing chunks field

dannywillems:
Following recent change

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -1397,9 +1397,8 @@ where
 
                     let chunks = runtime_comm
                         .blinders
-                        .chunks
-                        .iter()
-                        .map(|blinding| *joint_combiner * blinding + base_blinding)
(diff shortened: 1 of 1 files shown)


kimchi/src/prover.rs
```diff
@@ -1397,9 +1397,8 @@ where
 
                     let chunks = runtime_comm
                         .blinders
-                        .chunks
-                        .iter()
-                        .map(|blinding| *joint_combiner * blinding + base_blinding)
+                        .into_iter()
```
querolita, line 1400: I am wondering if consuming the blinders here instead of iterating over the references could be a problem elsewhere 🤔

querolita (approved):

kimchi/src/prover.rs
dannywillems: Discussed offline.

PR #2668, opened 2024-10-04
Poly-commitment/IPA: additional tests regarding expected nb chunks

dannywillems:
While working on the prover for o1vm/pickles, I was wondering the number of chunks `commit_non_hiding` was supposed to return in different cases, in particular when the polynomial has a degree higher than the number of chunks requested. This test clarifies it.
I hope it doesn't duplicate another test - I didn't look for.

As @querolita is (was?) working on chunking, cc'ing her as she might be interested in.

### DIFF
--- a/poly-commitment/src/ipa.rs
+++ b/poly-commitment/src/ipa.rs
@@ -584,11 +584,6 @@ where
         })
     }
 
-    /// This function commits a polynomial using the SRS' basis of size `n`.
-    /// - `plnm`: polynomial to commit to with max size of sections
-    /// - `num_chunks`: the number of commitments to be included in the output polynomial commitment
-    /// The function returns an unbounded commitment vector
-    /// (which splits the commitment into several commitments of size at most `n`).
     fn commit_non_hiding(
         &self,
         plnm: &DensePolynomial<G::ScalarField>,
--- a/poly-commitment/src/ipa.rs
+++ b/poly-commitment/src/ipa.rs
@@ -564,7 +564,6 @@ where
         self.mask_custom(comm, &blinders).unwrap()
     }
 
-    /// Same as [SRS::mask] except that you can pass the blinders manually.
     fn mask_custom(
         &self,
         com: PolyComm<G>,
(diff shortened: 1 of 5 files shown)


poly-commitment/src/ipa.rs
```diff
@@ -584,11 +583,6 @@ where
         })
     }
 
-    /// This function commits a polynomial using the SRS' basis of size `n`.
```
querolita, line 587: Not sure if this is intentional, but it seems like the docs for these functions have been removed.

querolita: Oh, it looks like you're moving the comments to the lib instead

poly-commitment/src/lib.rs
```diff
@@ -60,13 +63,24 @@ pub trait SRS<G: CommitmentCurve>: Clone + Sized {
...
-    /// The function returns an unbounded commitment vector (which splits the
-    /// commitment into several commitments of size at most `n`).
-    /// It is analogous to [SRS::commit_evaluations_non_hiding] but for
-    /// polynomials.
+    /// - `plnm`: polynomial to commit to. The polynomial can be of any degree,
+    /// including higher than `n`.
+    /// - `num_chunks`: the minimal number of commitments to be included in the
+    /// output polynomial commitment.
+    ///
+    /// The function returns the commitments to the chunks (of at most `n`) of
```
querolita, line 71: I guess there's a missing word in that sentence. Didn't you mean? 
```suggestion
    /// The function returns the commitments to the chunks (of size at most `n`) of
```

poly-commitment/src/pbt_srs.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // Polynomial of exactly a multiple of the SRS size, i.e degree is k *
+    // srs_size - 1.
+    {
+        let k = rng.gen_range(2..5);
+        let poly_degree = k * srs_size - 1;
+        let poly = DensePolynomial::<G::ScalarField>::rand(poly_degree, &mut rng);
+        // if we request a number of chunks smaller than the multiple, we will
+        // still get a number of chunks equals to the multiple.
```
querolita, line 61: This is a nice test

querolita (approved):

PR #2671, opened 2024-10-07
Add selectors to prover

Fizzixnerd:


dannywillems (changes requested):
See comments on Slack.

o1vm/src/pickles/prover.rs
```diff
@@ -72,50 +75,65 @@ where
...
             scratch,
             instruction_counter,
             error,
-            // FIXME
-            selector: _,
+            selector,
         } = evaluations;
+
+        // Build the selectors
+        let selector: [Vec<<<G as AffineRepr>::Group as Group>::ScalarField>; N_MIPS_SEL_COLS] =
```
dannywillems, line 91: :+1:

o1vm/src/pickles/prover.rs
```diff
@@ -72,50 +75,65 @@ where
...
+            array::from_fn(|i| {
+                let mut s_i = Vec::with_capacity(DOMAIN_SIZE);
```
dannywillems, line 93: The domain size should be the one given in parameter, i.e. `domain`.

o1vm/src/pickles/prover.rs
```diff
@@ -11,6 +13,7 @@ use kimchi::{
     groupmap::GroupMap,
     plonk_sponge::FrSponge,
 };
+use kimchi_msm::DOMAIN_SIZE;
```
dannywillems, line 16: Use the domain size from the parameter.

o1vm/src/pickles/proof.rs
```diff
@@ -1,15 +1,17 @@
...
 
-pub struct WitnessColumns<G> {
+use crate::interpreters::mips::column::N_MIPS_SEL_COLS;
+
+pub struct WitnessColumns<G, S> {
     pub scratch: [G; crate::interpreters::mips::witness::SCRATCH_SIZE],
     pub instruction_counter: G,
     pub error: G,
-    pub selector: G,
+    pub selector: S,
```
dannywillems, line 10: I'm not a big fan of this, but I do not have a better idea for now. Maybe `Vec<G>`? It would `ScalarField` when we do have the single value for the row, and `Vec<ScalarField>` when we deal with the evaluations of all selectors? The only thing is that the number of selectors would be "unknown". Which, I think, is fine.

dannywillems (changes requested):
Good start! There is also a change in "column_env.rs", see [here](https://github.com/o1-labs/proof-systems/blob/master/o1vm/src/pickles/column_env.rs#L60).

o1vm/src/pickles/proof.rs
Fizzixnerd: Changing this to `Vec<G>` would, I believe, make it impossible to fix the previous comment, as then `G` would need to be `usize`.

dannywillems: You can use `Fp::from(x as usize)`.

dannywillems (approved):

PR #2672, opened 2024-10-08
Make lagrange basis handling safe

mrmr1993:
This PR adds the concept of `HashMapCache` to remove the potential for errors / panics in code. In particular, this addresses https://github.com/o1-labs/o1js/issues/1411.

### DIFF
--- a/poly-commitment/src/hash_map_cache.rs
+++ b/poly-commitment/src/hash_map_cache.rs
@@ -16,11 +16,11 @@ impl<Key: Hash + std::cmp::Eq, Value> HashMapCache<Key, Value> {
         }
     }
 
-    pub fn get_or_generate<'a, F: FnOnce() -> Value>(
-        &'a self,
+    pub fn get_or_generate<F: FnOnce() -> Value>(
+        &self,
         key: Key,
         generator: F,
-    ) -> &'a Value {
+    ) -> &Value {
         let mut hashmap = self.contents.lock().unwrap();
         let entry = (*hashmap).entry(key);
         let inner_ptr = match entry {
--- a/poly-commitment/src/hash_map_cache.rs
+++ b/poly-commitment/src/hash_map_cache.rs
@@ -16,11 +16,7 @@ impl<Key: Hash + std::cmp::Eq, Value> HashMapCache<Key, Value> {
         }
     }
 
-    pub fn get_or_generate<F: FnOnce() -> Value>(
-        &self,
-        key: Key,
-        generator: F,
-    ) -> &Value {
+    pub fn get_or_generate<F: FnOnce() -> Value>(&self, key: Key, generator: F) -> &Value {
         let mut hashmap = self.contents.lock().unwrap();
         let entry = (*hashmap).entry(key);
         let inner_ptr = match entry {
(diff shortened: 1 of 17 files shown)


45930:
@mrmr1993, I am working through some linting errors.  It looks like you removed the function `add_lagrange_basis`, but it is still referenced by some srs (in test cases, AFAICT).  e.g.

```rs
error[E0599]: no method named `add_lagrange_basis` found for struct `poly_commitment::srs::SRS` in the current scope
    --> kimchi/src/circuits/expr.rs:3012:17
     |
3012 |             srs.add_lagrange_basis(constraint_system.domain.d1);
     |                 ^^^^^^^^^^^^^^^^^^ help: there is a method with a similar name: `get_lagrange_basis`

```

If that function is meant to be removed, what should be done with these references?

45930:
@mrmr1993 is there a test case explicitly for the case for which we are implementing this?  Would it be easy to whip one up?

querolita:
@45930 One of the references to the old `add_lagrange_basis` was in a comment, and the other was inside circuit-construction which is still not a member of the `Cargo.toml` (work in progress). Either way, I updated both in [a70e020](https://github.com/o1-labs/proof-systems/pull/2672/commits/a70e020d15d9feaf513f12fb35c92fad1e8be0ce).

querolita:
@45930 Because the issue was found in o1js, we probably want to create any remaining PRs for the mina / bindings / o1js sides with a test in o1js (probably just one of the few that were raised in https://github.com/o1-labs/o1js/issues/1411) to see the effects of this fix.

querolita (approved):

45930:
@querolita , thanks for the review!

Is it possible to test this branch in o1js locally?  There is a PR up for o1js: https://github.com/o1-labs/o1js/pull/1857 but I'm not sure how to get these changes into that branch to test with.

querolita:
@45930 Oh thanks for pointing me to the other PR, I was not aware that it was already being done. I will take a look at it locally to try to test the changes before merging 😄

poly-commitment/src/hash_map_cache.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[derive(Debug, Clone, Default)]
+pub struct HashMapCache<Key: Hash, Value> {
+    contents: Arc<Mutex<HashMap<Key, Value>>>,
+}
+
+impl<Key: Hash + std::cmp::Eq, Value> HashMapCache<Key, Value> {
+    pub fn new() -> Self {
+        HashMapCache {
+            contents: Arc::new(Mutex::new(HashMap::new())),
```
volhovm, line 15: Cool! I did something like this in parallel in my working branch: https://github.com/o1-labs/proof-systems/commit/444e07f4b19207096852e6d9beed09586dc471fd#diff-c89c45dc84ecd1f409b94d69ca374ed88678982f04f419b79435fd5ebff85b4fR362

volhovm: @mrmr1993 I'd suggest using `RWLock` though. Unless I'm misunderstanding, `Mutex` won't allow two concurrent reads, locking several parallel uses (which is the case in mina), and the upgrade to `RWLock` is easy. Any reasons not to?

PR #2673, opened 2024-10-08
Poly-commitment: rename elems into chunks

dannywillems:
Backporting from master
See https://github.com/o1-labs/proof-systems/pull/2658

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -1153,7 +1153,7 @@ where
 
             PolyComm {
                 // blinding_f - Z_H(zeta) * blinding_t
-                elems: vec![
+                chunks: vec![
                     blinding_f - (zeta_to_domain_size - G::ScalarField::one()) * blinding_t,
                 ],
             }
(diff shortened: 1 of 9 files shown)


volhovm (approved):
Looks good. There are just a few other lines that could be /also/ renamed:
```
ї> git grep elems
poly-commitment/src/commitment.rs:        let elems_size = Iterator::max(com.iter().map(|c| c.chunks.len())).unwrap();
poly-commitment/src/commitment.rs:        let mut chunks = Vec::with_capacity(elems_size);
poly-commitment/src/commitment.rs:        for chunk in 0..elems_size {
poly-commitment/src/srs.rs:    let t = G::BaseField::from_base_prime_field_elems(&base_fields).unwrap();
poly-commitment/src/srs.rs:        let num_elems = (n + srs_size - 1) / srs_size;
poly-commitment/src/srs.rs:        let mut chunks = Vec::with_capacity(num_elems);
poly-commitment/src/srs.rs:        for i in 0..num_elems {
poly-commitment/src/srs.rs:            // Append the 'partial Langrange polynomials' to the vector of elems chunks
```

PR #2684, opened 2024-10-09
Arrabiata: define correctly the different challenges used in the protocol

dannywillems:


kimchi/src/circuits/expr.rs
```diff
@@ -2870,6 +2848,32 @@ where
...
+    fn latex(&self, _cache: &mut HashMap<CacheId, Self>) -> String {
+        let col = self.col.latex(&mut HashMap::new());
+        match self.row {
+            Curr => col,
+            Next => format!("\\tilde{{{col}}}"),
+        }
+    }
+
+    fn text(&self, _cache: &mut HashMap<CacheId, Self>) -> String {
+        let col = self.col.text(&mut HashMap::new());
```
marcbeunardeau88, line 2869: why isn't the cache in the arg of the function given as arg to the recursive call of text ?

kimchi/src/circuits/expr.rs
```diff
@@ -3075,7 +3079,7 @@ where
                 });
                 res
             }
-            Atom(Cell(v)) => v.latex(),
+            Atom(Cell(v)) => v.latex(&mut HashMap::new()),
```
marcbeunardeau88, line 3082: why isn't the cache in the arg of the function given as arg to the recursive call ?

kimchi/src/circuits/expr.rs
```diff
@@ -3122,7 +3126,7 @@ where
                 });
                 res
             }
-            Atom(Cell(v)) => v.text(),
+            Atom(Cell(v)) => v.text(&mut HashMap::new()),
```
marcbeunardeau88, line 3129: why isn't the cache in the arg of the function given as arg to the recursive call ?

marcbeunardeau88 (approved):

kimchi/src/circuits/expr.rs
dannywillems: Not the same type, and also, it is unused in any case.

kimchi/src/circuits/expr.rs
dannywillems: Not the same type, and also, it is unused in any case.

kimchi/src/circuits/expr.rs
dannywillems: Not the same type, and also, it is unused in any case.

PR #2689, opened 2024-10-09
Arrabiata: gather all constraints for the IVC

dannywillems:
See comments. It might not be complete, but providing an interface for now.

### DIFF
--- a/arrabiata/src/constraints.rs
+++ b/arrabiata/src/constraints.rs
@@ -1,7 +1,7 @@
 use super::{columns::Column, interpreter::InterpreterEnv};
 use crate::{
     columns::{Gadget, E},
-    interpreter::Side,
+    interpreter::{self, Instruction, Side},
     MAX_DEGREE, NUMBER_OF_COLUMNS, NUMBER_OF_PUBLIC_INPUTS,
 };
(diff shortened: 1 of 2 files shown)


arrabiata/src/constraints.rs
```diff
@@ -318,3 +319,69 @@ impl<Fp: PrimeField> InterpreterEnv for Env<Fp> {
...
+        let mut env = self.clone();
+        // Resetting before running anything
+        env.reset();
+
+        let mut constraints = vec![];
+
+        // Poseidon constraints
+        // The constraints are the same for all the value given in parameter,
+        // therefore picking 0
+        interpreter::run_ivc(&mut env, Instruction::PoseidonNextRow(0));
```
marcbeunardeau88, line 349: why do we need to run the interpreter to gather the constraints ?
Does the result of this function depends on self ?

marcbeunardeau88 (approved):

arrabiata/src/constraints.rs
dannywillems: The steps of the IVC circuit is abstracted in the interpreter.
It depends on `self` only because there is a dependency on the Poseidon parameters. I would like to change that later (in particular, we need a setup).

PR #2690, opened 2024-10-09
Arrabiata/main: add multiple FIXME to remind the next steps

dannywillems:
Might not be complete. Only to keep it in the master branch for external contributors.

Eventually, all steps will be moved into a single method in the witness environment.

### DIFF
--- a/arrabiata/src/main.rs
+++ b/arrabiata/src/main.rs
@@ -123,6 +123,9 @@ pub fn main() {
         // Compute the accumulation of the public inputs/selectors
 
         // FIXME:
+        // Compute the accumulation of the blinders for the PCS
+
+        // FIXME:
         // Compute the accumulated error
 
         debug!(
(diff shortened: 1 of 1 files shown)


marcbeunardeau88:
I didn't check the exhaustivity of the fixme

arrabiata/src/main.rs
```diff
@@ -92,6 +92,42 @@ pub fn main() {
...
         env.compute_and_update_previous_commitments();
 
+        // FIXME:
+        // Absorb all commitments in the sponge.
+
+        // FIXME:
+        // Coin chalenges β and γ for the permutation argument
+
+        // FIXME:
+        // Compute the accumulator for the permutation argument
```
dannywillems, line 102: Adding also the absorption of the accumulator of the permutation argument + coining alpha.
@marcbeunardeau88 I know we discussed having only one global permutation argument, but it only works if we do have the same circuit over time, which might not be always the case.

PR #2692, opened 2024-10-09
o1vm/pickles: Remove linearization and add quotient polynomial evaluations

Fizzixnerd:
To Reviewer: I didn't add the quotient polynomial to the OpeningProof.  Not sure if it's needed, but will if it is.


Fizzixnerd:
I ended up adding it after talking with @marcbeunardeau88 on Slack.

dannywillems (approved):

PR #2697, opened 2024-10-10
o1vm/constraint : add a missing constraint for sel

marcbeunardeau88:
the selectors were only constrained to be boolean
We add the constraint that one and only one
of them is activated.


dannywillems (approved):

dannywillems (changes requested):
Removing my approval. Too fast apparently.

dannywillems (approved):

PR #2702, opened 2024-10-14
Arrabiata: prepare constraints for cross-terms computation

dannywillems:
This pull request prepares the constraints to compute the cross-terms after the different columns and arguments have been computed. Challenges must be treated as a variable, and will be in a future PR.

Previous attempt: https://github.com/o1-labs/proof-systems/pull/2695

### DIFF
--- a/arrabbiata/src/columns.rs
+++ b/arrabbiata/src/columns.rs
@@ -36,6 +36,22 @@ pub enum Column {
     X(usize),
 }
 
+/// Convert a column to a usize. This is used by the library [mvpoly] when we
+/// need to compute the cross-terms.
+/// For now, we only support the conversion of the private inputs, i.e. the `X`
+/// variant.
+/// Also, for now, the [mvpoly] library expects the columns to be mapped to a
+/// prime number as the prime representation is used under the hood. This must
+/// be changed.
+impl From<Column> for usize {
+    fn from(val: Column) -> usize {
+        match val {
+            Column::X(i) => i,
+            _ => unimplemented!("Only the private inputs are supported for now"),
+        }
+    }
+}
+
 pub struct Challenges<F: Field> {
     /// Challenge used to aggregate the constraints
     pub alpha: F,
--- a/arrabbiata/src/columns.rs
+++ b/arrabbiata/src/columns.rs
@@ -8,6 +8,8 @@ use std::{
 };
 use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
 
+use crate::NUMBER_OF_COLUMNS;
+
 /// This enum represents the different gadgets that can be used in the circuit.
 /// The selectors are defined at setup time, can take only the values `0` or
 /// `1` and are public.
@@ -38,16 +40,21 @@ pub enum Column {
 
 /// Convert a column to a usize. This is used by the library [mvpoly] when we
 /// need to compute the cross-terms.
-/// For now, we only support the conversion of the private inputs, i.e. the `X`
-/// variant.
-/// Also, for now, the [mvpoly] library expects the columns to be mapped to a
-/// prime number as the prime representation is used under the hood. This must
-/// be changed.
+/// For now, only the private inputs and the public inputs are converted,
+/// because there might not need to treat the selectors in the polynomial while
+/// computing the cross-terms (FIXME: check this later, but pretty sure it's the
+/// case).
+///
+/// Also, the [mvpoly::monomials] implementation of the trait [mvpoly::MVPoly]
+/// will be used, and the mapping here is consistent with the one expected by
+/// this implementation, i.e. we simply map to an increasing number starting at
+/// 0, without any gap.
 impl From<Column> for usize {
     fn from(val: Column) -> usize {
         match val {
             Column::X(i) => i,
-            _ => unimplemented!("Only the private inputs are supported for now"),
+            Column::PublicInput(i) => NUMBER_OF_COLUMNS + i,
+            Column::Selector(_) => unimplemented!("Selectors are not supported. This method is supposed to be called only to compute the cross-term and an optimisation is in progress to avoid the inclusion of the selectors in the multi-variate polynomial."),
         }
     }
 }
--- a/mvpoly/src/lib.rs
+++ b/mvpoly/src/lib.rs
@@ -254,6 +254,45 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
         u2: F,
     ) -> HashMap<usize, F>;
 
+    /// Compute the cross-terms of the given polynomial, scaled by the given
+    /// scalar.
+    ///
+    /// More explicitly, given a polynomial `P(X1, ..., Xn)` and a scalar α, the
+    /// method computes the the cross-terms of the polynomial `Q(X1, ..., Xn, α)
+    /// = α * P(X1, ..., Xn)`. For this reason, the method takes as input the
+    /// two different scalars `scalar1` and `scalar2` as we are considering the
+    /// scaling factor as a variable.
+    ///
+    /// This method is particularly useful when you need to compute a
+    /// (possibly random) combinaison of polynomials `P1(X1, ..., Xn), ...,
+    /// Pm(X1, ..., Xn)`, like when computing a quotient polynomial in the PlonK
+    /// PIOP, as the result is the sum of individual "scaled" polynomials:
+    /// ```text
+    /// Q(X_1, ..., X_n, α_1, ..., α_m) =
+    ///   α_1 P1(X_1, ..., X_n) +
+    ///   ...
+    ///   α_m Pm(X_1, ..., X_n) +
+    /// ```
+    ///
+    /// The polynomial must not necessarily be homogeneous. For this reason, the
+    /// values `u1` and `u2` represents the extra variable that is used to make
+    /// the polynomial homogeneous.
+    ///
+    /// The homogeneous degree is supposed to be the one defined by the type of
+    /// the polynomial `P`, i.e. `D`.
+    ///
+    /// The output is a map of `D` values that represents the cross-terms
+    /// for each power of `r`.
+    fn compute_cross_terms_scaled(
+        &self,
+        eval1: &[F; N],
+        eval2: &[F; N],
+        u1: F,
+        u2: F,
+        scalar1: F,
+        scalar2: F,
+    ) -> HashMap<usize, F>;
+
     /// Modify the monomial in the polynomial to the new value `coeff`.
     fn modify_monomial(&mut self, exponents: [usize; N], coeff: F);
 
--- a/arrabbiata/src/main.rs
+++ b/arrabbiata/src/main.rs
@@ -102,6 +102,12 @@ pub fn main() {
         // Compute the accumulator for the permutation argument
 
         // FIXME:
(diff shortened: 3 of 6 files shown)


arrabiata/src/columns.rs
```diff
@@ -36,6 +38,23 @@ pub enum Column {
     X(usize),
 }
 
+/// Convert a column to a usize. This is used by the library [mvpoly] when we
```
dannywillems, line 41: Oopsie, old comment. Updating.

mvpoly/src/lib.rs
```diff
@@ -261,3 +261,62 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
...
+/// Compute the cross terms of a list of polynomials. The polynomials are
+/// linearly combined using the power of a combiner, often called `α`.
+/// Powers of the combiner are considered as an additional variable of the
+/// multi-variate polynomials, therefore increasing the degree of the sum of all
+/// polynomials by one. In other words, if we combine (M + 1) polynomials
+/// `P_1(X_1, ..., X_N)`, ..., `P_{M + 1}(X_1, ..., X_N)`, we will compute the
+/// cross-terms of the polynomial:
+///
+/// ```text
+/// P(X_1, ..., X_N, α_1, ..., α_M) =     P_1(X_1, ..., X_N)
```
marcbeunardeau88, line 274: is it alpha_i or alpha^i ?
The text says alpha^i and the formula alpha_i

mvpoly/src/lib.rs
```diff
@@ -261,3 +261,62 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
...
+/// We notice that we simply do have a shifted linear combination of the
+/// polynomials. Based on the property that the cross-terms of a sum of
+/// polynomials is the sum of the cross-terms of the individual polynomials, and
+/// that shifting a polynomial with a new variable simply requires to shift the
+/// power of cross-terms by one and multiply by the evaluation at the new point,
+/// we get a simple algorithm.
+///
+/// All polynomials are supposed to be given as multivariate polynomials of the
+/// same degree and the same number of variables, even if they do not have the
+/// same number of variables nor the same degree.
```
marcbeunardeau88, line 289: is this an invariant the caller needs to respect ?

mvpoly/src/lib.rs
```diff
@@ -261,3 +261,62 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
...
+    combiner2: F,
+    eval1: [F; N],
+    eval2: [F; N],
+    u1: F,
+    u2: F,
+) -> HashMap<usize, F> {
+    let mut cross_terms = HashMap::new();
+    polys.into_iter().enumerate().for_each(|(i, poly)| {
+        let cross_terms_poly: HashMap<usize, F> = poly.compute_cross_terms(&eval1, &eval2, u1, u2);
+        cross_terms_poly.into_iter().for_each(|(p, value)| {
```
marcbeunardeau88, line 312: I would change the name of the p variable

mvpoly/src/lib.rs
```diff
@@ -261,3 +261,62 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
     /// variable in each monomial is of maximum degree 1.
     fn is_multilinear(&self) -> bool;
 }
+
+/// Compute the cross terms of a list of polynomials. The polynomials are
+/// linearly combined using the power of a combiner, often called `α`.
+/// Powers of the combiner are considered as an additional variable of the
+/// multi-variate polynomials, therefore increasing the degree of the sum of all
+/// polynomials by one. In other words, if we combine (M + 1) polynomials
```
marcbeunardeau88, line 269: I don't see why the degree is only increased by one
deg(\sum alpha^i p_i) != deg (p_i) +1

mvpoly/src/lib.rs
```diff
@@ -261,3 +261,62 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
...
+    const D: usize,
+    T: MVPoly<F, N, D>,
+>(
+    polys: Vec<T>,
+    combiner1: F,
+    combiner2: F,
+    eval1: [F; N],
+    eval2: [F; N],
+    u1: F,
+    u2: F,
```
marcbeunardeau88, line 307: I m not sure I understand what is computed here ?
the text mentioned combining the polys with one combiner, and we have two of them here.
I understand that the two combiners are the alpha from each instances we fold ? 
But then I don't get the assymetrical role ?

arrabiata/tests/constraints.rs
```diff
@@ -164,3 +166,46 @@ fn test_gadget_elliptic_curve_scaling() {
 
     helper_check_gadget_activated(instr, Gadget::EllipticCurveScaling);
 }
+
+// This test is mostly an example to show to compute the cross-terms when we do
+// have the expressions, and some evaluations.
+// It doesn't test anything in particular. It is mostly an "integration" test.
```
marcbeunardeau88, line 171: A way to make this test test thing would be to checked the cross term property
poly(a +r*b) = poly(a) + r^n poly(b) + \sum_i r^i cross_term_i(a,b)
Also a nit: I wouldn't call that an integration test

mvpoly/src/utils.rs
```diff
@@ -272,3 +272,35 @@ pub fn compute_indices_nested_loop(nested_loop_sizes: Vec<usize>) -> Vec<Vec<usi
...
         .collect()
 }
+
+/// Same than `compute_indices_nested_loop` but with an upper bound on the sum
+/// of the indices.
+///
+/// It is useful when the number of nested loop is large and the size of the
+/// loops is large as well. `filter_map` is used to avoid computing the indices
+/// that are not needed and avoid a `capacity_overflow` error.
+pub fn compute_indices_nested_loop_upper_bound(
```
marcbeunardeau88, line 282: I would not expose this function, keep this implementation, and expose compute_indices_nested_loop implemented with that implementation, while computing the upper bound on the fly with a max.
Then we have one function instead of two, which is the efficient variant, and only need one argument

marcbeunardeau88 (changes requested):
I'd like a test on the defining property of cross terms to convince me of the correctness of the algorithm, which I wasn't able to check manually

mvpoly/src/lib.rs
```diff
@@ -309,13 +311,19 @@ pub fn compute_combined_cross_terms<
...
-            *entry += alpha_i * value;
+            // Handling 0^0 = 1
+            let entry = cross_terms.entry(p).or_insert(F::zero());
+            if combiner1 != F::zero() {
+                let alpha_i = combiner1.pow([i as u64]);
+                *entry += alpha_i * value;
+            };
             // Computing α'^i * value, and adding it to the next power of r
-            let alpha_i = combiner2.pow([i as u64]);
+            // Handling 0^0 = 1
```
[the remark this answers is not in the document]
dannywillems: Yes, but it was monstly for the tests (see test directory). I didn't handle it initially. Even if the case is negligible, I think it is ok to have it. Maybe we should fail instead?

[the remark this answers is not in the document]
marcbeunardeau88: I'd ignore it to simplify the code, and remove the testing with zero, but do as you feel

arrabiata/tests/constraints.rs
dannywillems: "Also a nit: I wouldn't call that an integration test"

That's the reason "integration" is double quoted.

mvpoly/src/lib.rs
dannywillems: See the comment just above:
```
/// Powers of the combiner are considered as an additional variable of the
/// multi-variate polynomials, therefore increasing the degree of the sum of all
/// polynomials by one.
```

Instead of having a power, we do consider them as a different variable.

mvpoly/src/lib.rs
dannywillems: See https://github.com/o1-labs/proof-systems/pull/2702/files#r1820537584

mvpoly/src/lib.rs
dannywillems: See https://hackmd.io/@dannywillems/Syo5MBq90#Computation-of-cross-terms-for-shifted-polynomials

dannywillems:
> I could not convinced my self of the correctness of the algorithm, and there is no test that reassure me.

See https://hackmd.io/@dannywillems/Syo5MBq90#Computation-of-cross-terms-for-shifted-polynomials

mvpoly/src/lib.rs
dannywillems: `p` is for `power`, which is the power of `r`.

arrabbiata/src/columns.rs
```diff
@@ -38,16 +40,21 @@ pub enum Column {
...
+/// Also, the [mvpoly::monomials] implementation of the trait [mvpoly::MVPoly]
+/// will be used, and the mapping here is consistent with the one expected by
+/// this implementation, i.e. we simply map to an increasing number starting at
+/// 0, without any gap.
 impl From<Column> for usize {
     fn from(val: Column) -> usize {
         match val {
             Column::X(i) => i,
-            _ => unimplemented!("Only the private inputs are supported for now"),
+            Column::PublicInput(i) => NUMBER_OF_COLUMNS + i,
```
marcbeunardeau88, line 56: Doesn't having (at least) a full columns for PI can create issue on the perfs of the verifier.
More importantely,  this translate in the folder verifier which will handle a big PI.
We can use the PLONK way to handle PI to avoid a PI being linear in the number of rows

Since this is not the topic of the PR, we can see that later

dannywillems: Yes, I want to investigate other solutions to handle PI.

mvpoly/src/lib.rs
```diff
@@ -254,6 +254,45 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
...
+    ///
+    /// The polynomial must not necessarily be homogeneous. For this reason, the
+    /// values `u1` and `u2` represents the extra variable that is used to make
+    /// the polynomial homogeneous.
+    ///
+    /// The homogeneous degree is supposed to be the one defined by the type of
+    /// the polynomial `P`, i.e. `D`.
+    ///
+    /// The output is a map of `D` values that represents the cross-terms
+    /// for each power of `r`.
```
marcbeunardeau88, line 285: I'd add in the description of the funciton something like the following to summarize
let res = compute_cross_terms_scaled(P,eval1,eval2,u1,u2,scalar1,scalar2)
let P'[X_1, ..., X_n,U,Alpha] be P homogenised by U and scaled by Alpha
Then for all r, 
  P'(eval_1,u1,scalar1) + r^{D+1} P'(eval_2,u2,scalar2) + \sum_{i=1}^{D} r^i res[i] = 
   P'(eval_1 +r*eval_2, u1 + r*u_2,scalar1+r*scalar2)

dannywillems: Sure, I'll do it in a follow-up.

marcbeunardeau88 (approved):
I did not check the maths but the test is convincing enough

mvpoly/src/lib.rs
dannywillems: Updated in the HackMD.

PR #2703, opened 2024-10-14
Kimchi: move more berkeley expr related code into berkeley_columns

dannywillems:
This must be done as part of the follow-up of https://github.com/o1-labs/proof-systems/pull/2566#discussion_r1764597368. cc @marcbeunardeau88 to continue

The feature flags should also be moved into berkeley_columns.rs


dannywillems:
Closing this as old. And there is a non negligible probability we're gonna change the design in the future.

PR #2711, opened 2024-10-15
CI: enforcing setup-ocaml version fixing darcs issue

dannywillems:


dannywillems:
Closing as 3 should target the latest version already. Fast propagation

PR #2710, opened 2024-10-14
MVPoly: make the expression generic over the challenge term

dannywillems:

PR #2713, opened 2024-10-15
MVPoly: improve the library + fix sub in monomials

dannywillems:
See individual commits.
It is mostly from https://github.com/o1-labs/proof-systems/pull/2702/ to split it in small target chunks.
Built on top of https://github.com/o1-labs/proof-systems/pull/2710.

### DIFF
--- a/mvpoly/src/monomials.rs
+++ b/mvpoly/src/monomials.rs
@@ -1,5 +1,12 @@
 use ark_ff::{One, PrimeField, Zero};
-use kimchi::circuits::expr::{ConstantExpr, Expr};
+use kimchi::circuits::{
+    expr::{
+        ConstantExpr, ConstantExprInner, ConstantTerm, Expr, ExprInner,
+        Operations::{self, *},
+        Variable,
+    },
+    gate::CurrOrNext,
+};
 use num_integer::binomial;
 use rand::RngCore;
 use std::{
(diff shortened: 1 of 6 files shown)


mvpoly/src/monomials.rs
```diff
@@ -427,12 +434,72 @@ impl<const N: usize, const D: usize, F: PrimeField> MVPoly<F, N, D> for Sparse<F
...
+        let exponents: [usize; N] = std::array::from_fn(|i| if i == var_usize { 1 } else { 0 });
+        monomials.insert(exponents, F::one());
+        Self { monomials }
+    }
+
     fn from_expr<Column: Into<usize>, ChallengeTerm: Clone>(
         expr: Expr<ConstantExpr<F, ChallengeTerm>, Column>,
     ) -> Self {
-        prime::Dense::from_expr(expr).into()
+        match expr {
```
dannywillems, line 448: I'm refactoring this for info. Coming in a small future PR. It is the same than prime.rs. We can generalize it. We only need to deal with the variables.

dannywillems: See https://github.com/o1-labs/proof-systems/pull/2714

mvpoly/src/prime.rs
```diff
@@ -239,15 +242,26 @@ impl<F: PrimeField, const N: usize, const D: usize> MVPoly<F, N, D> for Dense<F,
                     let degree = naive_prime_factors(*idx, &mut prime_gen)
                         .iter()
                         .fold(0, |acc, (_, d)| acc + d);
-                    if degree > max_degree {
+                    if degree > max_degree || rng.gen_range(0..10) == 0 {
+                        // Adding zero coefficients one every 10 monomials
                         F::zero()
                     } else {
                         F::rand(rng)
```
marcbeunardeau88, line 249: I'd simply replace this line with a generator biased toward 0

mvpoly/tests/monomials.rs
```diff
@@ -381,6 +381,36 @@ fn test_mvpoly_compute_cross_terms_degree_six() {
     assert_eq!(lhs, rhs);
 }
 
+// The cross-terms of a sum of polynomials is the sum of the cross-terms, per
+// power.
+#[test]
+fn test_mvpoly_pbt_cross_terms_addition() {
```
marcbeunardeau88, line 340: I'd check p_1 + lambda p_2 to check the full linearity property

mvpoly/src/prime.rs
dannywillems: > with a generator biased toward 0

How would you do that?

mvpoly/tests/monomials.rs
dannywillems: Good idea!

mvpoly/src/prime.rs
marcbeunardeau88: generate a nb between 0-10, if that number is 10 return 0, otherwise generate uniformly a field elment

dannywillems: But that's what the code does, doesn't?

mvpoly/tests/monomials.rs
dannywillems: Well no. It isn't true.

PR #2715, opened 2024-10-15
MVPoly: improve from_variable and support "next row" in from_expr

dannywillems:


marcbeunardeau88:
I understand you are using a new variable to handle next row columns
shouldn't you use the same variable multiplied by omega ?

dannywillems:
> I understand you are using a new variable to handle next row columns shouldn't you use the same variable multiplied by omega ?

No. `mvpoly` contains the evaluation, not the point we evaluate at.

marcbeunardeau88 (approved):

dannywillems:
Just discussed with @marcbeunardeau88: we should have something more descriptive than just integers, something like `Atoms` that contains the description `Challenge`, `Variable` (with curr and next row).

PR #2719, opened 2024-10-17
Arrabiata: introduce trait ArrabiataCurve to englobe all data a curve must implement to be used with Arrabiata

dannywillems:


arrabiata/src/curve.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use ark_ec::short_weierstrass::Affine;
+use kimchi::curve::{pallas_endos, vesta_endos};
+use mina_curves::pasta::curves::{pallas::PallasParameters, vesta::VestaParameters};
+use mina_poseidon::poseidon::ArithmeticSpongeParams;
+use poly_commitment::commitment::{CommitmentCurve, EndoCurve};
+
+/// Represents additional information that a curve needs in order to be used
+/// with Arrabiata.
+pub trait ArrabiataCurve: CommitmentCurve + EndoCurve {
```
marcbeunardeau88, line 31: Is it necessarry to introduce another trait ?
I don't see anything that isn't in KimchiCurve
(Which is not a surprise, as both Kimchi and Arrabiata use a cycle of curve for plonk verification operations)

marcbeunardeau88 (approved):

PR #2722, opened 2024-10-18
WIP: remove ExprOps

dannywillems:


dannywillems:
Closing as it doesn't give any value to the code atm. I guess rewrite it with the new interpreter trait would be better, but leaving it for future work, if we want.

dannywillems:
We should have something like `BerkeleyInterpreter`.

PR #2605, opened 2024-09-24
o1vm/pickles: unit tests from ethereum-optimism

dannywillems:

PR #2726, opened 2024-10-21
WIP: compile other MIPS programs

dannywillems:
Also see https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2
And for testing https://github.com/o1-labs/proof-systems/pull/2605
Use [elf](https://docs.rs/elf/latest/elf/) to read initial memory/register state.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,4 +128,7 @@ generate-doc:
 		@echo "The documentation is available at: ./target/doc"
 		@echo ""
 
+o1vm-test:
+		@mips-linux-gnu-as -o ./o1vm/resources/programs/hello_world.bin ./o1vm/resources/programs/hello_world.mips
+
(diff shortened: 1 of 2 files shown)


dannywillems:
For however who will work on this, see an ELF loader [here](https://github.com/o1-labs/proof-systems/blob/8fe31f4b347999fe8bc5775ee4f643f86752d8ac/o1vm/src/elf_loader.rs)

dannywillems:
Note that the ELF loader has been merged. Should work for small programs, but still need to be polished.
It will be part of the improvements for the RISC-V interpreter.

dannywillems:
Closing in favor of https://github.com/o1-labs/proof-systems/pull/2906 and https://github.com/o1-labs/proof-systems/pull/2911 and https://github.com/o1-labs/proof-systems/pull/2886

PR #2727, opened 2024-10-21
o1vm/riscv32i

dannywillems:
Drafting a RISC-V (32i version) flavor of o1vm, as the community is asking for it. Following https://riscv.org/wp-content/uploads/2019/12/riscv-spec-20191213.pdf

Gonna be a fun project with @svv232 to learn more about arithmetisation and PlonK and proof-systems.


dannywillems:
Note: we should get rid of meta.json. It is simply the list of symbols with their names in string. It is used for debugging while running the VM. We can use the ELF rust library to preprocess this.
See `test_riscv_elf` for examples

dannywillems:
We should think more about the memory layout. Leaving this comment for the future.

dannywillems:
TODO: add a long program (i.e. huge binary) to verify long traces, in particular when a lot of memory is used.
The memory layout used at the moment is the same as in Cannon. We should change that.
We can create long programs by using [o1vm-test](https://github.com/dannywillems/o1vm-test/) and adding dummy functions.

dannywillems:
The interpreter is now implemented, and available in master: https://github.com/o1-labs/proof-systems/tree/master/o1vm/src/interpreters/riscv32im

PR #2732, opened 2024-10-26
Adding new range check32 table implementation 

svv232:


o1vm/src/lookups.rs
```diff
@@ -275,4 +289,13 @@ impl<F: Field> FixedLookupTables<F> for LookupTable<F> {
                 .collect(),
         }
     }
+
+    fn table_range_check_32() -> Self {
```
dannywillems, line 293: You really want to keep 2^32 * 4 bytes in memory? This is madness!

o1vm/src/interpreters/mips/interpreter.rs
```diff
@@ -554,6 +554,24 @@ pub trait InterpreterEnv {
         self.lookup_2bits(value);
     }
 
+    // Adds a lookup to the ByteLookup table for each byte of a 32-bit value
+    fn lookup_32bits(&mut self, value: &Self::Variable) {
```
dannywillems, line 558: What are you trying to achieve here?

dannywillems (changes requested):
Too much memory used when checking entry

o1vm/src/interpreters/mips/interpreter.rs
volhovm: + yeah, why is it not enough to have two 16bit tables?...

dannywillems:
Might be useful to have a look at https://github.com/o1-labs/proof-systems/pull/2395/

PR #2733, opened 2024-10-26
range32 checking for added to mips

svv232:
I have yet to add the contraints for this PR, I am currently unsure what to constrain as 32 and 64 bits.

### DIFF
--- a/o1vm/src/interpreters/mips/constraints.rs
+++ b/o1vm/src/interpreters/mips/constraints.rs
@@ -168,6 +168,10 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         Self::Variable::constant(Operations::from(Literal(Fp::from(x))))
     }
 
+    fn constant64(x: u64) -> Self::Variable {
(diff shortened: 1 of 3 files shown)


o1vm/src/interpreters/mips/constraints.rs
```diff
@@ -168,6 +168,10 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
         Self::Variable::constant(Operations::from(Literal(Fp::from(x))))
     }
 
+    fn constant64(x: u64) -> Self::Variable {
+        Self::Variable::constant(Operations::from(Literal(Fp::from(x))))
```
dannywillems, line 172: Note that you can always use `Self::Variable::from(x as u64)`.

dannywillems (changes requested):
This seems wrong. The SRS is not big enough, in particular.

dannywillems:
See https://github.com/o1-labs/proof-systems/pull/2732#issuecomment-2520130536

PR #2737, opened 2024-10-31
Move tests to the subdirectory (develop)

volhovm:
This has been already done in (numerous PRs in) `master`, which I can't cherry-pick because there are too many & they're unfortunately pre-arkworks changes, so I'd have to edit every commit, which is more work than to redo a lot of `git mv`. 

This PR is moving the tests in develop.


volhovm:
If we're starting to use proof-system's `master` in `mina`, this PR is pretty much obsolete.

PR #2739, opened 2024-11-01
Compute the necessarry inversions of the witness in batch after it's creation

marcbeunardeau88:
Follow-up of https://github.com/o1-labs/proof-systems/pull/2047
but we proceed differently.
The scratch state now contains field element which are marked to be inversed or not, using a variant.
The Batch inversion is then done when creating the ProofInput.
We first create an intermediary type NotInversedProofInput, containing the same variant as the scratch state, and implement a conversion to ProofInput  which does the batch inversion

### DIFF
--- a/o1vm/src/interpreters/mips/witness.rs
+++ b/o1vm/src/interpreters/mips/witness.rs
@@ -65,6 +65,25 @@ impl SyscallEnv {
         }
     }
 }
+// This enum will populate the scracth state
+// The purpose is to perform a batch inversion at the end
+// We symbolically mark the inversion instead of computing them
+#[derive(Clone, Copy)]
+pub enum ToInverseOrNot<Fp> {
+    ToInverse(Fp),
+    NotToInverse(Fp),
+}
+
+impl<Fp: Field> ToInverseOrNot<Fp> {
+    pub fn to_field(self) -> Fp {
+        match self {
+            ToInverseOrNot::NotToInverse(x) => x,
+            ToInverseOrNot::ToInverse(x) => {
+                Fp::inverse(&x).expect("A zero value was marked as ToInverse")
+            }
+        }
+    }
+}
 
 /// This structure represents the environment the virtual machine state will use
 /// to transition. This environment will be used by the interpreter. The virtual
@@ -80,7 +99,7 @@ pub struct Env<Fp, PreImageOracle: PreImageOracleT> {
     pub registers: Registers<u32>,
     pub registers_write_index: Registers<u64>,
     pub scratch_state_idx: usize,
-    pub scratch_state: [Fp; SCRATCH_SIZE],
+    pub scratch_state: [ToInverseOrNot<Fp>; SCRATCH_SIZE],
     pub halt: bool,
     pub syscall_env: SyscallEnv,
     pub selector: usize,
@@ -92,8 +111,8 @@ pub struct Env<Fp, PreImageOracle: PreImageOracleT> {
     pub hash_counter: u64,
 }
 
-fn fresh_scratch_state<Fp: Field, const N: usize>() -> [Fp; N] {
-    array::from_fn(|_| Fp::zero())
+fn fresh_scratch_state<Fp: Field, const N: usize>() -> [ToInverseOrNot<Fp>; N] {
+    array::from_fn(|_| ToInverseOrNot::NotToInverse(Fp::zero()))
 }
 
 const KUNIT: usize = 1024; // a kunit of memory is 1024 things (bytes, kilobytes, ...)
@@ -952,7 +971,9 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> Env<Fp, PreImageOracle> {
 
     pub fn write_field_column(&mut self, column: Column, value: Fp) {
         match column {
-            Column::ScratchState(idx) => self.scratch_state[idx] = value,
+            Column::ScratchState(idx) => {
+                self.scratch_state[idx] = ToInverseOrNot::NotToInverse(value)
+            }
             Column::InstructionCounter => panic!("Cannot overwrite the column {:?}", column),
             Column::Selector(s) => self.selector = s,
         }
--- a/o1vm/src/interpreters/mips/witness.rs
+++ b/o1vm/src/interpreters/mips/witness.rs
@@ -375,7 +375,7 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
             self.write_column(position, 0);
             0
         } else {
-            self.write_field_column(position, Fp::from(*x).inverse().unwrap());
+            self.write_field_inverse_column(position, Fp::from(*x));
             1 // Placeholder value
         }
     }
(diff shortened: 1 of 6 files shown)


o1vm/src/interpreters/mips/witness.rs
```diff
@@ -80,7 +99,7 @@ pub struct Env<Fp, PreImageOracle: PreImageOracleT> {
     pub registers: Registers<u32>,
     pub registers_write_index: Registers<u64>,
     pub scratch_state_idx: usize,
-    pub scratch_state: [Fp; SCRATCH_SIZE],
+    pub scratch_state: [ToInverseOrNot<Fp>; SCRATCH_SIZE],
```
dannywillems, line 102: What do you think about using another type of column which cannot be used in further computations, as it is the case for here?
I would approach it with something like `allocate_auxiliary_variable` (or any better name that can include the idea that it is only to satisfy a single constraint, and should not be used for computation) in the `InterpreterEnv` interface and add a new column next to `ScratchState`.

dannywillems: I do also think, but I need to dig more into the idea, that we can simply commit to the polynomial whose evaluations are not the inverse. I do not know if it will improve the efficiency though. We can discuss it offline.

querolita: @dannywillems 
> add a new column next to ScratchState
That is precisely what I suggested in https://github.com/o1-labs/proof-systems/pull/2047

o1vm/src/legacy/main.rs
```diff
@@ -159,7 +159,7 @@ pub fn main() -> ExitCode {
         for i in 0..N_MIPS_REL_COLS {
             match i.cmp(&SCRATCH_SIZE) {
                 Ordering::Less => mips_trace.trace.get_mut(&instr).unwrap().witness.cols[i]
-                    .push(mips_wit_env.scratch_state[i]),
+                    .push(mips_wit_env.scratch_state[i].to_field()),
```
querolita, line 162: Basically here, it looks incremental to me, as opposed to batch inverses. Let's wait for some benchmarks perhaps, but my intuition says it will be similar to what we currently have but just more intricate 🤔

o1vm/src/pickles/proof.rs
```diff
@@ -10,11 +11,87 @@ pub struct WitnessColumns<G, S> {
...
+        // Perform the inversion
+        // FIXME: we go trough vec to get the array required by arkworks
+        // This is ugly
+        let mut to_inverse_vec: Vec<F> = Vec::with_capacity(to_inverse_map.len());
+        for v in to_inverse_map.values() {
+            to_inverse_vec.push(*v);
+        }
+        let to_inverse_slice = to_inverse_vec.as_mut_slice();
+
+        ark_ff::batch_inversion::<F>(to_inverse_slice);
```
querolita, line 59: Oh, so here's the batch inversion... Then I am not sure about what `to_field()` is being used for.

o1vm/src/pickles/proof.rs
```diff
@@ -10,11 +11,87 @@ pub struct WitnessColumns<G, S> {
...
+    // and returns a usable proof input
+    fn from(not_inversed: NotInversedProofInputs<F>) -> ProofInputs<F> {
+        // Initialising the result
+        let mut res = ProofInputs::new(not_inversed.scratch[1].len());
+        res.evaluations.error = not_inversed.error;
+        res.evaluations.instruction_counter = not_inversed.instruction_counter;
+        res.evaluations.selector = not_inversed.selector;
+
+        // Collecting the values to inverse
+        let to_inverse_map: &mut HashMap<_, F> = &mut HashMap::new();
```
querolita, line 37: Okay, I kind of like the idea of having a hash map to avoid repetitions of inverses computations, but at the same time I wonder if the memory overhead is worth the small chance of repeating field elements to be inverted (unless it's very obvious that we always invert the same set of "trivial" values, and by that I mean much smaller sample size than the actual field domain).

querolita (commented):
In general I would say that the same idea can be achieved with a less complex approach. I don't like the verbosity added to represent plain fields vs "to inverse" fields and would prefer to keep it simpler. I wonder why the additional column alias did not seem like a good idea, to me that one had an acceptable amount of "code overhead". I do like the hash map though, if it really compensates the memory consumption (we would need to see examples of real executions). I would feel more inclined to iterate over these two approaches and perform a quick benchmark on the result.

o1vm/src/interpreters/mips/witness.rs
dannywillems: Oopsie. I didn't check before :see_no_evil:. I do have a high preference for @querolita solution.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -65,6 +65,25 @@ impl SyscallEnv {
         }
     }
 }
+// This enum will populate the scracth state
+// The purpose is to perform a batch inversion at the end
+// We symbolically mark the inversion instead of computing them
+#[derive(Clone, Copy)]
+pub enum ToInverseOrNot<Fp> {
```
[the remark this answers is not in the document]
marcbeunardeau88: The reason I don't like adding columns, is that I feel that it mixes up concept.
Marking the elements to be inverted seems cleaner to me.
If you do not like the variant type (which I admit is verbose), I can create a set keeping track of the element to be inverted.
What do you think ?

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -65,6 +65,25 @@ impl SyscallEnv {
...
+    ToInverse(Fp),
+    NotToInverse(Fp),
+}
+
+impl<Fp: Field> ToInverseOrNot<Fp> {
+    pub fn to_field(self) -> Fp {
+        match self {
+            ToInverseOrNot::NotToInverse(x) => x,
+            ToInverseOrNot::ToInverse(x) => {
```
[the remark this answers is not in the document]
marcbeunardeau88: It does not indeed, this was to have something working easily.
A later commit removes the to_field function (not in the legacy folder though) and performs the batch inverse

o1vm/src/legacy/main.rs
marcbeunardeau88: Indeed it's not bacthed here, but we do not maintain the legacy folder (or should we ?)

o1vm/src/pickles/proof.rs
marcbeunardeau88: For the legacy, but I can change that and remove to field

o1vm/src/pickles/proof.rs
marcbeunardeau88: The chance of inverting the same element depends heavily on the use case, so it's hard to say.
Honnestly, I was using a hash map to have a key value mapping more than anything else.
As for the memory overhead, which one are you referring to ?

o1vm/src/legacy/main.rs
querolita: Ah I see, I didn't realize of the directories

o1vm/src/pickles/proof.rs
querolita: I was referring to the actual hash map structure. Not sure how large it can take in memory.

o1vm/src/interpreters/mips/witness.rs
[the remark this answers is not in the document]
querolita: I think I am biased with the column approach. Maybe it is true that adding a column alias just for to-be-inverted values is a mix of concept. But to me that seemed like a quick way to mark the elements that needed to be inverted. If there's a simpler way to do it without needing an extra column and without that wrapping interface, I would love to look at it. But if it is not straightforward, I wouldn't spend too much time trying to make it look "nice", if a simple enough solution works for now. At least so we can start benchmarking. Perhaps, after doing so, we can come up with alternatives to make it more performant.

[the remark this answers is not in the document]
dannywillems: >  Marking the elements to be inverted seems cleaner to me.

No, it is not cleaner!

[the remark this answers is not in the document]
dannywillems: Please, let's keep the version @querolita started.

dannywillems (changes requested):
Please come back to @querolita initial idea.

dannywillems:
Can we close this @marcbeunardeau88?

PR #2741, opened 2024-11-01
o1vm: implement a simple ELF loader

dannywillems:
This is only a first draft. In future PR, FIXME will be implemented.
It is mostly to start moving commits from the pull request https://github.com/o1-labs/proof-systems/pull/2727


dannywillems:
The stack pointer must be set to a different address. The loader decides where to locate the first address. With @svv232 we think about having two continuous array, one for the heap, and one for the stack. It would be a follow-up of https://github.com/o1-labs/proof-systems/pull/2042.

shimkiv (approved):

dannywillems:
I'm taking the responsibility to improve it in the near future.

PR #2745, opened 2024-11-04
o1vm/riscv32i: introduce column module

dannywillems:


o1vm/src/interpreters/riscv32i/mod.rs
```diff
@@ -1,2 +1,15 @@
+/// The minimal number of columns required for the VM
+// FIXME: the value will be updated when the interpreter is fully
+// implemented. Using a small value for now.
+pub const SCRATCH_SIZE: usize = 20;
+
+/// Number of instructions in the ISA
+// FIXME: the value might not be correct. It will be updated when the
+// interpreter is fully implemented.
+pub const INSTRUCTION_SET_SIZE: usize = 47;
```
svv232, line 9: These values are currently too small to run with the interpreter, dynamic testing yielded an oob when running tests that are yet to be merged. I made them larger in https://github.com/o1-labs/proof-systems/pull/2754/files . We should fix these to be correct in a follow up later down the stack as also mentioned in #2754

svv232 (approved):

PR #2754, opened 2024-11-12
adding no-action test, some i instructions, and single syscall impl

svv232:


o1vm/src/interpreters/riscv32i/mod.rs
```diff
@@ -1,12 +1,12 @@
...
 // FIXME: the value will be updated when the interpreter is fully
 // implemented. Using a small value for now.
-pub const SCRATCH_SIZE: usize = 20;
+pub const SCRATCH_SIZE: usize = 80;
 
 /// Number of instructions in the ISA
 // FIXME: the value might not be correct. It will be updated when the
 // interpreter is fully implemented.
-pub const INSTRUCTION_SET_SIZE: usize = 47;
+pub const INSTRUCTION_SET_SIZE: usize = 90;
```
dannywillems, line 9: I guess we can set the right value now?

o1vm/src/interpreters/riscv32i/mod.rs
```diff
@@ -1,12 +1,12 @@
 /// The minimal number of columns required for the VM
 // FIXME: the value will be updated when the interpreter is fully
 // implemented. Using a small value for now.
-pub const SCRATCH_SIZE: usize = 20;
+pub const SCRATCH_SIZE: usize = 80;
```
dannywillems, line 4: I guess we can try to set a value the no-action program requires atm?

o1vm/src/interpreters/riscv32i/interpreter.rs
```diff
@@ -1077,8 +1076,157 @@ pub fn interpret_rtype<Env: InterpreterEnv>(_env: &mut Env, _instr: RInstruction
...
+            env.set_next_instruction_pointer(next_instruction_pointer + Env::constant(4u32));
+        }
+        IInstruction::JumpAndLinkRegister => {
+            let addr = env.read_register(&rs1);
+            // jalr:
+            //  t  = pc+4;
+            //  pc = (x[rs1] + sext(offset)) & ∼1; <- NOT NOW
+            //  pc = (x[rs1] + sext(offset)); <- PLEASE FIXME
+            //  x[rd] = t
+            // copying mips for now to match deugger
```
dannywillems, line 1213: I guess we can remove this?

o1vm/src/interpreters/riscv32i/interpreter.rs
```diff
@@ -1077,8 +1076,157 @@ pub fn interpret_rtype<Env: InterpreterEnv>(_env: &mut Env, _instr: RInstruction
...
+            + v0
+    };
+
+    /* fetch opcode from instruction bit 0 - 6 for a total len of 7 */
+    let opcode = {
+        let pos = env.alloc_scratch();
+        unsafe { env.bitmask(&instruction, 7, 0, pos) }
+    };
+    /* verify opcode is 7 bits */
+    env.range_check8(&opcode, 7);
```
dannywillems, line 1102: I think we should have a lookup table with the opcode. Not any 7 bits value is a correct opcode.

o1vm/src/interpreters/riscv32i/interpreter.rs
```diff
@@ -1077,8 +1076,157 @@ pub fn interpret_rtype<Env: InterpreterEnv>(_env: &mut Env, _instr: RInstruction
...
+
+    /* decode and parse bits from the full 32 bit instruction in accordance with the Rtype riscV spec
```
dannywillems, line 1104:
```suggestion
    /* decode and parse bits from the full 32 bits instruction in accordance with the Rtype riscV spec
```

dannywillems, line 1104:
```suggestion
    /* decode and parse bits from the full 32 bit instruction in accordance with the Rtype RISC-V32i spec
```

dannywillems (approved):
Would be nice to have the changes in a follow-up. The lookup check on the opcode is important.

dannywillems:
Conflicts should be solved.

o1vm/src/interpreters/riscv32i/mod.rs
svv232: i just chose this value because it worked during testing, a size of 20 was too small.

PR #2757, opened 2024-11-13
Logup for o1vm

Fizzixnerd:

PR #2758, opened 2024-11-14
o1vm/pickles: regression test for proof size

dannywillems:
Gonna use serialize just after. Only a draft.


dannywillems:
I'm leaving this in draft for now, as #2757 will add the fixed tables in the proofs, and will therefore change the Proof structure.

PR #2760, opened 2024-11-19
Compute the necessarry inversions of the witness in batch after it's creation

marcbeunardeau88:
Follow-up of https://github.com/o1-labs/proof-systems/pull/2047
but we proceed differently.
The witness env now contains a set marking which element are to be inverted.
The Batch inversion is then done when creating the ProofInput.
We first create an intermediary type NotInversedProofInput, and implement a conversion to ProofInput which does the batch inversion

### DIFF
--- a/o1vm/src/interpreters/mips/witness.rs
+++ b/o1vm/src/interpreters/mips/witness.rs
@@ -962,6 +962,22 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> Env<Fp, PreImageOracle> {
         }
     }
 
+    pub fn write_inverse_column(&mut self, column: Column, value: Fp) {
+        assert!(value != Fp::zero());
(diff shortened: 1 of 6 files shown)


o1vm/src/interpreters/mips/witness.rs
```diff
@@ -90,6 +91,7 @@ pub struct Env<Fp, PreImageOracle: PreImageOracleT> {
     pub preimage_key: Option<[u8; 32]>,
     pub keccak_env: Option<KeccakEnv<Fp>>,
     pub hash_counter: u64,
+    pub scratch_to_inverse: HashSet<usize>,
```
dannywillems, line 94: What about using an index instead?

dannywillems:
Can we close this @marcbeunardeau88?

PR #2787, opened 2024-11-20
o1vm/riscv32: sketch interpretation of mtype instruction

dannywillems:


svv232:
I verified that the docs match the spec

PR #2790, opened 2024-11-20
o1vm/riscv32: implement mul_hi_signed

dannywillems:


o1vm/src/interpreters/riscv32im/witness.rs
```diff
@@ -441,6 +441,21 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        x: &Self::Variable,
+        y: &Self::Variable,
+        position: Self::Position,
+    ) -> Self::Variable {
+        let x: u32 = (*x).try_into().unwrap();
+        let y: u32 = (*y).try_into().unwrap();
+        let res = (((x as i32) as i64) * ((y as i32) as i64)) as u64;
+        let res = (res >> 32) as u32;
+        let res = res as u64;
+        self.write_column(position, res);
```
svv232, line 455: why not 

```suggestion
        let x: i32 = (*x).try_into().unwrap();
        let y: i32 = (*y).try_into().unwrap();
        let res = (x as i64) * (y as i64);
        let res = (res >> 32) as i32;
        let res = res as u64;
        self.write_column(position, res);
```

svv232: i'm just preserving the signed bit for as long as possible through each operation.

dannywillems: I updated.

dannywillems: See 75503032b4306e3743f936c413c3c4e4857abced

svv232 (approved):

PR #2794, opened 2024-11-20
o1vm/riscv32: implement M type instruction Mul

dannywillems:


svv232 (approved):

o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -2146,7 +2146,17 @@ pub fn interpret_mtype<Env: InterpreterEnv>(env: &mut Env, instr: MInstruction)
 
     match instr {
         MInstruction::Mul => {
-            unimplemented!("Mul")
```
svv232, line 2149: it would be nice to comment the implementation from https://msyksphinz-self.github.io/riscv-isadoc/html/rvm.html

PR #2799, opened 2024-11-20
o1vm/riscv32: implement div_signed

dannywillems:


o1vm/src/interpreters/riscv32im/witness.rs
```diff
@@ -486,6 +486,20 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        &mut self,
+        x: &Self::Variable,
+        y: &Self::Variable,
+        position: Self::Position,
+    ) -> Self::Variable {
+        let x: u32 = (*x).try_into().unwrap();
+        let y: u32 = (*y).try_into().unwrap();
+        let res = ((x as i32) / (y as i32)) as u32;
+        let res = res as u64;
+        self.write_column(position, res);
```
svv232, line 499:
```suggestion
        let x: i32 = (*x).try_into().unwrap();
        let y: i32 = (*y).try_into().unwrap();
        let res = x / y;
        let res = res as u64;
        self.write_column(position, res);
```

dannywillems: See 6f24e1e1c06e9377abbf600fc8847631aba36c1d

svv232 (approved):

PR #2803, opened 2024-11-20
o1vm/riscv32: implement M type instruction Rem

dannywillems:


svv232 (approved):

o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -2184,7 +2184,17 @@ pub fn interpret_mtype<Env: InterpreterEnv>(env: &mut Env, instr: MInstruction)
             unimplemented!("Divu")
         }
         MInstruction::Rem => {
-            unimplemented!("Rem")
```
svv232, line 2187: it would be nice to comment the implementation from https://msyksphinz-self.github.io/riscv-isadoc/html/rvm.html

PR #2807, opened 2024-11-20
o1vm/riscv32: implement M type instruction divu

dannywillems:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -2193,7 +2193,17 @@ pub fn interpret_mtype<Env: InterpreterEnv>(env: &mut Env, instr: MInstruction)
             env.set_next_instruction_pointer(next_instruction_pointer + Env::constant(4u32));
         }
         MInstruction::Divu => {
-            unimplemented!("Divu")
```
svv232, line 2196: it would be nice to comment the implementation from https://msyksphinz-self.github.io/riscv-isadoc/html/rvm.html

svv232: we can follow up with this documentation in a future pr

svv232 (approved):

PR #2809, opened 2024-11-20
o1vm/riscv32: implement mul_hi_signed_unsigned to be used for Mulhsu

dannywillems:


o1vm/src/interpreters/riscv32im/witness.rs
```diff
@@ -486,6 +486,21 @@ impl<Fp: Field> InterpreterEnv for Env<Fp> {
...
+        &mut self,
+        x: &Self::Variable,
+        y: &Self::Variable,
+        position: Self::Position,
+    ) -> Self::Variable {
+        let x: u32 = (*x).try_into().unwrap();
+        let y: u32 = (*y).try_into().unwrap();
+        let res = (((x as i32) as i64) * (y as i64)) as u64;
+        let res = (res >> 32) as u32;
+        let res = res as u64;
```
svv232, line 499:
```suggestion
       let x: i32 = (*x).try_into().unwrap();
        let y: i32 = (*y).try_into().unwrap();
        let res = ((x as i64) * (y as i64)) as u64;
        let res = (res >> 32) as u32;
        let res = res as u64;

```

svv232: this is fine after discussing, it is really just my preference.

svv232 (approved):

PR #2810, opened 2024-11-20
o1vm/riscv32: implement M type instruction Mulhsu

dannywillems:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -2178,7 +2178,17 @@ pub fn interpret_mtype<Env: InterpreterEnv>(env: &mut Env, instr: MInstruction)
             env.set_next_instruction_pointer(next_instruction_pointer + Env::constant(4u32));
         }
         MInstruction::Mulhsu => {
-            unimplemented!("Mulhsu")
+            let rs1 = env.read_register(&rs1);
```
svv232, line 2181: it would be nice if the implementation from https://msyksphinz-self.github.io/riscv-isadoc/html/rvm.html was commented below the instruction.

dannywillems: I'm adding this in a follow-up PR right now.

svv232 (approved):

PR #2811, opened 2024-11-20
Refactoring: Improve poly-commitment docs & decrease confusing variable names

volhovm:
This is a **no-op** docs/naming refactoring commit that does the following:
- Moves `combine_polys` to `poly-commitment/src/utils.rs` (there was a comment left by someone that this is a good idea; I agree)
- Improves docs for `combine_polys` and related structs,  like `PolynomialsToCombine`.
- Renames some variables inside `combine_polys`: "omegas" was used which is (1) only used once in the whole repo, and (2) is absolutely not descriptive, since it has no r. Was replaced by 
- Fixed some typos/bad argument names for `hiding/non_hiding` functions, that are actually operating over the integer expressing the number of chunks, not the size of the domain as the previous variable implied.
- Importantly: gets rid of `"xi"`, which was a very rarely used term to call `polyscale`, also known as `u` (not to be confused with the three other `u`s int he code). Used `polyscale` instead.
    -  `xi` is used more on the ocaml side, but my plan is to remove it from there too. See the justification below.
- Improves docs/disambiguates variable names in IPA commitment creation and verification.

Parent `mina` PR:
- https://github.com/MinaProtocol/mina/pull/16376
    - _(strictly speaking, these are independent; mina's PR targets p-s develop, while p-s PR is towards p-s master)_

On `xi` and justification for not using it:
- Throughout the codebase, we use `xi = polyscale = v` and `r = evalscale = u` which is a /pain/ for someone trying to make sense of what's happening. What's even worse, the term `u` means at least two other things: a generator in the IPA procedure, and the IPA challenge.
- I decided to get rid of the `xi/r` pair of these three, for the following reasons:
    - The original plonk paper uses `u` and `v`.
    - Our o1labs book on pickles uses `u` and `v` (with references to `polyscale` and `evalscale`).
    - Most of `proof-systems` except some 10 lines uses `u/v` or `polyscale/evalscale`.
    - `r` is a terribly overloaded variable name, which probably means about 5 things across the codebase already (e.g. `r` in `schnorr.ml` or `signature.ml`, e.g. `l` and `r` in `plonk_constraint_system.ml`). So I'm strongly against using it as a global identifier if possible. 
    - I think it would be better to stick with globally identifiable handles, like `polyscale`. I think it's _acceptable_ to keep two notations, one concise-mathematical, one globally-identifiable and code-friendly. So one mathematical notation should  go, and `xi / r` is the most ambiguous of the two and least used.

### DIFF
--- a/poly-commitment/src/ipa.rs
+++ b/poly-commitment/src/ipa.rs
@@ -745,7 +745,7 @@ impl<G: CommitmentCurve> SRS<G> {
         let r_delta = <G::ScalarField as UniformRand>::rand(rng);
 
         // Compute delta, the commitment
-        // delta = G0^d * U_base^{b0*d} * H^r_delta   (as a group element, in multiplicative notation)
+        // delta = [d] G0 + [b0*d] U_base + [r_delta] H^r   (as a group element, in additive notation)
         let delta = ((g0.into_group() + (u_base.mul(b0))).into_affine().mul(d)
             + self.h.mul(r_delta))
         .into_affine();
--- a/poly-commitment/src/ipa.rs
+++ b/poly-commitment/src/ipa.rs
@@ -183,8 +183,8 @@ impl<G: CommitmentCurve> SRS<G> {
         {
             sponge.absorb_fr(&[shift_scalar::<G>(*combined_inner_product)]);
 
-            let t = sponge.challenge_fq();
-            let u: G = {
+            let u_base: G = {
+                let t = sponge.challenge_fq();
                 let (x, y) = group_map.to_group(t);
                 G::of_coordinates(x, y)
             };
@@ -240,7 +240,7 @@ impl<G: CommitmentCurve> SRS<G> {
             // TERM
             // -rand_base_i * (z1 * b0 * U)
             scalars.push(neg_rand_base_i * (opening.z1 * b0));
-            points.push(u);
+            points.push(u_base);
 
             // TERM
             // rand_base_i c_i Q_i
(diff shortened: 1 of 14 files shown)


poly-commitment/src/ipa.rs
```diff
@@ -730,13 +744,16 @@ impl<G: CommitmentCurve> SRS<G> {
         let d = <G::ScalarField as UniformRand>::rand(rng);
         let r_delta = <G::ScalarField as UniformRand>::rand(rng);
 
-        // delta = (g0 + u*b0)*d + h*r_delta
-        let delta = ((g0.into_group() + (u.mul(b0))).into_affine().mul(d) + self.h.mul(r_delta))
-            .into_affine();
+        // Compute delta, the commitment
+        // delta = G0^d * U_base^{b0*d} * H^r_delta   (as a group element, in multiplicative notation)
```
dannywillems, line 748: Nit: I would maybe keep the additive notation to avoid confusion?

dannywillems:
> Fixed some typos/bad argument names for hiding/non_hiding functions, that are actually operating over the integer expressing the number of chunks, not the size of the domain as the previous variable implied.

Note that this happens in other places too.

poly-commitment/src/ipa.rs
volhovm: Changed to additive.

volhovm:
>Note that this happens in other places too.

Found it only in IVC prover (fixed it too) and expression framework (there this naming seems to make sense).

volhovm:
The CI passes except for codecov -0.19% coverage due to renaming :) Merging.

PR #2813, opened 2024-11-21
o1vm/mips: introduce scratch_state_inverse

dannywillems:
These fields will be used to keep track of the elements that should be inverted while generating the proof.


o1vm/src/pickles/tests.rs
```diff
@@ -121,3 +125,31 @@ fn test_small_circuit() {
     );
     assert!(verif, "Verification fails");
 }
+
+#[test]
```
dannywillems, line 129: Feel free to disagree with these changes. It is a single commit we can simply delete. I only want to keep a regression test.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -50,10 +50,17 @@ pub const NUM_INSTRUCTION_LOOKUP_TERMS: usize = 5;
 pub const NUM_LOOKUP_TERMS: usize =
     NUM_GLOBAL_LOOKUP_TERMS + NUM_DECODING_LOOKUP_TERMS + NUM_INSTRUCTION_LOOKUP_TERMS;
 // TODO: Delete and use a vector instead
+// FIXME: since the introduction of the scratch size inverse, the value below
```
dannywillems, line 53: This is fixed in https://github.com/o1-labs/proof-systems/pull/2815.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -50,10 +50,17 @@ pub const NUM_INSTRUCTION_LOOKUP_TERMS: usize = 5;
...
+// can be decreased. It implies to change the offsets defined in [column]. At
+// the moment, it incurs an overhead we could avoid as some columns are zeroes.
 // MIPS + hash_counter + byte_counter + eof + num_bytes_read + chunk + bytes
 // + length + has_n_bytes + chunk_bytes + preimage
 pub const SCRATCH_SIZE: usize = 98;
 
+/// Number of columns used by the MIPS interpreter to keep values to be
```
dannywillems, line 60: To verify it was the minimal value, I did try to use 11, and it failed.
From there, I run for some millions instructions the op-program, and it never failed.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -106,6 +115,12 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
         Column::ScratchState(scratch_idx)
     }
 
+    fn alloc_scratch_inverse(&mut self) -> Self::Position {
```
querolita, line 118: It seems like this function for the mips witness is the same as the one in the mips constraints. I wonder if it could be generically defined in the interpreter itself, but probably not directly given that it uses the fact that it accesses a `self.scratch_state_idx_inverse`. Maybe environments can be defined to implement a trait that returns it, but not sure if it is worth the code duplication.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -314,17 +329,17 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
...
+        let res = {
+            let pos = self.alloc_scratch();
+            unsafe { self.test_zero(x, pos) }
+        };
         // write the non deterministic advice inv_or_zero
-        let pos = self.alloc_scratch();
-        let inv_or_zero = if *x == 0 {
-            Fp::zero()
+        let pos = self.alloc_scratch_inverse();
+        if *x == 0 {
```
querolita, line 338: I am not sure why there's this `if-else` distinction if it always writes in the inverse scratch state. Is it to avoid a conversion in case `Fp::zero()` is faster than `Fp:from(0)`? Either way I suppose the batch inversion algorithm will just ignore zeros, I believe I checked that in the arkworks code.

o1vm/src/interpreters/mips/witness.rs
```diff
@@ -339,15 +354,11 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
...
-        let _to_zero_test_inv_or_zero = {
-            let pos = self.alloc_scratch();
-            let inv_or_zero = if to_zero_test == Fp::zero() {
-                Fp::zero()
-            } else {
-                Fp::inverse(&to_zero_test).unwrap()
-            };
-            self.write_field_column(pos, inv_or_zero);
-            1 // Placeholder value
+        let pos = self.alloc_scratch_inverse();
```
querolita, line 357: Less lines of code for the same thing, I like it 👍🏻

o1vm/src/pickles/column_env.rs
```diff
@@ -36,8 +39,9 @@ pub struct ColumnEnvironment<'a, F: FftField> {
 }
 
 pub fn get_all_columns() -> Vec<Column> {
-    let mut cols = Vec::<Column>::with_capacity(SCRATCH_SIZE + 2 + N_MIPS_SEL_COLS);
-    for i in 0..SCRATCH_SIZE + 2 {
+    let mut cols =
```
querolita, line 42: Perhaps use `N_MIPS_REL_COLS`, or you want to be super explicit here?

o1vm/src/pickles/prover.rs
```diff
@@ -110,9 +111,17 @@ where
         };
         // Doing in parallel
         let scratch = scratch.into_par_iter().map(eval_col).collect::<Vec<_>>();
+        let scratch_inverse = scratch_inverse
+            .into_par_iter()
```
querolita, line 115: Is this as efficient (or better) than having all of the scratch inverses in a single list and doing a single batch inversion of it all, instead of `n` parallel of shorter ones? Wondering for performance.

o1vm/src/pickles/tests.rs
querolita: These are good to visualize the behavior when zeroes are involved.

o1vm/src/pickles/prover.rs
dannywillems: I do not know. I'm wondering too. Might be good to investigate later. Opening a follow-up issue.

dannywillems: Opened https://github.com/o1-labs/proof-systems/issues/2818

o1vm/src/interpreters/mips/witness.rs
dannywillems: No reason. Seems to be an ugly code :sweat_smile:. For a follow-up.

PR #2814, opened 2024-11-21
o1vm/mips: remove inverse_or_zero from interpreter trait

dannywillems:
It was only used by `constraints.rs`, not in the witness nor the interpreter. Inline the method.

### DIFF
--- a/o1vm/src/interpreters/mips/witness.rs
+++ b/o1vm/src/interpreters/mips/witness.rs
@@ -312,20 +312,6 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
         res
     }
 
-    unsafe fn inverse_or_zero(
-        &mut self,
-        x: &Self::Variable,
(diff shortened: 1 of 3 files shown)


o1vm/src/interpreters/mips/witness.rs
```diff
@@ -312,20 +312,6 @@ impl<Fp: Field, PreImageOracle: PreImageOracleT> InterpreterEnv for Env<Fp, PreI
         res
     }
 
-    unsafe fn inverse_or_zero(
```
querolita, line 315: At first sight, it feels surprising that the `inverse_or_zero` was not being used in the witness code. But I suppose it makes sense. IIRC, a short time ago Marc removed a call to it when fixing the equals by inlining parts of it in `is_zero`, so probably that was the only place where it was being used before.

querolita (approved):

svv232 (approved):
Aside from the implementation being previously used, it’s generally ok to delete dead code.

PR #2815, opened 2024-11-21
o1vm: decrease scratch size by 40%

dannywillems:
It seems that most of the state was empty.


o1vm/src/interpreters/mips/column.rs
```diff
@@ -1,40 +1,44 @@
...
 };
 use std::ops::{Index, IndexMut};
 use strum::EnumCount;
 
-pub use super::{
-    witness::SCRATCH_SIZE_INVERSE, ITypeInstruction, JTypeInstruction, RTypeInstruction,
-};
+use super::{ITypeInstruction, JTypeInstruction, RTypeInstruction};
 
+pub(crate) const SCRATCH_SIZE_WITHOUT_KECCAK: usize = 45;
```
dannywillems, line 11: Test with 44, and you will see that the e2e tests fail at a certain point.

PR #2819, opened 2024-11-25
o1vm/riscv32im: test decoding add

dannywillems:
Unit tests for each instruction is coming. At the end of the PR train, we should be confident that the decode instruction is injective.


svv232 (approved):

dannywillems:
I did use `git merge` in this case because I do think I totally messed up the history with previous pull requests. Sorry.
I hope this git merge will be the only one required for the future PR.

PR #2821, opened 2024-11-25
o1vm/riscv32im: test decoding sll

dannywillems:


svv232 (approved):

dannywillems:
Damn it. Again conflicts.
I see it is because we keep adding in the PR the tests at the end of the file, where on master we have already a test at the end of the file, which is not present here.
Cool git.

PR #2827, opened 2024-11-25
implementation for load half instruction in riscv32im

svv232:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1652,7 +1652,25 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
             unimplemented!("LoadByte")
         }
         IInstruction::LoadHalf => {
-            unimplemented!("LoadHalf")
+            // lh:  x[rd] = sext(M[x[rs1] + sext(offset)][15:0])
```
dannywillems, line 1655: Nit: I would add documentation reg. "sext" and "M" like [here](https://github.com/o1-labs/proof-systems/pull/2817/files). Not blocking.

dannywillems (approved):

PR #2832, opened 2024-11-25
o1vm/riscv32im: decode M instruction type

dannywillems:
The decoding is more verbose on purpose, as I would like to have a regression test checking an exception is raised. Unit tests checking the decoding are coming, mimicking the previous ones.


svv232 (approved):
perfecto! I checked against the spec as well. https://www.cs.cornell.edu/courses/cs3410/2024fa/assignments/cpusim/riscv-instructions.pdf

PR #2833, opened 2024-11-25
implementation for load half unsigned riscv32im

svv232:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1732,7 +1732,24 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
             env.set_next_instruction_pointer(next_instruction_pointer + Env::constant(4u32));
         }
         IInstruction::LoadHalfUnsigned => {
-            unimplemented!("LoadHalfUnsigned")
+            // lhu: x[rd] = M[x[rs1] + sext(offset)][15:0]
```
dannywillems, line 1735: Let's keep in mind unaligned addresses. We should read more: https://github.com/riscv/riscv-isa-manual/blob/main/src/rv32.adoc

dannywillems (approved):

PR #2836, opened 2024-11-25
implementation for shift left logical immediate riscv32im

svv232:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1752,7 +1752,19 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
             env.set_next_instruction_pointer(next_instruction_pointer + Env::constant(4u32));
         }
         IInstruction::ShiftLeftLogicalImmediate => {
-            unimplemented!("ShiftLeftLogicalImmediate")
+            // slli: x[rd] = x[rs1] << shamt
+            let local_rs1 = env.read_register(&rs1);
+            let shamt = {
+                let pos = env.alloc_scratch();
+                unsafe { env.bitmask(&imm, 4, 0, pos) }
```
dannywillems, line 1759: Is it not 5?

```
    /// Format: `slli rd, rs1, shamt`
    ///
    /// Description: Performs logical left shift on the value in register rs1 by
    /// the shift amount held in the lower 5 bits of the immediate
```

o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1752,7 +1752,19 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
...
+            };
+            // parse shamt from imm as 20-24 of instruction and 0-4 wrt to imm
+            let rd_scratch = env.alloc_scratch();
```
dannywillems, line 1762: Nit: I would simply keep the same syntax as the others to allocate variables, like:
```rust
let local_rd = {
  unsafe { env.shift_left(&local_rs1, &shamt, rd_scratch) }
};
```
But it is a very small detail.

o1vm/src/interpreters/riscv32im/interpreter.rs
dannywillems, line 1759: There must also be constraints that it is actually 5 bits, and also the 5 lowest bits of the immediate.

dannywillems (changes requested):
This is incorrect. See comments above.

dannywillems:
I would recommend to write some tests on top of https://github.com/o1-labs/proof-systems/pull/2866. It requires to implement `lui`, available [here](https://github.com/o1-labs/proof-systems/pull/2852) and syscall. `lui` seems to be the first implementation to implement, modulo `jal` (even though it can be avoided), as it requires to set some registers to some initial values for the syscall.

o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1752,7 +1752,21 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
             env.set_next_instruction_pointer(next_instruction_pointer + Env::constant(4u32));
         }
         IInstruction::ShiftLeftLogicalImmediate => {
-            unimplemented!("ShiftLeftLogicalImmediate")
+            // slli: x[rd] = x[rs1] << shamt
+            let local_rs1 = env.read_register(&rs1);
+            let shamt = {
```
dannywillems, line 1757: Can you add some FIXME reg. [this comment](https://github.com/o1-labs/proof-systems/pull/2836#discussion_r1866012015) please? Can be done later.

dannywillems: Note that a test is ready [in master](https://github.com/o1-labs/proof-systems/blob/master/o1vm/tests/test_riscv_elf.rs#L81) reg. sll, see [assembly code](https://github.com/o1-labs/proof-systems/blob/master/o1vm/resources/programs/riscv32im/src/sll.S). Only [lui](https://github.com/o1-labs/proof-systems/pull/2852) should be implemented.

o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1752,7 +1752,21 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
...
+                let pos = env.alloc_scratch();
+                unsafe { env.bitmask(&imm, 5, 0, pos) }
+            };
+            // parse shamt from imm as 20-24 of instruction and 0-4 wrt to imm
```
dannywillems, line 1761: what is this comment for?

svv232: it was removed, but it was explaining where in the instruction and immediate the shamt was parsed from.

dannywillems (approved):

PR #2838, opened 2024-11-25
implementation for shift right logical immediate riscv32im

svv232:


dannywillems (changes requested):
See same comments than on https://github.com/o1-labs/proof-systems/pull/2836

PR #2841, opened 2024-11-25
implementation for shift right arithmetic immediate for riscv32im

svv232:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -1781,7 +1781,21 @@ pub fn interpret_itype<Env: InterpreterEnv>(env: &mut Env, instr: IInstruction)
...
+            let shamt = {
+                let pos = env.alloc_scratch();
+                unsafe { env.bitmask(&imm, 4, 0, pos) }
+            };
+            // parse shamt from imm as 20-24 of instruction and 0-4 wrt to imm
+            // sign extend shamt for arithmetic shift
+            let shamt = env.sign_extend(&shamt, 4);
+
+            let rd_scratch = env.alloc_scratch();
+            let local_rd = unsafe { env.shift_right_arithmetic(&local_rs1, &shamt, rd_scratch) };
```
dannywillems, line 1795: Nit: As in other places, I would like to keep the same shape:
```
let local_rd = {
  unsafe { env.shift_right_arithmetic(&local_rs1, &shamt, pos) }
};
```
Just to keep the scope of the position short.

dannywillems (changes requested):
See same comments than https://github.com/o1-labs/proof-systems/pull/2836

dannywillems (approved):

PR #2853, opened 2024-11-25
implementation for add upper immediate in riscv32im

svv232:


o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -2205,7 +2205,21 @@ pub fn interpret_utype<Env: InterpreterEnv>(env: &mut Env, instr: UInstruction)
...
         UInstruction::AddUpperImmediate => {
-            unimplemented!("AddUpperImmediate")
+            // auipc: x[rd] = pc + sext(immediate[31:12] << 12)
+            let local_imm = {
+                let pos = env.alloc_scratch();
+                let shifted_imm = unsafe { env.shift_left(&imm, &Env::constant(12), pos) };
+                env.sign_extend(&shifted_imm, 32)
+            };
+            let local_pc = instruction_pointer.clone();
+            let pos = env.alloc_scratch();
```
dannywillems, line 2215: A bit picky, but I would like keeping the same scope conventions.

o1vm/src/interpreters/riscv32im/interpreter.rs
```diff
@@ -2205,7 +2205,22 @@ pub fn interpret_utype<Env: InterpreterEnv>(env: &mut Env, instr: UInstruction)
...
         }
         UInstruction::AddUpperImmediate => {
-            unimplemented!("AddUpperImmediate")
+            // auipc: x[rd] = pc + sext(immediate[31:12] << 12)
+            let local_imm = {
+                let pos = env.alloc_scratch();
+                let shifted_imm = unsafe { env.shift_left(&imm, &Env::constant(12), pos) };
+                env.sign_extend(&shifted_imm, 32)
+            };
+            let local_pc = instruction_pointer.clone();
```
dannywillems, line 2214: Nit: you don't need this clone. If you can simply remove this in a follow-up, please.

dannywillems (approved):
Not entire convinced without test. But let's have it in master to move on.

PR #2761, opened 2024-11-19
fix: typos in documentation files

leopardracer:
This pull request contains changes to improve clarity, correctness and structure.

**Description correction:**
Corrected "each of th $15$" to "each of the $15$".
Corrected "loookup" to "lookup".

Please review the changes and let me know if any additional changes are needed.

PR #2755, opened 2024-11-12
Mina-signer: make secret field of KeyPair public

dannywillems:
Can be used by the upstream OpenMina. See https://github.com/o1-labs/proof-systems/pull/2750/files.

PR #2863, opened 2024-12-04
CI/o1vm: deactivating path dependent execution

dannywillems:
As we enforce some GH actions to pass when targeting master, it can make PR stalled and unmergeable. I do not see any options in GH settings at the moment to include these conditions.

cc @shimkiv: this blocks https://github.com/o1-labs/proof-systems/pull/2755 and https://github.com/o1-labs/proof-systems/pull/2761. Do you have a better option?

### DIFF
--- a/.github/workflows/o1vm-ci.yml
+++ b/.github/workflows/o1vm-ci.yml
@@ -3,37 +3,9 @@ name: o1vm CI
 on:
   workflow_dispatch:
   pull_request:
-    paths:
-      [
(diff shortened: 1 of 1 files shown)


dannywillems:
> @dannywillems please keep in mind though that jobs that we execute on self-hosted runners won't the executed for PRs that authored by external contributors.

Ah. Good to know.

shimkiv (approved):

PR #2867, opened 2024-12-05
Improve your Makefile with this one neat trick

martyall:
I have used this same `Makefile` help command in several previous projects, maybe other's find it useful? Here is what the printout looks like

image


dannywillems (approved):
Love it.

PR #2872, opened 2024-12-11
mvpoly: introduce optional param upper bound to compute nested loops

dannywillems:
Using filter_map speeds up the computation by avoiding creating a bigger map than expected if a low upper bound is given with large number of indices.

Cherry-picked from https://github.com/o1-labs/proof-systems/pull/2702/


querolita (approved):
looks like this is carrying the other changes (random generation and flaky test removal) as well though

dannywillems:
> looks like this is carrying the other changes (random generation and flaky test removal) as well though

Oopsie. Lemme rebase. Nice catch.

PR #2874, opened 2024-12-12
CI: dropping support for Rust 1.71

dannywillems:
Mina does use 1.72, therefore we have to support at minima 1.72


marcbeunardeau88 (approved):

dannywillems:
Dropped requirements for 1.71 on master.

PR #2876, opened 2024-12-14
add plonk-wasm

martyall:
Move over plonk-wasm package from mina


martyall:
@dannywillems tests are passing but lint check is failing

```
error: `extern` fn uses type `ark_ff::Fp<ark_ff::MontBackend<mina_curves::pasta::fields::FqConfig, 4>, 4>`, which is not FFI-safe
   --> plonk-wasm/src/projective.rs:147:9
    |
147 |         WasmPastaFp,
    |         ^^^^^^^^^^^ not FFI-safe
    |
    = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
    = note: this struct has unspecified layout

```

Since there was no clippy check in mina, this could have been present for quite a while. The `WasmPastFp` type does have a `#[repr(C)]` attribute, but it wraps a newtype in arkworks that has no such instance. Clearly things are working anyway, what do you think we should do?

dannywillems:
Closing as old. We have now plonk-wasm in proof-systems.

PR #2882, opened 2024-12-18
More robust o1vm cli

martyall:
I was working on adding more tests and functionality to the mips/cannon but I needed a more robust (i.e. enterprise) CLI. I factored out this work as part of that effort

## Changes
- use clap derive to clearly defined and document the cli arguments
- separate a `cannon` namespace (as opposed to `riscv`) so that we can keep the functionality separate
-  combine executables and remove the `test_optimism_preimage_read`


o1vm/src/cli/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Parser, Debug, Clone)]
+#[command(
+    name = "o1vm",
+    version = "0.1",
+    about = "o1vm - a generic purpose zero-knowledge virtual machine"
+)]
+pub enum Commands {
+    #[command(subcommand)]
+    Cannon(cannon::Cannon),
```
dannywillems, line 13: I guess "non-cannon" version will appear here. Waiting to see the other subcommands :).

dannywillems (approved):

PR #2883, opened 2024-12-19
mvpoly: compute the cross terms of a list of polynomials

dannywillems:


mvpoly/src/lib.rs
```diff
@@ -300,3 +300,39 @@ pub trait MVPoly<F: PrimeField, const N: usize, const D: usize>:
...
+    // It also makes the code cleaner as we do not need to handle 0^0
+    assert!(combiner1 != F::zero());
+    assert!(combiner2 != F::zero());
+    assert!(u1 != F::zero());
+    assert!(u2 != F::zero());
+    polys
+        .into_iter()
+        .enumerate()
+        .fold(HashMap::new(), |mut acc, (i, poly)| {
+            let scalar1 = combiner1.pow([i as u64]);
```
marcbeunardeau88, line 330: you can keep the alpha^i as part as the accumulator of the fold, to compute just one multiplication instead of the full exponentation
(Actually the way we want to use it, we could compute the combiner once for every row, but that would change the interface too much)

marcbeunardeau88 (approved):

PR #2887, opened 2024-12-21
MVPoly: additional checks reg. multiplication of polynomials

dannywillems:
The degree is in the type, but we suppose the user will be careful when using the library, which is a bad assumption.

It has already been discussed to remove the type dependency. Let's do it later.

### DIFF
--- a/mvpoly/src/monomials.rs
+++ b/mvpoly/src/monomials.rs
@@ -102,6 +102,9 @@ impl<const N: usize, const D: usize, F: PrimeField> Mul for Sparse<F, N, D> {
 
     fn mul(self, other: Self) -> Self {
         let mut monomials = HashMap::new();
+        let degree_lhs = unsafe { self.degree() };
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):

mvpoly/src/monomials.rs
```diff
@@ -102,6 +102,9 @@ impl<const N: usize, const D: usize, F: PrimeField> Mul for Sparse<F, N, D> {
 
     fn mul(self, other: Self) -> Self {
         let mut monomials = HashMap::new();
+        let degree_lhs = unsafe { self.degree() };
+        let degree_rhs = unsafe { other.degree() };
+        assert!(degree_lhs + degree_rhs <= D, "The degree of the output is expected to be maximum {D}, but the resulting output would be larger than {D} ({res})", res=degree_lhs + degree_rhs);
```
mrmr1993, line 107: Note that this will crash the WebAssembly instance once we propagate it to JS

PR #2895, opened 2024-12-23
o1vm/riscv32im: implement swap using add and sub

dannywillems:


o1vm/tests/test_riscv_elf.rs
```diff
@@ -196,3 +196,20 @@ fn test_addi_negative() {
...
+    ));
+    let state = o1vm::elf_loader::parse_riscv32(&path).unwrap();
+    let mut witness = Env::<Fp>::create(PAGE_SIZE.try_into().unwrap(), state);
+
+    while !witness.halt {
+        witness.step();
+    }
+
+    assert_eq!(witness.registers[T0], 10);
+    assert_eq!(witness.registers[T1], 5);
```
svv232, line 214: I think it makes sense to add an assertion for T2 here.

svv232 (changes requested):

o1vm/tests/test_riscv_elf.rs
dannywillems: Done in 84989741c03571fc6c1f0c2a4b17cab8465d2702

svv232 (approved):

PR #2897, opened 2024-12-23
o1vm/riscv32im: add one test for addi with negative values only

dannywillems:


o1vm/tests/test_riscv_elf.rs
```diff
@@ -195,6 +195,7 @@ fn test_addi_negative() {
     assert_eq!(witness.registers[T0], 100);
     assert_eq!(witness.registers[T1], 50);
     assert_eq!(witness.registers[T2], (-50_i32) as u32);
```
svv232, line 197: i think it makes sense to add a t3 assertion here as well for -1000

o1vm/tests/test_riscv_elf.rs
```diff
@@ -195,6 +195,7 @@ fn test_addi_negative() {
...
+    assert_eq!(witness.registers[T4], (-1500_i32) as u32);
```
svv232, line 198: i think it makes sense to add a t3 assertion here as well for -1000

svv232 (changes requested):

o1vm/tests/test_riscv_elf.rs
dannywillems: Addressed in 2c55e6ba2a31c62e6e6f4c2edabd2134550be340

svv232 (approved):

PR #2906, opened 2024-12-23
Makefile: target to import all programs from Optimism

dannywillems:
Import all MIPS programs from Optimism, and tweak it to make it compatible with our ELF loader. Requested by @martyall to have more small test programs for the o1vm MIPS edition.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ state.json
 # Directory for the RISC-V 32bits toolchain
 _riscv32-gnu-toolchain
 o1vm/resources/programs/riscv32im/bin/*.o
+o1vm/resources/programs/mips/bin/*.o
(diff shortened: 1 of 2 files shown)


martyall (changes requested):
This  is super useful! My only comment is that the linked files shouldn't be committed, so basically the entire `resources/programs/mips/bin` directory should be ignored (not just the object files). It seems like this was also your intention since `make clean` blows this directory away.

dannywillems:
> This is super useful! My only request is that the linked files shouldn't be committed, i.e. the entire `resources/programs/{mips,riscv}/bin` directory should be ignored (not just the object files). It seems like this was also your intention since `make clean` blows this directory away.

The only issue I see by not committing them is that the CI (and users) will have to build them, and therefore would need to add the mips toolchain on their machine.

martyall:
> The only issue I see by not committing them is that the CI (and users) will have to build them, and therefore would need to add the mips toolchain on their machine.

That's a fair points about its annoying to require it for devs, it just seems weird that running `make clean` creates a diff by deleting them.

BTW I can confirm that these are working with #2911 .

martyall:
@dannywillems I would suggest that we use a github workflow with a `workflow_dispatch` trigger to be able to run this job in ci and tar/upload the artifacts. This way we can still ignore the `bin` directory and `make clean` is really clean (as opposed to "dirty"). 

You can see what this looks like [here](https://github.com/o1-labs/proof-systems/actions/runs/12508373496/workflow), and here is an [example upload](https://github.com/o1-labs/proof-systems/actions/runs/12508373496/job/34896394289#step:7:20)

dannywillems:
I did remove the commit https://github.com/o1-labs/proof-systems/commit/49235d338ff233dc29a2809b0d179f603b657301 to clean the history and avoid unnecessary commits. https://github.com/o1-labs/proof-systems/pull/2911 would therefore need a rebase.

martyall (approved):

PR #2910, opened 2024-12-23
no-op MIPS

dannywillems:
cc @martyall


martyall:
This seems to be working now (largely thanks to @marcbeunardeau88). 

I added a script you can use to test any of our example mips programs (with the exception of the oracle programs, see [this comment](https://github.com/o1-labs/proof-systems/pull/2945#issuecomment-2589082330). For example, from the root of the repository you can run

```
RUST_LOG=debug ./o1vm/test-mips-program-prover.sh o1vm/resources/programs/mips/bin/and srs/test_vesta.srs
```

to test the `and` program (assuming you have the mips bin files). I didn't add this to CI, but we can if you want

o1vm/src/interpreters/mips/interpreter.rs
```diff
@@ -970,9 +972,34 @@ pub fn interpret_instruction<Env: InterpreterEnv>(env: &mut Env, instr: Instruct
         Instruction::RType(instr) => interpret_rtype(env, instr),
         Instruction::JType(instr) => interpret_jtype(env, instr),
         Instruction::IType(instr) => interpret_itype(env, instr),
+        Instruction::NoOp => interpret_noop(env),
     }
 }
 
+pub fn interpret_noop<Env: InterpreterEnv>(env: &mut Env) {
```
dannywillems, line 979:
```suggestion
// FIXME: the noop should not be used in production. The interpreter semantic should be refined.
pub fn interpret_noop<Env: InterpreterEnv>(env: &mut Env) {
```

o1vm/src/interpreters/mips/interpreter.rs
```diff
@@ -31,6 +31,7 @@ pub enum Instruction {
     RType(RTypeInstruction),
     JType(JTypeInstruction),
     IType(ITypeInstruction),
+    NoOp,
```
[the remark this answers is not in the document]
dannywillems: Committing to the repo directly.

o1vm/src/interpreters/mips/interpreter.rs
```diff
@@ -970,9 +974,37 @@ pub fn interpret_instruction<Env: InterpreterEnv>(env: &mut Env, instr: Instruct
...
+            + (v1 * Env::constant(1 << 16))
+            + (v2 * Env::constant(1 << 8))
+            + v3
+    };
+    let opcode = {
+        let pos = env.alloc_scratch();
+        unsafe { env.bitmask(&instruction, 32, 26, pos) }
+    };
+
+    env.range_check8(&opcode, 6);
```
dannywillems, line 1001: This check can even be removed as we check opcode is zero below.

martyall (approved):

dannywillems:
Approving also. Most important part is to avoid using it in production, as written in the comment.

martyall:
> Approving also. Most important part is to avoid using it in production, as written in the comment.

The comment says

```
// FIXME: the noop should not be used in production. The interpreter semantic
// should be refined. The padding is only for testing purposes when padding is
// required to reach the size of the domain.
```

hmm, can we just maybe just do that now? I don't see why we would merge if we can't use it ?

PR #2911, opened 2024-12-24
gen-state-json

martyall:
This PR is meant to test #2906 and also the `elf_loader` module. Additionally, we  generalize the `elf_loader` to work with both riscv and mips

## Changes
- generalize `elf_loader` to work with mips/ simplify some indices during memory-page construction.
- remove the generated dir `o1vm/resources/programs/mips/bin` and add to gitignore
- add a test script `o1vm/test-gen-state-json.sh` which tests the elf loader / `state.json` generator over all of the example mips programs.
- added a `MIPS Build and Package`  github workflow that runs on dispatch trigger. This workflow:
  1. builds the mips tools
  2. runs `make build-mips-programs` to build all of the example mips program binaries.
  3. checks that our elf loader can parse these programs
  4. uploads a tar file and provides a URL for downloads (intended for devs who don't have the tools to build them locally)
- makes `meta.json` file optional in `pickles_o1vm` cli. This simple change is not needed for the above -- it was picked over by mistake. Still, the change is trivial and needed for later work so I'm hoping we can keep it.


o1vm/meta_test.json
```diff
@@ -0,0 +1,37 @@
+{
```
dannywillems, line 1: Can you remove this file, please?

.github/workflows/mips-build.yml
```diff
@@ -0,0 +1,55 @@
+name: MIPS Build and Package
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - 'o1vm/resources/programs/mips/**'
```
dannywillems, line 7: It is too restrictive. For instance, when we modify the code of o1vm, we should also run the job.
I would remove any path dependent workflows. We had some in the past but [removed them](https://github.com/o1-labs/proof-systems/pull/2863) because it was making some PR stalled as we enforce some jobs to run.

.github/workflows/mips-build.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Create tar archive
+      run: |
+        cd o1vm/resources/programs/mips
+        tar -czf mips-binaries.tar.gz bin/
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: mips-binaries
+        path: o1vm/resources/programs/mips/mips-binaries.tar.gz
```
dannywillems, line 55: Not blocking, but would be good to have it in a follow-up if not in this PR:
I would add a line in the README of o1vm mentioning how to download and run the test.

.github/workflows/mips-build.yml
```diff
@@ -0,0 +1,55 @@
+name: MIPS Build and Package
```
dannywillems, line 1: If you add new jobs, add in the [top level README](https://github.com/o1-labs/proof-systems/blob/master/README.md#ci) what this file does, for consistency.

o1vm/src/cli/cannon.rs
```diff
@@ -104,9 +99,25 @@ pub struct RunArgs {
...
+    pub output: String,
+}
+
 #[derive(Subcommand, Clone, Debug)]
 pub enum Cannon {
     Run(RunArgs),
     #[command(name = "test-optimism-preimage-read")]
     TestPreimageRead(RunArgs),
+    #[command(name = "gen-state-json")]
+    GenStateJson(GenStateJsonArgs),
```
dannywillems, line 122: I'm fine with an additional subcommand to test the generation of `state.json`. However, this will only be useful for Cannon programs, not for generic programs. `state.json` won't be used for other MIPS/RISC-V programs.
MIPS/RISC-V program state should be loaded directly from the ELF, see [this example](https://github.com/o1-labs/proof-systems/blob/master/o1vm/tests/test_riscv_elf.rs#L113).

o1vm/src/elf_loader.rs
```diff
@@ -1,24 +1,19 @@
...
-// We should return a structure specifically built for the o1vm, and not tight
-// to Cannon. It will be done in a future PR to avoid breaking the current code
-// and have a huge diff.
-pub fn parse_riscv32(path: &Path) -> Result<State, String> {
-    debug!("Start parsing the ELF file to load a RISC-V 32i compatible state");
-    let file_data = std::fs::read(path).expect("Could not read file.");
-    let slice = file_data.as_slice();
-    let file = ElfBytes::<LittleEndian>::minimal_parse(slice).expect("Open ELF file failed.");
+pub enum Architecture {
+    Mips,
```
dannywillems, line 11: +1

o1vm/src/elf_loader.rs
```diff
@@ -1,24 +1,19 @@
...
+    RiscV,
```
dannywillems, line 12: Nit: maybe `RiscV32` to stress it is only 32 bits?

.github/workflows/mips-build.yml
dannywillems: Note: not blocking for this PR, but would love to have it at least in a follow-up.

o1vm/src/cli/cannon.rs
martyall: :ok_hand: -- we can remove it in the future as we want, but right now I am using it heavily for local development/testing

dannywillems (approved):

PR #2914, opened 2024-12-26
o1vm/riscv32im: decrease the number of required columns

dannywillems:
Try with 38, and fibonacci will fail.
It might need to be increased in the future if you see that the current testsuite doesn't cover the whole instruction set.

### DIFF
--- a/o1vm/src/interpreters/riscv32im/mod.rs
+++ b/o1vm/src/interpreters/riscv32im/mod.rs
@@ -1,7 +1,5 @@
 /// The minimal number of columns required for the VM
-// FIXME: the value will be updated when the interpreter is fully
-// implemented. Using a small value for now.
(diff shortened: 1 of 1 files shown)


o1vm/src/interpreters/riscv32im/mod.rs
```diff
@@ -1,7 +1,5 @@
 /// The minimal number of columns required for the VM
-// FIXME: the value will be updated when the interpreter is fully
-// implemented. Using a small value for now.
-pub const SCRATCH_SIZE: usize = 80;
+pub const SCRATCH_SIZE: usize = 39;
```
svv232, line 2: does the fibb 7 test still work after this change?

svv232 (approved):

PR #2920, opened 2025-01-02
Optional PreImageOracle

martyall:
Not every mips program requires this, but we currently throw an exception if it isn't provided


dannywillems (approved):
Approving for now. It is not the best solution as we discussed, but it works, and help to get something working.

We will also need to change it when we integrate the prover/verifier for RISC-V.

Also, we need a better solution to handle MIPS programs that are not targeting OP (i.e. no oracle for instance). We might want to have in the future a different subinterpreter for the syscalls..

PR #2922, opened 2025-01-03
o1vm: remove legacy flavor

dannywillems:


o1vm/src/legacy/tests.rs
```diff
@@ -1,742 +0,0 @@
...
-
-        let assert_num_constraints = |instr: &Instruction, num: usize| {
-            assert_eq!(
-                mips_circuit.trace.get(instr).unwrap().constraints.len(),
-                num
-            )
-        };
-
-        let mut i = 0;
-        for instr in Instruction::iter().flat_map(|x| x.into_iter()) {
```
dannywillems, line 145: We should add it back in interpreters/mips

mrmr1993 (approved):

PR #2926, opened 2025-01-06
add open mips tests 

martyall:
- added `halt_address` field to `VmConfiguration` as a way to test for program exit (same trick used by `cannon` to run open mips tests)
- added test file `test_mips_elf.rs` and run all the (non oracle) tests in CI that aren't currently failing
- added build mips action, change workflows to use this action since mips artifacts are needed when running complete test suite

Current (ignored) test failures that need to be addressed:
- `mul.asm`: We have an overflow exception because we are trying to do signed integer multiplication with casting that leads to overflow
- `exit_group.asm`: it's unclear to me what this program means, but the use of `syscall` is suspect


martyall:
There is an annoying problem here with the way we are using `--all-features` for testing and what I had in mind here. I really only want to run these tests for o1vm since it requires installing the linux tooling and also has the push artifacts step. Any ideas ?

dannywillems (approved):

PR #2930, opened 2025-01-07
Make clippy happy for 1.75

dannywillems:


dannywillems:
Interesting error... Checking later.

utils/src/field_helpers.rs
```diff
@@ -84,9 +84,8 @@ pub trait FieldHelpers<F> {
     where
         F: PrimeField,
     {
-        big.clone()
-            .try_into()
-            .map_err(|_| FieldHelpersError::DeserializeBytes)
+        let x = F::from_biguint(big);
```
dannywillems, line 87: Coming from here I guess

dannywillems:
Better now...

querolita (approved):

PR #2939, opened 2025-01-08
riscv batch inverse

svv232:


o1vm/src/interpreters/riscv32im/mod.rs
```diff
@@ -1,5 +1,6 @@
 /// The minimal number of columns required for the VM
 pub const SCRATCH_SIZE: usize = 39;
```
dannywillems, line 2: Follow-up: this can be decreased then.

dannywillems (approved):

PR #2940, opened 2025-01-09
added srs_cache cli arg

martyall:
This has been floating around for a while but is extremely useful to save time on loading the SRS

### DIFF
--- a/o1vm/src/pickles/main.rs
+++ b/o1vm/src/pickles/main.rs
@@ -1,6 +1,6 @@
 use ark_ff::UniformRand;
 use clap::Parser;
-use kimchi::circuits::domains::EvaluationDomains;
+use kimchi::{circuits::domains::EvaluationDomains, precomputed_srs::TestSRS};
 use log::debug;
 use mina_curves::pasta::{Fp, Vesta, VestaParameters};
 use mina_poseidon::{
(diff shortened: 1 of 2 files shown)


o1vm/src/pickles/main.rs
```diff
@@ -48,10 +48,37 @@ pub fn cannon_main(args: cli::cannon::RunArgs) {
     let start = Start::create(state.step as usize);
 
     let domain_fp = EvaluationDomains::<Fp>::create(DOMAIN_SIZE).unwrap();
-    let srs: SRS<Vesta> = {
-        let srs = SRS::create(DOMAIN_SIZE);
-        srs.get_lagrange_basis(domain_fp.d1);
-        srs
+    let srs: SRS<Vesta> = match &args.srs_cache {
+        Some(cache) => {
+            debug!("Loading SRS from cache {}", cache);
```
dannywillems, line 53: I would check that the SRS is of the expected size (`DOMAIN_SIZE`).

o1vm/src/pickles/main.rs
```diff
@@ -1,6 +1,6 @@
 use ark_ff::UniformRand;
 use clap::Parser;
-use kimchi::circuits::domains::EvaluationDomains;
+use kimchi::{circuits::domains::EvaluationDomains, precomputed_srs::TestSRS};
```
dannywillems, line 3: Out of the scope of this PR: it would be nice to move this structure `TestSRS` in `poly_commitment` in `ipa.rs`, next to `SRS`. We should also have a unified structure for `TestSRS` and `SRS`. It is mostly the encoding the difference. I guess we can have a parameter or something to serde.

o1vm/src/pickles/main.rs
martyall: Isn't the requirement that `srs.size() >= DOMAIN_SIZE` ?  because I'm looking that the size of `Vesta` srs and it's actually `2^16`, even though our domain size is half that

martyall: The mismatch in the hard-coded `DOMAIN_SIZE` and what you can load from cache was responsible for a pretty big bug we were experiencing when trying to verify proofs for small programs. 

I made changes to remove the `const` and instead to use the info coming from the cache (if provided). I also changed the `1<<15` to `1<<16` to match what I was finding in the cache generated by the repo.

marcbeunardeau88 (approved):

PR #2945, opened 2025-01-10
Add Oracle tests for MIPS

martyall:
This PR consists of two commits
- [d98442c](https://github.com/o1-labs/proof-systems/pull/2945/commits/d98442c56af3282b252ae1ceecd3f482324ab48e): Add cannon mips tests for oracle reads/writes:
 -- `oracle.asm`
 -- `oracle_kzg.asm`
 -- `oracle_unaligned_read.asm`
 -- `oracle_unaligned_write.asm`
 - [c5095ad](https://github.com/o1-labs/proof-systems/pull/2945/commits/c5095ad5e04e7249bbedf09cd9510600ea415f5d): Make modifications to `SyscallWritePreimage` handler to get tests to pass
 
The test was adapted from the [cannon codebase version](https://github.com/ethereum-optimism/optimism/blob/4a487b8920daa9dc4b496d691d5f283f9bb659b1/cannon/mipsevm/state_test.go). **None of these oracle based tests passed with the current implementation.** The problem was with the `SyscallWritePreimage` syscall handler, which if you look at the test logs for [d98442c](https://github.com/o1-labs/proof-systems/pull/2945/commits/d98442c56af3282b252ae1ceecd3f482324ab48e) you can see the errors of the form:

```
[2025-01-13T16:18:51Z DEBUG o1vm::elf_loader] The executable code starts at address 4194512, has size 304 bytes, and ends at address 4194815.
[2025-01-13T16:18:51Z DEBUG test_mips_elf::test_oracle] Asking oracle for key: 000000000000000000000000000000000000000000000000000000004a29921f
thread 'tests::open_mips_tests' panicked at 'Unknown key type prefix: 0', o1vm/tests/test_mips_elf.rs:62:22
```

If you want to run them locally, you can use the one-liner

```
> RUST_LOG=debug RUST_BACKTRACE=1 cargo test --features open_mips --test test_mips_elf -- --nocapture
```

The problem is that the current write handler isn't writing to the appropriate places in the register. I came up with _a solution_ that fixes the problem with the cost of increasing the scratch space and the number of registers. If we're looking to "pay less" for the correct implementation, we can use this as a starting point.

### DIFF
--- a/o1vm/src/cannon.rs
+++ b/o1vm/src/cannon.rs
@@ -316,6 +316,7 @@ pub const HINT_CLIENT_WRITE_FD: i32 = 4;
 pub const PREIMAGE_CLIENT_READ_FD: i32 = 5;
 pub const PREIMAGE_CLIENT_WRITE_FD: i32 = 6;
 
+#[derive(Clone)]
 pub struct Preimage(Vec<u8>);
 
 impl Preimage {
--- a/o1vm/src/interpreters/mips/column.rs
+++ b/o1vm/src/interpreters/mips/column.rs
@@ -8,7 +8,7 @@ use strum::EnumCount;
 
 use super::{ITypeInstruction, JTypeInstruction, RTypeInstruction};
 
-pub(crate) const SCRATCH_SIZE_WITHOUT_KECCAK: usize = 45;
+pub(crate) const SCRATCH_SIZE_WITHOUT_KECCAK: usize = 93;
 /// The number of hashes performed so far in the block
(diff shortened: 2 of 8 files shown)


martyall:
As of now there is no interest in using this -- there is confidence that the original implementation will use this codepath and work if you run the optimism example long enouhg/to completion. 

As a warning, the current CI job which runs the OP program doesn't run long enough to test `SyscallWritePreimage`, and none of the existing unit tests will test it through vm execution. If there are future problems in this domain, we can revisit.

PR #2944, opened 2025-01-10
Martin/add oracle tests

martyall:

PR #2956, opened 2025-01-17
Add padding and proving for smaller programs

martyall:
Builds on #2910 and uses the no-op selector to do the padding. 

You can run against any test mips program using 

```
 RUST_LOG=debug ./o1vm/test-mips-program-prover.sh o1vm/resources/programs/mips/bin/nor srs/test_vesta.srs
```

where you can replace `nor` with any test mips program (except the oracle programs, see #2944) . I didn't add this to CI, but we can if we want.


dannywillems:
> I didn't add this to CI, but we can if we want.

I would be in favor of adding it in the CI.

o1vm/src/pickles/main.rs
```diff
@@ -158,38 +159,87 @@ pub fn cannon_main(args: cli::cannon::RunArgs) {
...
+    curr_proof_inputs
+        .evaluations
+        .scratch
+        .iter_mut()
+        .for_each(|x| x.resize(x.capacity(), zero));
+    curr_proof_inputs
+        .evaluations
+        .scratch_inverse
+        .iter_mut()
+        .for_each(|x| x.resize(x.capacity(), zero));
```
dannywillems, line 228: Even though we use `Vec::with_capacity(domain_size)`, I would be in favor of using the domain size, as this piece of code, even though we manage it in the same codebase, is not supposed to know how the structure `ProofInputs` is initialized.

o1vm/test-mips-program-prover.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+trap cleanup EXIT
+
+# Generate state JSON
+echo "Generating state JSON for $PROGRAM_NAME..."
+cargo run --bin pickles_o1vm -- cannon gen-state-json \
+    -i "$PROGRAM_FILEPATH" \
+    -o "$STATE_JSON"
+
+if [ $? -ne 0 ]; then
```
dannywillems, line 30:
```

In test-mips-program-prover.sh line 30:
if [ $? -ne 0 ]; then
     ^-- SC2181 (style): Check exit code directly with e.g. 'if ! mycmd;', not indirectly with $?.


In test-mips-program-prover.sh line 42:
if [ $? -ne 0 ]; then
     ^-- SC2181 (style): Check exit code directly with e.g. 'if ! mycmd;', not indirectly with $?.

For more information:
  https://www.shellcheck.net/wiki/SC2181 -- Check exit code directly with e.g...

```

dannywillems:
```
output=$(cargo run --bin pickles_o1vm -- cannon gen-state-json \
    -i "$PROGRAM_FILEPATH" \
    -o "$STATE_JSON")

if [ ! "${output}" ]; then
    echo "Error: Failed to generate state JSON"
    exit 1
fi
```
Something like this should do the job

dannywillems (approved):
Nothing blocking, but I would appreciate having the comments addressed in a follow-up if aggreed. In addition to that, I would add in the CI the test.

A documentation should be added in the README reg. the new shell script.

dannywillems:
Merging master into it for the CI.

PR #2957, opened 2025-01-21
Saffron e2e tests in CI

martyall:
Add CI for saffron e2e test (bash scripts using the CLI)

### DIFF
--- a/.github/workflows/saffron.yml
+++ b/.github/workflows/saffron.yml
@@ -1,0 +1,40 @@
+name: Saffron CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  run:
+    name: Run saffron e2e tests
+
+    runs-on: ["ubuntu-latest"]
+
+    strategy:
+      matrix:
+        rust_toolchain_version: ["1.74"]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


saffron/.gitignore
```diff
@@ -1,3 +1 @@
 proptest-regressions
-fixtures/lorem.bin
```
martyall, line 2: I just modified the test script to delete these files if the test passes.

.github/workflows/saffron.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Apply the Rust smart cacheing
+        uses: Swatinem/rust-cache@v2
+
+      - name: Build the saffron cli binary
+        run: |
+          cargo build --release --bin saffron
+
+      - name: Run the saffron e2e encoding tests
+        run: |
+          ./saffron/test-encoding.sh saffron/fixtures/lorem.txt
```
dannywillems, line 40: You're building in release mode above, but the script doesn't use the release flag. Therefore, the command above is useless.

dannywillems (changes requested):
LGTM for the shell, with a small change requested. I can enforce the job in the CI after it has been merged.

.github/workflows/saffron.yml
martyall: :man_facepalming:

dannywillems (approved):

PR #2962, opened 2025-01-23
CI: check that Cargo.toml is ordered alphabetically

dannywillems:
Only for convenience.

### DIFF
--- a/.github/workflows/cargo-sort.yml
+++ b/.github/workflows/cargo-sort.yml
@@ -1,0 +1,32 @@
+name: Check Cargo.toml Sorting
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+
+jobs:
+  cargo-sort:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust_toolchain_version: ["stable"]
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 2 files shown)


martyall:
Nice, makes sense. It seems like we should add this to `make format` so that we don't ever have to do fix this manually. Also it seems like this could be part of the [run_formatting](https://github.com/o1-labs/proof-systems/blob/master/.github/workflows/ci.yml#L45-L64) job instead of it's own workflow?

dannywillems:
> Nice, makes sense. It seems like we should add this to `make format` so that we don't ever have to do fix this manually. Also it seems like this could be part of the [run_formatting](https://github.com/o1-labs/proof-systems/blob/master/.github/workflows/ci.yml#L45-L64) job instead of it's own workflow?
> 
> I would have the same comment for all of #2963 #2964

Wdyt about https://github.com/o1-labs/proof-systems/pull/2986?

dannywillems:
https://github.com/o1-labs/proof-systems/pull/2964 and https://github.com/o1-labs/proof-systems/pull/2963 will be changed if #2986 is accepted, and this one closed.

dannywillems:
Closing. Old.

PR #2965, opened 2025-01-23
factor out utils module

martyall:
Factor out some utils to a separate module, came up in #2965 when trying to share some code


Fizzixnerd (approved):
I wish `decode_into` wasn't required, but I guess rust move semantics aren't as advanced as I'd like!

marcbeunardeau88 (approved):

PR #2966, opened 2025-01-23
Saffron commitments

martyall:
## Summary
We want the ability for both the user and the storage provider to compute commitments to the data. In the case of the user, there is no reason for them to work on the interpolated polynomials, so this commitment should be computed directly over the evaluation domain using the lagrange basis commitments. In the case of the storage provider, they need to work with interpolated polynomials. This PR adds this functionality and thoroughly tests that they get the same result.

## Changes
- add function to compute commitment over evaluation domain (user uses on client side)
- add function to compute commitment over monomial basis (used by storage provider)
- add srs arguments to main functions, load from cache if possible

## How is it tested
- Add property based test to assert that user and storage provider commitments are the same
- added assertions to the e2e test script to enforce that commitments are the same


marcbeunardeau88 (commented):
I'm not sure why we are folding commitments here ?

marcbeunardeau88 (approved):

PR #2967, opened 2025-01-25
Store commitments in blob

martyall:
## Summary
This PR changes the way we (de)serialize blobs to use macros rather than doing it manually -- there should be no net effect. I also started storing the commitments in the blob metadata to save us from having to recompute them for reads. 

Finally there is a small change to the "sanity check" to make sure the user and storage provider agree on the data. This is not strictly part of the protocol, but rather it helps with the testing. 

## Changes
- use the serde_as package to derive the serialization instances: previously defined serialization instances by hand using arkworks CanonicalSerialize trait. I found other places in the codebase that used this simpler approach.
- store the commitments in the `FieldBlob` struct: otherwise we will need to recompute them for each read request. This means `FieldBlob` is now parameterized by the group instead of the scalar field.
- stop returning the "randomized sum of commitments" from store request -- they are not used until later anyway. We can just return a hash for now for the "sanity check". @marcbeunardeau88  Now I understand this is what you were saying [here](https://github.com/o1-labs/proof-systems/pull/2966#pullrequestreview-2572210365)


saffron/src/commitment.rs
```diff
@@ -26,18 +25,6 @@ pub fn commit_to_field_elems<G: CommitmentCurve>(
         .collect()
 }
 
-#[instrument(skip_all)]
-pub fn commit_to_blob<G: CommitmentCurve>(
```
martyall, line 30: Not deleted -- this and the tests below just got moved to the `blob` module

Fizzixnerd (approved):
looks good

PR #2968, opened 2025-01-25
Proof of storage

martyall:
- add opening proof
- add verification for opening proof
- add test

depends on #2967


martyall:
closing in favor of #2971

PR #2969, opened 2025-01-27
Performance: parallelising kimchi prover & verifier

volhovm:
A bunch of performance improvements to the kimchi prover aiming to utilise more parallelism where available.

As of now, the save-up seems to be as follows (approximate):
screen_2025-02-12_001

TODOs:
- [x] Check how much faster it is on mina side. Adjust the benchmark if necessary (more columns? lookups?)
    - There is a 7% improvement as opposed to the 22% on just prover side. Maybe the benchmark circuit is too simplistic (not enough lookups?). Maybe it's because mina runs some other threads in parallel (like two provers in parallel?).
- [x] Make sure there are no regressions for the verifier side.
    - Parallelising constraints seems to add 3% to the verifier for 1024 gates (and nothing when there are 16k gates?)?? generally this is hard to verify: even running every verifier benchmark for 5 minutes I still get sometimes 0.5% noise. Really hard to tell.
    - Conclusion: 
        - Measuring verification accurately seems to be hard for some reason I do not understand. I checked every single test parameter -- each verification bench runs for about 8 minutes (that's 5-6k iterations in total), and it's averaged over multiple SRS setups, over multiple proofs, and the benchmark type is "linear" which according to the criterion docs is the best way to do it. While in theory this seems more than enough to give reproducible results, I still see average noise around 0.5% (and sometimes up to 2%!!! out of nowhere) on both my laptop and on my server. What's even weirder is that in these cases the distribution has low variance still. I suspect this might be of some system-level threading: maybe due to the highly parallelised verification even 1-2 threads used by some system process causes this noise? I suspect this should be better on a "pristine clean" machine, or if we didn't rely heavily on parallelisation.
        - I _think_ that there was a tiny regression (1-2%: 89ms instead of 86ms) for small circuits because of constraint parallelisation. Not important in any case, but I removed parallelisation for smaller (<4k) circuits. I do not observe any reproducible difference right now.
        - In the end I added another optimisation parallelising the IPA MSM a bit, so in the end verification is faster than before anyway.
- [x] Check if parallel evaluations can be faster -- do we need to clone the vector and then interpolate?
    - Does not matter much. Interpolating directly is only 1% faster, not worth the change.
- [x] Check that `perm_aggreg` optimisation preserves correctness
    - I rewrote this change so that it's now looking much more similar to the old variant. I'm pretty sure the correctness holds because the new code is much more clean.
- [x] Check that optimisations apply to both curves
    - I ran the threading profiler on Pallas too and it looks very uniform too.
- [x] Last step: run the threading profiler once again to make sure I didn't miss anything
- [x] Extra double check: make sure that mina prover (run on actual data using mina benchmarks, not kimchi benchmarks) is parallelised well
    - Yes it is. I dumped inputs to the kimchi prover into a file (altering rust code in bindings), and then read them on the proof-systems side, and re-run the benches. Average thread utilisation is still quite high after my optimisations.

Notion page on how to run this:
https://www.notion.so/o1labs/Profiling-Multithreading-Utilisation-183e79b1f910800c8db9effda569efcd


poly-commitment/src/ipa.rs
```diff
@@ -408,18 +408,36 @@ where
...
         let is_zero = plnm.is_zero();
 
-        let coeffs: Vec<_> = plnm.iter().map(|c| c.into_bigint()).collect();
-
         // chunk while committing
-        let mut chunks = vec![];
-        if is_zero {
-            chunks.push(G::zero());
+        let mut chunks: Vec<_> = if is_zero {
+            vec![G::zero()]
```
mrmr1993, line 413: This breaks the invariant that the length of this vector is equal to `num_chunks`. Likewise below.

volhovm:
For the prover, this is the improvement, visually:

Before:
screen_2025-02-12_004

After:
screen_2025-02-11_001

For the verifier, that's the profile after parallelising the last MSM (10 verifications). MSM takes 95% of the whole computation.

screen_2025-02-12_005

volhovm:
When `transaction-snark-profiler` benchmark is run on the `mina` side with this PR, we see only 8% improvement, not 22%. The summary of "why" seems to be the design of the benchmarks, and proofs are indeed sped up. There are many suspicious 0-performance gaps between proving attempts for the `mina` program which I `perf`-profiled.

Before (this PR):

screen_2025-02-20_004

After (this PR):

screen_2025-02-20_003

poly-commitment/benches/ipa.rs
```diff
@@ -8,6 +8,268 @@ use poly_commitment::{
     commitment::CommitmentCurve, ipa::SRS, utils::DensePolynomialOrEvaluations, PolyComm, SRS as _,
 };
 
+fn benchmark_msm(c: &mut Criterion) {
```
dannywillems, line 11: Can you extract this in a separate PR? It does seem unrelated to the parallelization. It is a benchmark.

poly-commitment/benches/ipa.rs
```diff
@@ -8,6 +8,69 @@ use poly_commitment::{
...
```
dannywillems, line 11: Nit: I would add in the name that it is with Vesta.

poly-commitment/benches/ipa.rs
```diff
@@ -71,6 +71,52 @@ fn benchmark_msm(c: &mut Criterion) {
     }
 }
 
+fn benchmark_msm_parallel(c: &mut Criterion) {
```
dannywillems, line 74: Nit: I would add Vesta in the name.

utils/src/evaluations.rs
```diff
@@ -45,12 +45,10 @@ impl<F: FftField> ExtendedEvaluations<F> for Evaluations<F, Radix2EvaluationDoma
     }
 
     fn shift(&self, len: usize) -> Self {
-        let len = len % self.evals.len();
```
dannywillems, line 48: LGTM.

poly-commitment/src/ipa.rs
```diff
@@ -408,18 +408,36 @@ where
...
+        } else if plnm.len() < self.g.len() {
+            vec![G::Group::msm(&self.g[..plnm.len()], &plnm.coeffs)
+                .unwrap()
+                .into_affine()]
+        } else if plnm.len() == self.g.len() {
+            // when processing a single chunk, it's faster to parallelise vertically in 2 threads
```
dannywillems, line 437: > it's faster to parallelise vertically in 2 threads

Ah? Did you check experimentally?
I would mostly say that it depends on the size.

poly-commitment/src/ipa.rs
```diff
@@ -408,18 +408,36 @@ where
     ) -> PolyComm<G> {
```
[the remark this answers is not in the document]
dannywillems: Here the benchmarks, before and after:

| Benchmark                          | Previous Time (ms)                | New Time (ms)                      | Change (%)                         |
|------------------------------------|----------------------------------|----------------------------------|----------------------------------|
| IPA Commit (SRS 2^{8}, 1 chunk)   | [918.02 µs 970.59 µs 1.0224 ms]  | [664.75 µs 668.64 µs 672.81 µs]  | [-26.897% -23.736% -20.595%]  |
| IPA Commit (SRS 2^{8}, 2 chunks)  | [1.6086 ms 1.6796 ms 1.7612 ms]  | [934.17 µs 942.57 µs 950.52 µs]  | [-45.484% -43.682% -41.931%]  |
| IPA Commit (SRS 2^{8}, 4 chunks)  | [4.9738 ms 5.1678 ms 5.3490 ms]  | [1.5180 ms 1.5297 ms 1.5408 ms]  | [-71.497% -70.485% -69.330%]  |
| IPA Commit (SRS 2^{8}, 8 chunks)  | [10.414 ms 10.734 ms 11.031 ms]  | [2.5372 ms 2.5453 ms 2.5548 ms]  | [-76.934% -76.287% -75.544%]  |
| IPA Commit (SRS 2^{12}, 1 chunk)  | [7.9266 ms 8.2365 ms 8.5341 ms]  | [4.0860 ms 4.1039 ms 4.1232 ms]  | [-51.917% -50.174% -48.231%]  |
| IPA Commit (SRS 2^{12}, 2 chunks) | [15.977 ms 16.508 ms 17.008 ms]  | [8.8777 ms 9.0516 ms 9.2324 ms]  | [-47.067% -45.168% -43.082%]  |
| IPA Commit (SRS 2^{12}, 4 chunks) | [30.793 ms 31.951 ms 33.087 ms]  | [12.561 ms 12.609 ms 12.659 ms]  | [-61.894% -60.538% -59.053%]  |
| IPA Commit (SRS 2^{12}, 8 chunks) | [64.389 ms 66.402 ms 68.297 ms]  | [22.031 ms 22.126 ms 22.229 ms]  | [-67.612% -66.678% -65.633%]  |
| IPA Commit (SRS 2^{16}, 1 chunk)  | [60.901 ms 61.823 ms 62.868 ms]  | [55.346 ms 56.439 ms 57.561 ms]  | [-10.960% -8.7077% -6.2384%]  |
| IPA Commit (SRS 2^{16}, 2 chunks) | [125.20 ms 126.39 ms 127.81 ms]  | [87.581 ms 88.943 ms 90.364 ms]  | [-30.912% -29.627% -28.283%]  |
| IPA Commit (SRS 2^{16}, 4 chunks) | [249.73 ms 251.14 ms 252.61 ms]  | [153.96 ms 155.80 ms 157.66 ms]  | [-38.749% -37.965% -37.188%]  |
| IPA Commit (SRS 2^{16}, 8 chunks) | [500.42 ms 503.06 ms 505.87 ms]  | [262.38 ms 263.91 ms 265.50 ms]  | [-47.966% -47.539% -47.099%]  |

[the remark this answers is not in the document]
dannywillems: The only line we are interested in are:

```
IPA Commit (SRS 2^{16}, 1 chunk)	[60.901 ms 61.823 ms 62.868 ms]	[55.346 ms 56.439 ms 57.561 ms]	[-10.960% -8.7077% -6.2384%]
IPA Commit (SRS 2^{16}, 8 chunks)	[500.42 ms 503.06 ms 505.87 ms]	[262.38 ms 263.91 ms 265.50 ms]	[-47.966% -47.539% -47.099%]
```
as we use a SRS of 2^16. 1 chunk is the one for each wire, and 7 is for the quotient polynomial.

poly-commitment/benches/ipa.rs
```diff
@@ -8,6 +8,37 @@ use poly_commitment::{
     commitment::CommitmentCurve, ipa::SRS, utils::DensePolynomialOrEvaluations, PolyComm, SRS as _,
 };
 
+fn benchmark_ipa_commit(c: &mut Criterion) {
+    let mut group = c.benchmark_group("IPA Commit");
+    let mut rng = o1_utils::tests::make_test_rng(None);
+
+    for srs_size_log in [8, 12, 16].into_iter() {
```
dannywillems, line 246: Should we focus only on the benchmarks we are interested in? i.e. 15 and 16?

poly-commitment/benches/ipa.rs
```diff
@@ -8,6 +8,37 @@ use poly_commitment::{
     commitment::CommitmentCurve, ipa::SRS, utils::DensePolynomialOrEvaluations, PolyComm, SRS as _,
 };
 
+fn benchmark_ipa_commit(c: &mut Criterion) {
```
dannywillems, line 242: Nit: no strong opinion, but it took me a minute or so to find/remember how to run this benchmark only. Would it be worth to add a comment saying how to run this?
```
cargo bench -- "IPA Commit/commitment"
```

poly-commitment/src/ipa.rs
```diff
@@ -408,18 +408,36 @@ where
...
         let is_zero = plnm.is_zero();
 
-        let coeffs: Vec<_> = plnm.iter().map(|c| c.into_bigint()).collect();
-
         // chunk while committing
-        let mut chunks = vec![];
-        if is_zero {
-            chunks.push(G::zero());
+        let mut chunks: Vec<_> = if is_zero {
```
dannywillems, line 430: LGTM

kimchi/src/circuits/constraints.rs
```diff
@@ -373,25 +374,71 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
-            E::<F, D<F>>::from_vec_and_domain(
-                (0..self.domain.d4.size)
-                    .map(|j| w8[i].evals[2 * j as usize])
-                    .collect(),
-                self.domain.d4,
-            )
-        });
+        // the idea is to have threading minimised below a certain threshold.
+        let min_len = {
+            let threads_to_use = if w[0].len() <= 2048 {
```
dannywillems, line 379: Can you explain this value? Why 2048 in particular?

poly-commitment/benches/ipa.rs
```diff
@@ -8,6 +8,69 @@ use poly_commitment::{
...
+    use ark_ec::{AffineRepr, VariableBaseMSM};
+    use ark_ff::PrimeField;
+
+    let mut group = c.benchmark_group("MSM");
+    let mut rng = o1_utils::tests::make_test_rng(None);
+
+    let srs = SRS::<Vesta>::create(1 << 16);
+    srs.get_lagrange_basis_from_domain_size(1 << 16);
+
+    for msm_size_log in [8, 10, 12, 14, 16].into_iter() {
```
dannywillems, line 21: As for ae1034f6f5a5b84ab302becc4da8613d72a488dc, I would simply focus on sizes we care for our usecases, i.e. 16.

poly-commitment/benches/ipa.rs
```diff
@@ -71,6 +71,52 @@ fn benchmark_msm(c: &mut Criterion) {
...
+    use ark_ec::{AffineRepr, VariableBaseMSM};
+    use rayon::prelude::*;
+
+    let mut group = c.benchmark_group("MSM");
+    let mut rng = o1_utils::tests::make_test_rng(None);
+
+    let srs = SRS::<Vesta>::create(1 << 16);
+    srs.get_lagrange_basis_from_domain_size(1 << 16);
+
+    for msm_size_log in [8, 10, 12, 14, 16].into_iter() {
```
dannywillems, line 86: As for ae1034f6f5a5b84ab302becc4da8613d72a488dc, I would focus on sizes we care, i.e. 16.

poly-commitment/benches/ipa.rs
dannywillems: Nit: same than https://github.com/o1-labs/proof-systems/pull/2969#discussion_r1965248308

dannywillems: Nit: same than https://github.com/o1-labs/proof-systems/pull/2969#discussion_r1965248308

poly-commitment/benches/ipa.rs
dannywillems: Same comment as https://github.com/o1-labs/proof-systems/pull/2969#discussion_r1965248308

poly-commitment/benches/ipa.rs
```diff
@@ -105,6 +105,45 @@ fn benchmark_msm_parallel(c: &mut Criterion) {
...
+                            })
+                        },
+                        BatchSize::LargeInput,
+                    )
+                },
+            );
+        }
+    }
+
+    for msm_size_log in [8, 10, 12, 14, 16].into_iter() {
```
dannywillems, line 125: Same comment as https://github.com/o1-labs/proof-systems/pull/2969#discussion_r1965245759

kimchi/src/circuits/constraints.rs
```diff
@@ -373,25 +374,71 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
+                1
+            } else {
+                rayon::max_num_threads()
```
dannywillems, line 382: I would also add some documentation somewhere in the README or in the code that this value can be overwritten by setting `RAYON_NUM_THREADS`.

kimchi/src/circuits/constraints.rs
```diff
@@ -373,25 +374,71 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
 impl<F: PrimeField> ConstraintSystem<F> {
     /// evaluate witness polynomials over domains
     pub fn evaluate(&self, w: &[DP<F>; COLUMNS], z: &DP<F>) -> WitnessOverDomains<F> {
```
dannywillems, line 376: Can you provide a bench for this method, please? And give us a before + after result?
It would be good for regressions and reproducibility.

kimchi/src/circuits/constraints.rs
```diff
@@ -373,25 +374,71 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
+
+        // compute shifted witness polynomials and z8, all in parallel
+        let (w8, z8): ([E<F, D<F>>; COLUMNS], _) = {
+            let mut res = w
+                .par_iter()
+                .chain(rayon::iter::once(z))
+                .with_min_len(min_len)
+                .map(|elem| elem.evaluate_over_domain_by_ref(self.domain.d8))
+                .collect::<Vec<_>>();
+            let z8 = res[COLUMNS].clone();
```
dannywillems, line 399: Is it worth doing this? It implies an unnecessary cloning + truncate, for, I think, a negligible gain to evaluate z in parallel with the columns.
Cloning will allocate new memory and truncate will call an unnecessary drop_in_place. On the web-assembly side, it might not be negligible.
What about a `rayon.join`? Or simply a sequential call to `w.par_iter()` without `chain(rayon::iter::once(z))` followed by a simple call to `z.evaluate_over_domain_by_ref(self.domain.d8)`.

kimchi/src/circuits/constraints.rs
```diff
@@ -373,25 +374,71 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
+                )
+            })
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
+
         let z4 = DP::<F>::zero().evaluate_over_domain_by_ref(D::<F>::new(1).unwrap());
+        let z8_shift8 = z8.shift(8);
+
+        let d4_next_w: [_; COLUMNS] = w4
```
dannywillems, line 422: LGTM

poly-commitment/src/commitment.rs
```diff
@@ -332,19 +333,29 @@ impl<C: AffineRepr> PolyComm<C> {
...
+        let chunks = (0..elems_size)
+            .map(|chunk| {
+                let (points, scalars): (Vec<_>, Vec<_>) = com
+                    .iter()
+                    .zip(&all_scalars)
+                    // get rid of scalars that don't have an associated chunk
+                    .filter_map(|(com, scalar)| com.chunks.get(chunk).map(|c| (c, scalar)))
+                    .unzip();
+
+                // splitting into 2 chunks seems optimal; but in
```
dannywillems, line 346: > splitting into 2 chunks seems optimal
Is there a bench to confirm this?

poly-commitment/src/commitment.rs
```diff
@@ -332,19 +333,29 @@ impl<C: AffineRepr> PolyComm<C> {
         let all_scalars: Vec<_> = elm.iter().map(|s| s.into_bigint()).collect();
 
         let elems_size = Iterator::max(com.iter().map(|c| c.chunks.len())).unwrap();
```
dannywillems, line 335: The code LGTM in terms of functionality.
However, I do not see any unit test and bench for this particular function (only some using `msm` directly).
I think it would be nice to have to compare before and after, and for regression (bencharmking) tests later (as I feel the performance team will have later?).

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -257,32 +257,45 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
             // (w[1](x) + gamma + x * beta * shift[1]) * ...
             // (w[6](x) + gamma + x * beta * shift[6])
             // in evaluation form in d8
```
dannywillems, line 259: Same comment than https://github.com/o1-labs/proof-systems/pull/2969/commits/8e13b285588c68c903cb8bb106f83c793bc9e1bc#r1965351096

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -257,32 +257,45 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
-            let mut shifts = lagrange.d8.this.z.clone();
-            for (witness, shift) in lagrange.d8.this.w.iter().zip(self.cs.shift.iter()) {
-                let term =
-                    &(witness + gamma) + &self.cs.precomputations().poly_x_d1.scale(beta * shift);
-                shifts = &shifts * &term;
```
dannywillems, line 264: Interestingly, it seems that the old version might be doing unnecessary cloning by using `s = s * t` instead of `s *= t`.
`=` is implemented by `Mul` on `EvaluationDomain<F>`. The code is (in arkworks/univariate.rs):
```rust
    #[inline]
    fn mul(self, other: &'a Evaluations<F, D>) -> Evaluations<F, D> {
        let mut result = self.clone();
        result *= other;
        result
    }
```
where the new code you're using is `*=`, which seems to be the syntax that would be semantically translated into [MulAssign](https://doc.rust-lang.org/std/ops/trait.MulAssign.html), is:
```rust
impl<'a, F: FftField, D: EvaluationDomain<F>> MulAssign<&'a Evaluations<F, D>>
    for Evaluations<F, D>
{
    #[inline]
    fn mul_assign(&mut self, other: &'a Evaluations<F, D>) {
        assert_eq!(self.domain, other.domain, "domains are unequal");
        ark_std::cfg_iter_mut!(self.evals)
            .zip(&other.evals)
            .for_each(|(a, b)| *a *= b);
    }
}
```

Cool.

kimchi/src/prover.rs
```diff
@@ -297,35 +297,46 @@ where
         //~    Note: since the witness is in evaluation form,
         //~    we can use the `commit_evaluation` optimization.
         internal_tracing::checkpoint!(internal_traces; commit_to_witness_columns);
```
dannywillems, line 299: Could you provide a bench before and after this commit, with the command to run? It would be nice for the reviewer.

kimchi/src/prover.rs
```diff
@@ -297,35 +297,46 @@ where
...
+            Some(blinders_arr) => blinders_arr
+                .into_iter()
+                .map(|blinder_el| match blinder_el {
+                    None => PolyComm::new(vec![UniformRand::rand(rng)]),
+                    Some(blinder_el_some) => blinder_el_some,
+                })
+                .collect(),
+        };
+        let w_comm_opt_res: Vec<Result<_>> = witness
+            .clone()
```
dannywillems, line 314: Q: can we avoid the clone by passing a ref later?

kimchi/src/prover.rs
```diff
@@ -297,35 +297,46 @@ where
...
-            // witness coeff -> witness eval
-            let witness_eval =
-                Evaluations::<G::ScalarField, D<G::ScalarField>>::from_vec_and_domain(
-                    witness[col].clone(),
-                    index.cs.domain.d1,
-                );
+        // generate blinders if not given externally
+        let blinders_final: Vec<PolyComm<G::ScalarField>> = match blinders {
+            None => (0..COLUMNS)
+                .map(|_| PolyComm::new(vec![UniformRand::rand(rng)]))
```
dannywillems, line 303: Even though the idea is the same (generating a random value when no blinder is given), it is not calling exactly the same code with the previous version.
It is a change indirectly related to the commit itself, which is titled:
```
Kimchi prover: parallelise witness creation (+5% prover but 1% verifier regression?)
```

In particular, I am not sure it would work in the case of multiple chunks for the witness. In this change, you are returning a single blinder, where in the previous version, when there is no given blinder, we call `commit_evaluations`, which calls:
```
        self.mask(self.commit_evaluations_non_hiding(domain, plnm), rng)
```
where `self.mask` is:
```
        let blinders = comm.map(|_| G::ScalarField::rand(rng));
        self.mask_custom(comm, &blinders).unwrap()
```
i.e. multiple chunks are handled as the blinders definition is mapping over the chunks.

Do we have any regression test for multiple chunks?

kimchi/src/prover.rs
```diff
@@ -340,13 +351,18 @@ where
...
-            Evaluations::<G::ScalarField, D<G::ScalarField>>::from_vec_and_domain(
-                witness[i].clone(),
-                index.cs.domain.d1,
-            )
-            .interpolate()
-        });
+        let witness_poly: [DensePolynomial<G::ScalarField>; COLUMNS] = (0..COLUMNS)
+            .into_par_iter()
+            .map(|i| {
+                Evaluations::<G::ScalarField, D<G::ScalarField>>::from_vec_and_domain(
```
dannywillems, line 357: For a follow-up optimisation: we should have a method somewhere that allows use to interpolate directly using a ref on the Vec<F> directly, instead of a ref on `Evaluations` (we do have `interpolate_by_ref` already, but the ref is on `Evaluations`).
```
    /// Interpolate a polynomial from a list of evaluations
    pub fn interpolate_by_ref(&self) -> DensePolynomial<F> {
        DensePolynomial::from_coefficients_vec(self.domain.ifft(&self.evals))
    }
```
gives directly the evals.
We can avoid cloning `witness[i]` L358.

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -481,27 +478,58 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
-            z[j + 1] = witness
-                .iter()
-                .zip(self.column_evaluations.permutation_coefficients8.iter())
-                .map(|(w, s)| w[j] + (s[8 * j] * beta) + gamma)
-                .fold(F::one(), |x, y| x * y);
-        }
+
+        // We compute z such that:
+        // z[0] = 1
+        // z[j+1] = \Prod_{i=0}^{COLUMNS}(wit[i][j] + (s[i][8*j] * beta) + gamma)     for j ∈ 0..n-1
```
dannywillems, line 484: `COLUMNS` is not the correct variable. It is `PERMUTS`.
```
witness
            .par_iter()
            .zip(self.column_evaluations.permutation_coefficients8.par_iter())
```
will take the size of smaller iterator.

`witness` is of size `COLUMNS` (15) - the fifteen wires of Kimchi - where `[...permutation_coefficients8]` is of size `PERMUTS` (8) - the eight wires that are used by the permutation argument.

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -481,27 +478,58 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
+        //
+        // We compute every product batch separately first (one batch
+        // per i∈[COLUMNS]), and then multiply all batches together.
+        let mut z: Vec<F> = witness
+            .par_iter()
+            .zip(self.column_evaluations.permutation_coefficients8.par_iter())
+            .map(|(w_i, perm_coeffs8_i)| {
+                let mut output_vec: Vec<_> = vec![F::one(); 1];
```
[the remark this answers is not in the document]
dannywillems: Oopsie. I think it is `PERMUTS`.

poly-commitment/src/ipa.rs
volhovm: Yes. I need to refer to a benchmark.

kimchi/src/prover.rs
volhovm: I did check that and we could write this function, but it is same or 1% slower because `ifft_in_place` is a bit faster than `ifft` (by reference), so not worth adding more code.

kimchi/src/circuits/constraints.rs
dannywillems: As discussed, it seems that you did it manually, and you got better number when the circuit is smaller than 2^11. However, we don't care about circuits smaller than, let's say, 14 or 15 (our real usecases are 15 and 16). I would suggest to remove this code as we should only care about our cases.

kimchi/src/circuits/constraints.rs
dannywillems: It seems yes, as cloning + deallocating does not seem to cost much.
Makes sense!

kimchi/src/circuits/constraints.rs
dannywillems: As discussed, it is not required for this PR, but if we want to provide continuous bench (as I think the team is working on?), that would be nice.

poly-commitment/src/ipa.rs
```diff
@@ -278,9 +278,27 @@ impl<G: CommitmentCurve> SRS<G> {
             sg_rand_base_i *= &sg_rand_base;
         }
 
-        // verify the equation
-        let scalars: Vec<_> = scalars.iter().map(|x| x.into_bigint()).collect();
-        G::Group::msm_bigint(&points, &scalars) == G::Group::zero()
+        // verify the equation in two chunks, which is optimal
+        // according to our MSM benches. But don't parallelise at all
+        // if the MSM is too small (heuristically).
```
dannywillems, line 283: Note that most of the time, we will have sizes equal to `1 << 15` or `1 << 16`.

dannywillems (changes requested):
Ending this review cycle :exploding_head: 
Only requesting a change reg. the blinders. The other comments are not blocking, but nitpicks.

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -481,27 +478,58 @@ impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
...
         //~
         for j in 0..n - 1 {
             if j != n - zk_rows && j != n - zk_rows + 1 {
                 let x = z[j];
-                z[j + 1] *= witness
-                    .iter()
-                    .zip(self.cs.shift.iter())
-                    .map(|(w, s)| w[j] + (self.cs.sid[j] * beta * s) + gamma)
-                    .fold(x, |z, y| z * y);
+                z[j + 1] *= z_prefolded[j + 1] * x;
```
dannywillems, line 532: I'm still confused by this `j + 1` (instead of `j`).

kimchi/src/circuits/constraints.rs
volhovm: Removed!

poly-commitment/benches/ipa.rs
volhovm: I kind of want to keep it here because it's cheap, hope it's OK -- it's good to see at which value the optimisation starts to make sense, this gives a rough feeling of how big the chunk should be IMO

kimchi/src/prover.rs
volhovm: Nice catch, fixed! 

Re test we have something like:
```
#[test]
fn heavy_test_2_to_17_chunked_generic_gate_pub() {
    test_generic_gate_with_srs_override(17, Some(1 << 16))
}
```

I guess the previous code worked because ... well, just some chunks were not blinded, which is not a soundness issue, but a ZK one. Now it should work as expected.

kimchi/src/prover.rs
volhovm: the command is `cargo criterion -p kimchi --bench proof_criterion`. I'll re run it now, will post the results later. It's actually not introducing a big verifier regression -- I think it was benchmark noise after all. The prover speedup is indeed a few % I think.

kimchi/src/circuits/constraints.rs
volhovm: I'll post the summary of all the prover criterion benchmarks tomorrow per commit, it should be OK for now? NB: even the current benchmarks still have A LOT of verifier noise (~1%, sometimes up to 3%, maybe it's just my machine), so the prover data is much more reliable.

kimchi/src/circuits/polynomials/permutation.rs
volhovm: the link seems broken, not sure what you meant?

volhovm:
@dannywillems I think I addressed almost all the comments, see the commits from today. All the changes are small :)

If you're happy I'd suggest proceeding as follows:
- I'll squash the "fix comment" commits and fixes with the original commits so that commits are atomic again.
- Then I'll split the branch into two separate PRs: one with benches and one with performance improvements
    - And then you can auto-accept and we merge those?

kimchi/src/circuits/polynomials/permutation.rs
dannywillems: Mmh, I don't remember :confused:. I guess it was while reviewing commit by commit, and the comment has not been added at the right place :confused:

kimchi/src/prover.rs
```diff
@@ -304,8 +304,15 @@ where
...
-                .map(|blinder_el| match blinder_el {
-                    None => PolyComm::new(vec![UniformRand::rand(rng)]),
+                .zip(witness.iter())
+                .map(|(blinder_el, witness)| match blinder_el {
+                    None => {
+                        // `(a + b - 1) / b` is `ceil(a / b)`, gives number of chunks in the
+                        // witness vector
+                        let wit_chunks = (witness.len() + index.cs.domain.d1.size() - 1)
+                            / index.cs.domain.d1.size();
+                        PolyComm::new(vec![UniformRand::rand(rng); wit_chunks])
```
dannywillems, line 314: Just noticed there was a simpler solution using `num_chunks` defined L192

dannywillems: I tried [this code](https://github.com/o1-labs/proof-systems/pull/3069).

kimchi/src/prover.rs
volhovm: Here's this particular witness optimisation improvements:
```
Proof creation/proof creation (SRS size 2^{16}, 1014 gates)                                                                    
                        time:   [588.00 ms 588.67 ms 589.37 ms]                                                                
                        change: [-2.0525% -1.8817% -1.7093%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.              
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                                                                   
                        time:   [3.3926 s 3.3952 s 3.3979 s]                                                                   
                        change: [-7.0408% -6.9277% -6.8115%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                                                            
                        time:   [87.873 ms 87.912 ms 87.952 ms]                                                                
                        change: [-3.6469% -3.5316% -3.4324%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                                                                                         
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                                                           
                        time:   [87.957 ms 87.987 ms 88.018 ms]                                                                
                        change: [-3.7976% -3.7273% -3.6568%] (p = 0.00 < 0.05)                                                 
                        Performance has improved. 
```

kimchi/src/circuits/constraints.rs
volhovm: will post in the main thread

volhovm:
Here's another summary run for every commit. Note that verifier benchmarks regularly have 1% of noise, sometimes up to 3%, so... the final time is better than the starting time, that's what's most important I think (fluctuations are harder to judge):

```
$ for i in $(seq 0 9); do cargo criterion -p kimchi --bench proof_criterion; git rebase --continue; done                     

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)   
                        time:   [601.39 ms 602.01 ms 602.62 ms]                                                                                               
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)  
                        time:   [4.2620 s 4.2634 s 4.2648 s]                                                                                                  
                                                                                                                                                              
Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                                                           
                        time:   [84.537 ms 84.567 ms 84.596 ms]                                                                                               
Benchmarking Proof verification/proof verification (SRS size 2^
                        time:   [88.006 ms 88.042 ms 88.075 ms]                                                                                               

===== Stopped at 39ae7379da...  Parallelise IPA commit function                                                                      

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)                                                                     
                        time:   [601.89 ms 602.47 ms 603.05 ms]                                                                
                        change: [-0.4876% -0.3485% -0.2157%] (p = 0.00 < 0.05)                                                 
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                                    
                        time:   [3.9782 s 3.9800 s 3.9819 s]                                                                   
                        change: [-4.8482% -4.7885% -4.7302%] (p = 0.00 < 0.05)                                                 
                                                                                                                               
Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                             
                        time:   [88.842 ms 88.867 ms 88.892 ms]                                                                
                        change: [-3.3113% -3.2213% -3.1282%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                            
                        time:   [88.699 ms 88.726 ms 88.751 ms]                                                                
                        change: [-3.2735% -3.2073% -3.1427%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.

====== Stopped at a4817fc6d3...  Parallelise constraints creation                                                                     

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)     
                        time:   [604.35 ms 604.88 ms 605.41 ms]                                                                
                        change: [+0.2683% +0.3998% +0.5272%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                        
                        time:   [3.9551 s 3.9589 s 3.9628 s]                                                                   
                        change: [-0.6385% -0.5285% -0.4219%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                        
                        time:   [87.750 ms 87.775 ms 87.801 ms]                                                                
                        change: [-1.1939% -1.1456% -1.0970%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                        
                        time:   [88.426 ms 88.448 ms 88.469 ms]                                                                
                        change: [-0.3134% -0.2482% -0.1744%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)     
                        time:   [602.52 ms 603.03 ms 603.55 ms]                                                                
                        change: [-0.4330% -0.3054% -0.1831%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                        
                        time:   [3.7116 s 3.7149 s 3.7183 s]                                                                   
                        change: [-6.2870% -6.1644% -6.0364%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                        
                        time:   [86.751 ms 86.785 ms 86.819 ms]                                                                
                        change: [-1.1974% -1.1431% -1.0922%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                        
                        time:   [90.222 ms 90.267 ms 90.308 ms]                                                                
                        change: [+1.9158% +1.9913% +2.0679%] (p = 0.00 < 0.05)                                                 
                        Performance has regressed.                                                                             

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)     
                        time:   [600.86 ms 601.34 ms 601.83 ms]                                                                
                        change: [-0.3927% -0.2803% -0.1644%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                        
                        time:   [3.6501 s 3.6530 s 3.6559 s]                                                                   
                        change: [-1.7842% -1.6664% -1.5471%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                        
                        time:   [90.913 ms 90.962 ms 91.010 ms]                                                                
                        change: [+4.7226% +4.8022% +4.8846%] (p = 0.00 < 0.05)                                                 
                        Performance has regressed.                                                                             
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                        
                        time:   [91.103 ms 91.162 ms 91.234 ms]                                                                
                        change: [+1.1296% +1.2190% +1.3157%] (p = 0.00 < 0.05)                                                 
                        Performance has regressed.                                                                             

Stopped at 2a594be570...  Kimchi prover: parallelise witness creation (+5% prover but 1% verifier regression?)


Proof creation/proof creation (SRS size 2^{16}, 1014 gates)     
                        time:   [584.88 ms 585.46 ms 586.05 ms]                                                                
                        change: [-2.7607% -2.6403% -2.5105%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                        
                        time:   [3.3984 s 3.4007 s 3.4031 s]                                                                   
                        change: [-7.0056% -6.9057% -6.8070%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                        
                        time:   [87.472 ms 87.501 ms 87.530 ms]                                                                
                        change: [-3.8774% -3.8066% -3.7349%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                        
                        time:   [88.001 ms 88.055 ms 88.107 ms]                                                                
                        change: [-3.5620% -3.4581% -3.3614%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              

Stopped at 6095e0032e...  Parallelise perm_aggreg sub arrays                                                                   

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)     
                        time:   [588.50 ms 589.16 ms 589.91 ms]                                                                
                        change: [+0.4854% +0.6309% +0.7912%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                        
                        time:   [3.3283 s 3.3303 s 3.3324 s]                                                                   
                        change: [-2.1609% -2.0709% -1.9787%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                        
                        time:   [89.053 ms 89.085 ms 89.119 ms]                                                                
                        change: [+1.8083% +1.8716% +1.9378%] (p = 0.00 < 0.05)                                                 
                        Performance has regressed.                                                                             
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                        
                        time:   [88.934 ms 88.981 ms 89.034 ms]                                                                
                        change: [+0.9164% +1.0012% +1.0890%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         


Stopped at 3238ccc5ec...  Parallelise last MSM in IPA check                                                                    

Proof creation/proof creation (SRS size 2^{16}, 1014 gates)     
                        time:   [589.05 ms 589.72 ms 590.48 ms]                                                                
                        change: [-0.0840% +0.0957% +0.2587%] (p = 0.28 > 0.05)                                                 
Proof creation/proof creation (SRS size 2^{16}, 65526 gates)                        
                        time:   [3.3406 s 3.3428 s 3.3450 s]                                                                   
                        change: [+0.2793% +0.3739% +0.4663%] (p = 0.00 < 0.05)                                                 
                        Change within noise threshold.                                                                         

Proof verification/proof verification (SRS size 2^{16}, 1024 gates)                        
                        time:   [82.467 ms 82.599 ms 82.744 ms]                                                                
                        change: [-7.4204% -7.2890% -7.1561%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                                                                              
Proof verification/proof verification (SRS size 2^{16}, 16384 gates)                        
                        time:   [82.665 ms 82.837 ms 83.017 ms]                                                                
                        change: [-7.4201% -7.2035% -6.9956%] (p = 0.00 < 0.05)                                                 
                        Performance has improved.                         
```

dannywillems:
> Then I'll split the branch into two separate PRs: one with benches and one with performance improvements
> 
>     * And then you can auto-accept and we merge those?

The PR is well-organized. I am fine with merging all of it, as it seems to be straightforward to revert commits if required.

dannywillems (approved):

PR #2970, opened 2025-01-27
Saffron CI Speedups and Cleanups

martyall:
## Summary
Collects some improvements to running tests / CI that came up while working on #2968 , felt like it was better to pull them out and apply them in a separate PR

## Changes
- Add SRS caching in CI which reduces the e2e test time from [~3 min](https://github.com/o1-labs/proof-systems/actions/runs/12995728856/job/36243038553?pr=2967) to [20 seconds](https://github.com/o1-labs/proof-systems/actions/runs/12998174831/job/36250749999?pr=2970) :exploding_head:
- Refine the random data generation for the blob unit tests
- add console logging subscriber to tests (still controlled by `RUST_LOG` env var)

### DIFF
--- a/saffron/src/blob.rs
+++ b/saffron/src/blob.rs
@@ -22,7 +22,7 @@ pub struct FieldBlob<G: CommitmentCurve> {
     pub data: Vec<DensePolynomial<G::ScalarField>>,
 }
 
-#[instrument(skip_all)]
+#[instrument(skip_all, level = "debug")]
 fn commit_to_blob_data<G: CommitmentCurve>(
     srs: &SRS<G>,
     data: &[DensePolynomial<G::ScalarField>],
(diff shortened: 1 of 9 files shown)


saffron/src/blob.rs
```diff
@@ -92,28 +92,102 @@ impl<G: CommitmentCurve> FieldBlob<G> {
...
+    }
+
+    // TODO: This is fine for now, but figure out what the realistic expectations are.
+    impl DataSize {
+        fn size_range_bytes(&self) -> (usize, usize) {
+            let c = 1_000_000;
+            match self {
+                Self::Small => (1, 2 * c),
+                Self::Medium => (5 * c, 10 * c),
+                Self::Large => (50 * c, 75 * c),
```
Fizzixnerd, line 116: I'm a little worried about the spaces in between where we don't test.  Should I be?

Fizzixnerd (approved):

saffron/src/blob.rs
martyall: Yeah this is pretty ad-hoc, what do you think about https://github.com/o1-labs/proof-systems/pull/2970/commits/b815f49d3efdd5424907e7b55ed5c319137fd664 ? Or any other ideas?

PR #2972, opened 2025-01-28
[saffron] Property based test for query

martyall:
- move the `blob_test_utils` module to `utils.rs` and use as standard way for all data generation. Renamed the generator from `BlobData` to `UserData` to reflect that it has nothing to do with blobs apriori

- reworked @marcbeunardeau88 's test to generate and test multiple queries for every user data generation round

### DIFF
--- a/saffron/src/blob.rs
+++ b/saffron/src/blob.rs
@@ -93,85 +93,12 @@ impl<G: CommitmentCurve> FieldBlob<G> {
 }
 
 #[cfg(test)]
-mod blob_test_utils {
-    use proptest::prelude::*;
-
-    #[derive(Debug)]
-    pub struct BlobData(pub Vec<u8>);
-
-    #[derive(Clone, Debug)]
-    pub enum DataSize {
-        Small,
-        Medium,
-        Large,
-    }
-
-    impl DataSize {
-        const KB: usize = 1_000;
-        const MB: usize = 1_000_000;
-
-        fn size_range_bytes(&self) -> (usize, usize) {
-            match self {
-                // Small: 1KB - 1MB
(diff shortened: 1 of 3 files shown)


saffron/src/blob.rs
```diff
@@ -92,86 +92,13 @@ impl<G: CommitmentCurve> FieldBlob<G> {
     }
 }
 
-#[cfg(test)]
-mod blob_test_utils {
```
martyall, line 96: this got copy/pasted over to `utils.rs` since it is useful for generating data in pretty much all cases

saffron/src/utils.rs
```diff
@@ -57,17 +56,6 @@ pub struct QueryBytes {
     pub len: usize,
 }
 
-/// For testing purposes
-impl QueryBytes {
```
martyall, line 61: we don't need this anymore since we can use the combinators from proptest to generate exactly what we want

saffron/src/utils.rs
```diff
@@ -230,33 +299,27 @@ mod tests {
...
-                prop_assert!(base * multiplied != 0);
-            });
+        fn test_query(
+            (UserData(xs), queries) in UserData::arbitrary()
+                .prop_flat_map(|xs| {
+                    let n = xs.len();
+                    let query_strategy = (0..(n - 1)).prop_flat_map(move |start| {
+                        ((start + 1)..n).prop_map(move |end| QueryBytes { start, len: end - start})
+                    });
+                    let queries_strategy = prop::collection::vec(query_strategy, 10);
```
martyall, line 312: generate and test 10 queries for every round of testing

marcbeunardeau88 (approved):

PR #2978, opened 2025-01-30
Arrabbiata: specialise E1::BaseField into a PrimeField

dannywillems:
In addition to that, a helper to get the curve parameters is requested into the implementation of the trait ArrabbiataCurve. This is to avoid adding additional type constraints when accessing the parameters is required.

### DIFF
--- a/arrabbiata/src/constraints.rs
+++ b/arrabbiata/src/constraints.rs
@@ -5,7 +5,6 @@ use crate::{
     interpreter::{self, Instruction, Side},
     MAX_DEGREE, NUMBER_OF_COLUMNS, NUMBER_OF_PUBLIC_INPUTS,
 };
-use ark_ec::{short_weierstrass::SWCurveConfig, CurveConfig};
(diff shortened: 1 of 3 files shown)


arrabbiata/src/constraints.rs
```diff
@@ -30,11 +31,12 @@ pub struct Env<C: ArrabbiataCurve> {
 
 impl<C: ArrabbiataCurve> Env<C>
 where
-    <<C as CommitmentCurve>::Params as CurveConfig>::BaseField: PrimeField,
+    C::BaseField: PrimeField,
 {
     pub fn new() -> Self {
         // This check might not be useful
-        let a: BigInt = <C as CommitmentCurve>::Params::COEFF_A.to_biguint().into();
```
dannywillems, line 37: The helper is to satisfy this line, and avoiding adding type constraints.

marcbeunardeau88 (approved):

PR #2982, opened 2025-01-30
Arrabbiata: implement absorb_state

dannywillems:


dannywillems:
For a follow-up: draw the diagram with the digest after and before execution.

PR #2989, opened 2025-02-03
Saffron: initial README

dannywillems:


martyall:
Great! I hope you don't mind, but I pushed a small change directly to this branch RE invoking the e2e test. It also seemed like a good time to name the script what it actually is :upside_down_face:

martyall (approved):

PR #3000, opened 2025-02-04
o1vm/lookup: add lookup prover until alpha

marcbeunardeau88:
first part of the prover that will prove that the sum of the lookup from one circuit goes from one public to one public output


marcbeunardeau88:
redoing in a slightly different way on another branch

PR #3001, opened 2025-02-04
o1vm/lookup: add expressions for lookup constraint

marcbeunardeau88:
Lay the ground work to use a specific challenge term and column type for the lookup prover.
Will allow not to reuse the msm columns and berkeley challenge as we usually do


marcbeunardeau88:
redoing in a slightly different way on another branch

PR #3002, opened 2025-02-04
Marc/track lookup arity

marcbeunardeau88:
We need to keep track of the arity of the lookup we creating, so that the prover can later correctly collapse term into beta+sum_i gamma^i*lookup[i] where gamma is the lookup combiner


mrmr1993 (approved):

dannywillems:
Reiterating approval of @mrmr1993

dannywillems (approved):

PR #3004, opened 2025-02-05
[Saffron] Add update function

martyall:
## Summary
The user needs to be able to update data with the storage provider. This is done by posting "diff" objects, not by simply posting the raw updated data. This PR implements this feature.

## Changes
- add `Diff` type and utils function for computing a `Diff` over binary data. 
- add update method on `FieldBlob` type. Add tests for random updates on random data.
- add two cli commands:
  1. `calculate-diff`: used by the user to calculate the `Diff` object using the `old` and `new` files as input https://github.com/o1-labs/proof-systems/commit/6eeb13eabe277b0bc978c8ede2e0dda16b6fcd75
  2. `update`: Ask the storage provider to update some given data using the `Diff` object calculated by `(1)` https://github.com/o1-labs/proof-systems/commit/61bb85ffdb1d50d0c559114a6fc34b644aa1c876
- update e2e test script to perform a diff operation and check that the data was updated correctly. https://github.com/o1-labs/proof-systems/commit/2ba1a3f3c3f8ab9ddc6ecd5f8febe6e9b6a66c21

### DIFF
--- a/saffron/e2e-test.sh
+++ b/saffron/e2e-test.sh
@@ -33,11 +33,39 @@ compare_files() {
    fi
 }
 
-perturb_bytes() {
+perturb_file_bytes() {
    local input_file=$1
    local output_file=$2
    local threshold=${3:-0.1}  # Default 10% chance
-   perl -pe 'rand() < '$threshold' and $_ = chr(rand(256)) for split ""' "$input_file" > "$output_file"
+
+   python3 -c '
+import sys
+import random
+
+threshold = float(sys.argv[3])
+
+with open(sys.argv[1], "rb") as f:
+   data = bytearray(f.read())
+
+changed = False
+for i in range(len(data)):
+   if random.random() < threshold:
+       data[i] = random.randrange(256)
+       changed = True
+
+# Force at least one change if none made
+if not changed:
+   pos = random.randrange(len(data))
+   data[pos] = random.randrange(256)
+
+with open(sys.argv[2], "wb") as f:
+   f.write(data)
+' "$input_file" "$output_file" "$threshold"
+
+   if cmp -s "$input_file" "$output_file"; then
+       echo "Error: Output file is identical to input file. Perturbation failed."
+       exit 1
+   fi
 }
 
 # Ensure input file exists
@@ -88,8 +116,8 @@ fi
 # Compare file to original
 compare_files "$INPUT_FILE" "$DECODED_FILE"
 
-# Usage example in your script:
-perturb_bytes "$INPUT_FILE" "$PERTURBED_FILE" 0.1
+# create a random update to the original data
+perturb_file_bytes "$INPUT_FILE" "$PERTURBED_FILE" 0.25
 
 echo "Calculating diff for upated $INPUT_FILE (stored updated data in $PERTURBED_FILE)"
 cargo run --release --bin saffron calculate-diff --old "$INPUT_FILE" --new "$PERTURBED_FILE" -o "$ENCODED_DIFF_FILE" $SRS_ARG
--- a/saffron/e2e-test.sh
+++ b/saffron/e2e-test.sh
@@ -13,6 +13,31 @@ if [ $# -eq 2 ]; then
 fi
 ENCODED_FILE="${INPUT_FILE%.*}.bin"
 DECODED_FILE="${INPUT_FILE%.*}-decoded${INPUT_FILE##*.}"
+PERTURBED_FILE="${INPUT_FILE%.*}_perturbed"
+ENCODED_DIFF_FILE="${ENCODED_FILE%.*}_diff.bin"
+
+compare_files() {
+   local file1="$1"
+   local file2="$2"
+   
+   echo "Comparing files..."
+   if cmp -s "$file1" "$file2"; then
+       echo "✓ Success: Files are identical"
+   else
+       echo "✗ Error: Files differ"
+       echo "Keeping files for inspection"
+       diff "$file1" "$file2"
+       exit 1
+   fi
+}
+
+perturb_bytes() {
+   local input_file=$1
+   local output_file=$2
+   local threshold=${3:-0.1}  # Default 10% chance
+   
+   perl -pe "rand() < $threshold and \$_ = chr(rand(256))" "$input_file" > "$output_file"
+}
 
 # Ensure input file exists
 if [ ! -f "$INPUT_FILE" ]; then
@@ -22,8 +47,6 @@ fi
 
 # Compute commitment and capture last line
 COMMITMENT=$(cargo run --release --bin saffron compute-commitment -i "$INPUT_FILE" $SRS_ARG | tee /dev/stderr | tail -n 1)
-
-
 
 # Run encode with captured commitment
 echo "Encoding $INPUT_FILE to $ENCODED_FILE"
@@ -61,16 +84,28 @@ if ! cargo run --release --bin saffron decode -i "$ENCODED_FILE" -o "$DECODED_FI
     exit 1
 fi
 
-# Compare files
-echo "Comparing original and decoded files..."
-if cmp -s "$INPUT_FILE" "$DECODED_FILE"; then
-    echo "✓ Success: Files are identical"
-    echo "Cleaning up temporary files..."
-    rm -f "$ENCODED_FILE" "$DECODED_FILE"
-    exit 0
-else
-    echo "✗ Error: Files differ"
-    echo "Keeping temporary files for inspection"
-    diff "$INPUT_FILE" "$DECODED_FILE"
+# Compare file to original
+compare_files "$INPUT_FILE" "$DECODED_FILE"
+
+# Usage example in your script:
+perturb_bytes "$INPUT_FILE" "$PERTURBED_FILE" 0.1
+
+echo "Calculating diff for upated $INPUT_FILE (stored updated data in $PERTURBED_FILE)"
+cargo run --release --bin saffron calculate-diff --old "$INPUT_FILE" --new "$PERTURBED_FILE" -o "$ENCODED_DIFF_FILE" $SRS_ARG
+
+echo "Updating file with Storage Provider"
+cargo run --release --bin saffron update -i "$ENCODED_FILE" --diff-file "$ENCODED_DIFF_FILE" --assert-commitment "$COMMITMENT" $SRS_ARG
+
+# Run decode
+echo "Decoding $ENCODED_FILE to $DECODED_FILE"
+if ! cargo run --release --bin saffron decode -i "$ENCODED_FILE" -o "$DECODED_FILE" $SRS_ARG; then
+    echo "Decoding failed"
     exit 1
 fi
+
+# Compare update file to perturbed file
+compare_files "$PERTURBED_FILE" "$DECODED_FILE"
+
+echo "Cleaning up temporary files..."
+rm -f "$ENCODED_FILE" "$DECODED_FILE" "$PERTURBED_FILE" "$ENCODED_DIFF_FILE"
+exit 0
--- a/saffron/e2e-test.sh
+++ b/saffron/e2e-test.sh
@@ -13,10 +13,10 @@ if [ $# -eq 2 ]; then
 fi
 
 ENCODED_FILE="${INPUT_FILE%.*}.bin"
-DECODED_FILE="${INPUT_FILE%.*}-decoded${INPUT_FILE##*.}"
-PERTURBED_FILE="${INPUT_FILE%.*}_perturbed${INPUT_FILE##*.}"
+DECODED_FILE="${INPUT_FILE%.*}_decoded.${INPUT_FILE##*.}"
+PERTURBED_FILE="${INPUT_FILE%.*}_perturbed.${INPUT_FILE##*.}"
 ENCODED_DIFF_FILE="${ENCODED_FILE%.*}_diff.bin"
-DECODED_PERTURBED_FILE="${PERTURBED_FILE}-decoded${INPUT_FILE##*.}"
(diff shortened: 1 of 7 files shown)


saffron/e2e-test.sh
```diff
@@ -11,8 +11,34 @@ SRS_ARG=""
...
+       echo "✓ Success: Files are identical"
+   else
+       echo "✗ Error: Files differ"
+       echo "Keeping files for inspection"
+       diff "$file1" "$file2"
+       exit 1
+   fi
+}
+
+perturb_bytes() {
```
dannywillems, line 36: Are you sure this works?

saffron/e2e-test.sh
```diff
@@ -11,8 +11,34 @@ SRS_ARG=""
 if [ $# -eq 2 ]; then
    SRS_ARG="--srs-filepath $2"
 fi
+
 ENCODED_FILE="${INPUT_FILE%.*}.bin"
 DECODED_FILE="${INPUT_FILE%.*}-decoded${INPUT_FILE##*.}"
+PERTURBED_FILE="${INPUT_FILE%.*}_perturbed${INPUT_FILE##*.}"
```
dannywillems, line 17: Nit: unify the usage of "-" and "_".

saffron/e2e-test.sh
```diff
@@ -61,16 +85,28 @@ if ! cargo run --release --bin saffron decode -i "$ENCODED_FILE" -o "$DECODED_FI
...
-    echo "Keeping temporary files for inspection"
-    diff "$INPUT_FILE" "$DECODED_FILE"
+# Compare file to original
+compare_files "$INPUT_FILE" "$DECODED_FILE"
+
+# Usage example in your script:
+perturb_bytes "$INPUT_FILE" "$PERTURBED_FILE" 0.1
+
+echo "Calculating diff for upated $INPUT_FILE (stored updated data in $PERTURBED_FILE)"
+cargo run --release --bin saffron calculate-diff --old "$INPUT_FILE" --new "$PERTURBED_FILE" -o "$ENCODED_DIFF_FILE" $SRS_ARG
```
dannywillems, line 95:
```suggestion
cargo run --release --bin saffron calculate-diff --old "$INPUT_FILE" --new "$PERTURBED_FILE" -o "$ENCODED_DIFF_FILE" "$SRS_ARG"
```

(and the others below, to make shellcheck happy)

saffron/src/blob.rs
```diff
@@ -110,11 +114,51 @@ impl<G: KimchiCurve> FieldBlob<G> {
...
+            .evaluation_diffs
+            .into_par_iter()
+            .enumerate()
+            .map(|(index, d)| {
+                let d_p = {
+                    let evals = (0..domain.size())
+                        .map(|i| {
+                            d.get(&i)
+                                .copied()
+                                .unwrap_or(<G as AffineRepr>::ScalarField::zero())
```
dannywillems, line 135:
```suggestion
                                .unwrap_or(G::ScalarField::zero())
```
And you can remove `7 | use ark_ec::AffineRepr;`

saffron/src/utils.rs
```diff
@@ -179,6 +183,41 @@ impl QueryBytes {
     }
 }
 
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(bound = "F: CanonicalDeserialize + CanonicalSerialize")]
+pub struct Diff<F> {
+    #[serde_as(as = "Vec<HashMap<_, o1_utils::serialization::SerdeAs>>")]
+    pub evaluation_diffs: Vec<HashMap<usize, F>>,
```
dannywillems, line 191: Note that a vector is more efficient, if you don't need to access by the modified index.

saffron/src/utils.rs
```diff
@@ -179,6 +183,41 @@ impl QueryBytes {
     }
 }
 
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(bound = "F: CanonicalDeserialize + CanonicalSerialize")]
+pub struct Diff<F> {
```
dannywillems, line 189: I guess we can restrict it to `PrimeField` directly.

saffron/src/utils.rs
```diff
@@ -179,6 +183,41 @@ impl QueryBytes {
...
+    new: &[u8],
+) -> Diff<F> {
+    let old_elems: Vec<Vec<F>> = encode_for_domain(domain, old);
+    let new_elems: Vec<Vec<F>> = encode_for_domain(domain, new);
+    Diff {
+        evaluation_diffs: new_elems
+            .par_iter()
+            .zip(old_elems)
+            .map(|(n, o)| {
+                n.iter()
```
dannywillems, line 206: If you want to optimize further by keeping the hashmap, you can also parallelize here.

saffron/src/blob.rs
```diff
@@ -161,4 +206,47 @@ mod tests {
             prop_assert_eq!(user_commitments, blob.commitments);
           }
         }
+
+    fn random_perturbation(threshold: f64, data: &[u8]) -> Vec<u8> {
+        let mut rng = rand::thread_rng();
+        data.iter()
```
dannywillems, line 212: `par_iter` also here I guess.

saffron/src/blob.rs
```diff
@@ -161,4 +206,47 @@ mod tests {
...
+
+    proptest! {
+    #![proptest_config(ProptestConfig::with_cases(10))]
+    #[test]
+        fn test_update((threshold, UserData(xs)) in (0.0..1.0f64).prop_flat_map(|t| {
+           UserData::arbitrary().prop_map(move |d| (t, d))
+        }))
+        {
+            // start with some random user data
+            let mut xs_blob = FieldBlob::<Vesta>::encode::<_, DefaultFqSponge<VestaParameters, PlonkSpongeConstantsKimchi>>(&*SRS, *DOMAIN, &xs);
```
dannywillems, line 232: I guess you can also add a type alias for ` DefaultFqSponge<VestaParameters, PlonkSpongeConstantsKimchi>` at the top to get this cleaner.

saffron/src/main.rs
```diff
@@ -140,6 +140,56 @@ pub fn verify_storage_proof(args: cli::VerifyStorageProofArgs) -> Result<()> {
...
+    }?;
+    let diff: Diff<Fp> = {
+        let file = File::open(args.diff_file)?;
+        rmp_serde::decode::from_read(file)
+    }?;
+    blob.update::<DefaultFqSponge<VestaParameters, PlonkSpongeConstantsKimchi>>(
+        &srs, &domain, diff,
+    );
+    args.assert_commitment
+        .into_iter()
```
dannywillems, line 177: Nit: `into_par_iter`

saffron/e2e-test.sh
```diff
@@ -11,8 +11,34 @@ SRS_ARG=""
 if [ $# -eq 2 ]; then
    SRS_ARG="--srs-filepath $2"
 fi
+
 ENCODED_FILE="${INPUT_FILE%.*}.bin"
 DECODED_FILE="${INPUT_FILE%.*}-decoded${INPUT_FILE##*.}"
```
dannywillems, line 16: Missing the dot also.
```suggestion
DECODED_FILE="${INPUT_FILE%.*}-decoded.${INPUT_FILE##*.}"
```

dannywillems, line 16: Also, this doesn't handle the missing dot. And what about file without dot?

saffron/src/main.rs
dannywillems, line 177: The type is `Option<T>`. I won't use a `into_ter()`, even if it works.

saffron/src/blob.rs
```diff
@@ -110,11 +114,51 @@ impl<G: KimchiCurve> FieldBlob<G> {
...
     }
+
+    #[instrument(skip_all, level = "debug")]
+    pub fn update<EFqSponge: Clone + FqSponge<G::BaseField, G, G::ScalarField>>(
+        &mut self,
+        srs: &SRS<G>,
+        domain: &Radix2EvaluationDomain<G::ScalarField>,
+        diff: Diff<G::ScalarField>,
+    ) {
+        let updates: Vec<(usize, PolyComm<G>, DensePolynomial<G::ScalarField>)> = diff
```
dannywillems, line 125: That's not what we want to do, it is not efficient. You only want to use the Lagrange polynomials.

dannywillems, line 125: Let's add also some timers, activated while in debug more. We want to know how fast it is.

dannywillems (changes requested):
See my comments.

saffron/src/utils.rs
martyall: I'm assuming that updates can be sparse enough that using a full vector here would be overkill?

saffron/src/utils.rs
martyall: I'm not a `rayon` pro, but I have a bad feeling about using par_iter with something that can be _really_ long (like thousands of elements) where the work is trivial (in this case subtracting field elems. On top of that, multiple nested `par_iter`.

I don't have a problem using `par_iter` on the chunks of size `domain_size` since even a `500MB` blob is like 260 chunks (i.e. polynomials) and the work per chunk can be substantial. 


If you know a reason I shouldn't worry about this, can you post a doc link? otherwise I would test it / time it before agreeing

saffron/src/blob.rs
martyall: see [this comment](https://github.com/o1-labs/proof-systems/pull/3004#discussion_r1945469219)

saffron/src/blob.rs
martyall: Yeah this is what I figured as well, but I saw while working on this that it isn't part of the `SRS` (just the commitments). I didn't see another type in the codebase that manages this cache, am i missing something? 

If we need to make something, this updat needs the lagrange polynomials in the monomial basis since this is what the data is stored in. If it's an expensive one time set up (seems like it to me), can I suggest leaving a TODO and a follow up PR which:

- creates a type that manages the cache of lagrange polynomials for a given domain
- provied (de)serialize instances
- add a cli arg to load the cache from a file, apply it where necessary 
- save the cache in CI.

martyall:
@dannywillems 

Thanks for the thorough review! Your comments made me realize I need to do a little work RE managing commitments on the user and server side. I will submit that work as a separate PR then come back to this one, but in the meantime can you address questions about `rayon` and the lagrange polynomials.

saffron/e2e-test.sh
martyall: wow, no it wasn't. I have replaced it with a python script instead because perl is like hieroglyphics to me. But actually this makes me realize that the user will should keep all the commitments on their end, not just the folded commitment with powers of `alpha`. As of right now, they cannot update them without recomputing which is dumb.

saffron/e2e-test.sh
```diff
@@ -11,8 +11,34 @@ SRS_ARG=""
...
+PERTURBED_FILE="${INPUT_FILE%.*}_perturbed${INPUT_FILE##*.}"
+ENCODED_DIFF_FILE="${ENCODED_FILE%.*}_diff.bin"
+DECODED_PERTURBED_FILE="${PERTURBED_FILE}-decoded${INPUT_FILE##*.}"
```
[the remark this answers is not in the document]
martyall: ok all the file names should be fixed now

saffron/src/main.rs
martyall: Yeah this is dumb, it's a haskell pattern but I forgot about `if let` in rust

martyall:
@dannywillems Proposing to do this first https://github.com/o1-labs/proof-systems/pull/3005

saffron/src/utils.rs
dannywillems: You can use a vector of type Vec<(idx, F)>, or a linear memory layout where idx and F are inlined (tuples are not, I think).

martyall:
Co

saffron/src/blob.rs
dannywillems: We don't want to store the polynomials in the monomial basis, this has to be fixed as said in DM. You keep the evaluations. From there, you simply commit using the SRS on the diff.

saffron/src/utils.rs
dannywillems: I would simply benchmark. I did this in this branch: https://github.com/o1-labs/proof-systems/commit/3e583bb1b467505928afaef0ad84caaa2bc9bfdb.

You can run using: `cargo run --bin benchmark_diff --release -p saffron -- 22` for polynomials of size 2^22.
You get the following results:
```
> cargo run --bin benchmark_diff --release -p saffron -- 20
Time for difference: 16.75714ms
Time for difference in parallel: 5.640703ms
```

```
> cargo run --bin benchmark_diff --release -p saffron -- 16
Time for difference: 1.151862ms
Time for difference in parallel: 2.718978ms
```

```
> cargo run --bin benchmark_diff --release -p saffron -- 18
Time for difference: 4.807731ms
Time for difference in parallel: 2.612072ms
```

```
> cargo run --bin benchmark_diff --release -p saffron -- 17
Time for difference: 2.232708ms
Time for difference in parallel: 2.420316ms
```

This shows that for a domain size of 16 or 17, it is less efficient.
However, the difference is not negligible from 2^18.

dannywillems: I would skip the inner parallelization for now as also, it might depend on how dense the polynomials/states are.

saffron/src/blob.rs
dannywillems: See [this comment](https://github.com/o1-labs/proof-systems/pull/3004#discussion_r1948110106)

PR #3005, opened 2025-02-07
[saffron] add commitment type shared between user and provider

martyall:
## Summary
The user and storage provider need to communicate about a common commitment type. When working on #3004 I realized that the user will also need to persist these commitments in order to apply updates to them when updating their data. It makes #3004 cleaner if I do this first

## Changes
- Add `Commitment` type ([5b99888](https://github.com/o1-labs/proof-systems/pull/3005/commits/5b99888600060c9734f7f4a988bb1b932f84c8c1)) and use it in the blob ([3940789](https://github.com/o1-labs/proof-systems/pull/3005/commits/394078986cf203f2a308d4a2828df4d81bab4339))
- Add a `-o [output-file]` arg to the `compute-commitment` CLI command to write the `Commitment` value to disk https://github.com/o1-labs/proof-systems/pull/3005/commits/9bb1b56bcaf812f6981fdb5b0fa03eda6e8126ec
- a few trivial cleanups
   - use common type alias for sponge in tests and cli main https://github.com/o1-labs/proof-systems/pull/3005/commits/0ab9cc08c807563e6128f06808cac8ebc29f46b6
   - selective exports from the `commitment` module https://github.com/o1-labs/proof-systems/pull/3005/commits/927ff5bdab36b44c953dcde27535ea614301cae9
   - apply [suggestion](https://github.com/o1-labs/proof-systems/pull/3004#pullrequestreview-2598582602) from #3004 https://github.com/o1-labs/proof-systems/pull/3005/commits/505b3b7dae2484f1a1d41ab2c17010ee7a8e4203

### DIFF
--- a/saffron/e2e-test.sh
+++ b/saffron/e2e-test.sh
@@ -11,6 +11,7 @@ SRS_ARG=""
 if [ $# -eq 2 ]; then
    SRS_ARG="--srs-filepath $2"
 fi
+COMMITMENT_FILE="${INPUT_FILE%.*}_commitment.bin"
 ENCODED_FILE="${INPUT_FILE%.*}.bin"
 DECODED_FILE="${INPUT_FILE%.*}-decoded${INPUT_FILE##*.}"
 
@@ -21,9 +22,7 @@ if [ ! -f "$INPUT_FILE" ]; then
 fi
 
 # Compute commitment and capture last line
-COMMITMENT=$(cargo run --release --bin saffron compute-commitment -i "$INPUT_FILE" $SRS_ARG | tee /dev/stderr | tail -n 1)
-
-
+COMMITMENT=$(cargo run --release --bin saffron compute-commitment -i "$INPUT_FILE" -o "$COMMITMENT_FILE" $SRS_ARG | tee /dev/stderr | tail -n 1)
 
 # Run encode with captured commitment
 echo "Encoding $INPUT_FILE to $ENCODED_FILE"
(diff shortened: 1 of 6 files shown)


saffron/e2e-test.sh
```diff
@@ -11,6 +11,7 @@ SRS_ARG=""
 if [ $# -eq 2 ]; then
    SRS_ARG="--srs-filepath $2"
 fi
+COMMITMENT_FILE="${INPUT_FILE%.*}_commitment.bin"
```
dannywillems, line 14: Same comment that [here](https://github.com/o1-labs/proof-systems/pull/3004#discussion_r1944461370).

saffron/e2e-test.sh
```diff
@@ -11,6 +11,7 @@ SRS_ARG=""
...
 ENCODED_FILE="${INPUT_FILE%.*}.bin"
 DECODED_FILE="${INPUT_FILE%.*}-decoded${INPUT_FILE##*.}"
```
dannywillems, line 16: Same comment than [here](https://github.com/o1-labs/proof-systems/pull/3004#discussion_r1944520834).

saffron/e2e-test.sh
```diff
@@ -21,9 +22,7 @@ if [ ! -f "$INPUT_FILE" ]; then
 fi
 
 # Compute commitment and capture last line
-COMMITMENT=$(cargo run --release --bin saffron compute-commitment -i "$INPUT_FILE" $SRS_ARG | tee /dev/stderr | tail -n 1)
-
-
+COMMITMENT=$(cargo run --release --bin saffron compute-commitment -i "$INPUT_FILE" -o "$COMMITMENT_FILE" $SRS_ARG | tee /dev/stderr | tail -n 1)
```
dannywillems, line 25: Why `tee /dev/stderr`?

martyall: This is so that we can capture the final output of the command in the `COMMITMENT` variable as well as display the logs that it generates in the console while it is running

dannywillems: Ah, I see. I thought you wanted to discard the logs and only get the commitment value.

PR #3006, opened 2025-02-08
[saffron] Add Diff type , methods, and tests

martyall:
This PR builds on #3005 

While working on #3004 I did not approach the `Diff` type carefully enough. This PR is meant to slow things down and make sure we have the spec fulfilled before implementing the update functionality that uses it. 

- clearly define the `Diff` type and it's create method using [sparse representation](https://github.com/o1-labs/proof-systems/pull/3004#discussion_r1946443619) suggested by @dannywillems.
- clearly define `Diff` spec. The assumptions are:
  1. You are allowed to construct a diff if the number of polynomials (chunks) required is less than or equal to the original: `test_allow_legal_updates`
  2. You are not allowed to construct a diff if the new number of chunks is strictly more -- in this case we should throw an error: `test_cannot_construct_bad_diff`
- add property based tests to prove the `Diff` spec

The storage provider will also have to enforce these invariants

### DIFF
--- a/saffron/src/diff.rs
+++ b/saffron/src/diff.rs
@@ -56,7 +56,7 @@ impl<F: PrimeField> Diff<F> {
     }
 
     #[instrument(skip_all, level = "debug")]
-    pub fn as_evaulations(
+    pub fn as_evaluations(
         &self,
         domain: &Radix2EvaluationDomain<F>,
     ) -> Vec<Evaluations<F, Radix2EvaluationDomain<F>>> {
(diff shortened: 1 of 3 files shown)


saffron/src/diff.rs
```diff
@@ -0,0 +1,176 @@
...
+                        .map(|(index, (a, b))| (index, *a - b))
+                        .filter(|(_, x)| !x.is_zero())
+                        .collect()
+                })
+                .collect(),
+        })
+    }
+
+    #[instrument(skip_all, level = "debug")]
+    pub fn as_evaulations(
```
dannywillems, line 59: Typo.

martyall:
@dannywillems https://github.com/o1-labs/proof-systems/pull/3006/commits/297bf5b0a446ed6b5702749d68dd74bf651334b3 was added since original review

dannywillems (approved):

PR #3008, opened 2025-02-10
Marc/lookup prover

marcbeunardeau88:
adds a prover for a delayed lookup argument
Proves that we can accumulate some lookup, so that the sum goes from a public input acc_init to a public output acc_final.
A lot of unecessarry cloning is happening, which will be handle in a follow-up.
A lot of vec are used instead of array, which will be handle in a foloow-up.


o1vm/src/pickles/lookup_prover.rs
```diff
@@ -0,0 +1,226 @@
...
+    // init at acc_init
+    let mut partial_sum = acc_init;
+    let mut acc = vec![];
+    for inner in inverses.iter_mut() {
+        for x in inner.iter_mut() {
+            partial_sum += *x;
+            acc.push(partial_sum)
+        }
+    }
+    let acc_final = acc[acc.len()];
```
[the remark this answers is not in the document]
dannywillems: @marcbeunardeau88 mentions it is fixed later. Opening a follow-up issue in case, to keep track of it.

o1vm/src/pickles/lookup_prover.rs
```diff
@@ -0,0 +1,226 @@
+use ark_ff::{One, PrimeField, Zero};
+use ark_poly::{univariate::DensePolynomial, Evaluations, Polynomial, Radix2EvaluationDomain};
+use kimchi::{circuits::domains::EvaluationDomains, curve::KimchiCurve, plonk_sponge::FrSponge};
+use mina_poseidon::FqSponge;
+use o1_utils::ExtendedDensePolynomial;
+use poly_commitment::{commitment::absorb_commitment, ipa::SRS, OpenProof, SRS as _};
+//TODO Parralelize
```
dannywillems, line 7: This is done in a follow-up. However, it has been built on top of other PRs, and we cannot simply cherry-pick the changes as it creates conflicts. For the sake of simplicity and to move on, we will accept and fix it later.

o1vm/src/pickles/lookup_prover.rs
```diff
@@ -0,0 +1,226 @@
...
+    // Interpolating
+    let interpolate_col = |evals: Vec<G::ScalarField>| {
+        Evaluations::<G::ScalarField, Radix2EvaluationDomain<G::ScalarField>>::from_vec_and_domain(
+            evals, domain.d1,
+        )
+        .interpolate()
+    };
+    let columns_poly = columns.my_map(interpolate_col);
+    // Commiting
+    // TODO avoid cloning
```
dannywillems, line 105: This will be done later. The idea is to use an iterator with the lifetime.

o1vm/src/pickles/lookup_prover.rs
```diff
@@ -0,0 +1,226 @@
...
+    mut fq_sponge: EFqSponge,
+    constraint: &ELookup<G::ScalarField>,
+    rng: &mut RNG,
+) -> (Proof<G>, G::ScalarField)
+where
+    G::BaseField: PrimeField,
+    RNG: RngCore + CryptoRng,
+{
+    // TODO check that
+    let num_chunk = 8;
```
dannywillems, line 38: Opening follow-up https://github.com/o1-labs/proof-systems/issues/3189

dannywillems (approved):

PR #3009, opened 2025-02-10
[saffron] Add update methods with tests

martyall:
- use `chunks` as field name for consistency https://github.com/o1-labs/proof-systems/pull/3009/commits/b83f4e2e7f0215531232994922ed9f15bc9a3068
- add update method for `Commitment` type https://github.com/o1-labs/proof-systems/pull/3009/commits/40851bd8cac1449641b988e7580ba76038f79116
- add update method for `FieldBlob` type with tests https://github.com/o1-labs/proof-systems/pull/3009/commits/dca3c7fb2a212e1c8161b8ee1b2dad3ee56da88a

We discussed moving away from using a CLI to drive the actions e2e test and instead using a client/server implementation with flows written in rust. Instead of implementing the e2e for this flow (which is complicated), I will leave it at the prop tests

### DIFF
--- a/saffron/src/blob.rs
+++ b/saffron/src/blob.rs
@@ -1,9 +1,12 @@
 use crate::{
     commitment::Commitment,
+    diff::Diff,
     utils::{decode_into, encode_for_domain},
 };
 use ark_ff::PrimeField;
-use ark_poly::{univariate::DensePolynomial, EvaluationDomain, Evaluations};
+use ark_poly::{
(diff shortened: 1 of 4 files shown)


dannywillems (approved):

dannywillems:
LGTM. Let's keep in mind:
- get rid of the coefficients form.
- add benchmarks for each step.

Let's focus on the read proofs.

PR #3012, opened 2025-02-12
Split out query module

martyall:
This is a trivial code shuffle PR in preparation of #3011. The point of this is to reduce the noise inside a larger, more important PR

- move `query` out of utils into its own module
- create a separate `proof` dir/module (now that we have storage proofs _and_ read proofs)


martyall:
closed in favor of #3014

PR #3014, opened 2025-02-12
[saffron] separate query module

martyall:
This is a mostly a trivial "code shuffle" PR in preparation for read proofs. The diff numbers look big, but nothing is really happening here.

### Changes
- move query out of utils into its own module
- Introduce more query types (e.g. `IndexQuery` and `QueryResult`). Separate out query `apply` from `decode`
- add query methods to blob type

### DIFF
--- a/saffron/src/query.rs
+++ b/saffron/src/query.rs
@@ -81,16 +81,14 @@ impl<F: PrimeField> QueryField<F> {
         QueryResult { chunks }
     }
 
-    pub fn result_decoder(self) -> impl Fn(&QueryResult<F>) -> Vec<u8> {
-        move |res: &QueryResult<F>| -> Vec<u8> {
-            let elems: Vec<F> = res
-                .chunks
-                .iter()
-                .flat_map(|x| x.iter().map(|x| x.1).collect::<Vec<_>>())
-                .collect();
-            let answer = decode_from_field_elements(&elems);
-            answer[(self.leftover_start)..(answer.len() - self.leftover_end)].to_vec()
-        }
+    pub fn result_decoder(self, res: &QueryResult<F>) -> Vec<u8> {
+        let elems: Vec<F> = res
+            .chunks
+            .iter()
+            .flat_map(|x| x.iter().map(|x| x.1).collect::<Vec<_>>())
+            .collect();
+        let answer = decode_from_field_elements(&elems);
+        answer[(self.leftover_start)..(answer.len() - self.leftover_end)].to_vec()
     }
 
     pub fn as_indices(&self) -> IndexQuery {
@@ -219,11 +217,10 @@ mod tests {
             let chunked = encode_for_domain(&*DOMAIN, &xs);
             for query in queries {
                 let expected = &xs[query.start..(query.start+query.len)];
-                let got_answer = {
+                let got_answer: Vec<u8> = {
                     let field_query: QueryField<Fp> = query.into_query_field(DOMAIN.size(), chunked.len()).unwrap();
-                    let res = field_query.apply(&chunked);
-                    let decode = field_query.result_decoder();
-                    decode(&res)
+                    let res: QueryResult<Fp> = field_query.apply(&chunked);
+                    field_query.result_decoder(&res)
                 };
                 prop_assert_eq!(expected, got_answer);
             }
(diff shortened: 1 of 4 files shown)


saffron/src/query.rs
```diff
@@ -0,0 +1,292 @@
...
+            .collect()
+    }
+}
+
+impl<F: PrimeField> QueryField<F> {
+    fn n_polys(&self) -> usize {
+        self.start.n_polys
+    }
+
+    pub fn apply(&self, data: &[Vec<F>]) -> QueryResult<F> {
```
martyall, line 72: This is the only substantive change in this PR. Now instead of parsing the result (e.g using the `leftover_start` and `leftover_end` fields), we delay parsing and simply return the list of field elems. This is because when constructing a read proof, we want to operate on field elems still, not on bytes.

saffron/src/query.rs
```diff
@@ -0,0 +1,292 @@
...
+#[derive(Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Debug)]
+/// We store the data in a vector of vector of field element
+/// The inner vector represent polynomials
+pub struct FieldElt {
+    /// the index of the polynomial the data point is attached too
+    pub poly_index: usize,
+    /// the index of the root of unity the data point is attached too
+    pub eval_index: usize,
+    domain_size: usize,
+    n_polys: usize,
```
dannywillems, line 25: I don't understand why we have this field here, if `FieldElt` represents a single field element. Same for domain_size.

saffron/src/query.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+/// Represents a query in term of Field element
+#[derive(Debug)]
+pub struct QueryField<F> {
+    pub start: FieldElt,
+    /// how many bytes we need to trim from the first chunk
+    /// we get from the first field element we decode
+    pub leftover_start: usize,
```
dannywillems, line 33: I guess this is going to be removed? It seems over-complicated, as we are supposed to talk in terms of field elements only, no?

martyall: If your looking to read a slice of bytes from encoded data, the slice might start in the middle of a field element. This is used to trim the excess bytes in that case.

This is something that doesn't affect proving at all, which is only communicated in terms of field elems. It's used by the user to parse the returned values, they can construct these limit values themselves without interacting with anyone else.

saffron/src/query.rs
martyall: I used these in order to construct an `Iterator` instance which drastically simplified how we interact with this module. You need to know when to stop a given chunk, or when to stop iterating.

saffron/src/query.rs
```diff
@@ -0,0 +1,292 @@
...
+            .into_iter()
+            .take_while(|x| x <= &self.end)
+            .for_each(|x| {
+                let value = data[x.poly_index][x.eval_index];
+                chunks[x.poly_index].push((x.eval_index, value));
+            });
+        QueryResult { chunks }
+    }
+
+    pub fn result_decoder(self) -> impl Fn(&QueryResult<F>) -> Vec<u8> {
```
dannywillems, line 84: Why this signature?

dannywillems: See https://github.com/o1-labs/proof-systems/pull/3017

martyall: @dannywillems Because all I ever wanted since I was a child was to become a functional programmer :upside_down_face:

saffron/src/query.rs
Fizzixnerd: Given that the demo is supposed to show the use of saffron on bytes, it seems like an important thing to include.  Do you disagree, @dannywillems ?  I will resolve if you're okay with it.

svv232: I think the demo is supposed to use field elements, in reference to https://github.com/o1-labs/proof-systems/tree/dw/demo-different-encodings-saffron/saffron/src/encoding

Fizzixnerd: Sorry, I am a little unclear on what you mean, @svv232.  Indeed, the saffron demo uses field elements, but user's data (for example, a lorem ipsum file) will not be preencoded to field elements.  Performing reads on the underlying byte data in the dense representation (the only representation currently implemented) would require this logic.  Even if we expect to use a sparse representation for some demos (which seems unlikely tbh), this would still be needed for the dense representation.

Fizzixnerd: For example, if I wanted to query bytes 4-8, I would need to perform this kind of logic.

PR #3020, opened 2025-02-12
Arrabbiata/witness: add field to accumulate challenges

dannywillems:
And keep also previous challenges that will be given as public input.


arrabbiata/src/witness.rs
```diff
@@ -142,6 +142,22 @@ pub struct Env<
     pub challenges: Vec<BigInt>,
 
     /// Keep the current executed instruction
```
dannywillems, line 144: Fixing this now.

marcbeunardeau88 (approved):

PR #3022, opened 2025-02-12
Arrabbiata: define Challenges structure

dannywillems:
The goal is to be more explicit on the challenges the protocol defines. It will be easier in the code to refer to a particular name instead of an index. Also, we fix the number of challenges, hoping the structure will be linear in memory and we can enjoy the optimsiation of the CPU cache.

### DIFF
--- a/arrabbiata/src/columns.rs
+++ b/arrabbiata/src/columns.rs
@@ -60,31 +60,43 @@ impl From<Column> for usize {
 }
 
 pub struct Challenges<F: Field> {
-    /// Challenge used to aggregate the constraints
+    /// Used to aggregate the constraints describing the relation. It is used to
+    /// enforce all constraints are satisfied at the same time.
     pub alpha: F,
 
-    /// Both challenges used in the permutation argument
+    /// Both challenges used in the permutation argument.
     pub beta: F,
     pub gamma: F,
 
-    /// Challenge to homogenize the constraints
-    pub homogenous_challenge: F,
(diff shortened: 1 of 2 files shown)


arrabbiata/src/columns.rs
```diff
@@ -59,42 +60,89 @@ impl From<Column> for usize {
...
-    pub homogenous_challenge: F,
+    /// Used to homogenize the constraints and allow the protocol to fold two
+    /// instances of the same relation into a new one.
+    /// Often noted `u` in the paper mentioning "folding protocols".
+    pub constraint_homogeniser: F,
+
+    /// Used by the accumulation protocol (folding) to perform a random linear
+    /// transformation of the witnesses and the public values.
+    /// Often noted `r` in the paper mentioning "folding protocols".
+    pub relation_randomiser: F,
```
dannywillems, line 81: Note that even though it is a challenge, it must be not accumulated by the accumulation scheme.

arrabbiata/src/columns.rs
```diff
@@ -104,6 +104,18 @@ impl<F> Index<usize> for Challenges<F> {
     }
 }
 
+impl<F: Zero> Default for Challenges<F> {
+    fn default() -> Self {
```
marcbeunardeau88, line 108: NIT, I'd call it dummy, as default, unlike dummy, implies a valid value.
eg I know a test calling dummy cannot test the dummy structure semantic

marcbeunardeau88 (approved):

PR #3023, opened 2025-02-12
[saffron] less punishing ci

martyall:
Simple fix for this https://github.com/o1-labs/proof-systems/actions/runs/13273640255/job/37058646902?pr=2969

The problem comes from the time spent computing commitments for test iterations generating "large" user data. Using the `DataSize::Medium` and reducing to 10 test cases still gives good coverage. `Medium` allows for larger than 1 chunk which is the most important thing here. On my machine this went from `1min` to `11s`, CI went back to the normal `60min` range.

While working on this I realized that while correct, the test was written awkwardly. I reordered the commitment checks to more closely match the lifecycle of the update

### DIFF
--- a/saffron/src/blob.rs
+++ b/saffron/src/blob.rs
@@ -212,32 +212,36 @@ mod tests {
         #[test]
 
         fn test_allow_legal_updates((UserData(xs), UserData(ys)) in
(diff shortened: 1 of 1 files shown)


dannywillems (approved):
Ok for the changes in the user input. I hope it unblocks the CI of other people.

PR #3024, opened 2025-02-12
Arrabbiata: move challenge semantic into its own module

dannywillems:
Simple move, only as the code is larger than initially, and should not be related to the column module.

### DIFF
--- a/arrabbiata/src/challenge.rs
+++ b/arrabbiata/src/challenge.rs
@@ -1,0 +1,114 @@
+use ark_ff::{Field, Zero};
+use core::{
+    fmt::{Display, Formatter, Result},
+    ops::Index,
+};
+use kimchi::circuits::expr::AlphaChallengeTerm;
+use serde::{Deserialize, Serialize};
+use strum::EnumCount;
+use strum_macros::EnumCount as EnumCountMacro;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, EnumCountMacro)]
+pub enum ChallengeTerm {
+    /// Used to aggregate the constraints describing the relation. It is used to
+    /// enforce all constraints are satisfied at the same time.
+    /// Often noted `α`.
+    ConstraintRandomiser,
+    /// Both challenges used in the permutation argument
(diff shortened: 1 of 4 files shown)


arrabbiata/src/challenge.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Often noted `u` in the paper mentioning "folding protocols".
+    pub constraint_homogeniser: F,
+
+    /// Used by the accumulation protocol (folding) to perform a random linear
+    /// transformation of the witnesses and the public values.
+    /// Often noted `r` in the paper mentioning "folding protocols".
+    pub relation_randomiser: F,
+}
+
+impl<F> Index<usize> for Challenges<F> {
```
marcbeunardeau88, line 63: I know it's outside the scope of the PR, but I don't like having Index<usize>, only Index<ChallengeTerm> should be used

arrabbiata/src/challenge.rs
```diff
@@ -0,0 +1,114 @@
...
+                ChallengeTerm::COUNT
+            )
+        }
+    }
+}
+
+impl<F: Zero> Default for Challenges<F> {
+    fn default() -> Self {
+        Self {
+            constraint_randomiser: F::zero(),
```
marcbeunardeau88, line 89: Nit
Also outside of the scope, but the alpha field would better be called contraint_combiner.
The term constraint_randomiser indicates a prover coined random to achieve some zk property
same remark for relation_randomiser

Given that the field are documented it's nonentheless understandable

marcbeunardeau88 (approved):

arrabbiata/src/challenge.rs
dannywillems: I don't understand the comment.

arrabbiata/src/challenge.rs
dannywillems: Agreed. Let me change this.

dannywillems: See https://github.com/o1-labs/proof-systems/pull/3031

PR #3025, opened 2025-02-12
Arrabbiata: use singular everywhere

dannywillems:
Dummy, but it has been bugging me for a long time!


dannywillems:
Everything else is singular, why not them!?!?!?

marcbeunardeau88 (approved):

PR #3027, opened 2025-02-12
Arrabbiata: coining alpha challenge + implement generic coin_challenge

dannywillems:


arrabbiata/src/witness.rs
```diff
@@ -1236,4 +1236,51 @@ where
...
+            let mut sponge = E2::create_new_sponge();
+            self.prover_sponge_state.iter().for_each(|x| {
+                E2::absorb_fq(
+                    &mut sponge,
+                    E2::BaseField::from_biguint(&x.to_biguint().unwrap()).unwrap(),
+                )
+            });
+            let verifier_answer = E2::squeeze_challenge(&mut sponge).to_biguint().into();
+            self.challenges[chal] = verifier_answer;
+            sponge.sponge.poseidon_block_cipher();
```
dannywillems, line 1283: I want to abstract it later.

arrabbiata/src/witness.rs
```diff
@@ -1236,4 +1236,51 @@ where
...
+            let mut sponge = E1::create_new_sponge();
+            self.prover_sponge_state.iter().for_each(|x| {
+                E1::absorb_fq(
+                    &mut sponge,
+                    E1::BaseField::from_biguint(&x.to_biguint().unwrap()).unwrap(),
+                )
+            });
+            let verifier_answer = E1::squeeze_challenge(&mut sponge).to_biguint().into();
+            self.challenges[chal] = verifier_answer;
+            sponge.sponge.poseidon_block_cipher();
```
marcbeunardeau88, line 1272: why do we need to add a pass of block_cipher there ?

arrabbiata/src/witness.rs
marcbeunardeau88, line 1290: same comment as line 1272

arrabbiata/src/witness.rs
```diff
@@ -1281,6 +1288,13 @@ where
...
             sponge.sponge.poseidon_block_cipher();
-        };
+            let state: Vec<BigInt> = sponge
+                .sponge
+                .state
+                .iter()
+                .map(|x| x.to_biguint().into())
+                .collect();
+            self.prover_sponge_state = state.try_into().unwrap();
+        }
```
marcbeunardeau88, line 1298: this can be factorised with the other half of the commit

marcbeunardeau88 (approved):

arrabbiata/src/witness.rs
dannywillems: We load the state in the sponge earlier in the code (the state is always kept in the environment), instead of the Sponge structure itself, to have a more granular control on how to manipulate the sponge state. In particular, I want to always run a permutation (i.e. change the state) when we coin a value, as the code is doing here. The current implementation of the Sponge structure does not do that: it allows squeezing two elements without running a state permutation (as verified in [this pull request](https://github.com/o1-labs/proof-systems/pull/2999)). It is not the behavior I want to have for the circuit. I want to always run the permutation, as it would also make the circuit easier to implement and the gadget easier to reuse.

dannywillems: Also, if we don't run the permutation, the next time we load the sponge state, it will give the same challenges, as we don't keep the information that a previous challenge has been already coined with this state.

PR #3028, opened 2025-02-13
Arrabbiata: accumulate program state

dannywillems:
The reviewer is always encouraged to run:
```
cargo nextest run "test_arrabbiata_binary" --release --nocapture -p arrabbiata
```
that starts the example file. This is also executed in the CI in `tests/cli.rs`.

### DIFF
--- a/arrabbiata/src/witness.rs
+++ b/arrabbiata/src/witness.rs
@@ -1297,4 +1297,59 @@ where
             self.prover_sponge_state = state.try_into().unwrap();
         }
     }
+
+    /// Accumulate the program state (or in other words,
+    /// the witness), by adding the last computed program state into the
+    /// program state accumulator.
+    ///
+    /// This method is supposed to be called after the program state has been
+    /// committed (by calling [self.commit_state]) and absorbed (by calling
+    /// [self.absorb_state]). The "relation randomiser" must also have been
+    /// coined and saved in the environment before, by calling
+    /// [self.coin_challenge].
+    ///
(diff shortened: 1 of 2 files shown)


arrabbiata/src/witness.rs
```diff
@@ -1297,4 +1297,59 @@ where
...
+                .map(|(evals_accumulator, evals_witness)| {
+                    evals_accumulator
+                        .iter()
+                        .zip(evals_witness.iter()) // This iterate over the rows
+                        .map(|(acc, w)| (acc + chal.clone() * w).mod_floor(&modulus))
+                        .collect()
+                })
+                .collect();
+        } else {
+            let modulus: BigInt = E2::ScalarField::modulus_biguint().into();
```
marcbeunardeau88, line 1340: A high level comment applying to the previous PR regarding all the duplication depending on the parity of the iteration (here L1237-1338 is the same as L1341-1352) : it should be easy enough to factorize and reduce the code size by 2, don't know if you were planning on this in a follow-up

marcbeunardeau88 (approved):

arrabbiata/src/witness.rs
dannywillems: Yes, I would prefer to remove the duplication. Until now, I have applied the [KISS methodology](https://en.wikipedia.org/wiki/KISS_principle) in the whole codebase.
I would like to have a more typed version of the full protocol, as mentioned in a few places in the codebase. However, I am worried that it would become unreadable.

Please, disagree and I would enjoy changing many parts.

PR #3034, opened 2025-02-13
MVPoly: speed up cross-terms computation when multiplicands are zero

dannywillems:
Can be useful if we use with selectors for instance, i.e. `q * P(X1, ..., Xn)`.


marcbeunardeau88 (changes requested):
Can we add the test case in the prop based testing, I'd sleep better

marcbeunardeau88 (approved):

PR #3035, opened 2025-02-13
Run Saffron in it's own workflow, restore OG CI for core lib only

martyall:
This _actually_ fixes the problem that #3023 failed to address. 

- Run all saffron unit tests in their own workflow
- Stop running saffron in the general CI workflow

## Post Mortem
The tests themselves weren't even the problem -- there were 2 other major factors:
 - The code coverage job was building everything with instrumentation flags that were causing something like a 100x slowdown :exploding_head: 
 - Since the tests were being run without the SRS cache, they were spending a lot of time computing one (likely once per test module)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,24 +96,24 @@ test-all-with-coverage:
 
 
 nextest: ## Test the project with non-heavy tests and using nextest test runner
-		cargo nextest run --all-features --release --profile ci -E "not test(heavy)" $(BIN_EXTRA_ARGS)
+		cargo nextest run --all-features --release $(CARGO_EXTRA_ARGS) --profile ci -E "not test(heavy)" $(BIN_EXTRA_ARGS)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -96,24 +96,24 @@ test-all-with-coverage:
 
 
 nextest: ## Test the project with non-heavy tests and using nextest test runner
-		cargo nextest run --all-features --release --profile ci -E "not test(heavy)" $(BIN_EXTRA_ARGS)
+		cargo nextest run --all-features --release $(CARGO_EXTRA_ARGS) --profile ci -E "not test(heavy)" $(BIN_EXTRA_ARGS)
```
dannywillems, line 99: LGTM. I checked it was not running the Saffron tests.

dannywillems (approved):

PR #3036, opened 2025-02-13
Marc/lookup verif

marcbeunardeau88:
add the lookup verifier
Still a lot of cloning to be removed in a follow-up
Note that it is untested so probably wrong
Using it in the main will come in a follow-up


o1vm/src/pickles/verifier.rs
```diff
@@ -120,7 +120,7 @@ where
     // -- Preparing for opening proof verification
     let zeta_chal = ScalarChallenge(fq_sponge.challenge());
     let (_, endo_r) = G::endos();
-    let zeta: G::ScalarField = zeta_chal.to_field(endo_r);
```
marcbeunardeau88, line 123: Sry, a nit got mixed up in the commit

dannywillems (approved):

PR #3043, opened 2025-02-17
WIP MVPoly: introduce selectors i.e. concept of circuit with gadgets

dannywillems:


dannywillems:
We won't use this solution for now, as we will fix the circuit for each fold for the first version.

dannywillems:
Closing for now.

Issue #3050, opened 2025-02-19
Arrabbiata: Poseidon gate - use 12 different gadgets instead of one with many public inputs


dannywillems:
The current design of the Poseidon gadget is to have a single set of polynomials for 5 rounds, and the round constants are given as "public inputs". This design does not make sense, and the round constants should simply be constants in the polynomial.
We have something like:
```
q_pos P_pos(X1, ..., X15, cst1, ..., cst1) = q_pos [(X3 + cst1)^5 ...]
``` 

By supposing we have 60 full rounds, I suggest to have the following design:
- We split the unique gadget in 12 gadgets, each having a different selector.
- Each gadget has its own set of 15 round constants encoded in the polynomials describing the gadget (15 because we have 5 rounds, and 3 elements in the permutation state).
- We add one gadget for absorbing inputs.
We end up with 13 gadgets, and the following constraints:

```
q_pos_1 * P_pos1(X1, ..., X15) // compute rounds from 1 to 5
q_pos_2 * P_pos2(X1, ..., X15) // compute rounds from 6 to 10
...
q_pos_12 * P_pos12(X1, ..., X15) // compute rounds from 55 to 60
q_absorb * P_absorb(X1, ..., X15) // Computes X3 = X1 + X2, X6 = X4 + X5, simulating the absorbtion of two values (X5 and X2), that would be public inputs.
```

We would end up with the constraint:

```
q_pos1    * P_pos1(X1, ..., X15)   + \
q_pos2    * P_pos2(X1, ..., X15)   + \
... + \
q_pos12  * P_pos12(X1, ..., X15)   + \
q_absorb * P_absorb(X1, ..., X15)
```

The degree stays the same.

dannywillems:
Another design would have been to have 15 "selectors" (like the `q_c` in vanilla PlonK), and it would be fixed at setup time.
Instead of having 15 selectors that every gadget could use, we go with the solutions that 13 selectors are available.

PR #3053, opened 2025-02-20
Arrabbiata: new poseidon gadget

dannywillems:
Following https://github.com/o1-labs/proof-systems/issues/3050

### DIFF
--- a/arrabbiata/src/column.rs
+++ b/arrabbiata/src/column.rs
@@ -18,12 +18,19 @@ pub enum Gadget {
     // Elliptic curve related gadgets
     EllipticCurveAddition,
     EllipticCurveScaling,
-    /// This gadget implement the Poseidon hash instance described in the
-    /// top-level documentation. This implementation does use the "next row"
-    /// to allow the computation of one additional round per row. In the current
-    /// setup, with [crate::NUMBER_OF_COLUMNS] columns, we can compute 5 full
+    /// The following gadgets implement the Poseidon hash instance described in
+    /// the top-level documentation. In the current setup, with
+    /// [crate::NUMBER_OF_COLUMNS] columns, we can compute 5 full
     /// rounds per row.
     Poseidon,
+    /// We provide a new Poseidon gadget that allows computing 5 rounds, without
+    /// using "public inputs".
+    ///
+    /// We split the Poseidon gadget in 13 sub-gadgets, one for each set of 5
+    /// permutations and one for the absorbtion. The parameteris the starting
+    /// round of Poseidon. It is expected to be a multiple of five.
+    PoseidonPermutation(usize),
+    PoseidonSpongeAbsorb,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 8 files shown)


arrabbiata/src/column.rs
```diff
@@ -18,12 +18,19 @@ pub enum Gadget {
...
     EllipticCurveScaling,
-    /// This gadget implement the Poseidon hash instance described in the
-    /// top-level documentation. This implementation does use the "next row"
-    /// to allow the computation of one additional round per row. In the current
-    /// setup, with [crate::NUMBER_OF_COLUMNS] columns, we can compute 5 full
+    /// The following gadgets implement the Poseidon hash instance described in
+    /// the top-level documentation. In the current setup, with
+    /// [crate::NUMBER_OF_COLUMNS] columns, we can compute 5 full
     /// rounds per row.
     Poseidon,
```
dannywillems, line 25: Note: the old Poseidon gadget will be removed later.
By introducing incrementally the gadget `PoseidonPermutation`, I hope it will ease the reviewer work.

arrabbiata/src/column.rs
```diff
@@ -18,12 +18,19 @@ pub enum Gadget {
...
+    /// We provide a new Poseidon gadget that allows computing 5 rounds, without
+    /// using "public inputs".
+    ///
+    /// We split the Poseidon gadget in 13 sub-gadgets, one for each set of 5
+    /// permutations and one for the absorbtion. The parameteris the starting
```
marcbeunardeau88, line 30: in my mind the term permutation is usually used for the full poseidon, not a round

dannywillems: Yes, good point!
Changing it in a quick follow-up (to avoid breaking the PR train you already approved).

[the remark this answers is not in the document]
dannywillems: Oopsie.

arrabbiata/src/column.rs
```diff
@@ -56,6 +63,19 @@ impl From<Column> for usize {
...
 
+impl From<Gadget> for usize {
+    fn from(val: Gadget) -> usize {
+        match val {
+            Gadget::App => 0,
+            Gadget::EllipticCurveAddition => 1,
+            Gadget::EllipticCurveScaling => 2,
+            Gadget::Poseidon => 3,
+            Gadget::PoseidonSpongeAbsorb => 4,
+            Gadget::PoseidonPermutation(starting_round) => 5 + starting_round / 5,
```
marcbeunardeau88, line 74: I don't like that we could have two different gadget with the same conversion to usize
I know it's not supposed to happens as Gadget::PoseidonPermutation(6) should not be created but it has the same int representation as Gadget::PoseidonPermutation(10) 
In general, as previously commented I'd be happy if we can avoid a conversion to usize

marcbeunardeau88 (approved):

arrabbiata/src/column.rs
dannywillems: See https://github.com/o1-labs/proof-systems/pull/3066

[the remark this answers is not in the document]
dannywillems: See https://github.com/o1-labs/proof-systems/pull/3066

PR #3055, opened 2025-02-24
Arrabbiata: implement gadget PoseidonSpongeAbsorb

dannywillems:
This gadget, with `PoseidonPermutation(usize)`, aims to replace the gadget `Poseidon`, as suggested in #3050.
In a future PR, the gadget `Poseidon` is going to be removed, and the implementation of the methods `fetch_values_to_absorb` and `round_constants` are going to be replaced by the newly added methods.

The gadget is essentially two additions + use of the permutation argument to fetch the values. These two constraints could be provided generically for the whole circuit, but we leave this for future works.

An important note is that the gadget supposes that the permutation argument is implemented, through the use of `load_poseidon_state`, `save_poseidon_state` and `fetch_value_to_absorb_in_sponge`. This will come later.

### DIFF
--- a/arrabbiata/src/witness.rs
+++ b/arrabbiata/src/witness.rs
@@ -557,6 +557,50 @@ where
         }
     }
 
+    unsafe fn fetch_value_to_absorb_in_sponge(&mut self, pos: Self::Position) -> Self::Variable {
+        let (col, curr_or_next) = pos;
+        // Purely arbitrary for now
+        let Column::X(_idx) = col else {
+            panic!("Only private inputs can be accepted to load the values to be absorbed")
+        };
+        assert_eq!(
+            curr_or_next,
+            CurrOrNext::Curr,
+            "Only the current row can be used to load the values to be absorbed"
+        );
+        // FIXME: for now, we only absorb the commitments to the columns
+        let idx = self.idx_values_to_absorb;
+        let res = if idx < 2 * NUMBER_OF_COLUMNS {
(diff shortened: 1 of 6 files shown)


arrabbiata/src/witness.rs
```diff
@@ -557,6 +557,50 @@ where
...
+            CurrOrNext::Curr,
+            "Only the current row can be used to load the values to be absorbed"
+        );
+        // FIXME: for now, we only absorb the commitments to the columns
+        let idx = self.idx_values_to_absorb;
+        let res = if idx < 2 * NUMBER_OF_COLUMNS {
+            let idx_col = idx / 2;
+            debug!("Absorbing the accumulator for the column index {idx_col}. After this, there will still be {} elements to absorb", NUMBER_OF_VALUES_TO_ABSORB_PUBLIC_IO - idx - 1);
+            if self.current_iteration % 2 == 0 {
+                let (pt_x, pt_y) = self.accumulated_committed_state_e2[idx_col]
```
marcbeunardeau88, line 577: I think we can only absorb x
that would be a nice speed up
The reasoning is that we have more than collision resistance, we cannot produce two pre-image which commitments are opposites

marcbeunardeau88 (approved):

arrabbiata/src/witness.rs
dannywillems: Follow-up open [here](https://github.com/o1-labs/proof-systems/issues/3064)

PR #3061, opened 2025-02-25
Arrabbiata: introduce zkApp registry

dannywillems:


dannywillems:
A specific selector must also be available for each application.

arrabbiata/src/setup.rs
```diff
@@ -168,98 +164,68 @@ where
...
-                        k,
-                        polynomials
-                            .into_iter()
-                            .map(|p| Sparse::from_expr(p, Some(NUMBER_OF_COLUMNS)))
-                            .collect(),
-                    )
-                })
-                .collect()
+        let circuit_fp: Vec<Gadget> = {
+            let app_circuit: Vec<Gadget> = zkapp_fp.setup(app_size);
```
dannywillems, line 176: Instead the zkApp defines its own size, and the setup method could:
1. return the actual circuit size.
2. could be changed into a function taking a generic zkApp (`fn setup<C: ArrabbiataCurve, App: ZkApp<C>>(app: &App) -> Vec<Gadget>`).

arrabbiata/src/zkapp_registry/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// An instruction is a more granular part of a gadget. It is applied to a
+    /// single row, and can be parametrized by a value that will be used to
+    /// define some behaviors of the gadget.
+    type Instruction = Instruction;
+
+    /// Provide a dummy witness, used to generate a first non-folded instance.
+    fn dummy_witness(&self, srs_size: usize) -> Vec<Vec<C::ScalarField>>;
+
+    /// Describe the control-flow of the ZkApp.
+    fn fetch_next_instruction(&self, current_instr: Self::Instruction) -> Self::Instruction;
```
dannywillems, line 68: Can we have something that is more "programming" friendly, like loops, etc?
Could we describe something like:
```
for i in (0..n):
  run gadget0 i times
  run gadget1 i^2 times
```

It would provide a more "standard" interface available in "traditional" programming language.
Also, do we want that? Most of the time, the circuits are pretty straightforward where we do X followed by Y followed by Z, etc.

arrabbiata/src/zkapp_registry/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+/// A ZkApp structure is also responsible to provide "up-front" a "setup" method
+/// that describe the shape of the execution trace.
+/// The shape is defined by a list of "selectors" that are used to select the
+/// columns that are used in the computation, and how constrained they are.
+/// The method "run" will be responsible to build the execution trace, using the
+/// selectors the setup phase defined.
+///
+/// The method "run" will use the interpreter environment to build the execution
+/// trace, based on previously computed values.
+pub trait ZkApp<C: ArrabbiataCurve, Gadget, Instruction: Into<Gadget>>
```
dannywillems, line 35: Idea of ZkApp to implement to check if the API is good enough:
- Poseidon2
- Rescue
- Foreign field arithmetic
- signature over pasta curves

For non-crypto applications:
- Mastermind -> note that it might mean we need to have an internal state.

arrabbiata/src/zkapp_registry/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// formed by a set of instructions [EllipticCurveScaling(bit)] where `bit`
+    /// defines the bit of the scalar that is being processed.
+    ///
+    /// All instructions in a gadget are expected to be of the same type, i.e.
+    /// described by the same set of multivariate polynomials.
+    type Gadget = Gadget;
+
+    /// The type of the instructions used in the ZkApp.
+    ///
+    /// An instruction is a more granular part of a gadget. It is applied to a
```
dannywillems, line 59: Explain way better what is the difference between `Instruction` and `Gadget`. It makes sense when we get it, but it is not totally obvious.

arrabbiata/src/zkapp_registry/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+use ark_ff::PrimeField;
+use poly_commitment::commitment::CommitmentCurve;
+
+use crate::{
+    column::Gadget,
+    curve::ArrabbiataCurve,
+    interpreter::{Instruction, InterpreterEnv},
+};
+
+/// A ZkApp for Arrabbiata is a program that is built over a generic interpreter
```
dannywillems, line 11: Do we want to have a state in the ZkApp?

PR #3067, opened 2025-02-25
Arrabbiata: cross terms - the renaissance

dannywillems:


arrabbiata/src/setup.rs
```diff
@@ -261,9 +268,36 @@ where
...
+        // The initial transcript state, i.e. all the messages the verifier and
+        // prover know before any computation, are all the commitments to the
+        // selectors for both circuit.
+        // We create an initial sponge state by instantiating a sponge on both
+        // curves and absorb the corresponding commitments.
+        // After that, we squeeze the state of one sponge to absorb in the
+        // other.
+        // Arbitrarily, we pick the sponge over second curve to output a digest,
+        // and we take the state of the sponge over the first curve.
+        let mut sponge_e1 = E1::create_new_sponge();
```
dannywillems, line 280: Note that we could simply use Blake2b with a concatenated bytes sequence with all the information.

arrabbiata/src/witness.rs
```diff
@@ -252,6 +281,93 @@ where
...
+                    .unwrap();
+                E1::ScalarField::from_biguint(&v).unwrap()
+            };
+            let u_right = {
+                let v: BigUint = challenges[ChallengeTerm::ConstraintHomogeniser]
+                    .to_biguint()
+                    .unwrap();
+                E1::ScalarField::from_biguint(&v).unwrap()
+            };
+            let activated_gadget = gadgets[row];
```
dannywillems, line 336: This is not a good approach. We should rely on selectors. The terms "selector" and "gadget" are cofounded because there is a one-to-one correspondence for now.

Bad design! Thinking about changing it while writing an example app like in #3061

arrabbiata/src/witness.rs
```diff
@@ -252,6 +281,93 @@ where
...
+                    .unwrap();
+                E1::ScalarField::from_biguint(&v).unwrap()
+            };
+            let u_left = {
+                let v: BigUint = self.accumulated_challenges[ChallengeTerm::ConstraintHomogeniser]
+                    .to_biguint()
+                    .unwrap();
+                E1::ScalarField::from_biguint(&v).unwrap()
+            };
+            let u_right = {
```
dannywillems, line 330: This is equal to one, as the right instance is not homogenized.

PR #3068, opened 2025-02-25
CI: use Rust 1.79 everywhere and drop older versions

dannywillems:
Related to https://github.com/MinaProtocol/mina/pull/16652/ is merged.


querolita (approved):

dannywillems:
Merging as now Mina uses Rust 1.79.

PR #3069, opened 2025-02-25
Kimchi/prover: simplify blinders and add missing chunks when none given

dannywillems:
Following-up on https://github.com/o1-labs/proof-systems/pull/2969/commits/a1377e90b8878ade9f44d345b4d666699ed4cb18#r1970593373

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -304,15 +304,8 @@ where
                 .collect(),
             Some(blinders_arr) => blinders_arr
                 .into_iter()
-                .zip(witness.iter())
-                .map(|(blinder_el, witness)| match blinder_el {
-                    None => {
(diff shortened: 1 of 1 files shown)


dannywillems:
> @dannywillems I'm not sure this is correct? It assumes that every single witness column has the same number of chunks, which might not be the case IIRC?
> 
> Or do you think it's not a realistic scenario? (then maybe we should assert it somewhere)

It is true that I mentioned earlier that we should have a different number of chunks, and therefore we did require a different number of chunks. I was wrong :confused:. As for now, we do suppose the same number of chunks are used for each wire, as for other polynomials in the codebase. For instance, [here](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/prover.rs#L1006). `num_chunks` is used everywhere.

volhovm:
OK I agree! I'll integrate this change into https://github.com/o1-labs/proof-systems/pull/2969.

PR #3070, opened 2025-02-26
Kimchi benches: use serialized mina circuits

volhovm:
This adds 235M of serialized circuits, half of randomly selected from the following mina run: `_build/default/src/app/cli/src/mina.exe transaction-snark-profiler --zkapps --k 1 --max-num-updates 2 --min-num-updates 2`


volhovm:
Danny suggests that maybe the circuits can be not part of the github repo but instead stored somewhere else?

Similarly to https://github.com/o1-labs/proof-systems/blob/master/.github/workflows/o1vm-ci.yml#L37

georgeee:
I suggest that we upload the files to GCP

kimchi/benches/proof_criterion_mina.rs
```diff
@@ -0,0 +1,97 @@
...
+            format!("proof creation (mina, vesta, circuit seed {})", seed),
+            |b| {
+                b.iter(|| {
+                    black_box(ProverProof::create_recursive::<
+                        BaseSpongeVesta,
+                        ScalarSpongeVesta,
+                        _,
+                    >(
+                        &group_map,
+                        witness.clone(),
```
Fizzixnerd, line 52: Recommend wrapping these `clone`s in `black_box`es, since the compiler might be able to infer it can skip them.

I also wonder if you meant for these to be cloned inside the benchmark?  It would affect the results.  It might be better to clone them upfront, if that's possible.

Fizzixnerd (approved):
~~non-binding~~ I thought there were other reviewers, but my eyes tricked me.  Consider it binding.

kimchi/benches/proof_criterion_mina.rs
volhovm: You are right that these are part of the benchmark, but tbh the prover takes 3-4 seconds, so 0.1ms of cloning won't affect it much...

volhovm:
Luis also said "lgtm" in DMs :) so I'll merge this now and happy to address any issues that arise.

PR #3073, opened 2025-02-26
Arrabbiata: move selectors into setup

dannywillems:


arrabbiata/src/witness.rs
```diff
@@ -1048,105 +1047,6 @@ where
...
-    /// E = E1 - r T1 - r^2 T2 - ... - r^d T^d + r^(d+1) E2
-    ///   = E1 - r (T1 + r (T2 + ... + r T^(d - 1)) - r E2)
-    /// ```
-    /// where (d + 1) is the degree of the highest gate.
-    ///
-    /// - We compute the next hash we give to the next instance
-    ///
-    /// ```text
-    /// hash' = H(i + 1, acc'_1, ..., acc'_17, z_0, z_(i + 1))
-    /// ```
```
marcbeunardeau88, line 1103: is the suppression of the comment voluntary ?

arrabbiata/src/setup.rs
```diff
@@ -199,6 +207,27 @@ where
         // suppose we have the same circuit on both curves for now.
         let app_size = srs_size - VERIFIER_CIRCUIT_SIZE;
 
+        // Build the selectors for both circuits.
+        // FIXME: we suppose we have the same circuit on both curve for now.
```
marcbeunardeau88, line 216: I understand the app only handle one gadget.
Note that a generalisation will be needed at some point

arrabbiata/src/lib.rs
```diff
@@ -23,8 +23,7 @@ pub mod witness;
 pub const MAX_DEGREE: usize = 5;
 
 /// The minimum SRS size required to use Nova, in base 2.
-/// Requiring at least 2^16 to perform 16bits range checks.
-pub const MIN_SRS_LOG2_SIZE: usize = 16;
+pub const MIN_SRS_LOG2_SIZE: usize = 8;
```
marcbeunardeau88, line 26: shouldn't this be derived from the verification circuit size ?

arrabbiata/src/column.rs
```diff
@@ -17,6 +17,13 @@ pub enum Gadget {
     /// A dummy gadget, doing nothing. Use for padding.
     NoOp,
 
+    /// The gadget defining the app.
+    ///
+    /// For now, the application is considered to be a one-line computation.
+    /// However, we want to see the application as a collection of reusable
```
marcbeunardeau88, line 23: ah ok this addresses my previous comment on commit 25febf98a4853fc9e906a4a0ddab15a166e8994b

arrabbiata/src/lib.rs
```diff
@@ -57,8 +57,8 @@ pub const MAXIMUM_FIELD_SIZE_IN_BITS: u64 = 255;
 /// verifier circuit.
 pub const NUMBER_OF_VALUES_TO_ABSORB_PUBLIC_IO: usize = NUMBER_OF_COLUMNS * 2;
 
-/// The number of selectors used in the circuit.
-pub const NUMBER_OF_SELECTORS: usize =
+/// The number of gadgets supported by the program
```
marcbeunardeau88, line 60: isn't this the max nb of gadgets ?

arrabbiata/src/setup.rs
```diff
@@ -228,6 +233,34 @@ where
             v
         };
 
+        let selectors_comm: (
+            [PolyComm<E1>; NUMBER_OF_GADGETS],
+            [PolyComm<E2>; NUMBER_OF_GADGETS],
+        ) = {
+            // We initialize to the blinder to avoid the identity element.
```
marcbeunardeau88, line 240: why avoid the identity
Also we shouldn't blind the commitment, which are public

arrabbiata/tests/witness.rs
```diff
@@ -287,7 +282,7 @@ fn test_regression_witness_structure_sizeof() {
     // thining about the memory efficiency of the codebase.
```
marcbeunardeau88, line 282:
```suggestion
    // thinking about the memory efficiency of the codebase.
```

marcbeunardeau88 (approved):

marcbeunardeau88:
Note that the last commit is not verified

arrabbiata/src/witness.rs
dannywillems: Yes. It was outdated and a bit wrong. I would prefer rewrite it properly.

arrabbiata/src/lib.rs
dannywillems: Yes. I'll change it after, good point.

arrabbiata/src/lib.rs
dannywillems: Yes.

arrabbiata/src/setup.rs
dannywillems: It is to avoid only zeroes, and commit to the zero polynomial.

dannywillems:
> Note that the last commit is not verified

It seems it is because I run a script in Mina/buildkite that changed my git username/email into `It's me, CI`. I forced-pushed to remove the old author after a `git commit --amend`, but it does not work.

PR #3075, opened 2025-02-27
Arrabbiata: introduce the structure Program

dannywillems:


arrabbiata/src/witness.rs
```diff
@@ -18,6 +18,242 @@ use crate::{
     setup, NUMBER_OF_COLUMNS, NUMBER_OF_VALUES_TO_ABSORB_PUBLIC_IO,
 };
 
+/// A running program that the (folding) interpreter has access to.
+pub struct Program<
+    Fp: PrimeField,
+    Fq: PrimeField,
```
marcbeunardeau88, line 24: as discussed offline, I would remove these fields

arrabbiata/src/witness.rs
```diff
@@ -18,6 +18,242 @@ use crate::{
...
+    E1: ArrabbiataCurve<ScalarField = Fp, BaseField = Fq>,
```
marcbeunardeau88, line 25: NIT: rename to E, E1 must be a copy paste from the preceeding code

arrabbiata/src/witness.rs
```diff
@@ -18,6 +18,242 @@ use crate::{
...
+    Program<Fp, Fq, E1>
+where
+    E1::BaseField: PrimeField,
+    <<E1 as CommitmentCurve>::Params as CurveConfig>::BaseField: PrimeField,
+{
+    pub fn new(srs_size: usize, blinder: E1) -> Self {
+        // Default set to the blinders. Using double to make the EC scaling happy.
+        let previous_committed_state: Vec<PolyComm<E1>> = (0..NUMBER_OF_COLUMNS)
+            .map(|_| PolyComm::new(vec![(blinder + blinder).into()]))
+            .collect();
```
marcbeunardeau88, line 69: I don't understand why we use blinders here, nor why we double them
It can easily be forgotten that we did that and break correctness later

More importantly, can we not implement zk blinding only the final snark ?

arrabbiata/src/witness.rs
```diff
@@ -18,6 +18,242 @@ use crate::{
...
+    pub previous_challenges: Challenges<BigInt>,
+}
+
+impl<Fp: PrimeField, Fq: PrimeField, E1: ArrabbiataCurve<ScalarField = Fp, BaseField = Fq>>
+    Program<Fp, Fq, E1>
+where
+    E1::BaseField: PrimeField,
+    <<E1 as CommitmentCurve>::Params as CurveConfig>::BaseField: PrimeField,
+{
+    pub fn new(srs_size: usize, blinder: E1) -> Self {
```
marcbeunardeau88, line 65: I believe the default value for the previous commited state should be provided by the user (at least for the app part).
We cannot derive generically a satisfying instance

marcbeunardeau88:
I didn't review in details the methods for the program struct, as I assume these were copy pasted from the env method, which have been reviewed

arrabbiata/src/witness.rs
```diff
@@ -18,6 +18,242 @@ use crate::{
...
+    ///               OR
+    /// C_(q, n + 1) = C_(q, n) + chal * comm
+    /// ```
+    ///
+    /// Note that the committed program state is encoded in
+    /// [crate::NUMBER_OF_COLUMNS] values, therefore we must iterate over all
+    /// the columns to accumulate the committed state.
+    pub fn accumulate_committed_state(&mut self, chal: BigInt) {
+        let chal: BigUint = chal.to_biguint().unwrap();
+        let chal: E1::ScalarField = E1::ScalarField::from_biguint(&chal).unwrap();
```
marcbeunardeau88, line 247: these conversions seems uneccessarry 
We should change the arg type, I see no reason that an output of poseidon that we want to use as a scalar goes trough bigint/biguint

marcbeunardeau88 (approved):

arrabbiata/src/witness.rs
```diff
@@ -18,6 +18,242 @@ use crate::{
...
+    /// sponge state, and squeezing a challenge from it, relying on the
+    /// implementation of the sponge. Usually, the challenge would be the first
+    /// N bits of the first element, but it is left as an implementation detail
+    /// of the sponge given by the curve.
+    pub fn coin_challenge(&self, sponge_state: Vec<BigInt>) -> (BigInt, Vec<BigInt>) {
+        let mut sponge = E1::create_new_sponge();
+        sponge_state.iter().for_each(|x| {
+            E1::absorb_fq(
+                &mut sponge,
+                E1::BaseField::from_biguint(&x.to_biguint().unwrap()).unwrap(),
```
marcbeunardeau88, line 170: this conversion seems supicious
I understand it is to go from one field to another ?
But if our challenge are 128 bits we should not need it.

arrabbiata/src/witness.rs
dannywillems: If I change (in `setup.rs`) to:
```
pub struct IndexedRelation<
    E1: ArrabbiataCurve<ScalarField = E2::BaseField, BaseField = E2::ScalarField>,
    E2: ArrabbiataCurve<ScalarField = E1::BaseField, BaseField = E1::ScalarField>,
> where
    E1::BaseField: PrimeField,
    E2::BaseField: PrimeField,
```
I get a circular dependency (which makes sense). The fields seem to be required to avoid the circular dependency.

arrabbiata/src/witness.rs
dannywillems: The blinder is not for zero-knowledge, but to avoid the need of handling the identity element. I do not remember rn why we double. I am opening a follow-up to rewrite it later.

arrabbiata/src/witness.rs
dannywillems: Agreed. I would like to revisit the conversion and the usage of `BigInt`. It might have been a bad initial decision.

arrabbiata/src/witness.rs
dannywillems: It doesn't go from one field to another. It converts into base field elements the state of the sponge that is passed as an argument.
However, I thought that absorbing three times would set the state to the values we passed without running the cipher, but it is wrong:
```
    fn absorb(&mut self, x: &[F]) {
        for x in x.iter() {
            match self.sponge_state {
                SpongeState::Absorbed(n) => {
                    if n == self.rate {
                        self.poseidon_block_cipher();
                        self.sponge_state = SpongeState::Absorbed(1);
                        self.state[0].add_assign(x);
                    } else {
                        self.sponge_state = SpongeState::Absorbed(n + 1);
                        self.state[n].add_assign(x);
                    }
                }
                SpongeState::Squeezed(_n) => {
                    self.state[0].add_assign(x);
                    self.sponge_state = SpongeState::Absorbed(1);
                }
            }
        }
    }
```
it does absorb twice and after that run the cipher.

I'm opening a follow-up (as unrelated to this PR as it is mostly moving code).
I will revisit the sponge state management in a future PR.

arrabbiata/src/witness.rs
dannywillems: > We cannot derive generically a satisfying instance

Agreed. I am facing this issue after https://github.com/o1-labs/proof-systems/pull/3067

PR #3079, opened 2025-03-10
WASM Memory boosting mode with lazy cache

querolita:
Closes https://github.com/o1-labs/o1js/issues/2006 (kimchi-side https://github.com/o1-labs/o1js/issues/2074)

This PR introduces a new enum called `LazyCache` which is meant to provide support for expensive evaluations that can be computed on-demand on the prover, instead of precomputed on the setup to save memory space for WASM. 

~`LazyCache` contains two variants: `Cached` and `Lazy`. By default, these evaluations (selector polynomials, precomputed constants, and lookups) will be cached in memory. But when the flag `lazy_cache = true`, then their computation is deferred from constraint system creation until the prover. This is done by passing an `Arc` closure that will obtain the evaluations when prompted. The API is quite simple, as it just requires to use `.get()` in these structures. This method will simply access the cached values in the `Cached` variant, and if it is a `Lazy` variant it will invoke the closure the first time, and then the result will be available in the `computed` field which is a `OnceCell`. When this is done, the closure is taken from memory to decrease the counter of the `Arc` and free the `Mutex`.~ This is old

`LazyCache` mimics the logic of `LazyLock` with a `FnOnce` that is lazily evaluated and then values are stored in memory. The API is quite simple as it just requires to use `get()` to either get the cached value or evaluate it for the first time.

The main change is in `kimchi/src/circuits/lazy_cache.rs`. The rest of the diff is about adapting the constraint system, prover index, and prover code to the new type.

This PR also includes an optional mode for `TestFramework` to log heap allocations across execution steps. The unit tests provided show the functioning of the `LazyCache` struct, and gives some benchmarks of the `lazy_mode` flag in use inside a circuit. For example, in a ~65k gate circuit, we are obtaining the following measurements:

| lazy_mode | false   | true   |
|--------------|---------|--------|
| Index setup | 1375 MB | 213 MB |
| Proving |  1319 MB | 1242 MB |
| Verify |  1319 MB  | 1242 MB  |

The fact that the largest difference lies within the setup phase is completely expected, as that was the goal of deferring the precomputation of parameters to the proving phase.

There will be follow-up PRs to address the following issues:
- ~Fix serialization of `LazyCache` so that WASM does not need to call `get()`, invalidating the laziness.~ Serialization should not be a problem for lazy mode propagation, as bindings do not use serialization for the types.
- Update the inner code of `LazyCache` in favor of `LazyLock` (>Rust 1.80).


kimchi/src/tests/foreign_field_add.rs
```diff
@@ -1061,29 +1061,31 @@ fn test_bad_bound() {
...
     let right_input = BigUint::from_bytes_be(&random_input(rng, foreign_mod.clone(), false));
-    let (mut witness, mut index) = test_ffadd(
+    let (mut witness, index) = test_ffadd(
         foreign_mod,
         vec![left_input, right_input],
         &[FFOps::Add],
         true,
     );
+
+    let mut cs = match Arc::try_unwrap(index.cs) {
```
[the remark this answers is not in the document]
querolita: Ah true

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,129 @@
...
+    // TODO: Cached(OnceCell<T>),?
+    Cached(T),
+
+    /// Deferred computation that is evaluated only once when accessed
+    #[serde(skip)]
+    Lazy {
+        // The value once computed
+        computed: OnceCell<T>,
+        // The function to compute the value
+        compute_fn: Mutex<Option<Arc<dyn Fn() -> T + Send + Sync>>>,
```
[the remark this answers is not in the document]
querolita: > compute_fn: F where F: FnOnce() -> T
Oh this makes sense! I was stupidly trying to achieve that behavior manually, but forgot I could use that instead of `Fn` with the ugly `lock()` and `take()` approach inside the `get()`. I will try and reply back in the thread.

kimchi/src/circuits/lazy_cache.rs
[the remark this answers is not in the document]
querolita: Replied in a different comment, it sure is a bit of re-inventing the wheel... Reason is I did not realize I could just use `FnOnce` instead of `Fn` and all that extra dance 🤦🏻

kimchi/src/circuits/constraints.rs
```diff
@@ -271,10 +272,11 @@ impl<F: PrimeField> ConstraintSystem<F> {
...
     /// - `precomputations: None`,
     /// - `disable_gates_checks: false`,
+    /// - `lazy_mode: false`,
     ///
     /// How to use it:
     /// 1. Create your instance of your builder for the constraint system using `crate(gates, sponge params)`
-    /// 2. Iterativelly invoke any desired number of steps: `public(), lookup(), runtime(), precomputations()``
+    /// 2. Iterativelly invoke any desired number of steps: `public(), lookup(), runtime(), precomputations(), lazy_mode()`
     /// 3. Finally call the `build()` method and unwrap the `Result` to obtain your `ConstraintSystem`
     pub fn create(gates: Vec<CircuitGate<F>>) -> Builder<F> {
```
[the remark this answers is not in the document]
querolita: Oh interesting, opening an issue for that right now https://github.com/o1-labs/proof-systems/issues/3086

kimchi/src/circuits/constraints.rs
```diff
@@ -949,19 +949,49 @@ impl<F: PrimeField> Builder<F> {
...
         let endo = F::zero();
 
-        let domain_constant_evaluation = OnceCell::new();
+        let precomputations = if !self.lazy_mode {
+            match self.precomputations {
+                Some(t) => LazyCache::cache(t),
+                None => LazyCache::cache(Arc::new(
+                    DomainConstantEvaluations::create(domain, zk_rows).unwrap(),
+                )),
+            }
```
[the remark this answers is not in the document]
querolita: Not sure if this is your question, but `precomputations` used to be an optional before this PR as well. That's because, if there were any precomputations already available from a previous circuit or for some other reason, we would like to load them instead of recomputing everything. So now in the lazy cache PR, if the flag is false (basically the old behavior) then it just creates cached variables: with either the precomputed available data, or it computes it all there if none.

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+{
+    // Create a new `LazyCache` with a cached computation
+    pub fn cache(value: T) -> Self {
+        let cell = OnceCell::new();
+        let _ = cell.set(value);
+        LazyCache::Cached(cell)
+    }
+
+    // Create a new `LazyCache` with a deferred computation
+    pub fn lazy(compute_fn: impl Fn() -> T + Send + Sync + 'static) -> Self {
```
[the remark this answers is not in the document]
querolita: Oh that's a nice catch. I didn't know there were significant differences between both approaches and thought they were completely interchangeable.

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn get(&self) -> &T {
+        match self {
+            LazyCache::Cached(value) => value.get().unwrap(),
+            LazyCache::Lazy {
+                computed,
+                compute_fn,
+            } => computed.get_or_init(|| {
+                compute_fn
+                    .lock()
+                    .unwrap()
```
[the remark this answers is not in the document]
querolita: Probably this will change when I switch from `Fn` to `FnOnce`, but will keep my eyes open for next time.

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+            LazyCache::Lazy {
+                computed,
+                compute_fn,
+            } => LazyCache::Lazy {
+                computed: computed.clone(),
+                compute_fn: Mutex::new(compute_fn.lock().unwrap().as_ref().map(Arc::clone)),
+            },
+        }
+    }
+}
```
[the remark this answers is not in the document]
querolita: Basically, the o1js developer should be able to turn on/off the lazy mode depending on the complexity of the program they want to write. So if it is too complex and WASM does not have sufficient memory to store the structures needed for the circuit, they set it to true to save some space. Whether or not that will be sufficient in practice (to support features like chunking for example) is still unknown (benchmarks will come when the change is fully integrated).

kimchi/src/circuits/constraints.rs
```diff
@@ -949,19 +949,49 @@ impl<F: PrimeField> Builder<F> {
...
+        } else {
+            // Lookup constraint creation does not fail, we discard the struct
+            // to prevent cacheing in memory and store in the lazy cache the
+            // function needed to recompute it afterwards when needed.
+            let gates = gates.clone();
+            LazyCache::lazy(move || {
+                LookupConstraintSystem::create(
+                    &gates,
+                    lookup_tables.clone(),
+                    runtime_tables.clone(),
```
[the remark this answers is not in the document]
querolita: The problem is that `LazyCache` is being used as part of structs which derive from `Clone`, so `LazyCache` needs to be clonable to compile.

kimchi/src/circuits/constraints.rs
```diff
@@ -804,11 +799,16 @@ impl<F: PrimeField> Builder<F> {
...
+        self
+    }
+
     /// Build the [ConstraintSystem] from a [Builder].
     pub fn build(self) -> Result<ConstraintSystem<F>, SetupError> {
         let mut gates = self.gates;
-        let lookup_tables = self.lookup_tables;
-        let runtime_tables = self.runtime_tables;
+        let lookup_tables = self.lookup_tables.clone();
+        let runtime_tables = self.runtime_tables.clone();
```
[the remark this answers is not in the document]
querolita: The problem is that if I do `self.lookup_tables.clone()` in the `compute_fn` of the lazy branch [here](https://github.com/o1-labs/proof-systems/blob/cc52479dfe81a6a9709a1825408333dbebf5c2c6/kimchi/src/circuits/constraints.rs#L961) then `self.lookup_tables` is moved:
```console
move occurs because `self.lookup_tables` has type `Vec<tables::LookupTable<F>>`, which does not implement the `Copy` trait
```
I couldn't find a way with fewer clones than the one committed 🤔

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+use serde::{ser, Deserialize, Serialize, Serializer};
+use std::{
+    fmt,
+    sync::{Arc, Mutex},
+};
+
+/// A memory-efficient container that caches or computes `T` on demand.
+pub enum LazyCache<T>
+where
+    T: Send + Sync,
```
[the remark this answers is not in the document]
querolita: I guess I can remove the `where` clause from the enum definition

kimchi/src/circuits/constraints.rs
```diff
@@ -286,19 +288,12 @@ impl<F: PrimeField> ConstraintSystem<F> {
             precomputations: None,
             disable_gates_checks: false,
             max_poly_size: None,
+            lazy_mode: false,
         }
     }
 
     pub fn precomputations(&self) -> &Arc<DomainConstantEvaluations<F>> {
```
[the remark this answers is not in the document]
querolita: Actually `get()` returns a pointer to the lazy cache. Now, what you're saying is: since it is an `Arc`, then cloning the result of the `get()` shouldn't be cloning the contents, but just increase the counter of the reference? I guess that makes sense 🤔

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+{
+    fn default() -> Self {
+        LazyCache::Lazy {
+            computed: OnceCell::new(),
+            compute_fn: Mutex::new(None),
+        }
+    }
+}
```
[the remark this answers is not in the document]
querolita: I cannot remember what I needed defaulting for, if I remove it I get an unhelpful error message now 
```console
error[E0277]: the trait bound `LazyCache<std::sync::Arc<DomainConstantEvaluations<F>>>: Default` is not satisfied
   --> kimchi/src/circuits/constraints.rs:203:5
    |
203 |     /// precomputes
    |     ^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `LazyCache<std::sync::Arc<DomainConstantEvaluations<F>>>`
```
I get your point about panicking if getting on that default value... I know I am not doing that because I know where I have explicitly used `get()`, but I get your point. I like the `LazyCache<()>` though. I will keep looking for the reason that I was asked to provide a default impl for it.

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+    {
+        match self {
+            LazyCache::Cached(cell) => cell
+                .get()
+                .ok_or_else(|| serde::ser::Error::custom("LazyCache::Cached is uninitialized"))?
+                .serialize(serializer),
+            LazyCache::Lazy { computed, .. } => computed.get().map_or_else(
+                || Err(ser::Error::custom("LazyCache:Lazy is not computed")),
+                |value| value.serialize(serializer),
+            ),
```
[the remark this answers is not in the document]
querolita: But hey, this hasn't panicked on tests, thus I suppose it is never trying to serialize an empty lazy variant. Thus it should be equivalent to just return the serialization of the value obtained through the `get()` of the lazy cache itself...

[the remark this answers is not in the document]
querolita: I changed it to prevent erring. It now loads the value in both cases as you suggested.

kimchi/src/circuits/constraints.rs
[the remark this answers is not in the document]
querolita: RE: after chatting on https://github.com/o1-labs/proof-systems/pull/3091, I understood we don't want `LazyCache` to be `Clone`, in the same way `LazyLock` is not in 1.80. 

Using `FnOnce` allows me to remove the clone in those two lines, as pointed out by @richardpringle 👍🏻

kimchi/src/circuits/lazy_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+            LazyCache::Cached(value) => value.get().unwrap(),
+            LazyCache::Lazy {
+                computed,
+                compute_fn,
+            } => computed.get_or_init(|| {
+                compute_fn
+                    .lock()
+                    .unwrap()
+                    .take()
+                    .expect("no function inside LazyCache::Lazy")()
```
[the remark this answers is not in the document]
querolita: I can have a `try_get`

kimchi/src/circuits/lazy_cache.rs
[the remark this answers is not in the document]
querolita: it was for skip(serde)

dannywillems:
> I suggested you use LazyLock but realized that it only recently landed in stable and we're a little further behind.

It seems LazyLock is available from 1.80.
I'm currently updating to 1.82 in Mina.

querolita:
`LazyLock` functionality has been replicated in the `LazyCache` within `o1_utils` (this allows implementing traits like Serialize and Deserialize)

kimchi/src/circuits/constraints.rs
```diff
@@ -200,15 +200,78 @@ pub struct ConstraintSystem<F: PrimeField> {
...
     /// precomputes
     #[serde(skip)]
-    precomputations: OnceCell<Arc<DomainConstantEvaluations<F>>>,
+    precomputations: Arc<LazyCache<Arc<DomainConstantEvaluations<F>>>>,
 
     /// Disable gates checks (for testing; only enables with development builds)
     pub disable_gates_checks: bool,
 }
 
+pub(crate) type LookupConstraintSystemCache<F> =
```
dannywillems, line 212: Why do you have the `deserialize` separated?

kimchi/src/tests/lazy_mode.rs
```diff
@@ -0,0 +1,85 @@
...
+};
+use ark_ff::Zero;
+use itertools::iterate;
+use mina_curves::pasta::{Fp, Vesta, VestaParameters};
+use mina_poseidon::{
+    constants::PlonkSpongeConstantsKimchi,
+    sponge::{DefaultFqSponge, DefaultFrSponge},
+};
+use rand::Rng;
+use std::array;
```
dannywillems, line 16: Can you replace by `core::array`? We try to use `core` as much as possible instead of `std`.

utils/src/lazy_cache.rs
```diff
@@ -0,0 +1,256 @@
+//! This is a polyfill of the `LazyLock` type in the std library as of Rust 1.80.
```
dannywillems, line 1: From which version of LazyLock did you add this interface? It does not match the implementation [here](https://doc.rust-lang.org/src/std/sync/lazy_lock.rs.html).

Or is it supposed to be a copy-paste of the interface?

kimchi/src/circuits/constraints.rs
querolita: It was @richardpringle suggestion to avoid implementing a `Default` for `LazyCache` 😅 (required for serde)

kimchi/src/tests/lazy_mode.rs
querolita: Oh that's true

utils/src/lazy_cache.rs
querolita: It is not meant to be an exact copy-paste because that one was using way too many types which were not accessible from outside rust itself. Meaning, types which were pub(crate). Thus, if I wanted an exact replication I would have had to bring too many lines of code, which I thought we would not need for a simpler feature. It is basically heavily inspired by the implementation, but I removed parts which didn't seem as essential to provide the basic functionality.

dannywillems: Makes sense.

kimchi/src/circuits/constraints.rs
dannywillems: Makes sense.

kimchi/src/tests/foreign_field_add.rs
```diff
@@ -1176,50 +1165,40 @@ fn test_random_bad_parameters() {
         &[FFOps::Add],
         false,
     );
+
+    let mut cs = match Arc::try_unwrap(index.cs) {
+        Ok(cs) => cs,
+        Err(_) => panic!("Multiple references of Arc cs"),
```
dannywillems, line 1171: I'm wondering why this can be an issue?

dannywillems: We're in a test, so I guess it is fine.

book/src/specs/kimchi.md
```diff
@@ -1732,7 +1732,7 @@ pub struct ProverIndex<G: KimchiCurve, OpeningProof: OpenProof<G>> {
     pub max_poly_size: usize,
 
     #[serde(bound = "ColumnEvaluations<G::ScalarField>: Serialize + DeserializeOwned")]
-    pub column_evaluations: ColumnEvaluations<G::ScalarField>,
+    pub column_evaluations: Arc<LazyCache<ColumnEvaluations<G::ScalarField>>>,
```
dannywillems, line 1735: More for a discussion (no changes requested): we're adding many indirection here, even if it seems needed.
I hope it won't affect performances too much. I don't think so as the computation that will happen with the data resulting to accessing the additional indirection are way more expensive than going through memory.

I'm wondering how this will affect o1js - I don't have any idea.

dannywillems: Is it three indirections?
Arc -> LazyCache -> Option (for the field value) -> T

`UnsafeCell` is `transparent` (see https://doc.rust-lang.org/src/core/cell.rs.html#2073), so there is no additional pointer.

dannywillems: Also, even if it is negligible, we add a (really small) additional memory footprint. Explicitly:
- Some for the additional data in `Arc` (i.e. `strong` and `weak` -> 2 * 64 bits).
- Some for the fields in LazyCache -> `once`, `value` and `init`.

Only a few words, but still interesting to notice.

dannywillems: We only have one `ProverIndex` per proof, so it is fine.

dannywillems (commented):
Holding approval until the counterparts in Mina and o1js are updated and CI says it is all green.
At a high-level, it LGTM.
I'm mostly worried about the memory management at the WebAssembly/JavaScript level.

dannywillems (commented):
I have not spent time thinking on how much this pull request closes https://github.com/o1-labs/o1js/issues/2006. It seems that in the issue, different modes are suggested, and other changes should happen in the WebAssembly backend mostly.

I do not see yet how this would fit into the bigger task described in 2006.

I am also wondering what the tests written in `src/tests/lazy_mode.rs` is used for. Is it to simply check that the code correctly loads the value when `lazy_mode` is set to `true`?

If it is to benchmark the differences in memory and/or CPU time, I would consider adding more actual benchmarks.
After discussing with @querolita, it seems benchmarking should arrive later.

querolita:
Regarding WASM potential issues with Arc management, probably I could remove Arcs at the cost of more clones in the code, which would be bad for performance in both cases: precomputed and deferred (as in both cases the closure is needed to init the LazyCache, regardless of whether it is dropped/evaluated right after when lazy_mode = true)

querolita:
@dannywillems 
This PR is mostly about enabling a mode in Kimchi to allow lazy computation at proving time of some parameters that would otherwise be precomputed at setup phase, which should have some effects on the size of the prover index. An actual benchmark hasn't been carried out, and would probably take place once all the chain of PRs is in place (@richardpringle shared some commands I could use to measure the WASM used space). 

At a later phase of the project, it was suggested that this lazy computation could happen in a more fine-grained fashion, enabling a custom precomputation of parameters (instead of all or nothing as it would be the case with the single-bool lazy_mode flag).

utils/src/lazy_cache.rs
```diff
@@ -0,0 +1,283 @@
...
```
dannywillems, line 1: FWIW: we use now 1.81.

dannywillems: Let's not change though in this PR :). Can be done in a follow-up.

dannywillems:
Restarted the 1.81 job

dannywillems:
Nitpick (for the future PRs): it would be nice to have a cleaner git history and prefix the commit message with the section it modifies (for instance `kimchi: [...]`).

kimchi/Cargo.toml
```diff
@@ -13,6 +13,10 @@ license = "Apache-2.0"
 path = "src/lib.rs"
 bench = false       # needed for criterion (https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options)
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+jemalloc-ctl = { workspace = true, optional = true }
```
dannywillems, line 17: I am a bit worried that the package seems to not be maintained anymore: https://github.com/gnzlbg/jemallocator.

dannywillems: It seems there is this maintained fork: https://github.com/tikv/jemallocator.

dannywillems: It seems to be the one recommended in the Rust book: https://nnethercote.github.io/perf-book/build-configuration.html#alternative-allocators

Cargo.toml
```diff
@@ -53,6 +53,8 @@ getrandom = { version = "=0.2.15", features = ["js"] }
 hex = { version = "0.4", features = ["serde"] }
 iai = "0.1"
 itertools = "0.12.1"
+jemalloc-ctl = { version = "0.5" }
+jemallocator = { version = "0.5" }
```
dannywillems, line 57: As mentioned [here](https://github.com/o1-labs/proof-systems/pull/3079/files#r2089285483), I think it would be better to use the maintained fork: 
```toml
[...]
[dependencies]
[...]
tikv-jemallocator = "0.5"
```

dannywillems: Even version 0.6 (latest release)

dannywillems (changes requested):
I am requesting a change for the crates for the memory allocator.

Cargo.toml
querolita: Actually I had to go for 0.5 because I cannot find `stats` in `tikv-jemalloc-ctl:0.6`...

dannywillems (approved):
Approving, but as discussed with @querolita, we leave it on hold until we have the confirmation that it gives an appropriate result in terms of memory usage.

kimchi/Cargo.toml
```diff
@@ -13,6 +13,10 @@ license = "Apache-2.0"
 path = "src/lib.rs"
 bench = false       # needed for criterion (https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options)
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
```
dannywillems, line 16: Nit/unrelated to this patch: maybe we should also use this instead of a feature `wasm-types` for the rest of the WebAssembly code. Just a thought, not sure if it would work.

dannywillems:
This should not have been merged as the counterpart on Mina hasn't been approved.

PR #3098, opened 2025-03-21
Arrabbiata: Implement MinRoot VDF

dannywillems:
Do not review.

Note that the implementation is not tested yet. I post-pone this after the witness environment and the indexed relation takes into account the new ZkApp interface.


dannywillems:
Closing. I would like to come up with a bigger PR.

PR #3100, opened 2025-03-24
Arrabbiata: use GAT instead of type parameters in ZkApp trait

dannywillems:
Do not review.

As suggested by @richardpringle in https://github.com/o1-labs/proof-systems/pull/3097#pullrequestreview-2708680045.


dannywillems:
Closing. I would like to come up with a bigger PR.

PR #3106, opened 2025-03-26
Deps: enforcing strict deps on arkworks

dannywillems:
As arkworks do not follow semver, it is worth having strict condition on the version we want. Also, this has to be replicated on the Mina side to be sure we have the same version.


dannywillems:
Closing. It is fine as we use bugfix.

PR #3107, opened 2025-03-27
CI: dropping versions before Rust 1.82

dannywillems:
This should be merged after https://github.com/MinaProtocol/mina/pull/16794 has been merged, with the counterparties in o1js.


dannywillems:
Before merging this PR, settings should be updated to enforce 1.82.0

PR #3108, opened 2025-03-28
hasher: use core instead of std

dannywillems:
Relying on [core](https://github.com/rust-lang/rust/tree/master/library/core) instead of [std](https://github.com/rust-lang/rust/tree/master/library/std) decreases the number of dependencies, and make it more portable through different platforms.


dannywillems:
> I've PR'd your branch to fully make `hasher` `no_std`. Might as well go all the way, otherwise using a mix of `core` and `std` isn't all that helpful (people can just add new code that relies on `std` without realizing that you're trying to remove dependency on `std`.
> 
> #3118

Totally agree with your comment. My apologies to not have done it directly. As discussed in DM, it wasn't the full intention right now to get rid of std, but I should have taken the time as it was a pretty small changes. Thanks!

PR #3112, opened 2025-03-28
Deps: remove colored

dannywillems:
I took the liberty of removing it as it does not give much functional value to the codebase.


Cargo.lock
```diff
@@ -2613,7 +2601,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.52.0",
+ "windows-sys",
```
marcbeunardeau88, line 2604: I don't understand why this happens

dannywillems: :thinking: 
I would say, with a quite high confidence, that it is because with `colored`, two different versions of windows-sys were required (one introduced by colored, see
```
[[package]]
name = "colored"
version = "2.1.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "cbf2150cce219b664a8a70df7a1f933836724b503f8a413af9365b4dcc4d90b8"
dependencies = [
 "lazy_static",
 "windows-sys 0.48.0",
]
```
and another one introduced by `tempfile`, `isterminal`, etc. This line seems to remove the strict version dependency.

dannywillems: See https://github.com/colored-rs/colored/blob/v2.1.0/Cargo.toml#L21

marcbeunardeau88 (approved):

PR #3115, opened 2025-03-28
folding/ivc/curves: use core instead of std

dannywillems:
It is only a first step through having a no_std support. It is not the end-goal of this PR, but only a support in case we want to have it in the future.


marcbeunardeau88:
Not sure I see the point, but I don't mind so I gave the approval

marcbeunardeau88 (approved):

PR #3120, opened 2025-03-31
Kimchi: explicitly activate parallel and asm features

dannywillems:
It is activated at the top-level in [dependencies], and the documentation says:
```
features declared in this table are additive with the features from [dependencies]
```
Source: https://doc.rust-lang.org/cargo/reference/workspaces.html

Wondering if adding it just in case is worth.


dannywillems:
Useless. Closing.

PR #3154, opened 2025-04-07
[plonk-wasm] Remove static mut from rayon

richardpringle:
Moving this from here: https://github.com/MinaProtocol/mina/pull/16758

I've swapped the unsafe use of `static mut` for the safe usage of `LocalKey<RefCell<T>>` such that we are guaranteed not to have any UB. This comes at a slight runtime overhead, but this should not be a problem as it will only happen on starting or stopping the thread-pool. `static mut` is dangerous, and safety has only really been upheld by the runtime environment outside of this crate. If that runtime environment were to change, the old code could cause UB.

### DIFF
--- a/plonk-wasm/src/rayon.rs
+++ b/plonk-wasm/src/rayon.rs
@@ -13,20 +13,22 @@
 
 use js_sys::Promise;
 use spmc::{channel, Receiver, Sender};
+use std::{cell::RefCell, thread_local};
 use wasm_bindgen::prelude::*;
 
 #[cfg(feature = "nodejs")]
 use js_sys::JsString;
 
-static mut THREAD_POOL: Option<rayon::ThreadPool> = None;
+thread_local! {
+    static THREAD_POOL: RefCell<Option<rayon::ThreadPool>> = RefCell::new(None);
+}
 
 pub fn run_in_pool<OP, R>(op: OP) -> R
 where
     OP: FnOnce() -> R + Send,
     R: Send,
 {
-    let pool = unsafe { THREAD_POOL.as_ref().unwrap() };
-    pool.install(op)
+    THREAD_POOL.with_borrow(|pool| pool.as_ref().unwrap().install(op))
 }
 
 #[wasm_bindgen]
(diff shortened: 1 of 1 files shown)


plonk-wasm/src/rayon.rs
```diff
@@ -79,25 +81,23 @@ impl PoolBuilder {
...
+            // https://bugs.chromium.org/p/chromium/issues/detail?id=1075645
+            .spawn_handler(move |thread| {
+                // Note: `send` will return an error if there are no receivers.
+                // We can use it because all the threads are spawned and ready to accept
+                // messages by the time we call `build()` to instantiate spawn handler.
+                self.sender.send(thread).unwrap_throw();
+                Ok(())
+            });
+
+        THREAD_POOL.with_borrow_mut(|pool| {
```
volhovm, line 98: Also: Consider moving the `thread_pool_builder` building process into the `with_borrow_mut(..)` maybe. Like this it can happen (can it?) that two concurrent calls to `build` will be out of order: call A, call B, building and setting pool B, building and setting pool A. Not sure if this is desirable (maybe I'm wrong assuming these two calls can be parallel at all).

plonk-wasm/src/rayon.rs
```diff
@@ -13,20 +13,22 @@
...
 use js_sys::Promise;
 use spmc::{channel, Receiver, Sender};
+use std::{cell::RefCell, thread_local};
 use wasm_bindgen::prelude::*;
 
 #[cfg(feature = "nodejs")]
 use js_sys::JsString;
 
-static mut THREAD_POOL: Option<rayon::ThreadPool> = None;
+thread_local! {
```
volhovm, line 22: Yeah come to think of that I _really_ don't understand the use of `thread_local!` here. It seems that this thing should be accessible only in a single thread manner, so maybe what you want is a static `Mutex`.

volhovm:
@richardpringle let's also give Danny the chance to look at this before you merge? just in case

dannywillems (changes requested):
To avoid issues like we're having right now (see https://github.com/o1-labs/proof-systems/pull/3157), I would like to request a counterpart PR in o1js (+ Mina + o1js-bindings) to be sure that it does not break anything.
The changes might look innocent, but we encountered an equivalent case with https://github.com/o1-labs/proof-systems/pull/3163.

A successful build can be created on top of https://github.com/o1-labs/proof-systems/pull/3162 (and its counterpart in o1js). @richardpringle: you can also wait for later today CET to have the whole set of PR merged in master/main/compatible of the different repositories, if you want to get something easier to build on top of.

dannywillems:
@richardpringle you should be able to update Mina + o1js from now. You can merge master in this branch, and use the latest compatible on Mina ([695c6ef](https://github.com/MinaProtocol/mina/commit/695c6ef6e2b051c44cc5de44239eb5cfe85c12c6)).

PR #3165, opened 2025-04-08
Making a call instead of accessing const directly

dannywillems:


dannywillems:
It didn't give anything relevant. Closing, keeping for history.

PR #3169, opened 2025-04-09
kimchi-stubs: stop generating dylib

dannywillems:
On MacOS arm64, ocaml-rs causes some build issues.


Trivo25 (approved):
Confirming that this works on my machine

PR #3174, opened 2025-04-10
Add GitHub Actions workflow for automated typo checking

dannywillems:
This workflow regularly checks the codebase for typos using codespell:
- Checks documentation files (.md) and Rust source code (.rs)
- Excludes generated files and directories (target, JSON, lock files)
- Ignores known false positives
- Provides a summary report and steps to fix issues locally

This automated check will help maintain higher documentation quality.

### DIFF
--- a/.github/workflows/typo-check.yml
+++ b/.github/workflows/typo-check.yml
@@ -1,0 +1,36 @@
+name: Check for typos
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


volhovm (dismissed):
That's cool. I have two suggestions on top:
* Let's not make it mandatory? maybe we could add an exception words list file, e.g. it does not like `wit`...
* We should fix all the existing typos...

PR #3191, opened 2025-04-24
Use David Tolnay's toolchain support

richardpringle:
This is step 1 in some more cleanup that I need to build `plonk-native` (a native version of `plonk-wasm`).

Changes
 - turn on rust-cache by default. It's easy to change the default behaviour but also easy to turn the cache off for specific jobs with `cache: false`
 - change the default behaviour to only install the local target unless otherwise specified with `targets: wasm32-unknown-unknown, <other>`
 - Use `taiki-e/install-action` which delegates to `cargo binstall`. `binstall` will give use a major speed up as we won't be building cargo-tools from source anymore. It still does a checksum check and falls back to `cargo install` in the case where the binary isn't available. 
 -  ~~The last thing that I did that might be a little more contencious is use the exact commands from the makefile instead of using make. Personally, I find it's a lot easier to debug when the commands are directly placed in the action. This comes at the cost of deviating from the makefile. In the past, I have always treated CI as the source of truth.~~

### DIFF
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -22,9 +22,9 @@ jobs:
         with:
           rust_toolchain_version: ${{ matrix.rust_toolchain_version }}
 
-      - name: Install cargo-sort
-        run: cargo install cargo-sort
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-sort
 
       - name: Check formatting related to Rust code
-        run: |
-          make check-format
+        run: make check-format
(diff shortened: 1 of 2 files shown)


.github/actions/toolchain-shared/action.yml
```diff
@@ -1,24 +1,27 @@
 name: "Shared Rust toolchain setting up steps"
 description: "Shared Rust toolchain setting up steps"
 inputs:
+  cache:
+    description: "to cache or not to cache, that is the question"
```
dannywillems, line 5: :smile:

dannywillems (approved):

PR #3196, opened 2025-04-29
Speedup CI with better parallelism and caching

richardpringle:
Instead of running everything as sequential steps, run the things that were previously steps as jobs such that they can be run in parallel. This cuts CI time in half. 

Further, run the PR CI once per day to clear any unused dependencies and set up the cache so that PR runs aren't starting from scratch.


.github/workflows/ci.yml
```diff
@@ -85,11 +143,6 @@ jobs:
       # Doc & Spec
       #
 
-      - name: Install cargo-spec for specifications
```
dannywillems, line 88: Nice.

.github/workflows/ci.yml
```diff
@@ -43,17 +51,63 @@ jobs:
           make deps
           make build
 
-  run_checks:
-    name: Run checks and tests
+  define-matrix:
```
dannywillems, line 54: Nice. I've always wanted to try it.

.github/workflows/ci.yml
```diff
@@ -43,17 +51,63 @@ jobs:
...
+        run: cargo clean
+
+      # else
+
+      - name: skip-refresh
+        if: ${{ github.event_name != 'workflow_run' }}
+        shell: bash
+        run: echo 'no need to refresh'
+
+  doc-and-spec:
```
dannywillems, line 103: Nit: Do we want to run it for each Rust version? I do believe we did remove this. We should have a single job running the one that will be run on `master` to update the docs, and use a single Rust version. Can be done in a follow-up.

dannywillems: See `./github/workflows/gh-page.yml`

dannywillems: > I do believe we did remove this.

I'm wrong, sorry. We do run it for each run, you're right. I was confused by the diff.

dannywillems (approved):
Nice. Updating the requirements now.

dannywillems (changes requested):
Blocking until I set the correct targets.

dannywillems (approved):
All good now. Thanks.

PR #3198, opened 2025-04-30
Update flake to support ocaml

volhovm:
This adds ocaml environment into the flake, which became necessary to compile the project since we brought the bindings into the repo. 

To test run `nix develop`.


Fizzixnerd (approved):

dannywillems (commented):
Follow-up suggestion: would be nice to check in CI that it builds correctly. We're adding a runner where Nix is installed, so we would be able to use it here.

PR #3199, opened 2025-04-30
Stop clippy ignoring this file in `snarky`

richardpringle:
This was easy to fix and the behaviour has not changed


dannywillems:
Note that snarky isn't used, and the code should be deleted. It has been added a long time ago.
To be honest, I would simply suggest to remove it.

dannywillems (commented):
I would be in favor of removing the whole library `snarky`.

dannywillems (approved):

PR #3200, opened 2025-05-01
Fix clippy & suppress plonk-wasm deprecated clippy warnings

volhovm:
This fixes some clippy warnings:
- Fixes saffron related warnings. Also un-excludes saffron from CI so that these checks are done before merging.
- Suppresses plonk-wasm deprication warnings (https://github.com/o1-labs/proof-systems/pull/3163). Un-excludes plonk-wasm from CI, same reasoning.


dannywillems (approved):
Awesome, thanks!

PR #3183, opened 2025-04-18
Saffron: Add simple read proof

volhovm:
Subj. Adding the read proof attesting to `d * q - a = 0`.

PR #3202, opened 2025-05-02
PU: Implement naive folding & prover/verifier

volhovm:
This adds simple interactive folding: two functions for folding specifically (`folding_prover`, `folding_verifier`), and prover/verifier for the relaxed statement. No IVC/verification-in-circuit.

### DIFF
--- a/saffron/src/folding.rs
+++ b/saffron/src/folding.rs
@@ -1,0 +1,694 @@
+//! Folding version of the read prover.
+//!
+//! The main gist of the protocol is to run a trivial sangria-like
+//! folding without recursive circuit (non-IVC variant),
+//! interactively, between the prover and the verifier.
+//!
+//! The original equation we want to prove is `d(X) * q(X) - a(X) = 0`.
+//!
+//! After homogenization it becomes `d(X) * q(X) - a(X) * u - e(X) =
+//! 0` where u is a homogenization factor (field element), and `e(X)`
+//! is an error polynomial.
+//!
+//! During the folding procedure we fold one non-relaxed instance with
+//! a relaxed instance (number 1 and 2 correspondingly), and the
+//! result of this is a relaxed instance. The recombination
+//! coefficient `r` is sampled as usual, using Fiat-Shamir, and then
+//! we recombine vectors and commitments as follows:
+//!
+//! `a_3(X) <- a_1(X) + r * a_2(X)` (same for `d, q`)
+//!
+//! To compute the new error term, we have to use the formula
+//!
+//! `e_3(X) <- r * cross_term(X) + r^2 * e_2(X)`
+//!
+//! where `cross_term(X) := a_2 + u_2 * a_1 - q_2 * d_1 - q_1 * d_2`.
+//!
+//! Finally, in this file we provide a prover and verifier attesting
+//! to the validity of a relaxed instance using plonk-like protocol
+//! with IPA.
+//!
+//! For mor details see the full version of the protocol in the whitepaper.
+
+use crate::{Curve, CurveFqSponge, CurveFrSponge, ScalarField, SRS_SIZE};
+use ark_ec::AffineRepr;
+use ark_ff::{Field, One, Zero};
+use ark_poly::{
+    univariate::DensePolynomial, EvaluationDomain, Evaluations, Polynomial,
+    Radix2EvaluationDomain as R2D,
+};
+use kimchi::{circuits::domains::EvaluationDomains, curve::KimchiCurve, plonk_sponge::FrSponge};
+use mina_poseidon::FqSponge;
+use poly_commitment::{
+    commitment::{combined_inner_product, BatchEvaluationProof, CommitmentCurve, Evaluation},
+    ipa::{OpeningProof, SRS},
+    utils::DensePolynomialOrEvaluations,
+    PolyComm, SRS as _,
+};
+use rand::{CryptoRng, RngCore};
+
+#[derive(PartialEq, Eq)]
+/// Non-relaxed instance attesting to `d * q - a = 0`
+pub struct CoreInstance {
+    /// Commitment to the data
+    comm_d: Curve,
+    /// Commitment to the query polynomial
+    comm_q: Curve,
+    /// Commitment to the answers
+    comm_a: Curve,
+}
+
+#[derive(PartialEq, Eq)]
+/// Relaxed instance variant.
+pub struct RelaxedInstance {
+    /// Non-relaxed part
+    core: CoreInstance,
+    /// Homogeneization term for folding
+    u: ScalarField,
+    /// Commitment to the error term for folding
+    comm_e: Curve,
+}
+
+impl CoreInstance {
+    pub fn relax(self) -> RelaxedInstance {
+        RelaxedInstance {
+            core: self,
+            u: ScalarField::one(),
+            comm_e: Curve::zero(),
+        }
+    }
+}
+
+#[derive(PartialEq, Eq)]
+/// Non-relaxed witness contains evaluations (field vectors) for data,
+/// query, and answers.
+pub struct CoreWitness {
+    d: Evaluations<ScalarField, R2D<ScalarField>>,
+    q: Evaluations<ScalarField, R2D<ScalarField>>,
+    a: Evaluations<ScalarField, R2D<ScalarField>>,
+}
+
+#[derive(PartialEq, Eq)]
+/// Relaxed witness extends the non-relaxed witness with evaluations
+/// of the error term.
+pub struct RelaxedWitness {
+    core: CoreWitness,
+    e: Evaluations<ScalarField, R2D<ScalarField>>,
+}
+
+impl CoreWitness {
+    pub fn relax(self, domain: R2D<ScalarField>) -> RelaxedWitness {
+        RelaxedWitness {
+            core: self,
+            e: Evaluations::from_vec_and_domain(vec![ScalarField::zero(); domain.size()], domain),
+        }
+    }
+}
+
+impl RelaxedInstance {
+    /// This function checks if the provided instance is valid regarding the
+    /// provided witness, by checking both the validity of the witness and its
+    /// consistency with the instance's commitments.
+    pub fn check_in_language(
+        &self,
+        srs: &SRS<Curve>,
+        domain: R2D<ScalarField>,
+        wit: &RelaxedWitness,
+    ) -> bool {
+        for i in 0..SRS_SIZE {
+            // todo can be parallelized
+            if self.u * wit.core.a[i] - wit.core.q[i] * wit.core.d[i] + wit.e[i]
+                != ScalarField::zero()
+            {
+                return false;
+            }
+        }
+        if self.core.comm_a
+            != srs
+                .commit_evaluations_non_hiding(domain, &wit.core.a)
+                .chunks[0]
+        {
+            return false;
+        }
+        if self.core.comm_d
+            != srs
+                .commit_evaluations_non_hiding(domain, &wit.core.d)
+                .chunks[0]
+        {
+            return false;
+        }
+        if self.core.comm_q
+            != srs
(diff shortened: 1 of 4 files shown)


saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+        u: new_u,
+        comm_d: comm_d3.into(),
+        comm_q: comm_q3.into(),
+        comm_a: comm_a3.into(),
+        comm_e: comm_e3.into(),
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ReadProof {
```
anne-laure-s, line 186: Nit for later : there is room for factorization with read_proof.rs.

After reading the code, it may be even cleaner to remove read_proofs.rs

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+    RNG: RngCore + CryptoRng,
+{
+    let mut fq_sponge = CurveFqSponge::new(Curve::other_curve_sponge_params());
+
+    // TODO we assume that (inst,wit) ∈ L, that is inst.comm_d = Com(wit.d), etc.
+
+    let data_poly: DensePolynomial<ScalarField> = wit.d.interpolate_by_ref();
+    let query_poly: DensePolynomial<ScalarField> = wit.q.interpolate_by_ref();
+    let answer_poly: DensePolynomial<ScalarField> = wit.a.interpolate_by_ref();
+    let error_poly: DensePolynomial<ScalarField> = wit.e.interpolate_by_ref();
```
anne-laure-s, line 221: As the commitments for these polynomials already exist in `inst`, I also expect these polynomial to have already been computed (in order to compute the commitment). Should we expect the polynomials as argument to avoid recomputation ?

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+    rng: &mut RNG,
+    inst: &RelaxedInstance,
+    wit: &RelaxedWitness,
+) -> ReadProof
+where
+    RNG: RngCore + CryptoRng,
+{
+    let mut fq_sponge = CurveFqSponge::new(Curve::other_curve_sponge_params());
+
+    // TODO we assume that (inst,wit) ∈ L, that is inst.comm_d = Com(wit.d), etc.
```
anne-laure-s, line 216: Use `check_in_language` ?
Or do you expect the prover to perform these checks before (or not check at all) ?

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+    for eval in [
+        data_eval,
+        query_eval,
+        answer_eval,
+        error_eval,
+        quotient_eval,
+    ]
+    .into_iter()
+    {
+        fr_sponge.absorb(&eval);
```
anne-laure-s, line 281: Nit for later : it may be nice to have the possibility to feed the sponge with a list of scalars, in the same spirit as `absorb_g`.

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+        let comm_a = SRS
+            .commit_non_hiding(
+                &Evaluations::from_vec_and_domain(answer.clone(), DOMAIN.d1).interpolate(),
+                1,
+            )
+            .chunks[0];
+
+        let relaxed_instance = RelaxedInstance {
+            u: ScalarField::one(),
+            comm_e: Curve::zero(),
```
anne-laure-s, line 498: This only outputs non-relaxed instances, maybe we also want to test folding with relaxed instances ?

You can also make it clear in the function name or documentation or types, and have one more round of folding in the test to create your relaxed instance to fold.

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+// the first instance/witness is supposed to be non-relaxed
+pub fn folding_prover(
+    srs: &SRS<Curve>,
+    inst1: &RelaxedInstance,
+    inst2: &RelaxedInstance,
+    wit1: &RelaxedWitness,
+    wit2: &RelaxedWitness,
+) -> (RelaxedInstance, RelaxedWitness, Curve) {
+    assert!(inst1.u == ScalarField::one());
+    assert!(inst1.comm_e == Curve::zero());
```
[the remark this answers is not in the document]
volhovm: I could do it, yeah. In the previous implementation of folding there was a non-relaxed instance type but it was making things more confusing smhw.

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+
+    let mut fq_sponge = CurveFqSponge::new(Curve::other_curve_sponge_params());
+
+    let error_term: Evaluations<ScalarField, R2D<ScalarField>> =
+        &(&(&wit2.a + &(&wit1.a * inst2.u)) - &(&wit2.q * &wit1.d)) - &(&wit1.q * &wit2.d);
```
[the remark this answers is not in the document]
volhovm: Yes, sorry, bad naming. It's a cross term indeed.

saffron/src/folding.rs
volhovm: We can remove `read_proof.rs` later indeed.

saffron/src/folding.rs
volhovm: I want to assume that the (inst,wit) passed here is a valid pair by construction. I'd avoid adding extra checks into the prover... I want tor remove these asserts too.

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
+//! Folding version of the read prover.
+
```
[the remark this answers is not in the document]
volhovm: I will redescribe it very shortly and refer to the whitepaper as an original source of truth, how does this sound?

saffron/src/folding.rs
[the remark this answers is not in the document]
volhovm: done

saffron/src/folding.rs
```diff
@@ -0,0 +1,590 @@
...
+            return false;
+        }
+        true
+    }
+}
+
+// the first instance/witness is supposed to be non-relaxed
+pub fn folding_prover(
+    srs: &SRS<Curve>,
+    inst1: &RelaxedInstance,
```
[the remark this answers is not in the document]
volhovm: Re (1) I'm against tuples as input types, it does look to me like an anti-pattern in rust.

Re (2) I think calling "full witness & its commitments" an instance is a confusing and much less used variant of the terminology. You could go for `Statement` or `Witness` for semantic consistency, but then I'm also against using too many types and overgeneralising. I think it's super clear rn.

@dannywillems @marcbeunardeau88 wanna insert a third opinion?

saffron/src/folding.rs
volhovm: Makes sense. I'll add another round of folding in the test.

saffron/src/folding.rs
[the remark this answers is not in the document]
volhovm: I will make the argument order `inst1 wit1 inst2 wit2` instead of `inst1 inst2 wit1 wit2` though, I agree this is a bit better.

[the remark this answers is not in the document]
anne-laure-s: I’m happy with the way you changed it

saffron/src/folding.rs
volhovm: As per https://github.com/o1-labs/proof-systems/pull/3226 my understanding is that interpolation takes only 2ms, which we have to do 4 times here.

Going with your suggestion would be equivalent to recombining `4` vectors with challenges every time we fold, which is in total going to accumulate to much more.

That being said, I think this optimisation is not worth it, and is very minor. 

Also I now fixed it to use `commit_evals` in this PR where I could.

saffron/src/folding.rs
```diff
@@ -0,0 +1,695 @@
...
+//! The original equation we want to prove is `d(X) * q(X) - a(X) = 0`.
+//!
+//! After homogenization it becomes `d(X) * q(X) - a(X) * u - e(X) =
+//! 0` where u is a homogenization factor (field element), and `e(X)`
+//! is an error polynomial.
+//!
+//! During the folding procedure we fold one non-relaxed instance with
+//! a relaxed instance (number 1 and 2 correspondingly), and the
+//! result of this is a relaxed instance. The recombination
+//! coefficient `r` is sampled as usually, using Fiat-Shamir, and then
```
marcbeunardeau88, line 16:
```suggestion
//! coefficient `r` is sampled as usual, using Fiat-Shamir, and then
```

saffron/src/folding.rs
```diff
@@ -0,0 +1,695 @@
...
+
+// the first instance/witness is supposed to be non-relaxed
+pub fn folding_prover(
+    srs: &SRS<Curve>,
+    domain: R2D<ScalarField>,
+    inst1: &CoreInstance,
+    wit1: &CoreWitness,
+    inst2: &RelaxedInstance,
+    wit2: &RelaxedWitness,
+) -> (RelaxedInstance, RelaxedWitness, Curve) {
```
marcbeunardeau88, line 162: Might be nicer to alias the type curve with Proof or FoldingProof, as we have here a prover which seemingly does not return a proof

saffron/src/folding.rs
```diff
@@ -0,0 +1,695 @@
...
+            if self.u * wit.core.a[i] - wit.core.q[i] * wit.core.d[i] + wit.e[i]
+                != ScalarField::zero()
+            {
+                return false;
+            }
+        }
+        if self.core.comm_a
+            != srs
+                .commit_evaluations_non_hiding(domain, &wit.core.a)
+                .chunks[0]
```
marcbeunardeau88, line 129: defining a Column<X> = {a:X, q: X ; d :X , e : X} and creating an iterator on it would allow to not repeat this part of the code, as well as in other places.
(Or maybe a Column={a;q;d}and a ColumnRelaxed = {Column; E})

marcbeunardeau88 (dismissed):

saffron/src/folding.rs
```diff
@@ -0,0 +1,695 @@
...
+        inst1.comm_q,
+        inst1.comm_a,
+        inst2.core.comm_d,
+        inst2.core.comm_q,
+        inst2.core.comm_a,
+        inst2.comm_e,
+        cross_term_comm,
+    ]);
+
+    let recombination_chal = fq_sponge.squeeze(2);
```
dannywillems, line 184: Nit: as mentioned in https://github.com/o1-labs/proof-systems/pull/3183, we can use `challenge`.

saffron/src/folding.rs
volhovm: Maybe for the future? Thinking of how it would work in practice with many lists iterating not just over `a q d e` but also query polynomial, and in folding you need to iterate over three first ones for linear recombination, and `e` is recombined differently... Right now it seems to add much more type noise than provide clarity in my opinion.

saffron/src/folding.rs
volhovm: I think `recombination_chal` is more descriptive: aside from it, we also have challenge seed for storage proof (`challenge_seed`), the challenge itself `challenge` (for recombining!), in other plonkish proofs we have evaluation challenge `zeta` (a point, I call it `evaluation_point`), and we have two challenges `polyscale` and `evalscale` for recombining polynomials. Everything verifier samples is a challenge, but I'm thinking if it's cheap to give it a distinct name, it makes it more clear?

saffron/src/folding.rs
```diff
@@ -0,0 +1,695 @@
...
+            evalscale,
+            evaluations: coms_and_evaluations,
+            opening: &proof.opening_proof,
+            combined_inner_product,
+        }],
+        rng,
+    )
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
volhovm: Yep, will do, should be easy.

marcbeunardeau88 (approved):

saffron/src/folding.rs
dannywillems: I meant
```
let recombination_chal = fq_sponge.challenge()
```

PR #3203, opened 2025-05-02
CI: Use xtask pattern for wasm-pack

richardpringle:
Now cargo manages the wasm-pack dependency instead of relying on
whatever is in your shell. This allows you to have one global install of
wasm-pack that's completely different from the `plonk-wasm` dependency

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -29,10 +29,6 @@ jobs:
           rust_toolchain_version: ${{ matrix.rust_toolchain_version }}
           targets: wasm32-unknown-unknown
 
-      - uses: taiki-e/install-action@v2
-        with:
-          tool: wasm-pack@0.12.1
-
       - shell: bash
         run: rustup component add rust-src
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if 1.0.0",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
 name = "ahash"
 version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


xtask/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+fn main() -> Result<()> {
+    let cli = Cli::parse();
+
+    match &cli.command {
+        Commands::BuildWasm { out_dir, target } => build_wasm(out_dir, *target),
+    }
+}
+
+fn build_wasm(out_dir: &str, target: Target) -> Result<()> {
+    const RUSTFLAGS: &str = "-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=--max-memory=4294967296";
```
dannywillems, line 52: I guess [this](https://github.com/o1-labs/proof-systems/blob/master/Makefile#L34) can be removed too then?

.github/workflows/wasm.yml
```diff
@@ -29,10 +29,6 @@ jobs:
           rust_toolchain_version: ${{ matrix.rust_toolchain_version }}
           targets: wasm32-unknown-unknown
 
-      - uses: taiki-e/install-action@v2
-        with:
-          tool: wasm-pack@0.12.1
```
dannywillems, line 34: IIUC, using xtask is mostly to have this as a dependency in the Cargo.toml?

dannywillems (approved):
First review LGTM. Do you mind checking that it doesn't change the build system on Mina + o1js? It should be ok, but we never know, and I don't want someone being locked if it is the case.

Cargo.lock
```diff
@@ -23,6 +23,17 @@ version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
+[[package]]
```
dannywillems, line 26: I don't like this big diff in Cargo.lock, but I guess faster CI comes with this cost.

.github/workflows/wasm.yml
dannywillems: Makes sense.

dannywillems (approved):

dannywillems:
I merged this too fast. Mina CI is broken, see https://buildkite.com/o-1-labs-2/mina-o-1-labs/builds/37715#0196a47e-d186-4155-9cb5-82e5fe2cc9da
```
File "src/lib/crypto/kimchi_bindings/js/web/dune", line 13, characters 0-395:
13 | (rule
14 |  (targets
15 |   plonk_wasm_bg.wasm.d.ts
....
29 |     "%{read:../../dune-build-root}/cargo_kimchi_wasm"
30 |     (run bash build.sh))
31 |    (write-file flags.sexp "()"))))
make: Entering directory '/workdir/_build/default/src/lib/crypto/proof-systems'
cargo +nightly xtask build-wasm \
	--target web \
	--out-dir /workdir/_build/default/src/lib/crypto/kimchi_bindings/js/web \
	--rust-version "nightly-2024-09-05"
error: no such command: `xtask`
```

PR #3204, opened 2025-05-05
Saffron/benches: add bench for read proof

dannywillems:
This will be used to compare with the upcoming optimization of read proofs.

### DIFF
--- a/saffron/benches/read_proof_bench.rs
+++ b/saffron/benches/read_proof_bench.rs
@@ -1,0 +1,117 @@
+//! Run this bench using `cargo criterion -p saffron --bench read_proof_bench`
+
+use ark_ff::{One, UniformRand, Zero};
+use ark_poly::{univariate::DensePolynomial, Evaluations};
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
+use kimchi::{circuits::domains::EvaluationDomains, groupmap::GroupMap};
+use mina_curves::pasta::{Fp, Vesta};
+use once_cell::sync::Lazy;
+use poly_commitment::{commitment::CommitmentCurve, ipa::SRS, SRS as _};
+use rand::rngs::OsRng;
+use saffron::{
+    env,
+    read_proof::{prove, verify},
+    ScalarField, SRS_SIZE,
+};
+
+// Set up static resources to avoid re-computation during benchmarks
+static SRS: Lazy<SRS<Vesta>> = Lazy::new(|| {
+    if let Ok(srs) = std::env::var("SRS_FILEPATH") {
+        env::get_srs_from_cache(srs)
+    } else {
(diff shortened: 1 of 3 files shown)


saffron/benches/read_proof_bench.rs
```diff
@@ -0,0 +1,117 @@
...
+use poly_commitment::{commitment::CommitmentCurve, ipa::SRS, SRS as _};
+use rand::rngs::OsRng;
+use saffron::{
+    env,
+    read_proof::{prove, verify},
+    ScalarField, SRS_SIZE,
+};
+
+// Set up static resources to avoid re-computation during benchmarks
+static SRS: Lazy<SRS<Vesta>> = Lazy::new(|| {
```
volhovm, line 18: Maybe using `pub fn get_srs_test<G>() -> SRS<G>` from `kimchi/src/precomputed_srs.rs` would be a more canonical way?

saffron/benches/read_proof_bench.rs
```diff
@@ -0,0 +1,117 @@
...
+                    &data_comm,
+                ))
+            },
+            BatchSize::NumIterations(10),
+        )
+    });
+}
+
+fn bench_read_proof_verify(c: &mut Criterion) {
+    let (data, query, answer, data_comm) = generate_test_data(SRS_SIZE);
```
volhovm, line 87: A word of caution: the default kimchi prover/verifier bench is quite inaccurate, so you might need to set up bench parameters manually if you want to achieve meaningfully low noise.

volhovm (approved):
cool stuff :+1:

saffron/benches/read_proof_bench.rs
```diff
@@ -0,0 +1,117 @@
+//! Run this bench using `cargo criterion -p saffron --bench read_proof_bench`
```
anne-laure-s, line 1:
```
$ cargo criterion -p saffron --bench read_proof_bench
error: no such command: `criterion`

        View all installed commands with `cargo --list`
        Find a package to install `criterion` with `cargo search cargo-criterion`
```
Is there something I am missing to run this command ?

dannywillems: cargo install cargo-criterion

saffron/benches/read_proof_bench.rs
dannywillems: Seems ok, but only because we have the same size. I was first thinking about making the change, but wondering now if it is the best as Saffron and Kimchi are too different protocols. We might want to differentiate them?

saffron/benches/read_proof_bench.rs
volhovm: I suspect it will largely generalise, only mentioned it as a particular example. I guess on average it's hard to measure verification precisely (because of multi-threading? I'm not sure why actually).

PR #3205, opened 2025-05-05
Taplo/format: set reorder_keys to true and update accordingly

dannywillems:


dannywillems:
> I'm not sure I agree with this one (I explicitly didn't set it).
> 
> The first thing I want to see when I look at a Cargo.toml is the name. While that leaves the files open to inconsistent ordering, you can see that the files are relatively consistent anyway. There usually aren't that many keys to re-order manually with the exception of "dependencies" which are ordered.
> 
> Thoughts?

I do not have a strong opinion, and I am not sure it is relevant to argue a long time over this. I would simply left it in draft and anyone can revamp it later if they want it.
This change is mostly to ease how I bisect through the code dependencies/fields when modifying/checking a file as I can go alphabetically.

dannywillems:
Closing as I don't think we're gonna accept it.

PR #3208, opened 2025-05-05
Saffron/benches: use get_srs_test from kimchi

dannywillems:
See https://github.com/o1-labs/proof-systems/pull/3204#discussion_r2073052855 for context.


dannywillems:
Also, tests do not use `get_srs_test`: https://github.com/o1-labs/proof-systems/blob/master/saffron/src/blob.rs#L164-L170.

PR #3210, opened 2025-05-05
plonk-wasm: Fix the warning until we can update

richardpringle:
`ark-ff` contains breaking changes so it might take a while to update.
On the current version of Rust, we get a bunch of annoying scoping
warnings that are easily fixed. I've got an open PR to
`arkworks-rs/algebra` (the workspace repo) to backport there changes
from `0.5.x` onto a 0.4.x branch. Hopefully they accept the PR and do a
new patch release. If not, we can update the dependency.

The `arkworks` team doesn't currrently have a branch to support patches, but maybe they'll consider it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,8 +222,7 @@ dependencies = [
 [[package]]
 name = "ark-ff-macros"
 version = "0.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


dannywillems:
Link to arkworks branch: https://github.com/arkworks-rs/algebra/pull/979

dannywillems:
Even though I like the idea of removing the warning, I would be in favor of not having a fork of arkworks used in this workspace. I would prefer to postpone until it is solved upstream.

PR #3217, opened 2025-05-07
Rust native with `neon-rs`

richardpringle:
- **plonk-wasm: factor out future common crates**
- **plonk-neon: add new crate for nodejs**

### DIFF
--- a/crates/arkworks/src/bigint_256.rs
+++ b/crates/arkworks/src/bigint_256.rs
@@ -29,7 +29,7 @@ pub struct WasmBigInteger256(pub BigInteger256);
 
 impl wasm_bindgen::describe::WasmDescribe for WasmBigInteger256 {
     fn describe() {
-        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe()
+        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe();
     }
 }
 
@@ -60,7 +60,7 @@ pub fn of_biguint(x: &BigUint) -> BigInteger256 {
     let mut bytes = x.to_bytes_le();
     bytes.resize(BIGINT256_NUM_BYTES, 0);
     let limbs = bytes.as_ptr();
-    let limbs = limbs as *const [u64; BIGINT256_NUM_LIMBS as usize];
+    let limbs = limbs.cast::<[u64; BIGINT256_NUM_LIMBS as usize]>();
     let limbs = unsafe { &(*limbs) };
     BigInt(*limbs)
 }
--- a/crates/arkworks/src/pasta_fq.rs
+++ b/crates/arkworks/src/pasta_fq.rs
@@ -41,7 +41,7 @@ impl<'a> From<&'a WasmPastaFq> for &'a Fq {
 
 impl wasm_bindgen::describe::WasmDescribe for WasmPastaFq {
     fn describe() {
-        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe()
+        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe();
     }
 }
 
--- a/crates/arkworks/src/pasta_fp.rs
+++ b/crates/arkworks/src/pasta_fp.rs
@@ -43,7 +43,7 @@ impl<'a> From<&'a WasmPastaFp> for &'a Fp {
 
 impl wasm_bindgen::describe::WasmDescribe for WasmPastaFp {
     fn describe() {
-        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe()
+        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe();
     }
 }
 


plonk-neon/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "plonk_neon"
```
dannywillems, line 2: Nit: I'd like to rename plonk-neon onto kimchi-neon. Same for plonk-wasm to be consistent.
I'll try to find time in the next few days to propagate this change everywhere.

crates/arkworks/src/bigint_256.rs
```diff
@@ -103,12 +106,16 @@ pub fn caml_bigint_256_compare(x: WasmBigInteger256, y: WasmBigInteger256) -> i8
     }
 }
 
-#[wasm_bindgen]
+// I don't think this is used anywhere
```
dannywillems, line 109: It is required in `pasta_bindings.ml` on the Mina side, see [here](https://github.com/MinaProtocol/mina/blob/compatible/src/lib/crypto/kimchi_bindings/stubs/pasta_bindings.ml#L18). When we compile Mina, we need to have this stub. The `print` statement might be used in the Caml code in different places.

plonk-neon/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+[lib]
+name = "plonk_neon"
+crate-type = ["cdylib"]
+
+################################# Dependencies ################################
+
+[dependencies]
+# Strictly enforcing 0.2.87
+
+libc = "0.2.62"
```
dannywillems, line 19: Any reason to not use the one from the workspace?

plonk-neon/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+once_cell = "1.10.0"
```
dannywillems, line 20: Any reason to not use the one from the workspace?

plonk-neon/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+ark-serialize.workspace = true
+arkworks.workspace = true
+
+# proof-systems
+mina-curves = { path = "../curves" }
+mina-poseidon = { path = "../poseidon" }
+o1-utils = { path = "../utils" }
+
+getrandom.workspace = true
+neon = { version = "1.0.0" }
```
dannywillems, line 35: Should we move this to the workspace?

crates/arkworks/src/group_affine.rs
```diff
@@ -15,17 +15,17 @@ use wasm_bindgen::prelude::*;
 #[wasm_bindgen]
 #[derive(Clone, Copy, Debug)]
 pub struct WasmGPallas {
-    pub x: WasmPastaFp,
-    pub y: WasmPastaFp,
-    pub infinity: bool,
+    pub(crate) x: WasmPastaFp,
```
dannywillems, line 18: I'm wondering if it has an impact on the generated JavaScript code. Did you check?

Cargo.toml
```diff
@@ -108,6 +111,7 @@ saffron = { path = "./poly-commitment", version = "0.1.0" }
 signer = { path = "./signer", version = "0.1.0" }
 turshi = { path = "./turshi", version = "0.1.0" }
 utils = { path = "./utils", version = "0.1.0" }
+wasm-types = { path = "./crates/wasm-types" }
```
dannywillems, line 114: Is there a goal to move all WebAssembly definitions also present in `kimchi` in this crate?

crates/arkworks/src/bigint_256.rs
```diff
@@ -123,7 +130,9 @@ pub fn caml_bigint_256_test_bit(x: WasmBigInteger256, i: i32) -> bool {
 
 #[wasm_bindgen]
 pub fn caml_bigint_256_to_bytes(x: WasmBigInteger256) -> Vec<u8> {
-    let mut serialized_bytes = vec![];
+    // This is the size of the implementation of BigInteger256 at the time of writing
+    // but even if it changes, we will be no worse off than the original code `vec![]`
+    let mut serialized_bytes = Vec::with_capacity(core::mem::size_of::<WasmBigInteger256>());
```
dannywillems, line 135: Should we not use `BIGINT256_NUM_BYTES`?

crates/arkworks/src/bigint_256.rs
```diff
@@ -25,7 +29,7 @@ pub struct WasmBigInteger256(pub BigInteger256);
 
 impl wasm_bindgen::describe::WasmDescribe for WasmBigInteger256 {
     fn describe() {
-        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe()
+        <Vec<u8> as wasm_bindgen::describe::WasmDescribe>::describe();
```
dannywillems, line 32: Nit: seems to be an inoffensive/useless change?

plonk-neon/Cargo.toml
dannywillems: There is a demand from different persons to update the naming conventions because it can become confusing - and it is confusing. `plonk` is too generic.

We will also make new crates/modules for Saffron related definitions, including the o1vm. Therefore, being explicit on what refers to what is important.

> I really don't think that's worth the trouble at this point.
I do not think it would create a huge trouble. It is simply renaming in different places. It can be done as a follow-up, agreed.

plonk-neon/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+edition = "2021"
+
+[lib]
+name = "plonk_neon"
+crate-type = ["cdylib"]
+
+################################# Dependencies ################################
+
+[dependencies]
+# Strictly enforcing 0.2.87
```
dannywillems, line 17: I guess it comes from a copy paste?

crates/arkworks/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
dannywillems, line 1: I would be in favor of adding a top-level description of what the crate is supposed to contain to guide the next people coming into the repository. In addition to that, we are continuously publishing the documentation and it would be useful for everyone to understand the monorepo structure.

crates/wasm-types/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#![no_std]
```
dannywillems, line 1: Same comment than [here](https://github.com/o1-labs/proof-systems/pull/3217/files#r2079218001).

plonk-neon/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+//! The `Marlin_plonk_stubs` crate exports some functionalities
```
dannywillems, line 1: This is incorrect. Can you update the description, please?

Related to [this comment](Same comment than [here](https://github.com/o1-labs/proof-systems/pull/3217/files#r2079218001).

plonk-neon/src/poseidon.rs
```diff
@@ -0,0 +1,57 @@
+use arkworks::{WasmPastaFp, WasmPastaFq};
```
dannywillems, line 1: Can you add a top-level documentation as it has been done for `plonk-wasm/src/poseidon.rs`, please? It was confusing for some people.

plonk-neon/Cargo.toml
dannywillems: > Is anyone that's working on o1js complaining about the name?

Yes.

plonk-neon/Cargo.toml
dannywillems: Not at all :)

plonk-neon/src/lib.rs
dannywillems: Agreed for the cleanup in a follow-up. I've been wanting to do it since I moved it here.

dannywillems:
> @dannywillems, do you mind if we defer the esthetic changes until after this PR lands? I left a couple of replies asking for your thoughts on semantic changes. Since this PR is quite large, I would prefer to fix all semantic issues, then do a cleanup pass immediately following this.

Not at all.

crates/arkworks/src/group_affine.rs
dannywillems: It seems to be used [here](https://github.com/o1-labs/o1js/blob/main/src/bindings/crypto/bindings/conversion-base.ts#L81).

crates/arkworks/src/pasta_fq.rs
```diff
@@ -0,0 +1,75 @@
...
+use wasm_bindgen::convert::{FromWasmAbi, IntoWasmAbi, OptionFromWasmAbi, OptionIntoWasmAbi};
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct WasmPastaFq(pub Fq);
+
+impl wasm_types::FlatVectorElem for WasmPastaFq {
+    const FLATTENED_SIZE: usize = core::mem::size_of::<Fq>();
+    fn flatten(self) -> Vec<u8> {
+        let mut bytes: Vec<u8> = Vec::with_capacity(Self::FLATTENED_SIZE);
```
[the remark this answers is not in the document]
dannywillems: Same comment than [here](https://github.com/o1-labs/proof-systems/pull/3217#discussion_r2089529495). It happened in some codebase that the field element was a pointer, and not the direct value, and therefore was giving a wrong size.

No requested changes. Mostly a comment.

plonk-neon/src/poseidon.rs
dannywillems: > If we do want references to code in `mina`, they should at the very least be github permalinks.

Agreed.

dannywillems: > Is it alrigth if I just put
> 
> ```rust
> //! This file defines wrappers for the Poseidon hash functions that are used in
> //! the [Mina](https://github.com/MinaProtocol/mina) codebase.
> ```
> 
> All the extra stuff
> 
> ```rust
> //! It is a wrapper around the Poseidon implementation in the `mina_poseidon` crate.
> //! It is required as the native OCaml implementation of Mina does use the Rust
> //! implementation defined in the crate `mina_poseidon` instead of defining its
> //! own natively in OCaml for performance reasons. The bindings in OCaml can be
> //! found in `src/lib/crypto/kimchi_bindings/pasta_fp_poseidon` and
> //! `src/lib/crypto/kimchi_bindings/pasta_fq_poseidon` in the Mina codebase.
> ```
> 
> is redundant and impossible to programmatically assure that the code locations stay up to date. We've already seen in this PR that documentation doesn't get kept up to date when referring to other code.

Yes sure.

dannywillems (approved):

dannywillems:
It would be nice to also have a way to add a target in the Makefile to build the project and add a job in the CI.

dannywillems (approved):
Approving again after master has been merged.

dannywillems:
Rebased and hope it will get merged this time.

PR #3228, opened 2025-05-13
Move precomputed_srs to poly-commitment

volhovm:
Before, `precomputed_srs` was in `kimchi` although it's not a `kimchi`-specific module. We might want to use precomputed srs it for `saffron` too (for testing, it saves time on generating srs+lagrange bases). 

This PR moves it to `poly-commitment`, creating a `NamedCurve` trait in `mina_curves`.


volhovm:
@dannywillems I'm suspecting `NAME` is not used in `mina` because we'd need bindings for that... not 100% sure but I think the chances are quite low so let's proceed and if it breaks I'll fix it asap.

(I tried to upgrade, but nix did not work out of the box in `mina` :( )

PR #3230, opened 2025-05-13
Revert "Cargo: remove config.toml defining alias"

richardpringle:
I don't think the alias actually causes any problems and is convenient for local development. We just can't use the alias in the makefile

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run --package xtask --"


dannywillems:
> I don't think the alias actually causes any problems and is convenient for local development. We just can't use the alias in the makefile

The issue was caused on the Mina side. It should be simple to solve it though.

dannywillems (approved):
LGTM. Agreed it is ok if we do not change in the Makefile as Mina will keep using the Makefile.

PR #3233, opened 2025-05-16
PU: Add benches for folding

volhovm:
Adds benchmarks for folding: interactive prover/verifier + plonkish prover/verifier for the relaxed statement.

Bench results on my machine (may be noisy):
```
folding_ip/folding_prover
                        time:   [37.932 ms 38.138 ms 38.363 ms]
folding_ip/folding_verifier
                        time:   [499.56 µs 500.24 µs 500.96 µs]
folding_ip/prover_relaxed
                        time:   [732.84 ms 736.66 ms 742.93 ms]
folding_ip/verifier_relaxed
                        time:   [80.869 ms 81.115 ms 81.389 ms]
```
where the first two are aggregation functions, and the last two are plonking proof prover/verifier.

### DIFF
--- a/saffron/benches/folding_bench.rs
+++ b/saffron/benches/folding_bench.rs
@@ -34,7 +34,7 @@ fn bench_folding(c: &mut Criterion) {
 
     assert!(relaxed_instance_2.check_in_language(&srs, domain.d1, &relaxed_witness_2));
 
-    let (relaxed_instance_3, relaxed_witness_3, error_term_1) = folding_prover(
+    let (relaxed_instance_3, relaxed_witness_3, cross_term_1) = folding_prover(
         &srs,
         domain.d1,
         &core_instance_1,
@@ -46,7 +46,7 @@ fn bench_folding(c: &mut Criterion) {
     assert!(relaxed_instance_3.check_in_language(&srs, domain.d1, &relaxed_witness_3));
 
     assert!(
-        folding_verifier(&core_instance_1, &relaxed_instance_2, error_term_1) == relaxed_instance_3
+        folding_verifier(&core_instance_1, &relaxed_instance_2, cross_term_1) == relaxed_instance_3
     );
 
     group.bench_function("folding_prover", |b| {
@@ -71,7 +71,7 @@ fn bench_folding(c: &mut Criterion) {
             || {
                 let (core_instance_4, core_witness_4) =
                     generate_random_inst_wit(&srs, domain, &mut rng);
-                let (_, _, error_term_2) = folding_prover(
+                let (_, _, cross_term_2) = folding_prover(
                     &srs,
                     domain.d1,
                     &core_instance_4,
@@ -79,13 +79,13 @@ fn bench_folding(c: &mut Criterion) {
                     &relaxed_instance_3,
                     &relaxed_witness_3,
                 );
-                (core_instance_4, error_term_2)
+                (core_instance_4, cross_term_2)
             },
-            |(core_instance_4, error_term_2)| {
+            |(core_instance_4, cross_term_2)| {
                 black_box(folding_verifier(
                     &core_instance_4,
                     &relaxed_instance_3,
-                    error_term_2,
+                    cross_term_2,
                 ))
             },
             BatchSize::LargeInput,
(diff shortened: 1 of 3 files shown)


saffron/benches/folding_bench.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let srs = poly_commitment::precomputed_srs::get_srs_test();
+    let group_map = <Vesta as CommitmentCurve>::Map::setup();
+
+    let domain: EvaluationDomains<ScalarField> =
+        EvaluationDomains::<ScalarField>::create(srs.size()).unwrap();
+
+    let (core_instance_1, core_witness_1) = generate_random_inst_wit(&srs, domain, &mut rng);
+    let (core_instance_2, core_witness_2) = generate_random_inst_wit(&srs, domain, &mut rng);
+    let relaxed_instance_2 = core_instance_2.relax();
```
marcbeunardeau88, line 32: Shouldn't you generate a random relaxed instance ?
By using .relax() you get a specific case, with u=1 and e =0 which can mess with the bench

saffron/benches/folding_bench.rs
```diff
@@ -0,0 +1,142 @@
...
+    let relaxed_witness_2 = core_witness_2.relax(domain.d1);
+
+    assert!(relaxed_instance_2.check_in_language(&srs, domain.d1, &relaxed_witness_2));
+
+    let (relaxed_instance_3, relaxed_witness_3, error_term_1) = folding_prover(
```
marcbeunardeau88, line 37: Maybe add a comment saying this creates a pseudo random relaxed instance

saffron/benches/folding_bench.rs
```diff
@@ -0,0 +1,142 @@
...
+    assert!(relaxed_instance_3.check_in_language(&srs, domain.d1, &relaxed_witness_3));
+
+    assert!(
+        folding_verifier(&core_instance_1, &relaxed_instance_2, error_term_1) == relaxed_instance_3
+    );
+
+    group.bench_function("folding_prover", |b| {
+        b.iter_batched(
+            || generate_random_inst_wit(&srs, domain, &mut rng),
+            |(core_instance_4, core_witness_4)| {
```
marcbeunardeau88, line 55: The different bench only change in the core instance. This could be changed by adding a relaxed instance generator, but I'd be surprise if the result of the bench changes because of that.

marcbeunardeau88 (approved):

saffron/src/folding.rs
```diff
@@ -51,22 +51,22 @@ use rand::{CryptoRng, RngCore};
...
+    pub comm_a: Curve,
 }
 
 #[derive(PartialEq, Eq)]
-/// Relaxed instance variant.
+/// Relaxed instance variant, attesting to `d * q - a * u - e = 0`
 pub struct RelaxedInstance {
     /// Non-relaxed part
-    core: CoreInstance,
+    pub core: CoreInstance,
```
dannywillems, line 65: I would be in favor of not making fields public, and instead use a new method.

saffron/src/folding.rs
```diff
@@ -51,22 +51,22 @@ use rand::{CryptoRng, RngCore};
 /// Non-relaxed instance attesting to `d * q - a = 0`
 pub struct CoreInstance {
     /// Commitment to the data
-    comm_d: Curve,
+    pub comm_d: Curve,
```
dannywillems, line 54: Same than below. I would be in favor of not making the fields public, and keep a clean user API.

saffron/src/folding.rs
```diff
@@ -83,17 +83,17 @@ impl CoreInstance {
 /// Non-relaxed witness contains evaluations (field vectors) for data,
 /// query, and answers.
 pub struct CoreWitness {
-    d: Evaluations<ScalarField, R2D<ScalarField>>,
-    q: Evaluations<ScalarField, R2D<ScalarField>>,
-    a: Evaluations<ScalarField, R2D<ScalarField>>,
+    pub d: Evaluations<ScalarField, R2D<ScalarField>>,
```
dannywillems, line 86: Same as above.

dannywillems (approved):
LGTM, modulo comments. It would be nice to have them in a follow-up. I prefer to not block.

saffron/src/folding.rs
volhovm: :thinking: How is creating a method with the same name "cleaner" in terms of public API? I guess what you want to avoid is to minimise API changes, but if we want to change the name of the field we would also want to change the name of the method, so it's the same... the `pub/pub(crate)` accessors are the same for fields and methods as far as I know.

Taking this example to maximum: we could follow java style and wrap every field into `set_field()` and `get_field()`, together with creating a `constructor()`, but I really don't see what it achieves in terms of maintainability (while it adds _a lot_ of visual noise).

PR #3242, opened 2025-05-30
Deps: Removing serde_json exact version pin, replaced with semver caret.

jk89:
Motivation:

Proof-systems needs to co-exist with other rust libraries. Pinning the version of serde_json (a very common library) to a specific (older) version is very restrictive. Requesting it be unpinned and replaced with caret.


dannywillems (approved):
Thanks! I would remove the strict condition on serde at the same time.

PR #3246, opened 2025-06-04
Deps: bump up arkworks to 0.5

dannywillems:
See changelog: https://github.com/arkworks-rs/algebra/blob/master/CHANGELOG.md

See https://github.com/arkworks-rs/algebra/pull/850.

This is a path forward supporting more recent Rust version, and also bumping ocaml-rs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,6 +56,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.21"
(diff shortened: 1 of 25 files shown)


volhovm (approved):
Looks good to me, the update is not introducing any big breaking changes except for renamings / refactorings indeed! :+1:

dannywillems:
Note: it is to progress towards fixing https://github.com/o1-labs/proof-systems/issues/3254.

Issue #3255, opened 2025-06-07
tikv dependencies break sp1 builder


Kirol54:
We currently need to use a fork due to an issue with tikv package. (you can see the error here https://github.com/jk89/proof-systems/commit/822970c1916a27b049fa7d81f820327e02c1fc85#commitcomment-158528116 )

 https://github.com/jk89/proof-systems/commit/822970c1916a27b049fa7d81f820327e02c1fc85 

we need to compile o1-utils in SP1 zkVM code which target is:
 TARGET=riscv32im-succinct-zkvm-elf

so I guess this macro doesn't catch our particular target
https://github.com/o1-labs/proof-systems/blob/master/utils/Cargo.toml#L12
https://github.com/o1-labs/proof-systems/blob/master/utils/src/lazy_cache.rs#L267

dannywillems:
I also open a nice to have follow-up: https://github.com/o1-labs/proof-systems/issues/3259

dannywillems:
@Kirol54 can you give the commands you run to compile for SP1? It would help to add a CI job for this.

querolita:
Hi!

Because this `riscv32im-succinct-zkvm-elf` is not a standard target, I cannot just add it to the macro as done with `wasm32`. 

I am working now on a custom feature to only include those dependencies when both the architecture is not `wasm32` (as currently done) and the feature is not active. That would mean you would not need to change your build command. I will go back to this conversation when I have a PR that is ready.

querolita:
Created draft PR. Waiting on CI. When all green, I will ping you in case you can try to run it on your end @Kirol54 to make sure it fixes it before merging.

https://github.com/o1-labs/proof-systems/pull/3261

querolita:
The PR is now ready for review @Kirol54

PR #3261, opened 2025-06-10
Add custom feature to enable jemalloc dependencies

querolita:
This PR adds a custom feature in `kimchi` and `o1-utils` called `diagnostics` that should be passed whenever we want to include the dependencies `tikv-jemalloc-ctl` and `tikv-jemallocator`. That means, tests using `jemalloc` to measure memory allocation will only perform the memory measurements if both run with `--features diagnostics` and the target architecture is not `wasm32` (this last condition was added in https://github.com/o1-labs/proof-systems/pull/3079 to pass CI). 

The reason I am doing this is that tikv packages seem to have issues with the custom build target `riscv32im-succinct-zkvm-elf` used in SP1. By adding an optional cargo feature we can still run those tests for internal benchmarking while preserving compatibility with external teams without needing them to change their current build commands. Otherwise they need to work on a fork of `proof-systems` that comments out the parts of the code using `jemalloc`-related functions.

Closes https://github.com/o1-labs/proof-systems/issues/3255

### DIFF
--- a/kimchi/Cargo.toml
+++ b/kimchi/Cargo.toml
@@ -14,8 +14,8 @@ path = "src/lib.rs"
 bench = false       # needed for criterion (https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options)
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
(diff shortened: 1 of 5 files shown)


Kirol54:
just compiled without any issues, many thanks for sorting that !

dannywillems:
Rebased + auto-merge activated.

PR #3271, opened 2025-06-26
CI: adding check that we do not infringe copyright licenses

dannywillems:


glyh:
Does it work with TS/OCaml?

PR #3291, opened 2025-07-10
feat: implement serde for ROInput

scaraven:
## Overview
This PR introduces `serialize` and `deserialize` functions for the `ROInput` structure within `hasher/roinput.rs`.

## Motivation
Currently, `ROInput` does not allow (de)serialization which decreases compatibility with applications which require byte array representations of Hashable structures. For example, the [following project](https://github.com/Raspberry-Devs/mina-multi-sig).

## Changes
Implement `serialization` and `deserialization` and additional tests which allow 1 to 1 conversion between byte arrays and `ROInput`. The main reason why I have not implemented a `from_bytes` method is due to the fact that `to_bytes` is not injective (i.e. different ROInput representations may produce the same byte array which makes decoding ambigious).


hasher/src/roinput.rs
```diff
@@ -3,13 +3,18 @@
 //! Definition of random oracle input structure and
 //! methods for serializing into bytes and field elements
 
+use core::fmt::Error;
```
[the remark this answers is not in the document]
dannywillems: No, we can ignore I think this suggestion from Copilot. We don't want to library to use `std` (requirement for DFST).

hasher/src/roinput.rs
```diff
@@ -172,6 +177,66 @@ impl ROInput {
...
+
+        // Check that the number of bytes is consistent with the expected lengths
+        let expected_len_bits = fields_len * Fp::MODULUS_BIT_SIZE as usize + bits_len;
+        // Round up to nearest multiple of 8
+        let expected_len = (expected_len_bits + 7) / 8 + SER_HEADER_SIZE;
+        if input.len() != expected_len {
+            return Err(Error);
+        }
+
+        // allocate space for exactly `fields_ctr` elements
```
querolita, line 220:
```suggestion
        // allocate space for exactly `fields_len` elements
```

hasher/src/roinput.rs
```diff
@@ -909,4 +974,198 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn serialize_multiple_bools_length() {
+        for i in 0..1024 {
+            let roi = ROInput::new().append_bool(i % 2 == 0);
+            let serialized = roi.serialize();
+
+            // Deserualize and check if it matches
```
querolita, line 1067:
```suggestion
            // Deserialize and check if it matches
```

hasher/src/roinput.rs
```diff
@@ -3,13 +3,18 @@
...
+
 use super::Hashable;
 use alloc::{vec, vec::Vec};
-use ark_ff::PrimeField;
+use ark_ff::{BigInteger, PrimeField};
 use bitvec::{prelude::*, view::AsBits};
 use mina_curves::pasta::{Fp, Fq};
 use o1_utils::FieldHelpers;
 
+const SER_HEADER_SIZE: usize = 8;
```
querolita, line 15:
```suggestion
const SER_HEADER_SIZE: usize = 8; // total number of bytes for the header of the serialized ROInput
```

hasher/src/roinput.rs
```diff
@@ -3,13 +3,18 @@
...
+const SINGLE_HEADER_SIZE: usize = 4;
```
querolita, line 16:
```suggestion
const SINGLE_HEADER_SIZE: usize = 4; // number of bytes for each part of the header of the serialized ROInput
```

querolita (approved):
Great contribution! I left a few suggestions regarding comments. Other than that, it looks good to me

dannywillems (approved):
Blindly approving for @querolita.

PR #3302, opened 2025-07-28
feat: add ocaml compatible nonce derivation function

boray:


dannywillems:
Nice, thanks!
Before reviewing, could you remove the merge commits and squash the format and linting commits into the first commit? We would like to enforce the policy that each commit should pass the CI and we can easily revert on-demand.

dannywillems:
On the Mina side, we should check the zkApp command implementation.

dannywillems (changes requested):
See above.

hasher/src/roinput.rs
```diff
@@ -122,6 +122,30 @@ impl ROInput {
         self.append_bytes(&x.to_le_bytes())
     }
 
+    /// Convert the random oracle input to a vector of packed field elements
+    pub fn to_packed_fields(&self) -> Vec<Fp> {
```
dannywillems, line 126: Can you describe what "packed fiel elements" means? It is not obvious.

signer/tests/signer.rs
```diff
@@ -255,3 +256,68 @@ fn sign_delegation_test_4() {
...
+        fields: vec![BaseField::from(1), BaseField::from(2), BaseField::from(3)],
+    };
+
+    let mut testnet_ctx = mina_signer::create_kimchi::<Input>(NetworkId::TESTNET);
+    let mut mainnet_ctx = mina_signer::create_kimchi::<Input>(NetworkId::MAINNET);
+
+    let testnet_sig = testnet_ctx.sign(&kp, &input, true);
+    let mainnet_sig = mainnet_ctx.sign(&kp, &input, true);
+
+    assert_eq!(
```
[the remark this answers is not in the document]
dfstio: They are from the o1js TypeScript code (for testnet)
```
const privateKey = PrivateKey.fromBase58(
      "EKFXH5yESt7nsD1TJy5WNb4agVczkvzPRVexKQ8qYdNqauQRA8Ef"
    );
const signature = Signature.create(privateKey, [1n, 2n, 3n].map(Field));
 console.log("signature:", signature.toJSON());
```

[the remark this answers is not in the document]
dannywillems: Thanks @dfstio. Can you add a small comment in the codebase @boray please? It is always good to keep track of where the values come from.

signer/src/schnorr.rs
```diff
@@ -58,8 +59,11 @@ impl<H: Hashable> Hashable for Message<H> {
 }
 
 impl<H: 'static + Hashable> Signer<H> for Schnorr<H> {
-    fn sign(&mut self, kp: &Keypair, input: &H) -> Signature {
-        let k: ScalarField = self.derive_nonce(kp, input);
+    fn sign(&mut self, kp: &Keypair, input: &H, packed: bool) -> Signature {
```
dannywillems, line 62: Note: it should be able to reproduce the same behavior as [signFields](https://github.com/o1-labs/o1js/blob/main/src/mina-signer/mina-signer.ts#L148) for @dfstio. Note that "devnet" is hardcoded for the network.

As part of this patch, we need to ensure that the exposed Rust API allow the user to reproduce the same signature as the one in mina-signer.

hasher/src/roinput.rs
```diff
@@ -98,10 +98,8 @@ impl ROInput {
 
     /// Append a scalar field element
     pub fn append_scalar(mut self, s: Fq) -> Self {
```
dannywillems, line 100: Changing the implementation of this method implies introducing a breaking changes in the legacy signature, and in particular it implies we must change the test vectors for the signature algorithm in 3c6c4aa5e0759cf37a00a520037db4a01e320e24.
This is not a behavior we would like to introduce. Can you create a separate method instead please?

signer/src/schnorr.rs
```diff
@@ -176,7 +176,7 @@ impl<H: 'static + Hashable> Schnorr<H> {
             .to_roinput()
             .append_field(kp.public.point().x)
             .append_field(kp.public.point().y)
-            .append_scalar(kp.secret.scalar().clone())
+            .append_scalar(*kp.secret.scalar())
```
dannywillems, line 179: Can you squash this with the commit actually introducing this change, please? We want to avoid this kind of split commits to ease reverts if necessary.

hasher/src/roinput.rs
```diff
@@ -327,14 +327,10 @@ mod tests {
             Fq::from_hex("18b7ef420128e69623c0c0dcfa28d47a029d462720deb769d7b5dd6f17444216")
                 .expect("failed to create scalar");
         let roi = ROInput::new().append_scalar(scalar);
-        assert_eq!(roi.bits.len(), 255);
+        assert_eq!(roi.fields.len(), 1);
```
dannywillems, line 331: As mentioned earlier, it would be better to not change the actual behavior of append_scalar, and add a different method instead. Both methods should be documented.

dannywillems: As a consequence, these tests should not be updated.

hasher/src/roinput.rs
```diff
@@ -445,11 +446,13 @@ mod tests {
 
     #[test]
     fn all_1() {
-        let scalar1 = Fq::from_hex("01d1755db21c8cd2a9cf5a3436178da3d70f484cd4b4c8834b799921e7d7a102")
-            .expect("failed to create scalar");
-        let scalar2 = Fq::from_hex("e70187e9b125524489d0433da76fd8287fa652eaebde147b45fa0cd86f171810")
-            .expect("failed to create scalar");
-        
+        let scalar1 =
```
dannywillems, line 449: As for the commits applying the linter, we should squash this commit the one introducing the actual changes. Formatting and linting should be done after each commit.

dannywillems:
Taking over this work now.

hasher/src/roinput.rs
```diff
@@ -388,6 +431,48 @@ mod tests {
...
+            reconstructed_bytes[31] & last_byte_mask,
+            expected_bytes[31] & last_byte_mask
+        );
+
+        // Test serialization to bytes
+        let serialized_bytes = roi.to_bytes();
+        assert_eq!(serialized_bytes.len(), 32); // 255 bits rounded up to 32 bytes
+
+        // Test that max scalar converts to proper field elements
+        let fields = roi.to_fields();
```
dannywillems, line 463: Because to_fields uses 255 - 1 bits.

dannywillems:
I kept the git history and reverted commits @boray added to keep contributions to the repository. I reverted because I wanted to suggest a simpler solution (see commits after). The commits add features and remove them after to keep @boray contributions.

signer/src/schnorr.rs
```diff
@@ -111,9 +114,163 @@ pub(crate) fn create_kimchi<H: 'static + Hashable>(domain_param: H::D) -> impl S
...
+                let modulus = BaseField::MODULUS.into();
+                if secret_biguint >= modulus {
+                    // Reduce modulo base field modulus
+                    let reduced_biguint: BigUint = secret_biguint - modulus;
+                    BaseField::from_biguint(&reduced_biguint)
+                        .expect("Reduced bigint should fit in base field")
+                } else {
+                    BaseField::from_biguint(&secret_biguint)
+                        .expect("Scalar bigint should fit in base field")
+                }
```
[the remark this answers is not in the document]
dannywillems: It is fine as the scalar field modulus is slightly over the base field modulus. Resolving comment.

querolita (dismissed):
I checked the implementation matches the description, and the compatibility tests match those in https://github.com/MinaProtocol/mina/pull/17688/files. Great additions to the documentation of this library.

signer/src/signature.rs
```diff
@@ -31,3 +132,68 @@ impl fmt::Display for Signature {
...
+mod tests {
+    use super::*;
+    use alloc::{format, string::ToString};
+
+    #[test]
+    fn test_signature_encoding() {
+        // Test with small values to verify big-endian encoding and rx || s
+        // order
+        let rx = BaseField::from(4u64);
+        let s = ScalarField::from(42u64);
```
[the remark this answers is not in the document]
dannywillems: https://www.youtube.com/watch?v=aboZctrHfK8

dannywillems (approved):

PR #3350, opened 2025-10-21
[DNM] [native-prover] feat: canary

bleepbloopsify:
adding a canary value so we can ensure the e2e flow works

### DIFF
--- a/plonk-napi/src/poseidon.rs
+++ b/plonk-napi/src/poseidon.rs
@@ -7,7 +7,7 @@ use wasm_types::{FlatVector, FlatVectorElem};
 
 // fp
 
-#[napi]
+#[napi(js_name = "caml_pasta_fp_poseidon_block_cipher")]
 pub fn caml_pasta_fp_poseidon_block_cipher(state: Uint8Array) -> Result<Uint8Array> {
     println!("from native rust");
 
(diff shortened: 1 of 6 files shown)


plonk-napi/src/poseidon.rs
```diff
@@ -7,7 +7,7 @@ use wasm_types::{FlatVector, FlatVectorElem};
 
 // fp
 
-#[napi]
+#[napi(js_name = "caml_pasta_fp_poseidon_block_cipher")]
```
querolita: Mmmm actually is this the right format or should it be camel case in the napi annotation?

bleepbloopsify: this tells `napi` what name to use when exporting to js-land

querolita: I was asking because when I was doing the plonk-wasm overrides for the first time, I had to create this `snakeToCamel` [function](https://github.com/MinaProtocol/mina/blob/d4ff2e968c9b3239e99f9c414afe2a24ac855a1d/src/lib/crypto/kimchi_bindings/js/node_js/node_backend.js#L17-L21) in `node_backend.js` so that the function names coming from rust (in snake case) were mapped to camel case. So that makes me wonder if we could replace that whole function by just using `#[napi(js_name = "camlPastaFpPoseidonBlockCipher")]` in there.

bleepbloopsify: yeah, we don't need the snakeToCamel function if we do this here

Napi chooses to "snakeCase" your rust symbols by default, if we explicitly tell `napi` what name to give it, it will end up with this one guaranteed

PR #3351, opened 2025-10-22
(Native prover) prover index

Trivo25:
https://github.com/MinaProtocol/mina/pull/17993
https://github.com/o1-labs/proof-systems/pull/3351


bleepbloopsify (approved):
Looks fine overall, (interesting serde in plonk_wasm

Should probably move towards the better names, but not blocking, we can revise that later, of course

PR #3357, opened 2025-10-27
chore: shuffle around type declarations so native prover generation works better

bleepbloopsify:
Bumping to 3.0+ napi because otherwise typegen doesn't work right

added pass-by-ref to `External`, otherwise napi doesn't compile right (this was fixed in 3.0+, we shouldn't be passing them by copy tbh

moved prover_index functions outside of the types.rs, since types.rs uses internal `Result`, and we try to rename napi_bindgen::prelude::Result to NapiResult, but the typegeneration script looks for explicitly `Result<T>`, it breaks the typegeneration. Bad library design

### DIFF
--- a/plonk-napi/src/types.rs
+++ b/plonk-napi/src/types.rs
@@ -17,7 +17,7 @@ struct SerializedProverIndex {
 
 // TOOD: remove incl all dependencies when no longer needed and we only pass napi objects around
 impl WasmPastaFpPlonkIndex {
-    pub fn serialize_inner(&self) -> Result<Vec<u8>, String> {
+    pub(crate) fn serialize_inner(&self) -> Result<Vec<u8>, String> {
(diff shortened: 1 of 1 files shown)


plonk-napi/src/types.rs
```diff
@@ -18,7 +17,7 @@ struct SerializedProverIndex {
 
 // TOOD: remove incl all dependencies when no longer needed and we only pass napi objects around
 impl WasmPastaFpPlonkIndex {
-    fn serialize_inner(&self) -> Result<Vec<u8>, String> {
+    pub fn serialize_inner(&self) -> Result<Vec<u8>, String> {
```
querolita: I think `pub(crate)` should be enough

querolita (approved):
Looks good to me, also it makes more sense if we have prover index from/to bytes inside prover index itself...

PR #3358, opened 2025-10-28
fix: wasm build on macOS

bleepbloopsify:
RUSTFLAGS was being applied to each step of the compilation process, so on slightly different linkers, libc was being compiled with --max-memory (which macOS hated).

This moves the flags to only be applied to `wasm32-unknown-unknown` targets so we don't pollute our flags (or our logs)

This also enables us to build clean from a mac without a strange first-time-only error

https://github.com/MinaProtocol/mina/pull/18034

Historically, `.cargo/config.toml` wasn't respected by `dune` because `.blah` folders are ignored by `source_tree`. I fixed that in `mina`, so `.cargo` is now respected by `mina`. (We can add others if necessary).

This is without adding `dune` files to `proof-systems`, which I believe is the point


plonk-wasm/config.toml
bleepbloopsify: I don't believe this file was applied anywhere, since cargo only respects `.cargo/config.toml` files, not `config.toml` files sitting in random project directories

.github/workflows/wasm.yml
```diff
@@ -10,12 +10,12 @@ on:
 jobs:
   build-wasm:
     name: Build WebAssembly code
-
-    runs-on: ["ubuntu-latest"]
-
     strategy:
       matrix:
-        rust_toolchain_version: ["nightly-2024-09-05"]
+        os: [ubuntu-latest, macos-15]
```
dannywillems: Nit: would be better to use `macos-latest`

PR #3367, opened 2025-11-11
Update Verifier Index napi conversion

Shigoto-dev19:
This PR adds the `From` and `FromNapiValue` trait implementation for some of the strcuts in the verifier index napi conversion; preparing for the oracles napi conversion PR (https://github.com/o1-labs/proof-systems/pull/3364).

The changes include updating the verifier index to use `NapiPasta` instead of `Vec<u8>` as well as `NapiSrs`.
It adds `From` trait implementation for  `LookupSelectors`, `LookupVerifierIndex`, and `NapPlonkVerifierIndex`/`DlogVerifierIndex` for what's needed in the oracles conversion PR :point_down: 
https://github.com/o1-labs/proof-systems/blob/d198659190d211b46e1d464c748553cd822df91a/plonk-napi/src/oracles.rs#L169

**Note**: We might need to use the macros again to refacor the napi verifier index fp/fq modules.

### DIFF
--- a/plonk-napi/src/plonk_verifier_index/fq.rs
+++ b/plonk-napi/src/plonk_verifier_index/fq.rs
@@ -225,88 +225,79 @@ pub fn caml_pasta_fq_plonk_verifier_index_shifts(
 
 impl From<NapiFqPlonkVerifierIndex> for DlogVerifierIndex<GAffine, OpeningProof<GAffine>> {
     fn from(index: NapiFqPlonkVerifierIndex) -> Self {
-        of_napi(index).0
+        let max_poly_size = index.max_poly_size;
+        let public_ = index.public_;
+        let prev_challenges = index.prev_challenges;
+        let log_size_of_group = index.domain.log_size_of_group;
+        let srs = &index.srs;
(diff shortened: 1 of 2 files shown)


plonk-napi/src/plonk_verifier_index/fq.rs
```diff
@@ -96,20 +213,138 @@ pub fn caml_pasta_fq_plonk_verifier_index_shifts(
...
+    let prev_challenges = index.prev_challenges;
+    let log_size_of_group = index.domain.log_size_of_group;
+    let srs = &index.srs;
+    let evals = &index.evals;
+    let shifts = &index.shifts;
+
+    let (endo_q, _endo_r) = poly_commitment::ipa::endos::<GAffineOther>();
+    let domain = Domain::<Fq>::new(1 << log_size_of_group).unwrap();
+
+    let feature_flags = compute_feature_flags(&index);
```
querolita: here instead, because it is quite long, it's okay to modularize it. Also, we might switch to macro-style later, avoiding having both fp and fq files.

PR #3371, opened 2025-11-13
Napi conversions of proof-related types

Shigoto-dev19:
Closes https://github.com/o1-labs/o1js/issues/2487.


plonk-napi/src/proof.rs
```diff
@@ -40,22 +41,491 @@ macro_rules! impl_proof {
...
+                }
+            }
+
+            impl FromNapiValue for [<Napi $field_name:camel ProofEvaluations>] {
+                unsafe fn from_napi_value(
+                    env: sys::napi_env,
+                    napi_val: sys::napi_value,
+                ) -> Result<Self> {
+                    let instance = <ClassInstance<[<Napi $field_name:camel ProofEvaluations>]> as FromNapiValue>::from_napi_value(env, napi_val)?;
+                    Ok((*instance).clone())
```
querolita: I wonder if this clone is necessary. It looks like it is being created twice then. This is just a performance concern, if this can be done more efficiently we can address it in a follow-up

plonk-napi/src/proof.rs
```diff
@@ -40,22 +41,491 @@ macro_rules! impl_proof {
...
+                }
+            }
+
+            impl FromNapiValue for [<Napi $field_name:camel LookupCommitments>] {
+                unsafe fn from_napi_value(
+                    env: sys::napi_env,
+                    napi_val: sys::napi_value,
+                ) -> Result<Self> {
+                    let instance = <ClassInstance<[<Napi $field_name:camel LookupCommitments>]> as FromNapiValue>::from_napi_value(env, napi_val)?;
+                    Ok((*instance).clone())
```
querolita: Similar here.

plonk-napi/src/proof.rs
```diff
@@ -40,22 +41,491 @@ macro_rules! impl_proof {
...
+                #[napi(setter, js_name="set_prev_challenges_comms")]
+                pub fn set_prev_challenges_comms(&mut self, prev_challenges_comms: NapiVector<$NapiPolyComm>) {
+                    self.prev_challenges_comms = prev_challenges_comms
+                }
+
+                #[napi]
+                #[allow(deprecated)]
+                pub fn serialize(&self) -> String {
+                    let (proof, _public_input) = self.into();
+                    let serialized = rmp_serde::to_vec(&proof).unwrap();
```
querolita: Reminds me of the Nori bug https://github.com/o1-labs/proof-systems/pull/3369/, let's keep an eye open with these ones 👀

plonk-napi/src/proof.rs
Shigoto-dev19: That's a good point, but I wonder how it can be done more efficiently :thinking: 
Because even for structs that are annotated with `#[napi(object)]`, they are also cloned and that gives me the idea that a `FromNapiValue` trait implementation for structs requires using clone!

image

PR #3373, opened 2025-11-25
Remove IVC create: it is unused and will never be used

dannywillems:
The code is old. It was supposed to be used for the o1vm. But the version using this IVC crate has been abandoned, and it will never be revamped.

Fix https://github.com/o1-labs/mina-rust/issues/1660


querolita (approved):
Sad 🫠

PR #3377, opened 2025-12-02
fix(curves): allow asm cfg from ark-ff derive macros

dannywillems:


curves/Cargo.toml
```diff
@@ -9,6 +9,11 @@ readme = "../README.md"
 edition = "2021"
 license = "Apache-2.0"
 
+[lints.rust]
```
iostat: Just curious on the motivation for adding this as a lint exemption as opposed to explicitly listing the features in the crate manifest. My understanding is that derive macros are producing these feature-gates which aren't explicitly listed and the linter isn't pleased with that. But from a tooling perspective, the lint is absolutely correct in that code is being produced which may never be enabled because the corresponding `feature`s aren't defined.

I believe the following would be the idiomatic way to resolve the warning which preserves the lint's intent.
```toml
[features]
asm = []
```

PR #3386, opened 2025-12-15
rp/poseidon arrays

richardpringle:
The main change of this PR is the `ArithmeticSpongeParams` type (found [here](https://github.com/o1-labs/proof-systems/blob/44409d01d4ab763f6cba37fbd9d53637c3355bfc/poseidon/src/poseidon.rs#L50)). 

What used to be a `Vec<Vec<F>>` is now `[[F; FULL_ROUNDS]]` where `FULL_ROUNDS` is a const generic of type `usize`. This means the data can now sit on either the stack or the heap, and is contiguous rather than spread across the heap. 

Making this change had wide spread affects to the all the dependent functions, types and traits. Now many of them are generic over a `usize`. There's really only one file that had logical changes and that's `poseidon/src/permutation.rs`. Extra allocations were removed and the code was cleaned up a little to now handle the arrays instead of `Vec`s. 

I benchmarked locally and there is small consistent speed up (though that was not the goal).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### [kimchi](./kimchi)
 
 #### Changed
-- (No changes in current release)
+- Update `KimchiCurve` trait to be generic over `const FULL_ROUNDS: usize`.
+- Update `verify_poseidon`, `generate_witness` and other sponge-dependent functions to be generic over the number of full rounds.
+  ([#3386](https://github.com/o1-labs/proof-systems/pull/3386))
 
 ### [kimchi-msm](./msm)
 
@@ -58,7 +60,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### [mina-poseidon](./poseidon)
 
 #### Changed
-- (No changes in current release)
+- Update `ArithmeticSpongeParams` to be generic over `const FULL_ROUNDS: usize`, replacing `Vec` fields with fixed-size arrays.
+- Update `Sponge` trait and `ArithmeticSponge` struct to be generic over the number of full rounds.
+  ([#3386](https://github.com/o1-labs/proof-systems/pull/3386))
 
 ### [mina-signer](./signer)
 


dannywillems:
Next steps:
- check benchmarks
  - this is the first thing to check because the hash function is a critical part of the whole protocol. If we get worse performances, we should abort this task immediately. However, there is a small probability that we get something worse. If we get better performances, we should absolutely documented it, and the CHANGELOG should contain the benchmarks, in addition to this PR.
- check if updating `MinaProtocol/Mina` with this version doesn't break
- as well as o1js
- fix the different lints
- clean the history
- rename `ROUNDS` into `FULL_ROUNDS`

Probably as a follow-up:
- add a new type parameter `PARTIAL_ROUNDS` as a Poseidon instance contains two types of rounds, so-called "partial" and "full". Our current Poseidon instance do not use any partial rounds, but if we change it in the future, we would need it. As discussed, if the parameter is not used right now, we should not have it in the codebase now.

kimchi-stubs/src/oracles.rs
richardpringle: Constant of `55` is always used. If you try to use a different number with `Pallas` or `Vesta` you'll get compiler errors

kimchi-stubs/src/pasta_fp_plonk_index.rs
```diff
@@ -160,9 +161,7 @@ pub fn caml_pasta_fp_plonk_index_read(
     }
 
     // deserialize the index
-    let mut t = ProverIndex::<Vesta, OpeningProof<Vesta>>::deserialize(
-        &mut rmp_serde::Deserializer::new(r),
-    )?;
+    let mut t = ProverIndex::<55, Vesta, Srs>::deserialize(&mut rmp_serde::Deserializer::new(r))?;
```
richardpringle: The reason this works out of the box is because the change from `Vec` to `array` in poseidon, only affects two static methods on the `KimchiCurve` trait.

kimchi/src/circuits/polynomials/generic.rs
```diff
@@ -314,8 +313,11 @@ pub mod testing {
         }
     }
 
-    impl<F: PrimeField, G: KimchiCurve<ScalarField = F>, OpeningProof: OpenProof<G>>
-        ProverIndex<G, OpeningProof>
+    impl<const ROUNDS: usize, F, G, Srs> ProverIndex<ROUNDS, G, Srs>
+    where
+        F: PrimeField,
+        G: KimchiCurve<ROUNDS, ScalarField = F>,
+        Srs: poly_commitment::SRS<G>,
```
richardpringle: As I commented above, this never needed to be generic over `OpenProof`, it was previously over-constrained.

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -103,7 +103,7 @@ impl<F: PrimeField> CircuitGate<F> {
         row: usize,
         // first and last row of the poseidon circuit (because they are used in the permutation)
         first_and_last_row: [GateWires; 2],
-        round_constants: &[Vec<F>],
+        round_constants: &[[F; 3]],
```
richardpringle: This could probably be an array of arrays instead of a slice of arrays. It's safer not to change that, however.

kimchi/src/circuits/polynomials/poseidon.rs
```diff
@@ -123,7 +123,7 @@ impl<F: PrimeField> CircuitGate<F> {
             // round constant for this row
             let coeffs = core::array::from_fn(|offset| {
                 let round = rel_row * ROUNDS_PER_ROW + offset;
-                core::array::from_fn(|field_el| round_constants[round][field_el])
+                round_constants[round]
```
richardpringle: We leverage the fact that `[F; 3]` is `Copy`

kimchi/src/curve.rs
```diff
@@ -60,12 +60,14 @@ pub fn pallas_endos() -> &'static (
     &PALLAS_ENDOS
 }
 
-impl KimchiCurve for Affine<VestaParameters> {
-    fn sponge_params() -> &'static ArithmeticSpongeParams<Self::ScalarField> {
+impl KimchiCurve<{ mina_poseidon::pasta::ROUNDS }> for Affine<VestaParameters> {
```
richardpringle: `mina_poseidon::pasta::ROUNDS` is hardcoded to 55 (you'll get a compiler error if you try to use a different number). 

Most of the places that `55` is hardcoded should also use this constant

kimchi/src/prover.rs
```diff
@@ -168,21 +170,21 @@ where
...
+        index: &ProverIndex<ROUNDS, G, OpeningProof::SRS>,
         prev_challenges: Vec<RecursionChallenge<G>>,
         blinders: Option<[Option<PolyComm<G::ScalarField>>; COLUMNS]>,
         rng: &mut RNG,
     ) -> Result<Self>
     where
-        VerifierIndex<G, OpeningProof>: Clone,
+        EFqSponge: Clone + FqSponge<G::BaseField, G, G::ScalarField, ROUNDS>,
+        EFrSponge: FrSponge<G::ScalarField>,
+        EFrSponge: From<&'static ArithmeticSpongeParams<G::ScalarField, ROUNDS>>,
```
richardpringle: This constraint is new and is a better practice than what was there before. Prior, the `EFrSponge` trait had a `new` function. To keep said function, I would have to parametrize the trait over `ROUNDS` because it was part of the `new` function signature. Unfortunately, generic associated constants cannot be used as part of method signatures. It doesn't actually make sense for a type to be able to implement `EFrSponge` for multiple different `usize` values. In other words, a const generic was the wrong solution here. Instead, I removed the constructor and implemented `From<T>`, where `T` was the only argument being passed in. Functionaly, there's is no change, I just simplified the `EFrSponge` trait and added a `From` implementation and have both bounds now. This is the _composable_ way of using traits.

kimchi/src/tests/generic.rs
```diff
@@ -8,12 +8,29 @@ use core::array;
...
+mod kzg {
+    pub(super) use ark_bn254::{g1::Config as Bn254G1Config, Fr};
+    pub(super) use ark_ec::bn::Bn;
+    pub(super) use poly_commitment::{
+        kzg::{KZGProof, PairingSRS},
+        SRS,
+    };
+}
+#[cfg(feature = "bn254")]
+use kzg::*;
```
richardpringle: It would be better if everything related to kzg was in its own module including tests but this was the less invasive change

plonk-wasm/src/oracles.rs
```diff
@@ -176,7 +177,7 @@ macro_rules! impl_oracles {
             ) -> Result<[<Wasm $field_name:camel Oracles>], JsError> {
                 // conversions
                 let result = crate::rayon::run_in_pool(|| {
-                    let index: DlogVerifierIndex<$G, OpeningProof<$G>> = index.into();
+                    let index: DlogVerifierIndex<FULL_ROUNDS, $G, <OpeningProof<$G, FULL_ROUNDS> as poly_commitment::OpenProof<$G, FULL_ROUNDS>>::SRS> = index.into();
```
richardpringle: I think I can use the fully qualified `OpenProof` here

poseidon/src/pasta/fp_kimchi.rs
```diff
@@ -830,7 +830,7 @@ pub fn params() -> ArithmeticSpongeParams<Fp> {
 }
 
 /// the fp sponge params
-pub fn static_params() -> &'static ArithmeticSpongeParams<Fp> {
-    static PARAMS: Lazy<ArithmeticSpongeParams<Fp>> = Lazy::new(params);
+pub fn static_params() -> &'static Params {
+    static PARAMS: Lazy<Params> = Lazy::new(params);
```
richardpringle: `Lazy` is actually no longer required here because there are no allocations. I didn't change this though as I'm sure the compiler can handle optimizing that part away. This should be fixed in a follow up

poseidon/src/pasta/fp_legacy.rs
```diff
@@ -1459,7 +1459,7 @@ fn params() -> ArithmeticSpongeParams<Fp> {
 }
 
 /// the legacy fp sponge params
-pub fn static_params() -> &'static ArithmeticSpongeParams<Fp> {
-    static PARAMS: Lazy<ArithmeticSpongeParams<Fp>> = Lazy::new(params);
+pub fn static_params() -> &'static Params {
+    static PARAMS: Lazy<Params> = Lazy::new(params);
```
richardpringle: `Lazy` no longer required. This should be fixed in a follow up

poseidon/src/pasta/fq_kimchi.rs
```diff
@@ -829,7 +830,7 @@ pub fn params() -> ArithmeticSpongeParams<Fq> {
 }
 
 /// the fq sponge params
-pub fn static_params() -> &'static ArithmeticSpongeParams<Fq> {
-    static PARAMS: Lazy<ArithmeticSpongeParams<Fq>> = Lazy::new(params);
+pub fn static_params() -> &'static Params {
+    static PARAMS: Lazy<Params> = Lazy::new(params);
```
richardpringle: `Lazy` no longer required, this should be fixed in a follow up

poseidon/src/permutation.rs
```diff
@@ -6,31 +6,42 @@ use crate::{
...
 
 fn apply_mds_matrix<F: Field, SC: SpongeConstants>(
-    params: &ArithmeticSpongeParams<F>,
-    state: &[F],
-) -> Vec<F> {
-    if SC::PERM_FULL_MDS {
-        params
-            .mds
+    mds: [[F; MDS_WIDTH]; MDS_WIDTH],
+    state: &mut [F],
```
richardpringle: `state` is now mutated directly

poseidon/src/permutation.rs
```diff
@@ -93,34 +104,40 @@ pub fn half_rounds<F: Field, SC: SpongeConstants>(
...
+    params: &ArithmeticSpongeParams<F, FULL_ROUNDS>,
+    state: &mut [F],
 ) {
     if SC::PERM_HALF_ROUNDS_FULL == 0 {
         if SC::PERM_INITIAL_ARK {
-            for (i, x) in params.round_constants[0].iter().enumerate() {
-                state[i].add_assign(x);
-            }
+            // maintaining previous invariants
+            assert!(params.round_constants[0].len() <= state.len());
```
richardpringle: Previously, if `round_constants[0]` was larger than `state`, `state[i]` would panic with an out-of-bounds access

poseidon/src/permutation.rs
richardpringle: This file contains the only logical change in the entire PR

plonk-wasm/src/oracles.rs
richardpringle: nvm, can do in a follow up

richardpringle:
Running CI on [github.com/mina](https://github.com/MinaProtocol/mina/pull/18264)

kimchi/src/tests/generic.rs
dannywillems: Agreed

hasher/src/poseidon.rs
```diff
@@ -46,22 +49,29 @@ impl<SC: SpongeConstants, H: Hashable> Poseidon<SC, H> {
 }
 
 /// Poseidon hasher type with legacy plonk sponge constants
-pub type PoseidonHasherLegacy<H> = Poseidon<PlonkSpongeConstantsLegacy, H>;
+pub type PoseidonHasherLegacy<H> = Poseidon<PlonkSpongeConstantsLegacy, H, 100>;
```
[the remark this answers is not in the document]
richardpringle: #3433

poseidon/src/pasta/fq_legacy.rs
```diff
@@ -1459,7 +1460,7 @@ fn params() -> ArithmeticSpongeParams<Fq> {
 }
 
 /// the legacy fq sponge params
-pub fn static_params() -> &'static ArithmeticSpongeParams<Fq> {
-    static PARAMS: Lazy<ArithmeticSpongeParams<Fq>> = Lazy::new(params);
+pub fn static_params() -> &'static Params {
+    static PARAMS: Lazy<Params> = Lazy::new(params);
```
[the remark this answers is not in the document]
richardpringle: #3432

poseidon/src/permutation.rs
```diff
@@ -93,34 +104,40 @@ pub fn half_rounds<F: Field, SC: SpongeConstants>(
...
+
+            state
+                .iter_mut()
+                .zip(params.round_constants[0].iter())
+                .for_each(|(s, x)| {
+                    s.add_assign(x);
+                });
```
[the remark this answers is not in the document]
richardpringle: @dannywillems, should we add an issue to add an assertion here? Could mark it with "good first issue" and hope someone else does it

poseidon/src/permutation.rs
```diff
@@ -6,31 +6,42 @@ use crate::{
     constants::SpongeConstants,
     poseidon::{sbox, ArithmeticSpongeParams},
 };
-use alloc::{vec, vec::Vec};
 use ark_ff::Field;
+const MDS_WIDTH: usize = 3;
```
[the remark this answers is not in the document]
richardpringle: #3434

PR #3387, opened 2025-12-16
Fix xtask linter

richardpringle:
Needed to remove a single character to fix this instead of explude it when linting


dannywillems (approved):
LGTM

PR #3391, opened 2025-12-17
fix(mina-curves): add asm feature for stable Rust compatibility

dannywillems:
## Summary
- Add `asm` feature to mina-curves for stable Rust compatibility
- The `MontConfig` derive macro from ark_ff generates code with `cfg(feature = "asm")`, requiring the consuming crate to declare this feature
- Enable `asm` by default to ensure assembly optimizations are used

## Crates enabled for stable Rust lint
- [x] `arkworks` (closes https://github.com/o1-labs/mina-rust/issues/1929)
- [x] `mina-curves` (closes https://github.com/o1-labs/mina-rust/issues/1931)
- [x] `wasm-types` (closes https://github.com/o1-labs/mina-rust/issues/1930)

## Changes
- `curves/src/lib.rs`: Add compile-time check for asm feature
- `.github/workflows/ci-lint-stable.yml`: Remove arkworks and mina-curves from exclude list

## Test plan
- [x] `cargo clippy -p mina-curves` passes on stable
- [x] `cargo clippy -p arkworks` passes on stable
- [x] Building without asm feature fails with helpful error

### DIFF
--- a/curves/Cargo.toml
+++ b/curves/Cargo.toml
@@ -19,6 +19,7 @@ asm = []
 ark-bn254.workspace = true
 ark-ec.workspace = true
 ark-ff.workspace = true
+ark-std.workspace = true
 num-bigint.workspace = true
 
 [dev-dependencies]
@@ -26,4 +27,3 @@ rand.workspace = true
 ark-test-curves.workspace = true
 ark-algebra-test-templates.workspace = true
 ark-serialize.workspace = true
-ark-std.workspace = true


curves/Cargo.toml
```diff
@@ -9,6 +9,14 @@ readme = "../README.md"
 edition = "2021"
 license = "Apache-2.0"
 
+[features]
+default = ["asm"]
+# The MontConfig derive macro from ark_ff generates code with cfg(feature = "asm").
+# This feature enables assembly optimizations in ark_ff.
+asm = []
+# Disable the compile-time check that ensures asm is enabled.
```
richardpringle: Just an observation, when you do clippy checks with --all-features, you turn this feature on which seems like it disables stuff.

Just make sure this is what you want.

PR #3410, opened 2025-12-18
ci: replace Rust 1.82 with stable and beta in build matrix

dannywillems:
See commit history


iostat (dismissed):
We might want to lock `1.92` as part of the matrix, since that's what stable on `mina-rust` is locked to as well. However, the wasm there is nightly, so perhaps it's not that big a deal.

PR #3424, opened 2025-12-23
Makefile: exclude plonk-wasm from doc generation

dannywillems:


Makefile
```diff
@@ -181,7 +198,14 @@ generate-doc: ## Generate the Rust documentation
 		@echo ""
 		@echo "Generating the documentation."
 		@echo ""
-		RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps --document-private-items --workspace --exclude xtask
+		RUSTDOCFLAGS="--enable-index-page -Zunstable-options" cargo +nightly doc \
```
richardpringle: Why do we need these nightly features?

Makefile
```diff
@@ -181,7 +198,14 @@ generate-doc: ## Generate the Rust documentation
...
+			--all-features \
+			--no-deps \
+			--document-private-items \
+			--workspace \
+			--exclude plonk_neon \
+			--exclude plonk_wasm \
+			--exclude xtask
```
richardpringle: Why is xtask excluded? I don't think there are docs here, but there's nothing gained by excluding it

Makefile
```diff
@@ -94,17 +94,31 @@ clean: ## Clean the project
...
+			--all-targets \
+			--all-features \
+			--exclude plonk_neon \
+			--exclude plonk_wasm \
+			--workspace
 
 
 release: ## Build the project in release mode
-		cargo build --release --all-targets --all-features --workspace \
-			--exclude plonk_neon --exclude plonk_wasm --exclude xtask
```
dannywillems: Note: we stop excluding `xtask` here.

PR #3427, opened 2025-12-30
feat: add mina-tx-type crate with coinbase transaction types

dannywillems:
## Summary

Implements transaction type extraction as specified in:
- o1-labs/mina-rust#1665 (main issue)
- o1-labs/mina-rust#1824 (coinbase types sub-issue)

This PR adds a new `mina-tx-type` crate providing standalone, `no_std` compatible transaction types for external use (hardware wallets, WASM, etc.).

### New Types

**Currency types:**
- `Amount` - Currency amount in nanomina (64-bit)
- `Fee` - Transaction fee in nanomina (64-bit)
- `Signed<T>` - Signed quantity with separate magnitude and sign
- `Sgn` - Sign enum (Pos/Neg)
- `Magnitude` trait - Core numeric operations
- `MinMax` trait - Min/max bounds

**Transaction types:**
- `Coinbase` - Block reward transaction
- `CoinbaseFeeTransfer` - Fee transfer to SNARK workers

### Features

- `no_std` compatible for embedded/constrained environments
- All fields documented with rustdoc comments
- Comprehensive test coverage (12 tests)

## Test plan

- [x] `cargo test -p mina-tx-type` - All 12 tests pass
- [x] `cargo clippy -p mina-tx-type` - No warnings
- [x] `cargo +nightly fmt -- --check` - Formatting passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -250,8 +250,7 @@ and this project adheres to
 - New crate providing `no_std` compatible Mina transaction types for external
   use (hardware wallets, WASM). Implements
   [o1-labs/mina-rust#1824](https://github.com/o1-labs/mina-rust/issues/1824).
-- Currency types: `Amount`, `Fee`, `Signed<T>`, `Sgn`, `Magnitude` trait,
-  `MinMax` trait
+- Currency types: `Amount`, `Fee`, `Signed<T>`, `Sign`, `Magnitude` trait
 - Transaction types: `Coinbase`, `CoinbaseFeeTransfer`
 
 ### [mvpoly](./mvpoly)


mina-tx-type/src/currency.rs
```diff
@@ -0,0 +1,317 @@
...
+//! in Mina Protocol transactions. All types are `no_std` compatible.
+
+use core::{fmt, ops::Neg};
+
+/// Sign of a value, either positive or negative.
+///
+/// Used in conjunction with [`Signed`] to represent signed quantities
+/// where the sign is tracked separately from the magnitude.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Hash)]
+pub enum Sgn {
```
richardpringle:
```suggestion
pub enum Sign {
```

mina-tx-type/src/currency.rs
```diff
@@ -0,0 +1,317 @@
...
+///
+/// This type represents signed quantities where the sign is tracked
+/// separately from the magnitude. This is useful for representing
+/// values that can be positive or negative without using two's complement.
+///
+/// Zero is always normalized to have a positive sign.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Signed<T: Magnitude> {
+    magnitude: T,
+    sgn: Sgn,
```
richardpringle:
```suggestion
    sign: Sign,
```

mina-tx-type/src/currency.rs
```diff
@@ -0,0 +1,317 @@
...
+            write!(f, "{}", self.magnitude)
+        }
+    }
+}
+
+impl<T: Magnitude> Neg for Signed<T> {
+    type Output = Self;
+
+    fn neg(self) -> Self::Output {
+        self.negate()
```
richardpringle: Same comment, there should only be one implementation

richardpringle (changes requested):
Personally, I think it would be a mistake to merge this as is. I think the `Magnitude` trait is a premature abstraction. I also think we could probably just use `i128` instead of `u64` with a separate `Sign` enum. We aren't using anything like `repr(packed)` so `i128` and type wrapping a magnitude and sign actually probably take up the same space...

Actually, instead of speculating, I took a look, here's the playground link:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=69628ead8c03a066fff61b7a10bc8b43

I personally don't think it makes sense to do all of this. It's just a maintenance burden, there are already types in the standard library that handle all of this. The `Fee` and `Amount` newtype wrappers make sense, but they should just wrap an i128 unless there's a specific reason not to (in which case that should be documented)

CHANGELOG.md
```diff
@@ -243,6 +243,17 @@ and this project adheres to
...
 
+### [mina-tx-type](./mina-tx-type)
+
+#### Added
+
+- New crate providing `no_std` compatible Mina transaction types for external
+  use (hardware wallets, WASM). Implements
+  [o1-labs/mina-rust#1824](https://github.com/o1-labs/mina-rust/issues/1824).
+- Currency types: `Amount`, `Fee`, `Signed<T>`, `Sgn`, `Magnitude` trait,
+  `MinMax` trait
```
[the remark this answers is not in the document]
dannywillems: Oopsie.

richardpringle:
> > Personally, I think it would be a mistake to merge this as is. I think the `Magnitude` trait is a premature abstraction. I also think we could probably just use `i128` instead of `u64` with a separate `Sign` enum. We aren't using anything like `repr(packed)` so `i128` and type wrapping a magnitude and sign actually probably take up the same space...
> > Actually, instead of speculating, I took a look, here's the playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=69628ead8c03a066fff61b7a10bc8b43
> > I personally don't think it makes sense to do all of this. It's just a maintenance burden, there are already types in the standard library that handle all of this. The `Fee` and `Amount` newtype wrappers make sense, but they should just wrap an i128 unless there's a specific reason not to (in which case that should be documented)
> 
> If it is fine for you, i’d prefer to keep close to the actual code of mina-rust, and when we change in mina-rust, we update and add regtest.

👍 @dannywillems, can you create an issue to fix this?

richardpringle (approved):

PR #3473, opened 2026-02-05
signer: implement TryFrom<SecKey> for Keypair

dannywillems:
## Summary

- Add idiomatic `TryFrom<SecKey>` trait implementation for `Keypair`
- Deprecate `from_secret_key` method (will be removed in 0.5.0)
- Use internal `from_secret_key_impl` to avoid deprecation warnings within the crate

## Motivation

Users can now write:
```rust
let keypair = Keypair::try_from(secret_key)?;
```

Instead of:
```rust
let keypair = Keypair::from_secret_key(secret_key)?;
```

The `TryFrom` trait is the standard Rust idiom for fallible type conversions and enables use with the `?` operator and other generic code that works with `TryFrom`.

## Changes

- Added `impl TryFrom<SecKey> for Keypair`
- Deprecated `from_secret_key` with notice it will be removed in 0.5.0
- Internal callers (`rand`, `from_bytes`, `from_hex`) use private `from_secret_key_impl`
- Includes documentation with example

## Test plan

- [x] All existing tests pass
- [x] New doc test for `TryFrom` implementation passes
- [x] Deprecation warnings appear when using `from_secret_key` in test code

### DIFF
--- a/signer/src/keypair.rs
+++ b/signer/src/keypair.rs
@@ -168,7 +168,7 @@ impl TryFrom<SecKey> for Keypair {
     /// let keypair = Keypair::try_from(secret).unwrap();
     /// ```
     fn try_from(secret_key: SecKey) -> Result<Self> {
-        Keypair::from_secret_key_impl(secret_key)
+        Self::from_secret_key_impl(secret_key)
     }
 }
 


signer/tests/signer.rs
```diff
@@ -612,7 +612,7 @@ pub fn test_testnet_compatibility_derive_nonce_compatible_multiple_fields() {
 #[test]
 pub fn test_testnet_compatibility_corner_case() {
     let privkey = -ScalarField::from(1u64);
-    let kp = Keypair::from_secret_key(SecKey::new(privkey)).unwrap();
+    let kp = Keypair::try_from(SecKey::new(privkey)).unwrap();
```
querolita: I wonder if we want to use ? instead of unwrap

querolita (dismissed):

querolita (approved):

PR #3478, opened 2026-02-06
feat: add mina-base58 crate

dannywillems:
Part of the review for mina-multisig. They reimplement base58, but it should be provided and maintained by us.

## Summary

- Add standalone `mina-base58` crate (`crates/mina-base58/`) providing generic base58check encoding/decoding
- Include all 30 known Mina version byte constants, matching the OCaml [`version_bytes.ml`](https://github.com/MinaProtocol/mina/blob/master/src/lib/base58_check/version_bytes.ml)
- `#![no_std]` compatible with `extern crate alloc`

## Public API

- `encode(version, payload)` / `decode(b58)` / `decode_version(b58, expected)` — standard base58check with version byte
- `encode_raw(raw)` / `decode_raw(b58)` — raw bytes with checksum (for cases where version bytes are already embedded in the payload)
- `checksum(data)` — double-SHA256 checksum utility
- `DecodeError` enum: `InvalidBase58`, `TooShort`, `InvalidChecksum`, `InvalidVersion`

## Test plan

- [x] 48 unit tests passing locally
- [x] OCaml test vectors from `mina/src/lib/base58_check/tests/test_base58_check.ml` (version `0x53`: empty, `"vectors"`, `"test"`)
- [x] State hash, ledger hash, and address vectors from mina-rust
- [x] Address and secret key roundtrip vectors from mina-signer
- [x] Roundtrip encode/decode across all 30 version bytes
- [x] Cross-API consistency (`encode` vs `encode_raw`, `decode` vs `decode_raw`)
- [x] Error cases: invalid base58 chars, too short, corrupted checksum, version mismatch, empty input
- [x] `cargo +nightly fmt -- --check` passes
- [x] `cargo clippy -p mina-base58 -- -W clippy::all -D warnings` passes
- [x] `cargo +nightly doc -p mina-base58` generates cleanly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,13 @@ and this project adheres to
 - Add `no_std` support
   ([#3415](https://github.com/o1-labs/proof-systems/pull/3415))
 
+### [mina-base58](./crates/mina-base58)
+
+#### Added
+
+- Add `mina-base58` crate with generic base58check encoding/decoding and all
+  known Mina version bytes, matching the OCaml `Base58_check` implementation
+
 ### [kimchi](./kimchi)
 
 #### Added


crates/mina-base58/src/version.rs
```diff
@@ -0,0 +1,41 @@
...
+pub const VERIFICATION_KEY: u8 = 0x1b;
+pub const TOKEN_ID_KEY: u8 = 0x1c;
+pub const TRANSACTION_HASH: u8 = 0x1d;
+
+/// Used for testing only.
+pub const LEDGER_TEST_HASH: u8 = 0x30;
+
+// Non-sequential version bytes. Existing user key infrastructure
+// depends on them -- do not change.
+pub const SECRET_KEY: u8 = 0x5a;
```
querolita: On mina these are called:

```ocaml
let private_key : t = '\x5A'

let non_zero_curve_point_compressed : t = '\xCB'

let signature : t = '\x9A'
```

So different order and private vs secret

Maybe lets at least have the same order

querolita (approved):

crates/mina-base58/src/lib.rs
```diff
@@ -0,0 +1,852 @@
...
+        // src/app/heap_usage/values.ml — state_body_hash field
+        let b58 = "3WuibKRQv4TmqEj48a39QehVueRp8fCZ1Ta4CHfCLdVGG1y2HvDy";
+        let (ver, payload) = decode(b58).unwrap();
+        assert_eq!(ver, version::STATE_BODY_HASH);
+        assert_eq!(encode(ver, &payload), b58);
+    }
+
+    #[test]
+    fn test_regression_v1_transaction_hash() {
+        // src/lib/transaction/transaction_hash.ml — V1 hash test vector
```
querolita: okay these comments are useful

PR #3481, opened 2026-02-06
feat: add transaction types to mina-tx-type crate

dannywillems:
In a follow-up, we will add hashing, regtest, etc. We go step by step.

## Summary

- Add signed command types (payments and stake delegations), zkApp command types (account updates, verification keys, events, actions), fee transfer types, and top-level `Transaction`/`UserCommand` enums to the `mina-tx-type` crate
- Add supporting types: currency (`Balance`, `Nonce`, `Slot`, `SlotSpan`, `Length`, `TxnVersion`), common utilities (`SetOrKeep`, `OrIgnore`, `ClosedInterval`, `OneOrTwo`), primitives (`TokenId`, `Memo`, `AccountId`), permissions (`AuthRequired`, `Permissions`), and preconditions (`NetworkPreconditions`, `AccountPrecondition`)
- All types are `no_std` compatible with no circuit traits or `mina-p2p-messages` dependency, only `mina-signer` and `mina-curves`

Builds on top of #3427 which introduced the crate with `Coinbase` and currency types.

## Test plan

- [x] `cargo check -p mina-tx-type` passes
- [x] `cargo clippy -p mina-tx-type -- -D warnings` passes clean
- [x] `cargo test -p mina-tx-type` passes (doc test)
- [ ] CI passes

### DIFF
--- a/mina-tx-type/src/lib.rs
+++ b/mina-tx-type/src/lib.rs
@@ -68,7 +68,7 @@ pub mod zkapp_command;
 pub use coinbase::{Coinbase, CoinbaseFeeTransfer};
 pub use common::{ClosedInterval, OneOrTwo, OrIgnore, SetOrKeep};
 pub use currency::{
-    Amount, Balance, Fee, Length, Magnitude, Nonce, Sgn, Signed, Slot, SlotSpan, TxnVersion,
+    Amount, Balance, Fee, Length, Magnitude, Nonce, Sign, Signed, Slot, SlotSpan, TxnVersion,
 };
 pub use fee_transfer::{FeeTransfer, SingleFeeTransfer};
 pub use permissions::{AuthRequired, Permissions, SetVerificationKey};


mina-tx-type/src/common.rs
```diff
@@ -0,0 +1,55 @@
...
+/// A closed interval `[lower, upper]`.
+///
+/// Used in numeric preconditions to specify acceptable ranges.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClosedInterval<T> {
+    /// The lower bound (inclusive).
+    pub lower: T,
+    /// The upper bound (inclusive).
+    pub upper: T,
+}
```
richardpringle: This type also already exists in the standard library:
https://doc.rust-lang.org/stable/std/ops/struct.RangeInclusive.html

mina-tx-type/src/signed_command.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// A stake delegation payload.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum StakeDelegationPayload {
+    /// Delegate stake to a new delegate.
+    SetDelegate {
+        /// The public key of the new delegate.
+        new_delegate: CompressedPubKey,
+    },
+}
```
richardpringle: Why an enum with a single variant? That _might_ make sense if we wanted to add `#[non_exhaustive]` to future proof things, but that isn't being done here.

mina-tx-type/src/common.rs
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Used in zkApp account updates to indicate which fields should be
+/// modified.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SetOrKeep<T> {
+    /// Set the field to this value.
+    Set(T),
+    /// Keep the field unchanged.
+    Keep,
+}
```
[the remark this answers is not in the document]
dannywillems: It does replicate the OCaml structure available [here](https://github.com/MinaProtocol/mina/blob/compatible/src/lib/mina_base/zkapp_basic.ml#L83).

I agree it is isomorphic to `Option<T>`.
I am not against changing it to something else later.

PR #3522, opened 2026-03-03
[no-std]: o1-utils

richardpringle:
Should be self explanatory


utils/src/lazy_cache.rs
```diff
@@ -5,15 +5,20 @@
 //! use the official `LazyLock`, and [`LazyCache`] as a wrapper around `LazyLock`
```
iostat: See https://github.com/o1-labs/proof-systems/issues/3535

please fix in follow-up

PR #3523, opened 2026-03-03
[no-std]: internal-tracing

richardpringle:
Self-explanatory


richardpringle:
admin mergin as @iostat already approved [here](https://github.com/o1-labs/proof-systems/pull/3523#pullrequestreview-3932293176)

PR #3534, opened 2026-03-10
fix lookup selectors to use `num_chunks` for chunking 

Trivo25:
lookup selector polynomials used `non_hiding(1)` instead of `non_hiding(num_chunks)`


querolita:
What about merging into master though?

querolita:
I think this should go to master. It is something that was found as part of native prover work but when we merge it eventually I believe it will be much easier to track if we had a dedicated changelog entry for example. So what about we merge it to native/napi but cherry-pick it as part of another PR against master? This could go in the following 0.6.0 or whatever.

PR #3569, opened 2026-04-28
Add support for mmap-backed prover indexes

mrmr1993:
This PR adds a new store / load path for prover indexes, backed by a POSIX [`mmap`ed file](https://man7.org/linux/man-pages/man2/mmap.2.html). This provides 2 practical improvements:
* the format contains only the information necessary for proving, which reduces the RAM usage for the prover index, and
* the OS can unload the backing memory from RAM when there is memory pressure, re-loading it from disk when needed.
The first of these is already an appealing improvement in practice, but the second allows loading more pickles rules into memory than the physical RAM of the machine can support.

This has been successfully used as part of an OCaml re-implementation of [Nori's proof-conversion](https://github.com/Nori-zk/proof-conversion), allowing all rules to be loaded into memory even though the system memory is insufficient to accommodate all of the proving keys.

This proving key is gated behind a flag, and the semantics of the existing flow are unchanged by this PR; this is a purely optional addition.


kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+    PoseidonSelector8 = 0x21,
+    /// Complete-add selector over domain d4.
+    CompleteAddSelector4 = 0x22,
+    /// Variable-base scalar-mul selector over domain d8.
+    MulSelector8 = 0x23,
+    /// Endo scalar-mul selector over domain d8.
+    EmulSelector8 = 0x24,
+    /// Endo-mul-scalar selector over domain d8.
+    EndomulScalarSelector8 = 0x25,
+    /// Permutation coefficients 0..6 over domain d8 (one tag per column).
```
querolita: 0..=6 according to `permutation_coefficient_tag`

```suggestion
    /// Permutation coefficients 0..=6 over domain d8 (one tag per column).
```

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+/// Tag values are stable across format versions within the same major
+/// version and must never be reused for a different meaning.
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum SectionTag {
+    /// `sid: Vec<F>`.
+    Sid = 0x01,
+    /// Packed `[PrunedGate]` array.
+    Gates = 0x02,
+    /// Coefficients 0..14 over domain d8 (one tag per column).
```
querolita: 0..=14 according to `coefficient_tag`

```suggestion
    /// Coefficients 0..=14 over domain d8 (one tag per column).
```

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+    "cached_prover_index requires a little-endian target; the on-disk \
+     layout stores field-element limbs as little-endian u64."
+);
+
+/// Tags identifying the different payload sections in the section table.
+/// Tag values are stable across format versions within the same major
+/// version and must never be reused for a different meaning.
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum SectionTag {
```
querolita, line 79: Maybe clarify that not all numeric tag values seem to be taken (sparse)

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+    identifier: &str,
+    path: &Path,
+    srs: Arc<Srs>,
+) -> Result<MmapProverIndex<FULL_ROUNDS, G, Srs>, CacheError>
+where
+    G: KimchiCurve<FULL_ROUNDS>,
+    Srs: SRS<G>,
+    G::BaseField: ark_ff::PrimeField,
+{
+    assert_bigint_layout::<G::ScalarField>()?;
```
querolita: Perhaps include the check for the base field as well because `verifier_index_digest` is one of them. 

```suggestion
    assert_bigint_layout::<G::ScalarField>()?;
    assert_bigint_layout::<G::BaseField>()?;
```

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+    identifier: &str,
+    index: &ProverIndex<FULL_ROUNDS, G, Srs>,
+    path: &Path,
+) -> Result<(), CacheError>
+where
+    G: KimchiCurve<FULL_ROUNDS>,
+    Srs: SRS<G>,
+    G::BaseField: PrimeField,
+{
+    assert_bigint_layout::<G::ScalarField>()?;
```
querolita: Perhaps include the check for the base field as well because `verifier_index_digest` is one of them (same comment for `read_cache`)

```suggestion
    assert_bigint_layout::<G::ScalarField>()?;
    assert_bigint_layout::<G::BaseField>()?;
```

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+    // be uninitialised, but since the Pasta Fp has no padding this is a
+    // full-bytes read.
+    unsafe { core::ptr::read(limbs as *const [u64; 4] as *const F) }
+}
+
+// Runtime assertion that the field-element layout assumption holds. Cheap
+// and only paid once per process; protects against silently picking up an
+// ark-ff upgrade that changes the representation.
+pub(crate) fn assert_bigint_layout<F: PrimeField>() -> Result<(), CacheError> {
+    if core::mem::size_of::<F::BigInt>() == FIELD_ELEMENT_BYTES {
```
querolita: In `read_cache` you get a vec from raw parts so I wonder if a safer check would include a memory alignment check for the 4 limbs of 64 bits. If the internal representation ever changed, probably the `from_raw_parts` could need changes.

```suggestion
    if core::mem::size_of::<F::BigInt>() == FIELD_ELEMENT_BYTES 
        && core::mem::align_of::<F>() == core::mem::align_of::<[u64; 4]>()
    {
```

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+                f,
+                "section {tag:#x} offset {offset} is not {SECTION_ALIGNMENT}-byte aligned"
+            ),
+            CacheError::TruncatedFile => write!(f, "cache file truncated before end of declared payload"),
+            CacheError::PayloadNotFieldAligned { tag, length } => write!(
+                f,
+                "section {tag:#x} payload length {length} is not a multiple of {FIELD_ELEMENT_BYTES}"
+            ),
+            CacheError::BigIntSizeMismatch => {
+                write!(f, "ark_ff::BigInt<4> is not 32 bytes on this build")
```
querolita: If we added the guard for 4 limbs of u64, then this error message could be adapted


```suggestion
                write!(f,
                    "field-element layout is not compatible with four 64-bit Montgomery limbs"
                )
```

kimchi/src/cached_prover_index.rs
```diff
@@ -0,0 +1,2049 @@
...
+
+    let cs = &index.cs;
+    // Grab the materialised LookupConstraintSystem (if any) once, up front.
+    // A `Some(_)` here triggers emission of sections 0x50..0x58; `None` or a
+    // `LookupError` leaves them out and keeps the file backward-compatible
+    // with non-lookup readers.
+    let lcs_result = cs.lookup_constraint_system.get();
+    let lcs: Option<&LookupConstraintSystem<G::ScalarField>> = match lcs_result {
+        Ok(opt) => opt.as_ref(),
+        Err(_) => None,
```
querolita: If we added that kind of error then we could have something more verbose here like 


```suggestion
        Err(e) => return Err(CacheError::LookupConstraintSystem(e.to_string())),
```

kimchi/src/tests/cached_index.rs
```diff
@@ -0,0 +1,337 @@
...
+    let path = tmpfile("badmagic");
+    std::fs::write(&path, vec![0u8; 1024]).unwrap();
+    let srs = std::sync::Arc::new(poly_commitment::ipa::SRS::<Vesta>::create(16));
+    let err = read_cache::<FULL_ROUNDS, Vesta, _>("anything", &path, srs).expect_err("should fail");
+    assert!(
+        matches!(err, CacheError::BadMagic { .. }),
+        "expected BadMagic, got {err}"
+    );
+    std::fs::remove_file(&path).ok();
+}
```
querolita, line 604: If we added the lookup error we could also have this as a unit test


```suggestion
}

/// Cache writes must preserve lookup setup failures instead of silently
/// serialising the index as if it had no lookup constraint system.
#[test]
fn cached_index_lookup_error_fails_write() {
    let (_, gates) = CircuitGate::<Fp>::create_multi_range_check(0);
    let index = new_index_for_test_with_lookups::<FULL_ROUNDS, Vesta>(
        gates,
        0,
        0,
        vec![crate::circuits::polynomials::range_check::gadget::lookup_table()],
        None,
        false,
        None,
        true,
    );

    let path = tmpfile("lookup_error");
    let err = write_cache("lookup-error", &index, &path).expect_err("should fail");
    assert!(
        matches!(err, CacheError::LookupConstraintSystem(_)),
        "expected lookup constraint system error, got {err}"
    );

    std::fs::remove_file(&path).ok();
}
```

querolita (dismissed):
Very interesting optional feature. I wonder if this will need a follow-up for full-stack integration. I added a few suggestions. They are not critical, but if you agree about the  new alignment checks, then I think it's worth applying those changes before merging.

I will try now to pass CI (format is off looks like) and update with more recent master. Then you can apply commits on top if you want (I believe my approval won't suffice if the latest commit is mine)

martyall (approved):
review was done on fork https://github.com/mrmr1993/proof-systems/pull/2

PR #3577, opened 2026-06-10
Preserve chunks of public evaluations inside oracles

### LINKED ISSUE #3576: Adapt oracle bindings for multiple chunks

querolita:
@martyall observed a potential bug in `kimchi-stubs` where the public evaluations of the oracles were being chopped when on chunking scenarios, those are vectors instead. We observed the same pattern on both `kimchi-wasm` and `kimchi-napi`.   If we want claims to attest for all the polynomials, then we should include the evaluations of the rest of chunks as well. Otherwise there could be cases where the mismatch could be exploited to ignore checks about higher chunks. This fix requires updating the mina counterpart as well. It is not super critical right now because the current proof system path is taking these elements from the proof itself where they are indeed vectors, and for dummy verifications (where this function is used) they are indeed single-chunked. But we think it's a good practice to preserve the `Vec` type for safety reasons regardless.

Closes https://github.com/o1-labs/proof-systems/issues/3576

mina side: https://github.com/MinaProtocol/mina/pull/18884
o1js side: https://github.com/o1-labs/o1js/pull/2887


martyall (approved):

martyall:
The core part of this PR matches some local patching I've done to make it work out

PR #3578, opened 2026-06-11
kimchi-stubs: use wNAF for native Pasta scalar multiplication

mrmr1993:
The OCaml-facing `caml_{pallas,vesta}_scale` stubs (via the shared `impl_projective!` macro) used arkworks' default `mul_bigint`, a plain bit-by-bit double-and-add (~255 doublings + ~127 additions for a 255-bit scalar). Switch to a windowed-NAF multiplication (`WnafContext::new(4)`), which cuts the additions substantially.

Measured on Pasta (Pallas, fixed 255-bit scalar over varying bases): ~25% faster (~67.4us -> ~50.7us per scalar mul; window 4 was fastest in a sweep over windows 2..8). The result is bit-identical (same curve point), so it is safe on the consensus path and speeds up every native value-level scalar mul (VRF evaluation, signatures, witness generation, ...).

**Edit:** The reason that this *isn't* the default is that wNAF setup does some additional allocations that aren't suitable for no-std environments, and that wNAF is faster only for scalars that are approximately the size of the field (so ~2^255, not e.g. ~2^64). The OCaml bindings aren't operating in a no-std environment, and most of the scalars we handle in the protocol are large (hashes or private keys), so this is almost universally a benefit.


mrmr1993:
Annoyingly, this runs into a pickles error (something depends on the projective coordinates, which are different under the 2 representations). Closing for now.

mrmr1993:
The pickles test error was a flake, not related to this. This passes the test suite. Reopening.

martyall (approved):

PR #3582, opened 2026-06-18
fix(utils): handle i32::MIN in i32_to_field

haoshengzhen:
## Summary

Fix `i32_to_field` for the `i32::MIN` edge case.

The previous implementation converted negative values with `-i as u64`. This overflows for `i32::MIN`, since its absolute value cannot be represented as an `i32`. With overflow checks enabled, this causes a panic before the value can be converted into the field.

This change uses `i32::unsigned_abs()` for the negative branch, preserving the existing mapping of negative integers to the negated field element while avoiding the intermediate signed overflow.

## Testing

- `git diff --check`
- `cargo fmt --check -- utils/src/field_helpers.rs utils/tests/field_helpers.rs`

I also verified the overflow root cause with a minimal `rustc -Coverflow-checks=y` reproduction: the old `-i` expression panics for `i32::MIN`, while the new `unsigned_abs()` conversion succeeds.

### DIFF
--- a/utils/src/field_helpers.rs
+++ b/utils/src/field_helpers.rs
@@ -243,7 +243,7 @@ pub fn i32_to_field<F: From<u64> + Neg<Output = F>>(i: i32) -> F {
     if i >= 0 {
         F::from(i as u64)
     } else {
-        -F::from(-i as u64)
+        -F::from(u64::from(i.unsigned_abs()))
     }
 }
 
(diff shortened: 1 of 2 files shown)


haoshengzhen:
@querolita @iostat @Trivo25 Hi, Could you please review this PR at your convenience? Thank you very much.

PR #3586, opened 2026-08-14
Remove FFTs from `DomainConstantEvaluations` and some in-memory precomputations

mrmr1993:
This PR changes the algorithm for generating `DomainConstantEvaluations`, using the underlying structure of the precomputed terms instead of brute-force FFT-ing them. In particular, the zk-rows vanishing polynomial is the low-degree polynomial `(x - w^(n-3))(x - w^(n-2))(x - w^(n-1))`, and it's significantly faster to compute this directly on the evaluations of `x` instead of materialising it via the coefficients.

### DIFF
--- a/kimchi/src/circuits/domain_constant_evaluation.rs
+++ b/kimchi/src/circuits/domain_constant_evaluation.rs
@@ -1,15 +1,29 @@
 //! This contains the [DomainConstantEvaluations] which is used to provide precomputations to a [ConstraintSystem](super::constraints::ConstraintSystem).
 
 use crate::circuits::domains::EvaluationDomains;
+use alloc::vec::Vec;
 use ark_ff::FftField;
 use ark_poly::{
-    univariate::DensePolynomial as DP, DenseUVPolynomial, Evaluations as E,
+    univariate::DensePolynomial as DP, EvaluationDomain, Evaluations as E,
     Radix2EvaluationDomain as D,
 };
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
-use super::polynomials::permutation::{permutation_vanishing_polynomial, vanishes_on_last_n_rows};
+use super::polynomials::permutation::permutation_vanishing_polynomial;
+
+/// Evaluate the polynomial `Π (x - root)` (given by its roots) at every point of
+/// `d8`, where `x_d8` holds the d8 domain points. For the low-degree vanishing
+/// polynomials this is far cheaper than padding to 8n coefficients and running
+/// an FFT, and it parallelises trivially over the points.
+fn eval_from_roots_over_d8<F: FftField>(x_d8: &[F], roots: &[F], d8: D<F>) -> E<F, D<F>> {
+    let evals: Vec<F> = o1_utils::cfg_iter!(x_d8)
+        .map(|z| roots.iter().map(|root| *z - *root).product())
+        .collect();
+    E::from_vec_and_domain(evals, d8)
+}
 
 #[serde_as]
 #[derive(Clone, Serialize, Deserialize, Debug)]
@@ -30,19 +44,36 @@ pub struct DomainConstantEvaluations<F: FftField> {
 
 impl<F: FftField> DomainConstantEvaluations<F> {
     pub fn create(domain: EvaluationDomains<F>, zk_rows: u64) -> Option<Self> {
-        let poly_x_d1 = DP::from_coefficients_slice(&[F::zero(), F::one()])
-            .evaluate_over_domain_by_ref(domain.d8);
-
-        let vanishes_on_zero_knowledge_and_previous_rows =
-            vanishes_on_last_n_rows(domain.d1, zk_rows + 1).evaluate_over_domain(domain.d8);
-
         assert!(domain.d1.size > zk_rows);
 
-        // x^3 - x^2(w1+w2+w3) + x(w1w2+w1w3+w2w3) - w1w2w3
+        let omega = domain.d1.group_gen;
+        let n = domain.d1.size;
+
+        // `x` over d8 is just the d8 domain points (g8^row); recover them from
+        // the domain rather than through an FFT.
+        let x_d8: Vec<F> = domain.d8.elements().collect();
+
+        // Vanishes on the last (zk_rows + 1) rows: roots omega^{n-(zk_rows+1)} ..
+        // omega^{n-1}.
+        let zk_roots: Vec<F> = ((n - (zk_rows + 1))..n).map(|i| omega.pow([i])).collect();
+        let vanishes_on_zero_knowledge_and_previous_rows =
+            eval_from_roots_over_d8(&x_d8, &zk_roots, domain.d8);
+
+        // x^3 - x^2(w1+w2+w3) + x(w1w2+w1w3+w2w3) - w1w2w3, with the three roots
+        // omega^{n-zk_rows}, omega^{n-zk_rows+1}, omega^{n-1}.
         let permutation_vanishing_polynomial_m =
             permutation_vanishing_polynomial(domain.d1, zk_rows);
-        let permutation_vanishing_polynomial_l =
-            permutation_vanishing_polynomial_m.evaluate_over_domain_by_ref(domain.d8);
+        let permutation_vanishing_polynomial_l = eval_from_roots_over_d8(
+            &x_d8,
+            &[
+                omega.pow([n - zk_rows]),
+                omega.pow([n - zk_rows + 1]),
+                omega.pow([n - 1]),
+            ],
+            domain.d8,
+        );
+
+        let poly_x_d1 = E::from_vec_and_domain(x_d8, domain.d8);
 
         Some(DomainConstantEvaluations {
             poly_x_d1,
--- a/kimchi/src/circuits/domain_constant_evaluation.rs
+++ b/kimchi/src/circuits/domain_constant_evaluation.rs
@@ -4,7 +4,7 @@ use crate::circuits::domains::EvaluationDomains;
 use alloc::{vec, vec::Vec};
 use ark_ff::FftField;
 use ark_poly::{
-    univariate::DensePolynomial as DP, EvaluationDomain, Evaluations as E,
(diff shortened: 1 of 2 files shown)


kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -36,8 +32,6 @@ impl<F: FftField> DomainConstantEvaluations<F> {
     pub fn create(domain: EvaluationDomains<F>, zk_rows: u64) -> Option<Self> {
         let poly_x_d1 = DP::from_coefficients_slice(&[F::zero(), F::one()])
```
martyall: for the record, I find the name `poly_x_d1` confusing, it's evaluated over `d8`

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -30,19 +44,36 @@ pub struct DomainConstantEvaluations<F: FftField> {
...
+            eval_from_roots_over_d8(&x_d8, &zk_roots, domain.d8);
 
-        // x^3 - x^2(w1+w2+w3) + x(w1w2+w1w3+w2w3) - w1w2w3
+        // x^3 - x^2(w1+w2+w3) + x(w1w2+w1w3+w2w3) - w1w2w3, with the three roots
+        // omega^{n-zk_rows}, omega^{n-zk_rows+1}, omega^{n-1}.
         let permutation_vanishing_polynomial_m =
             permutation_vanishing_polynomial(domain.d1, zk_rows);
-        let permutation_vanishing_polynomial_l =
-            permutation_vanishing_polynomial_m.evaluate_over_domain_by_ref(domain.d8);
+        let permutation_vanishing_polynomial_l = eval_from_roots_over_d8(
```
martyall: previously this was derived directly from `permutation_vanishing_polynomial_m`, now nothing enforces that. 

If 

```
        // x^3 - x^2(w1+w2+w3) + x(w1w2+w1w3+w2w3) - w1w2w3, with the three roots
        // omega^{n-zk_rows}, omega^{n-zk_rows+1}, omega^{n-1}.
```
is the closed form of `permutation_vanishing_polynomial_m`, the difference between evaluating it via horner fold vs `eval_from_roots_over_d8` is really in the wash at that point. This would relink `_m` and `_l` with marginal cost

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -1,15 +1,29 @@
...
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
-use super::polynomials::permutation::{permutation_vanishing_polynomial, vanishes_on_last_n_rows};
+use super::polynomials::permutation::permutation_vanishing_polynomial;
+
+/// Evaluate the polynomial `Π (x - root)` (given by its roots) at every point of
+/// `d8`, where `x_d8` holds the d8 domain points. For the low-degree vanishing
+/// polynomials this is far cheaper than padding to 8n coefficients and running
+/// an FFT, and it parallelises trivially over the points.
```
martyall, line 41: while this would still be faster, just fyi the original is not as bad as it looks:

https://github.com/arkworks-rs/algebra/blob/v0.5.0/poly/src/domain/radix2/mod.rs#L148-L149

kimchi/tests/test_domain_constant_evaluation.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// The construction each field used before it was given a closed form: build
+/// the polynomial in coefficient form and evaluate it over d8 by FFT.
+fn reference<F: FftField>(poly: &DP<F>, d8: D<F>) -> E<F, D<F>> {
+    poly.evaluate_over_domain_by_ref(d8)
+}
+
+/// Domain sizes and zk_rows worth exercising: small domains (where d8 is close
+/// to the number of zk rows) and larger ones, against every zk_rows the
+/// constraint system builds in practice.
```
martyall, line 27: mina has tests that go up to 8 chunks, so `(16 * 8 + 5) / 7 = 19`

kimchi/src/circuits/domain_constant_evaluation.rs
martyall: @mrmr1993 This is my only real comment, I'm not going to die on this hill but I'm not going to approve/merge until you resond

mrmr1993: This is true, but proofs won't verify if they delink. Up to you, I'm inclined to keep this as-is.

martyall (dismissed):

martyall:
@mrmr1993 the changes seem fine, but from the benches I'm looking at in terms of overall speed we're looking at minor improvements -- the places where improvements are made are just not where the bulk over prover time / peak memory is spent (total allocation seems good, I just didn't see it advertised). In particular the claim about the FFT time doesn't match what I'm seeing, this might be because of the "degree aware" codepath that arkworks takes. Did you see something different and if so can you post it? 

# Report

**Environment:** Apple M5 Pro (18 cores, 48 GB), macOS arm64, rustc 1.92.0 `--release`,
kimchi default features (`prover` → `parallel`), criterion 0.5, Pasta `Fp`.
Base `adc0540d26` (merge-base) vs PR head `f1a0d8a92e`, each in its own git worktree
with a separate `CARGO_TARGET_DIR`.

### 1. `perm_quot` — the only proving-path change

| Domain | Base | PR | Δ |
|---|---|---|---|
| 2^10 | 6.16 ms | 3.87 ms | **−37%** |
| 2^14 | 15.39 ms | 10.69 ms | **−31%** |
| 2^16 | 41.84 ms | 30.42 ms | **−27%** |

Confidence intervals do not overlap at any size.

**Share of total:** at 2^16, `perm_quot` is 41.8 ms of a 1990 ms proof (**2.1%**),
so −27% there is **~0.6% end-to-end** — below the ~10% run-to-run drift on this
machine. End-to-end wall-clock cannot resolve this change.

### 2. `DomainConstantEvaluations::create`

Output verified bit-exact vs the old implementation at every size.

| Domain | Old (FFT) | New (direct) | Speedup |
|---|---|---|---|
| 2^12 | 2.89 ms | 1.22 ms | **2.4×** |
| 2^14 | 3.82 ms | 2.04 ms | **1.9×** |
| 2^16 | 6.89 ms | 5.99 ms | **1.15×** |
| 2^18 | 23.6 ms | 23.3 ms | **parity** |

Measured with both variants in one binary, interleaved over two rounds
(an uninterleaved run showed a false regression at 2^18 from thermal drift).

**Decomposition at 2^18:**

| Component | Old | New | |
|---|---|---|---|
| `x` over d8 → `domain_points` | 3.50 ms | 2.65 ms | win at all sizes |
| 2 vanishing polys → `eval_from_roots_over_d8` ×2 | 16.4 ms | 18.0 ms | **slower** |
| 2 constant vectors | 0.76 ms | — | removed |

Both paths are memory-bandwidth bound at this size; the root-product moves more
bytes per output element than the FFT it replaces.

### 3. Memory (exact allocator counts, no statistics)

At 2^16:

| Metric | Base | PR | Δ |
|---|---|---|---|
| `create` retained | 75.50 MB | 50.33 MB | **−24.0 MiB (−33%)** |
| Prover index retained | 610.95 MB | 585.79 MB | −24.0 MiB (−4.1%) |
| Allocated per proof | 7.352 GB | 6.866 GB | **−464 MiB (−6.6%)** |
| Peak heap per proof | 1.659 GB | 1.627 GB | −29.9 MiB (−1.9%) |

−24.0 MiB is exactly `12n × 32 B`, the deleted `constant_1_d4` + `constant_1_d8`.
Scales linearly (−6 MiB at 2^14).

### Summary

- Net improvement; rewrite is equivalent.
- The **headline claim (FFT removal) is the weakest result** — 1.15× at 2^16,
  nothing at 2^18 — and half of it (`eval_from_roots_over_d8`) is a mild
  pessimization at large domains. Keeping `domain_points` but reverting the two
  vanishing polynomials to FFT would likely beat both variants.
- The **strongest wall-clock result (`perm_quot`, −27%) isn't mentioned** in the
  PR description, though it's only ~0.6% of a proof.
- The most defensible claim is the one not made at all: **−464 MiB allocated per
  proof, −24 MiB retained per index** at 2^16.

mrmr1993:
Yea, a lot of the commits were driven by memory pressure, so I could run more provers with 32GB of RAM. I thought the performance wins were more interesting for upstreaming, but you can imagine that ~0.5 GB of removed allocations was a big win when running 10+ provers.

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -1,31 +1,58 @@
...
 /// pre-computed polynomials that depend only on the chosen field and domain
 pub struct DomainConstantEvaluations<F: FftField> {
-    /// 1-st Lagrange evaluated over domain.d8
+    /// the polynomial `x` evaluated over domain.d8 (i.e. the d8 domain points)
     #[serde_as(as = "o1_utils::serialization::SerdeAs")]
     pub poly_x_d1: E<F, D<F>>,
-    /// 0-th Lagrange evaluated over domain.d4
-    // TODO(mimoo): be consistent with the paper/spec, call it L1 here or call it L0 there
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub constant_1_d4: E<F, D<F>>,
```
martyall: seems like no one uses this

kimchi/src/circuits/domain_constant_evaluation.rs
```diff
@@ -1,31 +1,58 @@
...
-    /// 0-th Lagrange evaluated over domain.d8
-    #[serde_as(as = "o1_utils::serialization::SerdeAs")]
-    pub constant_1_d8: E<F, D<F>>,
```
martyall: removing this and dealing with gamma as it comes makes way more sense

martyall (approved):
It seems that the comment I made about improvements was only relevant for the portion that already made it into #3589 

There is no observable benefit performance wise here, but also no degredation.

kimchi/src/circuits/domain_constant_evaluation.rs
martyall: There is no measurable difference

querolita (approved):

PR #3587, opened 2026-08-14
Improve prover peak RAM usage

mrmr1993:
This PR
* reduces peak RAM by releasing the d8 evaluations once the linearization is done, instead of holding them until the end of the computation
* improves prover speed and reduces peak memory, by using sub-sampling to iFFT a d1-domain polynomial instead of a d8-domain one.

### DIFF
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -484,20 +484,21 @@ where
 
             //~~ * Compute the lookup table values as the combination of the lookup table entries.
             let joint_lookup_table_d8 = {
-                let mut evals = Vec::with_capacity(d1_size);
+                // Each row combines its own table entry independently of every
+                // other row, so this is a pure map over the d8 domain.
+                let evals = (0..(d1_size * 8))
+                    .into_par_iter()
+                    .map(|idx| {
+                        let table_id = match lcs.table_ids8.as_ref() {
+                            Some(table_ids8) => table_ids8.evals[idx],
+                            None =>
+                            // If there is no `table_ids8` in the constraint system,
+                            // every table ID is identically 0.
+                            {
+                                G::ScalarField::zero()
+                            }
+                        };
 
-                for idx in 0..(d1_size * 8) {
-                    let table_id = match lcs.table_ids8.as_ref() {
-                        Some(table_ids8) => table_ids8.evals[idx],
-                        None =>
-                        // If there is no `table_ids8` in the constraint system,
-                        // every table ID is identically 0.
-                        {
-                            G::ScalarField::zero()
-                        }
-                    };
-
-                    let combined_entry =
                         if !lcs.configuration.lookup_info.features.uses_runtime_tables {
                             let table_row = lcs.lookup_table8.iter().map(|e| &e.evals[idx]);
 
@@ -525,9 +526,9 @@ where
                                 table_row,
                                 &table_id,
                             )
-                        };
-                    evals.push(combined_entry);
-                }
+                        }
+                    })
+                    .collect();
 
                 Evaluations::from_vec_and_domain(evals, index.cs.domain.d8)
             };
@@ -592,9 +593,10 @@ where
 
             // precompute different forms of the sorted polynomials for later
             // TODO: We can avoid storing these coefficients.
-            let sorted_coeffs: Vec<_> = sorted.iter().map(|e| e.clone().interpolate()).collect();
+            let sorted_coeffs: Vec<_> =
+                sorted.par_iter().map(|e| e.clone().interpolate()).collect();
             let sorted8: Vec<_> = sorted_coeffs
-                .iter()
+                .par_iter()
                 .map(|v| v.evaluate_over_domain_by_ref(index.cs.domain.d8))
                 .collect();
 
--- a/kimchi/src/prover.rs
+++ b/kimchi/src/prover.rs
@@ -1146,7 +1146,17 @@ where
                     lin.interpolate()
                 };
 
+                // The constraint environment and the d8 evaluations it borrows
+                // (`lagrange`, and the lookup table / sorted / aggregation
+                // evaluations) are not needed past this point -- everything
+                // below operates on d1-sized polynomials. Those d8 buffers
+                // dominate the prover's peak memory, so releasing them here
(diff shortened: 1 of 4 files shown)


martyall:
is there a way you can post a flame graph before/after ?

mrmr1993:
I've added `kimchi/examples/prove_lookup_circuit.rs`, which proves a circuit of multi-range-check gadgets — the shape foreign-field arithmetic produces, and what the workload that motivated this PR is made of. Fixed seed, so before/after runs are directly comparable.

## Memory

Peak RSS barely moves, and that's expected: the peak falls during quotient computation, *before* the point where these buffers are released. What changes more is how long the d8 evaluations are held. Sampling RSS every ~30ms through one 2^16 proof, on all 20 cores:

```mermaid
xychart-beta
    title "Resident memory during one proof (multi-range-check circuit, 2^16)"
    x-axis "seconds" [0.0, 0.2, 0.4, 0.6, 0.8, 1.0, 1.2, 1.4, 1.6, 1.8, 2.0, 2.2, 2.4, 2.6, 2.8, 3.0, 3.2, 3.4, 3.6, 3.8, 4.0, 4.2, 4.4, 4.6, 4.8, 5.0, 5.2, 5.4]
    y-axis "RSS (MB)" 0 --> 2500
    line [6, 27, 42, 179, 300, 427, 801, 803, 804, 806, 1089, 988, 997, 1060, 1372, 1399, 2201, 2329, 2313, 2377, 2201, 2202, 2178, 2178, 2178, 2178, 2178, 964]
    line [5, 27, 42, 184, 298, 425, 790, 802, 803, 807, 899, 926, 1034, 1357, 1365, 2060, 2174, 2142, 2190, 2126, 2126, 2102, 2102, 1216, 1137, 1137, 893, 893]
```

The upper line, flat at ~2178 MB right to the end, is `master`. The lower one, dropping at ~4.6s, is this PR: **~1040 MB less resident (≈48%) held through the entire closing phase** — the IPA opening — instead of being freed at end-of-proof. Peak RSS moves by only ~2%. What it lowers is *sustained* memory during the closing phase. Aggregated over the proving window:

| metric (2^16, 20 threads) | master | this PR | |
|---|---|---|---|
| peak RSS | 2409 MB | 2350 MB | −2.4% |
| mean RSS during proving | 1662 MB | 1504 MB | **−9.5%** |
| time spent above 1800 MB | 2.2 s | 1.7 s | **−23.5%** |
| memory×time integral | 7.0 GB·s | 5.9 GB·s | **−12.4%** |

## Runtime

8 interleaved runs at 2^16, alternating binaries to cancel drift:

| | mean | median | sd |
|---|---|---|---|
| master | 4.054 s | 4.040 s | 0.112 |
| this PR | 3.919 s | 3.910 s | 0.079 |

Faster in 7 of 8 pairs; **−3.3%**, with the gap larger than either standard deviation.

## Low thread counts

The deployment this came from runs many provers concurrently, each with a small rayon pool, so I repeated everything with `RAYON_NUM_THREADS=2`. It's proportionally smaller, though the absolute saving is unchanged:

| proving window, 2^16 | 20 threads | 2 threads |
|---|---|---|
| runtime | −3.3% | −4.2% |
| mean RSS during proving | −9.5% | −4.9% |
| memory×time integral | −12.4% | −8.9% |

The ~1 GB released early is the same in absolute terms either way; at 2 threads the phases *before* the release take a larger share of a longer proof, so it's a smaller fraction of the run. Runtime goes slightly the other way (−4.2% vs −3.3%).

Caveat: this is a synthetic circuit, and the balance between quotient, linearization and opening in a real proof-conversion circuit is not the same, so I would not read these percentages across to that workload. What does carry over is the absolute figure — roughly a gigabyte per prover, returned before the closing phase rather than after it — which is the number that decides how many provers fit in a box.

## Flame graphs specifically

I generated CPU flame graphs before/after and they are not informative for this change, so I haven't attached them. No single frame collapses: the profile is a wide forest of parallel rayon work, and the largest differential frame is `rayon::bridge_producer_consumer` moving *up* by 0.8pp, consistent with more work happening in parallel. Everything else is under 0.6pp. Happy to attach them if you'd still like to see them.

## Reproducing

```
cargo run --release --example prove_lookup_circuit -- 16 1
```

Peak RSS via `/usr/bin/time -v`; for the curve, sample `/proc/<pid>/status` externally, since the most interesting behaviour is the release *after* the peak.

kimchi/src/prover.rs
```diff
@@ -532,8 +532,22 @@ where
...
+            // constant (challenge) coefficients, so the joint table is a fixed
+            // linear combination of the columns -- each of which is the d8
+            // evaluation of a polynomial of degree < d1 -- and so has degree < d1
+            // itself. d1 is a subgroup of d8 (d8 = 8*d1, Radix2), so the joint
+            // table's d1 evaluations are exactly every 8th d8 evaluation, and a
+            // d1-sized iFFT over them recovers the coefficients exactly. This
+            // avoids the full d8 iFFT the interpolation here used to perform.
+            let joint_lookup_table = {
+                let d1_evals: Vec<G::ScalarField> = (0..d1_size)
+                    .map(|j| joint_lookup_table_d8.evals[8 * j])
```
martyall: nit:
`joint_lookup_table_d8.evals.iter().step_by(8).collect()`

kimchi/tests/test_domain.rs
```diff
@@ -16,3 +19,54 @@ fn test_create_domain() {
...
+    for log_n in 3..=10u32 {
+        let d = EvaluationDomains::<Fp>::create(1 << log_n).unwrap();
+
+        // Any polynomial of degree < d1 -- the bound the joint lookup table meets.
+        let coeffs: Vec<Fp> = (0..d.d1.size()).map(|_| Fp::rand(&mut rng)).collect();
+        let poly = DP::from_coefficients_vec(coeffs);
+
+        let evals_d8 = poly.evaluate_over_domain_by_ref(d.d8);
+
+        let subsampled: Vec<Fp> = (0..d.d1.size()).map(|j| evals_d8.evals[8 * j]).collect();
```
martyall: nit:

`evals.iter().step_by(8).collect()`

martyall (dismissed):
approve with nit comment

martyall (approved):

PR #3588, opened 2026-08-14
Performance: skip constraint checks on `d1` rows where we expensively calculate 0

mrmr1993:
This PR modifies the computations for the constraint polynomial that feeds into the quotient, omitting calculations for `d1` -- where any valid witness is identically 0, as immediately checked by division by the vanishing polynomial -- achieving a `1/8 = 12.5%` speed-up for this part of the prover.

This comes at a cost: an invalid witness will now no longer be detected by the prover's division by the vanishing polynomial, and instead an invalid proof will only be caught by running the verifier on that proof. If this is an issue in practice, it's possible to run the verifier's check on the constraint polynomial at the end -- still wasteful, but not too expensive -- or to surface a dedicated call to run only that check which can be run optionally by consumers (and can e.g. be toggled in pickles by OCaml's snarky `eval_constraints` mechanism).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,20 @@ and this project adheres to
   disk while allowing the OS page cache to evict unused pages
   ([#3569](https://github.com/o1-labs/proof-systems/pull/3569))
 
+#### Changed
+
+- Skip the d1 rows when evaluating constraints for the quotient. Every
+  constraint vanishes there for a satisfying witness, so the honest prover was
+  computing known zeros; the generic constraint's public-input rows are patched
+  back on afterwards. Proofs are unchanged
+  ([#3588](https://github.com/o1-labs/proof-systems/pull/3588))
+- As a consequence of the above, the prover's "rest of division by vanishing
+  polynomial" error can no longer fire: the quotient numerator is now divisible
+  by `Z_H` by construction. In release builds an unsatisfied witness produces a
+  proof that fails verification rather than an error from `create_recursive`.
+  Debug builds are unaffected, since `index.verify()` already rejects such a
+  witness earlier ([#3588](https://github.com/o1-labs/proof-systems/pull/3588))
+
 ### [kimchi-napi](./kimchi-napi)
 
 #### Fixed
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -1194,6 +1194,29 @@ impl<'a, F: FftField> EvalResult<'a, F> {
         }
     }
 
+    /// Apply `op` in place to the non-d1 rows of an evaluation over `domain`,
+    /// leaving the d1 rows untouched: they are already zero from the skipping
+    /// materialization (`init_`) and every constraint vanishes on them, so there
+    /// is no need to compute the in-place combination there. This is the
+    /// in-place analogue of the skip in `init_`.
+    fn update_non_d1<H: Sync + Send + Fn(usize, &mut F)>(
+        mut evals: Evaluations<F, D<F>>,
+        domain: Domain,
+        op: H,
+    ) -> Evaluations<F, D<F>> {
+        let stride = domain as usize;
+        let skip_d1 = stride > 1;
+        let mask = stride - 1;
+        o1_utils::cfg_iter_mut!(evals.evals)
+            .enumerate()
+            .for_each(|(i, e)| {
+                if !skip_d1 || (i & mask) != 0 {
+                    op(i, e);
+                }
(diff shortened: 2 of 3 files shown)


mrmr1993:
A github partial outage is causing CI to fail. When it resolves, @martyall can you try re-running? The latest commit fixes the genuine failure from the previous CI run.

martyall:
@mrmr1993 

I'm not detecting the same speed up that's being claimed here based on a bench suite interactively developed with claude, do you have a particular bench suite to run?

What's believable based on the report is that checking every index and deciding whether to zero out or evaluate + whatever memory traffic you incur might be more expensive than just evaluating small expressions. 

Given that this has a direct impact on completeness (or rather, prover side validation) and code complexity, I'm gonna push back to get a bench that we can objectively both measure.
 

# Benchmark report

**Setup:** Apple M5 Pro (18 cores), macOS arm64, rustc 1.92.0 `--release`, kimchi
default features (`prover` → `parallel`), Criterion 0.5, Pasta `Fp`.
Base `adc0540d26` vs PR head `490d2a9240`, separate worktrees and target dirs.

## Share of proving time

`Expr::evaluations` for the always-on gate constraints (CompleteAdd, VarbaseMul,
EndosclMul, EndomulScalar, Poseidon) at 2^16: **~490 ms of a ~1.9 s proof (~26%)**.

## Gate constraints, d8 (12.5% of rows skipped)

Real `constraint.evaluations(&env)` with a prover-identical `Environment`, 2^16.
Base/PR alternated across rounds to control drift.

| Round | Base | PR | Δ |
|---|---|---|---|
| 1 | 490.4 ms | 475.9 ms | −3.0% |
| 2 | 514.5 ms | 552.1 ms | +7.3% |
| 3 | 553.7 ms | 518.9 ms | −6.3% |

Median of 5 runs per side: base 490.4 ms, PR 479.6 ms — **−2.2%**, against ±12%
round-to-round drift.

## Generic constraint, d4 (25% of rows skipped)

Should show ~double the d8 effect. 2^16: base 9.41 / 8.56 ms, PR 8.86 / 9.01 ms — null.

## Changed loop shapes

Pre-PR and post-PR forms of each loop compiled into one binary and interleaved by
Criterion, over d8-length vectors of random field elements. Positive = PR slower.

| Loop shape | 2^14 | 2^16 |
|---|---|---|
| `init_` materialisation | −5.3% | +14.9% |
| in-place `+= constant` | +3.7% | +5.1% |
| in-place `+= Evals` | +7.1% | −3.9% |
| in-place `+= SubEvals` | +3.2% | −7.6% |

## Invalid-witness detection

Generic constraint evaluated against a uniformly random witness, 2^10.

| | Base | PR |
|---|---|---|
| d1 rows nonzero | 1014 / 1024 | 0 / 1024 |
| non-d1 rows nonzero | 3072 | 3072 |
| Caught by Z_H division | yes | no |

## Reading

- Constraint evaluation is ~26% of proving time, so the target is right.
- The 12.5% is not present: measured effect is ~−2%, inside the noise floor.
- No loop shape shows a consistent win, and signs flip between sizes — every element
  is still visited, so cache lines touched are unchanged while the `i & mask` test is
  paid per element.
- Scaling is 140 ms at 2^10 → 490 ms at 2^16 (3.5× for 64× data), so much of the cost
  is not row-proportional and cannot be reached by skipping rows.

martyall (commented):
After having done some extensive benching in the last days, I'm not convinced it's worth it to do the manual zero-ing of d1. There are other savings collected in #3600 from your work, but from what I can see on my end we're talking low single digit prover time savings at the cost of (in my opinion) valuable diagnostics. 

Plus this strategy can be layered on at a later time (together with some the same strategy from the fused constraint evaluator).

Instead I am going to turn the sketch at https://github.com/o1-labs/proof-systems/pull/3600 into a PR with correct commit attribution

martyall (changes requested):
@mrmr1993 This PR actually exposed the fact that no one ever updated the benchmark fixtures after the endomul breaking change (March 2026). The benchmark suite never asserted prover success (ie `unwrap`) so the benchmarks were just measuring how long took before the prover silently failed. Since this pr unsafely sets the d1 values to zero, it meant that the prover stoped failing and therefore the benchmarks were measuring the the proof creation time vs the proof failure time.

I updated the fixtures and this PR is now _meaningfully_ failing that check, and it also shows a measurable performance decrease on my machine. The best explanation I can have is that iterating through and canceling rather than doing cheap work is not worth it. Maybe some of this (eg the init) is salvageable on its own.

PR #3589, opened 2026-08-14
Performance: skip permutation argument constraint checks on d1 rows

mrmr1993:
This PR is the natural counterpart to #3588, which skips the d1 component of the permutation argument. The first commit in this series is shared with #3586.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ and this project adheres to
   disk while allowing the OS page cache to evict unused pages
   ([#3569](https://github.com/o1-labs/proof-systems/pull/3569))
 
+#### Changed
+
+- Skip the d1 rows in the permutation quotient. The permutation contribution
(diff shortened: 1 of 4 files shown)


martyall (approved):
The performance increase in terms of time is at the noise floor (this is just not in the hot path), but there are modest memory improvements coming from fewer d8 vectors floating around. 

| Source (per proof, SRS 2^16) | Old allocation | New allocation | Saving |
|---|---|---|---|
| Gamma broadcast (`constant_1_d8.scale(gamma)`) | 1 × 16 MB | 0 — scalar folded into the per-element closure | −16 MB |
| Shifts product, 7 witness columns | 3 temporaries × 16 MB × 7 = 336 MB | 1 `collect` × 16 MB × 7 = 112 MB | −224 MB |
| Sigmas product, 7 witness columns | 3 temporaries × 16 MB × 7 = 336 MB | 1 `collect` × 16 MB × 7 = 112 MB | −224 MB |

PR #3590, opened 2026-08-14
Fused constraint evaluator

mrmr1993:
This PR implements a 'fused' evaluator for expressions. This is used to compute the constraint evaluations for the quotient in a more efficient way (in terms of both memory and time).

The biggest win is to avoid creating a stack of d8 vectors while evaluating: instead of computing every intermediate value in the stack-based interpreter as a full-width vector, it evaluates the entire expression of a block of rows (currently hard-coded as size 8), and writes back the results to the final output vector. This performs better than the old algorithm even when ignoring the benefits to RAM consumption.

Embedded in this is the same optimisation as implemented in #3588: the d1 rows are omitted, since their evaluations will be calculated as 0 values there anyway (for a valid witness). The caveat around prover failures for invalid witnesses thus applies here too -- in the same way that a Groth16 prover will never catch a constraint failure unless they explicitly evaluate the constraints or check the proof, a prover using this option will produce an invalid proof for an invalid witness.

This is currently behind an env-var feature flag `KIMCHI_FUSED_EVAL`. This was done to allow it to be used for nori without changing the algorithm out from under Mina and o1js unilaterally, even while using the same compiled rust binary object. Reviewer-TODO: determine a better treatment for configurability if required.

his PR has been split out of the work on optimising nori native proving in OCaml for easy review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,15 @@ and this project adheres to
 - Treat public evaluations of oracles as vectors for chunking coherence
   ([#3577](https://github.com/o1-labs/proof-systems/pull/3577))
 
+### [o1-utils](./utils)
+
+#### Added
+
+- Add `cfg_chunks_mut!`, the mutable-chunks counterpart to the existing
+  `cfg_iter!` family, so callers can chunk in parallel without depending on
+  `rayon` unconditionally
+  ([#3586](https://github.com/o1-labs/proof-systems/pull/3586))
+
 ## 0.7.0
 
 ### [kimchi-napi](./kimchi-napi)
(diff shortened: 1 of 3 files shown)


martyall (commented):
This is great! I have a couple of comments:

1. Why make this optional instead of always-on
2. I have benchmarked against a baseline on master and I don't see the benefit of 

```rust
       let chunk = 1usize << 14; // multiple of B
```
Like I understand the reasoning RE L1, but fyi it seems like you can get just as good results (and significantly better on smaller domains eg 2^13,2^14) through a simple task distribution that's a function of `rayon::current_num_threads`

3. Again I'm not convinced that the e2e prover tme speedup is worth the manual zeroing at the expense of the diagnostics (based on what I've measured) and it's trivial to add that back in at a later point (together with the other recent work that does this)

Overall I approve. I have a sketch with the baseline bench + your work - zeroing that I will put up tomorrow and maybe we can go from there

martyall (dismissed):

martyall (approved):

querolita (approved):

PR #3591, opened 2026-08-14
Fat proofs: return the oracle challenges from the prover instead of recomputing them

mrmr1993:
This PR modifies the prover API to forward the oracle challenges that it computes instead of discarding them. This is of particular interest to pickles, where a locally-produced proof runs the prover -- who computes the challenges as a matter of course anyway -- and then has to run large portions of the verifier on the proof to recompute those challenges.

This PR forms the foundation to forward the oracle values all the way up to the pickles prover at proving time, removing some overhead from pickles' recursion.

### DIFF
--- a/kimchi-stubs/src/pasta_fp_plonk_proof.rs
+++ b/kimchi-stubs/src/pasta_fp_plonk_proof.rs
@@ -14,6 +14,7 @@ use kimchi::{
         lookup::runtime_tables::{caml::CamlRuntimeTable, RuntimeTable},
         polynomial::COLUMNS,
     },
+    oracles::caml::CamlOracles,
     proof::{
         PointEvaluations, ProofEvaluations, ProverCommitments, ProverProof, RecursionChallenge,
     },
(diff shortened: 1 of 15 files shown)


poly-commitment/src/lib.rs
```diff
@@ -251,6 +251,28 @@ type PolynomialsToCombine<'a, G: CommitmentCurve, D: EvaluationDomain<G::ScalarF
...
+/// [`OpenProof::open`] so the prover can surface the challenges -- which the
+/// verifier would otherwise recompute -- without changing the thin proof type.
+pub struct OpeningProofFat<Proof, F> {
+    pub proof: Proof,
+    pub challenges: Vec<F>,
+}
+
+impl<Proof, F: Clone> OpeningProofFat<Proof, F> {
+    /// Recover the thin (cryptographic) opening proof.
+    pub fn inner(self) -> Proof {
```
martyall, line 266: The `FatProverProof` type just uses `.proof`.  Now we have `inner()` as well. Suggestion, drop `inner` and just use proof

kimchi-stubs/src/pasta_fp_plonk_proof.rs
```diff
@@ -114,6 +115,87 @@ pub fn caml_pasta_fp_plonk_proof_create(
...
 
+/// Like [`caml_pasta_fp_plonk_proof_create`], but also returns the oracle
+/// challenges the prover already computed, so the caller (the wrap) can skip
+/// recomputing them via `caml_pasta_fp_plonk_oracles_create`.
+///
+/// The OCaml binding is hand-written (see `kimchi_bindings_extra.ml`): ocaml_gen
+/// cannot emit a signature mentioning `CamlOracles` from this module's scope, so
+/// `#[ocaml_gen::func]` is intentionally omitted.
+#[ocaml::func]
+pub fn caml_pasta_fp_plonk_proof_create_with_oracles(
```
martyall, line 126: this is just `caml_pasta_fp_plonk_proof_create` with a different return type, suggest that `caml_pasta_fp_plonk_proof_create` be implemented as a projection of this one

martyall (changes requested):

martyall:
looks good, just 2 suggestions that are about hygiene

PR #3592, opened 2026-08-14
kimchi_bindings: Add support for rayon worker pools with different sizes

mrmr1993:
This PR adds a `with_prover_pool` helper and a cache of rayon worker pools to access it, such the the prover invoked over the bindings can be run with different parallelism configurations. This is particularly useful when pipelining a tree of proofs -- for example, using a low rayon number while proving the leaves in parallel, all the way up to max cores for the tip.

Since the pool descriptions take up some memory, the number of cached rayon worked pools is governed by a cap set by the environment variable `KIMCHI_PROVE_POOL_CAP`. The pool that any prove job is initiated with is chosen via `KIMCHI_PROVE_THREADS`, which can be varied on the OCaml side by updating the contents of the environment variables.

This is used extensively in the nori optimised prover, and the maximum performance requires this. This is deliberately the least invasive mechanism from the perspective of the bindings consumers that do not use it, and also is designed such that e.g. the snark workers could harness it from a coordination layer without interfering with the snark worker internals.

### DIFF
--- a/kimchi-stubs/src/field_vector.rs
+++ b/kimchi-stubs/src/field_vector.rs
@@ -27,6 +27,16 @@ macro_rules! impl_vector_old {
                 (*v).push(x.into());
             }
 
+            // Empty the vector and release its backing allocation. OCaml only
+            // sees the pointer, so its GC never feels the (potentially large)
+            // Rust buffer; this lets a caller free it eagerly when done.
+            #[ocaml_gen::func]
+            #[ocaml::func]
+            pub fn [<$name:snake _clear>](mut v: $name) {
+                (*v).clear();
+                (*v).shrink_to_fit();
+            }
+
             #[ocaml_gen::func]
             #[ocaml::func]
             pub fn [<$name:snake _get>](
--- a/kimchi-stubs/src/lib.rs
+++ b/kimchi-stubs/src/lib.rs
@@ -123,11 +123,20 @@ pub(crate) fn with_prove_pool<R: Send>(f: impl FnOnce() -> R + Send) -> R {
             })
     };
     let result = pool.install(f);
-    prove_pool_cache()
-        .lock()
-        .unwrap()
-        .entry(n)
-        .or_default()
-        .push(pool);
+    // Return the warm pool for reuse, but bound the freelist (default 2 per N,
+    // override with KIMCHI_PROVE_POOL_CAP) so idle pools do not accumulate
+    // unbounded resident memory across a long conversion; overflow pools are
+    // dropped, joining their threads.
+    let cap = std::env::var("KIMCHI_PROVE_POOL_CAP")
+        .ok()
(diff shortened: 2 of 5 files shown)


kimchi-stubs/src/field_vector.rs
```diff
@@ -27,6 +27,16 @@ macro_rules! impl_vector_old {
                 (*v).push(x.into());
             }
 
+            // Empty the vector and release its backing allocation. OCaml only
+            // sees the pointer, so its GC never feels the (potentially large)
```
martyall:
```suggestion
            // sees the pointer, so its GC never frees the (potentially large)

```

kimchi-stubs/src/pasta_fp_plonk_proof.rs
```diff
@@ -99,15 +99,17 @@ pub fn caml_pasta_fp_plonk_proof_create(
...
-        let proof = ProverProof::create_recursive::<EFqSponge, EFrSponge, _>(
-            &group_map,
-            witness,
-            &runtime_tables,
-            index,
-            prev,
-            None,
-            &mut rand::rngs::OsRng,
-        )
+        let proof = crate::with_prove_pool(|| {
```
martyall: Other functions in this module which call `ProverProof::create_recursive` don't get this wrapper, some of which seem like examples but others no. Is this a potential problem or not?

kimchi-stubs/src/lib.rs
```diff
@@ -76,23 +76,58 @@ pub use {
...
-            .build()
-            .expect("KIMCHI_PROVE_THREADS thread pool")
-            .install(f),
-        _ => f(),
-    }
+        Some(n) if n >= 1 => n,
+        _ => return f(),
+    };
+    let pool = {
+        let mut cache = prove_pool_cache().lock().unwrap();
```
martyall: if you don't care about the `pool_cache` mutex being poisoned and want to proceed anyway, this code just throws an exception.

kimchi-stubs/src/lib.rs
```diff
@@ -75,3 +75,68 @@ pub use {
...
 };
+
+/// Per-thread-count freelist of warm rayon pools, reused across proves so a
+/// long-running worker does not pay thread-spawn + cold-cache cost on every
+/// compression proof.
+fn prove_pool_cache() -> &'static std::sync::Mutex<
+    std::collections::HashMap<usize, Vec<std::sync::Arc<rayon::ThreadPool>>>,
+> {
+    static CACHE: std::sync::OnceLock<
+        std::sync::Mutex<std::collections::HashMap<usize, Vec<std::sync::Arc<rayon::ThreadPool>>>>,
```
martyall: why use a hashmap here if you are only ever accessing one entry, namely whatever `KIMCHI_PROVE_THREADS` is?

martyall (approved):

PR #3594, opened 2026-08-17
Free the quotient polynomial in the prover early

mrmr1993:
This PR drops the quotient polynomial after its final use, reducing the memory usage at the tail of the prover function.

The quotient polynomial is committed to in `t_comm` and then only borrowed by `to_chunked_polynomial` when building `ft`, so it is dead from that point on. Rust drops locals at end of scope rather than at last use, so it otherwise stays resident across the evaluation and opening phases -- 14 MB at a 2^16 domain.

Proofs are unchanged; the compiler rejects any later use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to
   disk while allowing the OS page cache to evict unused pages
   ([#3569](https://github.com/o1-labs/proof-systems/pull/3569))
 
+#### Changed
+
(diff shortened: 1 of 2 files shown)


martyall:
@mrmr1993 Can  I suggest https://github.com/o1-labs/proof-systems/pull/3599 instead. We recently had #3587 which also manually dropped dome evaluations, and there was already some manual dropping and setting to `None` to free memory.

martyall:
closing as #3599 was merged

PR #3595, opened 2026-08-17
Add expression versions of the permutation terms

mrmr1993:
This PR updates the expression framework to support the primitives needed for the permutation contributions to the linearization and the constant term -- both of which are currently manually implemented.

This change is currently unused, but forms an obvious extension to the PRs #3581 and https://github.com/MinaProtocol/mina/pull/18939: using this we can erase the remaining hard-coded constraints from the pickles code, and drive everything from the expression framework in rust.

Those neglected PRs should be reviewed before considering this one; it's unlikely to be useful on its own.

### DIFF
--- a/kimchi/src/circuits/expr.rs
+++ b/kimchi/src/circuits/expr.rs
@@ -7,7 +7,8 @@ use crate::{
         gate::CurrOrNext,
         lookup::lookups::{LookupPattern, LookupPatterns},
         polynomials::{
-            foreign_field_common::KimchiForeignElement, permutation::eval_vanishes_on_last_n_rows,
+            foreign_field_common::KimchiForeignElement,
+            permutation::{coset_shifts, eval_vanishes_on_last_n_rows},
         },
     },
     collections::{HashMap, HashSet},
@@ -134,6 +135,32 @@ pub fn l0_1<F: FftField>(d: D<F>) -> F {
     d.elements()
         .skip(1)
         .fold(F::one(), |acc, omega_j| acc * (F::one() - omega_j))
+}
+
+/// The root of unity a row offset names, `omega^i`.
+pub fn root_of_unity<F: FftField>(domain: &D<F>, zk_rows: u64, i: &RowOffset) -> F {
+    let offset = if i.zk_rows {
+        -(zk_rows as i32) + i.offset
+    } else {
+        i.offset
+    };
+    if offset < 0 {
+        domain
+            .group_gen
+            .pow([-offset as u64])
+            .inverse()
+            .expect("a root of unity is invertible")
+    } else {
+        domain.group_gen.pow([offset as u64])
+    }
+}
+
+/// The polynomial vanishing on the given rows, `prod_j (x - omega^i_j)`,
+/// evaluated at `pt`.
+pub fn vanishing_on_rows<F: FftField>(domain: D<F>, zk_rows: u64, rows: &[RowOffset], pt: F) -> F {
+    rows.iter().fold(F::one(), |acc, i| {
+        acc * (pt - root_of_unity(&domain, zk_rows, i))
+    })
 }
 
 // Compute the ith unnormalized lagrange basis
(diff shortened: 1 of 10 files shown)


kimchi/src/circuits/expr.rs
```diff
@@ -136,6 +137,32 @@ pub fn l0_1<F: FftField>(d: D<F>) -> F {
...
 }
 
+/// The root of unity a row offset names, `omega^i`.
+pub fn root_of_unity<F: FftField>(domain: &D<F>, zk_rows: u64, i: &RowOffset) -> F {
+    let offset = if i.zk_rows {
+        -(zk_rows as i32) + i.offset
+    } else {
+        i.offset
+    };
+    if offset < 0 {
```
martyall, line 155: this is just

```rust
let row = if i.zk_rows { 
  domain.size - zk_rows + i.offset
  } else { 
  i.offset
};
domain.group_gen.pow([row])
```

kimchi/src/circuits/expr.rs
```diff
@@ -298,6 +325,21 @@ pub enum Operations<T> {
     Square(Box<Self>),
     Cache(CacheId, Box<Self>),
     IfFeature(FeatureFlag, Box<Self>, Box<Self>),
+    /// Divide by the polynomial vanishing on the given rows,
```
martyall, line 336: This comment reads like ai slop, I think you cna just remove it.

kimchi/src/circuits/expr.rs
```diff
@@ -608,6 +660,28 @@ pub enum ExprInner<C, Column> {
     /// UnnormalizedLagrangeBasis(i) is
     /// (x^n - 1) / (x - omega^i)
     UnnormalizedLagrangeBasis(RowOffset),
+    /// The polynomial vanishing on the zero-knowledge rows, evaluated at the
+    /// point the expression is evaluated at.
+    ///
+    /// Distinct from [`ExprInner::VanishesOnZeroKnowledgeAndPreviousRows`],
```
martyall, line 674: remove this block

kimchi/src/circuits/expr.rs
```diff
@@ -608,6 +660,28 @@ pub enum ExprInner<C, Column> {
...
+    /// which vanishes on one row more. This is the multiplier the permutation
+    /// argument uses to confine itself to the rows that carry witness data.
+    PermutationVanishingPolynomial,
+    /// The point the expression is evaluated at.
+    ///
+    /// Ordinarily implicit — an expression denotes a polynomial and the point
```
martyall, line 680: remove this block

kimchi/src/circuits/expr.rs
```diff
@@ -608,6 +660,28 @@ pub enum ExprInner<C, Column> {
...
+    /// is where you evaluate it. `ft(zeta)` is not such a polynomial: it is
+    /// something the verifier computes, and it mentions the point directly.
+    EvaluationPoint,
+    /// `x^n - 1`, the polynomial vanishing on the whole domain.
+    ///
+    /// [`ExprInner::UnnormalizedLagrangeBasis`] is this divided by a single
```
martyall, line 686: remove this block

kimchi/src/circuits/expr.rs
```diff
@@ -827,6 +910,15 @@ pub enum PolishToken<F, Column, ChallengeTerm> {
     Sub,
     VanishesOnZeroKnowledgeAndPreviousRows,
     UnnormalizedLagrangeBasis(RowOffset),
+    PermutationVanishingPolynomial,
+    EvaluationPoint,
+    VanishingPolynomial,
+    RootOfUnity(RowOffset),
+    Shift(usize),
+    /// Divide the top of the stack by the polynomial vanishing on these rows.
+    ///
```
martyall, line 927: remove

kimchi/src/circuits/expr.rs
```diff
@@ -2382,11 +2558,19 @@ where
             Double(e) => {
                 HashMap::from_iter(e.monomials(ev).into_iter().map(|(m, c)| (m, c.double())))
             }
+            DivideByVanishingOn(_, _) => {
```
martyall, line 2572: but it could be, you can just distribute it over `e.monomials(ev)` even if it's unreachable in practice. There are other instances already where we give an implementation that can never be called just because the math is coherent. (I think giving a degree impl ultimately has the same problem)

kimchi/src/circuits/polynomials/permutation.rs
```diff
@@ -67,6 +67,30 @@ use {
...
+    let mut shifts = [F::zero(); PERMUTS];
+
+    // first shift is the identity
+    shifts[0] = F::one();
+
+    // sample the other shifts
+    let mut i: u32 = 7;
+    for idx in 1..(PERMUTS) {
+        let mut shift = Shifts::sample(domain, &mut i);
+        // they have to be distincts
```
martyall, line 84: this is not the actual condition, they need to give distinct cosets no? the real check would be 

`there is no shift in shifts with (new_shift_canditate/shift)^n == 1`

granted the probability of messing this up is still tiny since we only need 7 of these things, but that's an even better reason to do it the right way -- you won't waste much time

kimchi/src/circuits/expr.rs
```diff
@@ -114,6 +114,14 @@ pub trait ColumnEnvironment<
 
     fn vanishes_on_zero_knowledge_and_previous_rows(&self) -> &'a Evaluations<F, D<F>>;
 
+    /// The polynomial vanishing on the zero-knowledge rows — one row fewer than
+    /// [`ColumnEnvironment::vanishes_on_zero_knowledge_and_previous_rows`].
+    ///
+    /// This is the multiplier the permutation argument uses. A protocol without
```
martyall, line 120: remove this block

martyall (changes requested):
looks good, just remove AI comments that have a way of building up. 

The only real feedback is around sampling the shifts, it looks "broken" and it also would imply that what was originally there was broken.

PR #3602, opened 2026-08-26
Optimize the IPA opening prover

martyall:
**Proving is now 1.5–2× faster: the IPA opening drops from 642 ms to 135 ms, cutting end-to-end proof creation by 32% at domain 2^16 and 53% at 2^14 — with byte-identical proofs.**

`SRS::open` dominated proving time (~60% of wall at domain 2^14 / SRS 2^16). Profiling (done with feature-gated instrumentation during development; removed before merge, along with the `libc` pin it required) showed the cost was mostly parallelism structure — fixed 4096-element chunks, per-element rayon inner loops, and `ark_ff::batch_inversion` forking per call — plus loop-invariant work recomputed inside the endo ladder.

Every change is a pure reordering/precompute: proof bytes were verified unchanged during development against a seeded-RNG digest (the development-time digest test itself is not kept). A new frozen benchmark (`poly-commitment/benches/fold_regression.rs`) grounds all numbers; each commit message carries its own A/B ledger.

Changes (one commit each):

- frozen regression bench
- `rayon::join` + de-alloc for the fold L/R MSMs
- borrow SRS bases in `open` instead of cloning (~4.5 MB/open)
- precompute the four ladder selector vectors `{±g2, ±φ(g2)}`
- single parallelism level: adaptive chunks, serial batch helpers, local serial Montgomery inversion
- ladder step as `2·acc + s` instead of `(acc + s) + acc`
- cap chunk size (env-tunable via `KIMCHI_IPA_FOLD_MAX_POINTS_PER_CHUNK`; doc comment carries the topology-free sizing rule) — insurance for small-cache/few-thread machines, inert on wide ones
- one chunk per thread (oversubscription factor swept 1×/2×/4×; 1× wins on all cases)

A design note for reviewers: re-expressing the chunked design with per-pass `ark_std::cfg_*` + grain floors was measured and rejected — it gives back ~85% of the win (`open` 156 → 572 ms) because chunking is what provides one fork per combine, core-to-data affinity across the ladder's ~130 passes, and cache blocking.

## Benchmarks

Apple M5 Pro (18 threads), medians, `cargo bench -p poly-commitment --bench fold_regression`:

| benchmark | before | after | Δ |
|---|---:|---:|---:|
| `combine_one_endo/1024` | 36.1 ms | 2.36 ms | −93% |
| `combine_one_endo/4096` | 49.6 ms | 4.20 ms | −92% |
| `combine_one_endo/32768` | 79.1 ms | 28.1 ms | −64% |
| `open/srs16` | 641.9 ms | 135.1 ms | −79% |

End-to-end proof creation (`proof_criterion`, SRS 2^16):

| workload | before | after | Δ |
|---|---:|---:|---:|
| domain 2^16 (matched) | 1.51 s | 1.03 s | −32% |
| domain 2^14 (padded) | ~1.00 s | ~0.47 s | −53% |


martyall:
https://github.com/o1-labs/proof-systems/actions/runs/33714413675/job/100520444146

poly-commitment/src/combine.rs
```diff
@@ -78,7 +78,7 @@ fn max_chunk() -> usize {
 /// all the batch helpers below run serially within one chunk.
 #[cfg(feature = "parallel")]
 fn chunk_size(n: usize) -> usize {
-    let num_chunks = (n / MIN_CHUNK).clamp(1, 2 * rayon::current_num_threads());
+    let num_chunks = (n / MIN_CHUNK).clamp(1, rayon::current_num_threads());
```
querolita, line 81: Nice observation, only 1 chunk per thread is better.

querolita (approved):
Excellent benchmarking and results from this optimization. Going forward.

PR #3605, opened 2026-08-30
Add an e2e prover memory profile under the diagnostics feature

martyall:
Timing regressions are covered by the criterion benches, but memory behavior is not: jemalloc's `stats::allocated` checkpoints (behind the `diagnostics` feature) only give point-in-time live bytes. This PR adds a standalone binary that profiles allocation over proof creation.

A `#[global_allocator]` is a process-wide singleton, so the profiler gets its own binary / process rather than living inside the test framework.

## Changes

- **`kimchi/src/bin/memory_profile.rs`** (`required-features = ["prover", "diagnostics"]`): proves the canonical benchmark circuit via `kimchi::bench::BenchmarkCtx` — the same circuit builder the criterion benches and the `flamegraph` bin use — under a counting wrapper around jemalloc. For the proof-creation window it reports: total bytes allocated, allocation count, peak live bytes (and its delta over bytes live at window open), plus jemalloc's peak resident set sampled every 25 ms by a background thread. Setup (index, verifier digest, Lagrange basis) completes before the window opens.
- `srs_log2` is a positional argument (default 16, validated to 4..=28).

The workload is deterministic, so a single run is the answer; to A/B a change, run at both revisions:

```
cargo run --release -p kimchi --bin memory_profile --features diagnostics -- 16
```

Allocated bytes are stable across runs; the allocation *count* is rayon-scheduling-noisy at 2^16+ — compare bytes.

## Sample output (Apple M5 Pro, 18 threads)

```
PROVER MEMORY PROFILE: domain/SRS 2^16
- setup time (index, verifier digest, lagrange basis): 528 ms
- proof creation allocation profile: allocated=6654.9 MB allocs=363946 peak_live=1340.8 MB peak_delta=758.0 MB peak_resident=1743.0 MB
- time to create proof: 1666 ms
```

## Sample output (CI, ubuntu-latest, 4 vCPU — via the manual `Prover memory profile` workflow)

```
PROVER MEMORY PROFILE: domain/SRS 2^16
- setup time (index, verifier digest, lagrange basis): 2426 ms
- proof creation allocation profile: allocated=6661.2 MB allocs=176269 peak_live=1294.7 MB peak_delta=712.1 MB peak_resident=1404.5 MB
- time to create proof: 7342 ms
```

Total allocated matches the M5 Pro run to 0.1%, while peak live/delta and alloc count differ — rayon scratch buffers scale with thread count. Compare runs from the same machine class only.

### DIFF
--- a/.github/workflows/memory-profile.yml
+++ b/.github/workflows/memory-profile.yml
@@ -1,0 +1,41 @@
+#
+# Manually-triggered end-to-end prover memory profile, running the
+# kimchi memory_profile binary. The byte metrics are deterministic for
+# a given runner class, but scale with thread count (rayon scratch
+# buffers) — compare runs from the same runner class only.
+#
+
+name: Prover memory profile
+
+on:
+  workflow_dispatch:
+    inputs:
+      srs_log2:
+        description: "log2 of the domain/SRS size (4..=28)"
+        required: false
+        default: "16"
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  memory_profile:
+    name: Run the prover memory profile
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Load versions
+        id: versions
+        uses: ./.github/actions/load-versions
+
+      - name: Use shared Rust toolchain setting up steps
+        uses: ./.github/actions/toolchain-shared
+        with:
+          rust_toolchain_version: ${{ steps.versions.outputs.rust-msrv }}
+
+      - name: Run the prover memory profile
+        run: |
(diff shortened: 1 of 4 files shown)


martyall:
Recently we were discussing RAM impact of recent PRs, but after looking into I found no profiling test or binary in the codebase to offer any baseline or point of comparison. This PR uses the standard `BenchmarkCtx` proof construction, I figured this is good enough but am open to suggestions.

cc @mrmr1993

iostat (approved):
✅Does the thing it claims to, by the looks of it. As an aside, given all the profiling/optimization work you're doing -- the Valgrind suite does work with Rust (and is just as molasses-slow as always) and may be handy when you want to drill deeper to find opts.

PR #3606, opened 2026-08-31
Fix the kimchi-stubs macOS cdylib link under CI's RUSTFLAGS

martyall:
Master CI's macOS tests jobs fail to link the `kimchi_stubs` cdylib with undefined `_caml_*` symbols (e.g. [this run](https://github.com/o1-labs/proof-systems/actions/runs/33406660958)).

**Root cause:** the cdylib (introduced in #3584) intentionally leaves the OCaml runtime symbols undefined for load-time resolution, which macOS's linker only accepts with `-undefined dynamic_lookup`. That flag lives in `.cargo/config.toml` under `[target.aarch64-apple-darwin] rustflags` — but cargo ignores config-file rustflags entirely whenever the `RUSTFLAGS` environment variable is set, and `ci.yml` sets `RUSTFLAGS: -Coverflow-checks=y -Cdebug-assertions=y` workflow-wide. Linux is unaffected because ELF shared objects permit undefined symbols by default. The break went unnoticed because every master CI run since #3584 died earlier (in `refresh-cache`); the first run to reach the macOS tests job exposed it.

**Fix:** emit the flag from a `kimchi-stubs/build.rs` via `cargo:rustc-link-arg-cdylib`, which is additive with `RUSTFLAGS` and scoped to the cdylib link only (the `lib`/`staticlib` targets and other platforms are untouched). This mirrors what napi-build already does for `kimchi-napi`'s cdylib, which links fine.

Verified locally on arm64 macOS: with CI's exact `RUSTFLAGS`, `cargo build -p kimchi-stubs` reproduces the CI failure at master and links cleanly with this change; plain builds and the release test target are unaffected.

The `.cargo/config.toml` entry is left in place — it still serves builds of other targets that link OCaml symbols when `RUSTFLAGS` is unset.

https://claude.ai/code/session_01QC5hVAkPFx6DCeghdALffj

**Update (review):** kimchi-wasm needed no flag after all — its host cdylib was only ever built because kimchi-napi carried a vestigial `kimchi_wasm` dependency (the shared helpers long since moved to `wasm-types`). Dropped the dependency and deleted `kimchi-wasm/build.rs`; only the kimchi-stubs fix remains.


kimchi-wasm/build.rs
```diff
@@ -0,0 +1,16 @@
+fn main() {
```
iostat: Why does this flag need to exist for kimchi-wasm? I don't believe that crate should be linked into any kind of macOS binary...

martyall: @iostat 

kimchi-napi depends on kimchi_wasm, and apparently cargo links a dependency's cdylib target even when only the rlib is used

iostat: I think the actual fix here might be to just remove kimchi-wasm from kimchi-napi's deps. Seems like at some point shared types moved to wasm-types, so this dependency (and accompanying build.rs) is vestigial.

martyall: Good call — it is vestigial: kimchi-napi imports everything it shares from wasm-types and never references kimchi_wasm. Dropped the dependency and deleted kimchi-wasm/build.rs in 98d08081c2; with the dep gone, kimchi_wasm's host cdylib is never built (it was only ever compiled as a dependency of kimchi-napi — nextest/make build --exclude it as a target), so the link flag has nothing left to fix. Verified with CI's exact RUSTFLAGS on arm64 macOS: cargo build -p kimchi-napi with the workspace features links clean and kimchi_wasm no longer appears in the build graph. kimchi-stubs keeps its build.rs — that cdylib is built as a workspace member in its own right.

iostat (approved):

PR #3608, opened 2026-09-01
memory_profile: run over the mina circuit fixtures

martyall:
The prover memory profiler (#3605) could only measure the synthetic `BenchmarkCtx` circuit. This teaches it to profile the real serialised mina circuits — the same `kimchi_inputs_CURVE_SEED.ser` fixtures `proof_criterion_mina` consumes (regenerated and re-uploaded after #3607) — and turns the CI workflow into a gated base-vs-head comparison.

**Binary** (clap-derived CLI, JSON-only output):
- `memory_profile synthetic [--srs-log2 16]` — the synthetic circuit
- `memory_profile fixture <path>` — loads a fixture via `bench_arguments_from_file`, parsing curve and seed from the filename exactly as the criterion bench does, and profiles `create_recursive` (setup outside the measurement window; proof creation unwrapped so a stale fixture fails loudly)
- One JSON object per run (byte counts, not MB); one fixture per process, since jemalloc retains pages and `peak_resident` is only trustworthy for the first proof

**Scripts** (the CI job is nothing but these, so the whole flow runs locally):
- `scripts/memory-profile-mina-circuits.sh <out.jsonl>` — profiles every fixture (GCS download, or `FIXTURES_DIR` override) plus a synthetic 2^16 row as a size anchor (no mina fixture reaches that domain)
- `scripts/memory-profile-diff.py <base.jsonl> <head.jsonl>` — joins on (workload, curve, seed) and renders a markdown matrix, cells `head (delta vs base, %)`

**Regression gate**: the diff exits nonzero if `allocated`, `peak_live`, or `peak_delta` grows by more than 1% on any row (offending cells flagged in the matrix). These metrics are deterministic per runner class, and 1% of peak is roughly one domain-sized d8 polynomial buffer, so the gate reads as "this change retains at least a polynomial's worth of extra memory". Allocation counts, resident set, and wall time are report-only — timing regressions remain the criterion bench's job.

**CI** (`memory-profile.yml`, dispatch-only): profiles the merge base with master and HEAD back to back on the same runner (the `benches-mina-prover.yml` shape, for the same reason), writes the matrix to the job summary, and fails on the gate.

Verified locally: both subcommands on both curves; the full pipeline base-vs-head over all fixtures for master vs #3602 (gate correctly silent on an improvement: allocated −1.8%, byte peaks bit-identical, and correctly firing with the sides reversed); identity diff all `=`.

Note: dispatching the workflow produces a comparison once the merge base contains this PR (same property as the criterion bench workflow).

https://claude.ai/code/session_01KcoXpnzQ2ufCzqzB1tZtfm

### DIFF
--- a/kimchi/src/bin/memory_profile.rs
+++ b/kimchi/src/bin/memory_profile.rs
@@ -13,31 +13,39 @@
 //! Usage:
 //!
 //! ```text
-//! cargo run --release -p kimchi --bin memory_profile --features diagnostics -- [srs_log2]
-//! cargo run --release -p kimchi --bin memory_profile --features diagnostics -- <kimchi_inputs_CURVE_SEED.ser>
+//! cargo run --release -p kimchi --bin memory_profile --features diagnostics -- synthetic [--srs-log2 16]
+//! cargo run --release -p kimchi --bin memory_profile --features diagnostics -- fixture <kimchi_inputs_CURVE_SEED.ser>
 //! ```
 //!
-//! An integer argument selects the synthetic benchmark circuit and sets its
-//! domain/SRS size (default 16); a path selects a mina fixture, whose curve
-//! and seed are parsed from the filename exactly as in `proof_criterion_mina`.
+//! `synthetic` proves the benchmark circuit at the given domain/SRS size;
+//! `fixture` proves a mina fixture, whose curve and seed are parsed from the
+//! filename exactly as in `proof_criterion_mina`. Output is a single JSON
+//! object on stdout (byte counts, not MB), meant to be collected across
+//! fixtures and diffed against a baseline run — see
+//! `scripts/memory-profile-mina-circuits.sh` and
+//! `scripts/memory-profile-diff.py`.
 //! One fixture per invocation: jemalloc retains pages across proofs, so
 //! `peak_resident` is only trustworthy for the first proof in a process.
 //! The resident-set sampler interval is `KIMCHI_MEMORY_PROFILE_SAMPLE_MS`
 //! (default 25).
 
+use ark_ff::PrimeField;
+use clap::Parser;
 use groupmap::GroupMap;
 use kimchi::{
     bench::{
         bench_arguments_from_file, BaseSpongePallas, BaseSpongeVesta, BenchmarkCtx,
         ScalarSpongePallas, ScalarSpongeVesta,
     },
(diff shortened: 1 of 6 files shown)


kimchi/src/bin/memory_profile.rs
```diff
@@ -123,50 +149,157 @@ mod mem_profile {
...
+        setup_ms,
+        prove_ms,
+        &metrics,
+    );
+}
+
+fn profile_mina_fixture(filename: &str) {
+    // Parse filename "kimchi_inputs_CURVENAME_SEED.ser" into two parameters,
+    // exactly as `proof_criterion_mina` does.
+    let (curve_name, seed): (&str, &str) = filename
```
martyall: make a datatype that represents the fixture and implement a separate parser for that data type, don't inline it

kimchi/src/bin/memory_profile.rs
```diff
@@ -123,50 +149,157 @@ mod mem_profile {
...
+        .next_back()
+        .unwrap()
+        .strip_prefix("kimchi_inputs_")
+        .and_then(|s| s.strip_suffix(".ser"))
+        .and_then(|s| s.split_once('_'))
+        .unwrap_or_else(|| {
+            panic!("fixture filename must look like kimchi_inputs_CURVE_SEED.ser, got {filename:?}")
+        });
+
+    macro_rules! profile_curve {
```
martyall: factor this out if you want a macro

kimchi/src/bin/memory_profile.rs
```diff
@@ -123,50 +149,157 @@ mod mem_profile {
...
-        Some(s) => s
-            .parse()
-            .unwrap_or_else(|_| panic!("srs_log2 must be an integer, got {s:?}")),
-    };
-    assert!(
-        (4..=28).contains(&srs_log2),
-        "srs_log2 must be in 4..=28, got {srs_log2}"
-    );
-    println!("PROVER MEMORY PROFILE: domain/SRS 2^{srs_log2}");
+fn print_record(head: &[(&str, String)], setup_ms: u128, prove_ms: u128, m: &mem_profile::Metrics) {
```
martyall: why not just use a format/pretty print instance for Mtrics insterad of deriving it yourself in an ad-hoc-fashion

iostat (approved):

PR #3628, opened 2026-09-05
kimchi-stubs: in-place Bigstring to_bytes/of_bytes

### LINKED ISSUE #3627: kimchi-stubs: in-place (bigstring + offset) variants of pasta field / bigint to_bytes and of_bytes to avoid per-element allocations in Mina serialization
## Summary

`caml_pasta_fp_to_bytes` / `caml_pasta_fq_to_bytes` / `caml_bigint_256_to_bytes` each return a freshly allocated 32-byte OCaml `bytes`, and the `of_bytes` counterparts take a `&[u8]` that the OCaml side has to cut out of its buffer first. On Mina's serialization hot paths (bin_prot writers and readers for field elements) this makes every field element cost one or two heap allocations plus a copy, in both directions. A pair of stubs that serialize into / deserialize from a caller-provided `Bigstring` at an offset would remove them.

## Where it shows up

In Mina (`o1-labs/mina`, `compatible` at 1e83915c08, proof-systems submodule at 89edaa3204), the field element `Binable` is `Of_binable (Bigint)`, and `Bigint.bin_write_t` is:

```ocaml
let bin_write_t buf ~pos t =
  let bytes = to_bytes t in                      (* caml_bigint_256_to_bytes: fresh Bytes *)
  ...
  Bigstring.From_bytes.blit ~src:bytes ~src_pos:0 ~len:length_in_bytes ~dst:buf ~dst_pos:pos ;
  pos + len
```

and `bin_read_t` does `Bigstring.To_bytes.sub` (allocate + copy) and then `of_bytes` (parse into a new element).

Measured on a dummy Pickles `Proofs_verified_2` proof (10,368 bytes, ~300 field elements) after Mina-side fixes that already removed the other allocation sources from the writer:

| operation | allocation per proof |
|---|---|
| `bin_write_t` | ~50 KB, of which ~46 KB is field-element writes (`to_bigint` custom block + `to_bytes` Bytes per element) |
| `bin_read_t` (decode) | ~108 KB (sub + bigint + `of_bigint` per element) |

Multiplied by the scan-state hash (hundreds of proofs per call, once per block on every node), the LMDB proof cache (every read decodes), and gossip, this is a large share of the daemon's steady-state garbage. Mina PR #19363 and follow-up profiling: https://github.com/MinaProtocol/mina/pull/19363.

## Request

Add, alongside the existing functions in `kimchi-stubs/src/arkworks/{pasta_fp,pasta_fq,bigint_256}.rs`:

```rust
/// Serialize `x` into `buf[pos .. pos + 32]`; same bytes as `to_bytes`.
pub fn caml_pasta_fp_to_bytes_into(x: ocaml::Pointer<CamlFp>, buf: ocaml::bigarray::Array1<u8>, pos: usize) -> Result<(), ocaml::Error>
/// Deserialize from `buf[pos .. pos + 32]`; same semantics as `of_bytes`.
pub fn caml_pasta_fp_of_bytes_from(buf: ocaml::bigarray::Array1<u8>, pos: usize) -> Result<CamlFp, ocaml::Error>
```

(and the `fq` / `bigint_256` equivalents), with bounds checks returning an error rather than panicking. The existing `to_bytes` body is `serialize_compressed(&mut res[..])` into a stack array; the new one would serialize into the slice of the caller's bigarray instead. Since `pasta_bindings.ml` is generated by `ocaml_gen`, exposing them is automatic once the functions exist.

With those, Mina's field `bin_write_t` becomes a single stub call with no temporary and no blit, and `bin_read_t` parses in place, taking a field element's serialization cost to zero allocations in both directions. The wire bytes are unchanged.

## Notes

- Not blocking: Mina can halve the write-side allocation on its own by calling `Fp.to_bytes` directly instead of going through `to_bigint` + `Bigint.to_bytes`; the in-place stubs are what removes the rest.
- Happy to open the PR if the API shape above is acceptable.

martyall:
Add `caml_pasta_{fp,fq}_{to_bytes_into,of_bytes_from}` and `caml_bigint_256_{to_bytes_into,of_bytes_from}`, which serialize into, or deserialize from, a caller-provided Bigstring at an offset. Same wire bytes as to_bytes / of_bytes, but with no intermediate `bytes` allocation, so Mina's bin_prot writer and reader for a field element can allocate nothing.

The buffer is passed as a new `CamlBigstring` wrapper around `ocaml::bigarray::Array1<u8>`, described to ocaml-gen as `(char, int8_unsigned_elt, c_layout) Bigarray.Array1.t`, i.e. `Core.Bigstring.t`. Out-of-range or negative offsets raise Invalid_argument instead of panicking across the FFI.

Fixes #3627

### DIFF
--- a/kimchi-stubs/src/arkworks/bigint_256.rs
+++ b/kimchi-stubs/src/arkworks/bigint_256.rs
@@ -1,4 +1,4 @@
-use crate::caml::caml_bytes_string::CamlBytesString;
+use crate::caml::{caml_bigstring::CamlBigstring, caml_bytes_string::CamlBytesString};
 use ark_ff::{BigInteger as ark_BigInteger, BigInteger256};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use core::{
@@ -228,6 +228,44 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<CamlBigInteger256, ocaml::Er
     Ok(CamlBigInteger256(result))
 }
 
+/// Serialize `x` into `buf[pos .. pos + 32]`, the same bytes as
+/// `caml_bigint_256_to_bytes`, without allocating. Raises `Invalid_argument`
+/// when the window does not fit in `buf`.
+#[ocaml_gen::func]
+#[ocaml::func]
+pub fn caml_bigint_256_to_bytes_into(
+    x: ocaml::Pointer<CamlBigInteger256>,
+    mut buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<(), ocaml::Error> {
+    let dst = buf.slice_mut(
+        pos,
+        core::mem::size_of::<BigInteger256>(),
+        "caml_bigint_256_to_bytes_into",
+    )?;
+    x.as_ref().0.serialize_compressed(dst).unwrap();
+    Ok(())
+}
+
+/// Deserialize from `buf[pos .. pos + 32]`, the same semantics as
+/// `caml_bigint_256_of_bytes`, without copying the bytes out first. Raises
+/// `Invalid_argument` when the window does not fit in `buf`.
+#[ocaml_gen::func]
+#[ocaml::func]
+pub fn caml_bigint_256_of_bytes_from(
+    buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<CamlBigInteger256, ocaml::Error> {
+    let src = buf.slice(
+        pos,
+        core::mem::size_of::<BigInteger256>(),
+        "caml_bigint_256_of_bytes_from",
+    )?;
+    let result = BigInteger256::deserialize_compressed(src)
+        .map_err(|_| ocaml::Error::Message("deserialization error"))?;
+    Ok(CamlBigInteger256(result))
+}
+
 #[ocaml_gen::func]
 #[ocaml::func]
 pub fn caml_bigint_256_deep_copy(x: CamlBigInteger256) -> CamlBigInteger256 {
@@ -253,5 +291,30 @@ mod tests {
         let x2: BigUint = y.into();
         assert!(x2 == x);
         println!("biguint.to_string: {}", x2);
+    }
+
+    /// The in-place path writes the same bytes as `to_bytes`, at the offset,
+    /// and reads them back.
+    #[test]
+    fn in_place_bytes_match_to_bytes() {
+        let x = BigInteger256::from(0xdead_beef_u64);
+        let len = core::mem::size_of::<BigInteger256>();
+
+        let mut expected = vec![0u8; len];
+        x.serialize_compressed(&mut expected[..]).unwrap();
+
+        let mut buf = vec![0xffu8; 3 * len];
+        let range =
+            crate::caml::caml_bigstring::checked_range(buf.len(), len as ocaml::Int, len, "t")
+                .unwrap();
+        x.serialize_compressed(&mut buf[range.clone()]).unwrap();
+
+        assert_eq!(&buf[range.clone()], &expected[..]);
+        assert!(buf[..len].iter().all(|&b| b == 0xff));
+        assert!(buf[2 * len..].iter().all(|&b| b == 0xff));
+        assert_eq!(
+            BigInteger256::deserialize_compressed(&buf[range]).unwrap(),
+            x
+        );
     }
 }
--- a/kimchi-stubs/src/arkworks/pasta_fp.rs
+++ b/kimchi-stubs/src/arkworks/pasta_fp.rs
@@ -1,4 +1,7 @@
-use crate::{arkworks::CamlBigInteger256, caml::caml_bytes_string::CamlBytesString};
+use crate::{
+    arkworks::CamlBigInteger256,
+    caml::{caml_bigstring::CamlBigstring, caml_bytes_string::CamlBytesString},
+};
 use ark_ff::{FftField, Field, One, PrimeField, UniformRand, Zero};
 use ark_poly::{EvaluationDomain, Radix2EvaluationDomain as Domain};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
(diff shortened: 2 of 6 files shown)


kimchi-stubs/src/arkworks/bigint_256.rs
```diff
@@ -228,6 +228,44 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<CamlBigInteger256, ocaml::Er
     Ok(CamlBigInteger256(result))
 }
 
+/// Serialize `x` into `buf[pos .. pos + 32]`, the same bytes as
+/// `caml_bigint_256_to_bytes`, without allocating. Raises `Invalid_argument`
+/// when the window does not fit in `buf`.
+#[ocaml_gen::func]
+#[ocaml::func]
+pub fn caml_bigint_256_to_bytes_into(
```
glyh: suggest rename to `caml_bigint_256_blit_to_bigstring`, same goes for funcs in pasta_fp & pasta_fq.

kimchi-stubs/src/arkworks/bigint_256.rs
```diff
@@ -228,6 +228,44 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<CamlBigInteger256, ocaml::Er
...
+    x.as_ref().0.serialize_compressed(dst).unwrap();
+    Ok(())
+}
+
+/// Deserialize from `buf[pos .. pos + 32]`, the same semantics as
+/// `caml_bigint_256_of_bytes`, without copying the bytes out first. Raises
+/// `Invalid_argument` when the window does not fit in `buf`.
+#[ocaml_gen::func]
+#[ocaml::func]
+pub fn caml_bigint_256_of_bytes_from(
```
glyh: suggests rename to `caml_bigint_256_of_bigstring`, same goes for funcs in pasta_fp & pasta_fq.

glyh (commented):
Reviewed for correctness, performance, and error handling.

Correctness and performance both look good to me. I traced the buffer handling through ocaml-rs: `Array1<T>` is `#[repr(transparent)]` over `Value`, a `Bigstring`'s bytes live in a malloc'd region outside the OCaml heap so neither the GC nor compaction relocates them, and no OCaml allocation happens while the `&[u8]` / `&mut [u8]` borrow is live — so no rooting is needed and none is missing. `checked_range` covers negative `pos`, `usize` overflow, and overrun, and the five unit tests on it are the load-bearing ones. `Validate::Yes` is preserved, so the rejection set for non-canonical values is identical to the old `BigInt256::deserialize` -> `from_bigint` path, and the wire bytes are unchanged. I also checked the fq file specifically for stray `Fp` copy-paste and found none. Per call the path is genuinely allocation-free; what's left is arkworks' Montgomery reduction, which is inherent to the format.

My only substantive comments are about error handling, inline below. All are cheap, and none are blocking.

One non-blocking note that doesn't fit on a line: every raise path in ocaml-rs allocates (a `CString`, or a `String` via `format!("{:?}\0", e)`) and then longjmps through `caml_failwith`, leaking it. For `of_bytes_from` that's per malformed element on a path a peer can trigger. It's identical to the existing `of_bytes`, so not a regression, and the amounts are tiny — just worth knowing it's there.

Separately, a test observation rather than an error-handling one: the per-field tests re-execute the stub body inline rather than calling the stub, since the real entry point needs a live OCaml runtime. They validate the approach well, but a wrong length constant or a swapped argument *inside* a stub would pass. Today that's only caught downstream in Mina's `test_serialization.ml`, which means this PR can't be verified standalone — probably worth saying so explicitly in the description.

kimchi-stubs/src/arkworks/pasta_fp.rs
```diff
@@ -323,8 +326,72 @@ pub fn caml_pasta_fp_of_bytes(x: &[u8]) -> Result<CamlFp, ocaml::Error> {
...
+    x: ocaml::Pointer<CamlFp>,
+    mut buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<(), ocaml::Error> {
+    let dst = buf.slice_mut(
+        pos,
+        core::mem::size_of::<Fp>(),
+        "caml_pasta_fp_to_bytes_into",
+    )?;
+    x.as_ref().0.serialize_compressed(dst).unwrap();
```
glyh: This `.unwrap()` contradicts the doc comment three lines up, which promises these "[raise] `Invalid_argument` ... instead of panicking across the FFI".

It isn't UB — ocaml-rs installs a global panic hook that calls `caml_failwith` from inside the hook — but that longjmps out of the hook, skipping every Rust destructor on the way, and surfaces as an opaque `Failure`.

It's also genuinely unreachable: `dst` is exactly `serialized_size(Compress::Yes)`, so the write cannot come up short. Which is the argument for spelling that out rather than leaving a panic in the binary:

```rust
x.as_ref().0
    .serialize_compressed(dst)
    .map_err(|_| ocaml::Error::Message("caml_pasta_fp_to_bytes_into: serialization failed"))?;
```

Zero cost on the success path, and it makes the doc comment true. Same in `pasta_fq.rs` and `bigint_256.rs`.

kimchi-stubs/src/arkworks/pasta_fq.rs
```diff
@@ -323,8 +326,72 @@ pub fn caml_pasta_fq_of_bytes(x: &[u8]) -> Result<CamlFq, ocaml::Error> {
...
+    x: ocaml::Pointer<CamlFq>,
+    mut buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<(), ocaml::Error> {
+    let dst = buf.slice_mut(
+        pos,
+        core::mem::size_of::<Fq>(),
+        "caml_pasta_fq_to_bytes_into",
+    )?;
+    x.as_ref().0.serialize_compressed(dst).unwrap();
```
glyh: Same as `pasta_fp.rs:344` — unreachable `.unwrap()` that the doc comment says won't be there. `.map_err(|_| ocaml::Error::Message(...))?` instead.

kimchi-stubs/src/arkworks/bigint_256.rs
```diff
@@ -228,6 +228,44 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<CamlBigInteger256, ocaml::Er
...
+    x: ocaml::Pointer<CamlBigInteger256>,
+    mut buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<(), ocaml::Error> {
+    let dst = buf.slice_mut(
+        pos,
+        core::mem::size_of::<BigInteger256>(),
+        "caml_bigint_256_to_bytes_into",
+    )?;
+    x.as_ref().0.serialize_compressed(dst).unwrap();
```
glyh: Same as `pasta_fp.rs:344` — unreachable `.unwrap()` that the doc comment says won't be there. `.map_err(|_| ocaml::Error::Message(...))?` instead.

kimchi-stubs/src/arkworks/pasta_fp.rs
```diff
@@ -323,8 +326,72 @@ pub fn caml_pasta_fp_of_bytes(x: &[u8]) -> Result<CamlFp, ocaml::Error> {
...
+pub fn caml_pasta_fp_of_bytes_from(
+    buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<CamlFp, ocaml::Error> {
+    let src = buf.slice(
+        pos,
+        core::mem::size_of::<Fp>(),
+        "caml_pasta_fp_of_bytes_from",
+    )?;
+    let x = Fp::deserialize_compressed(src)?;
```
glyh: The bare `?` here and the explicit `map_err` in `caml_bigint_256_of_bytes_from` raise two different exceptions for the same failure.

The blanket `impl<T: std::error::Error> From<T> for Error` in ocaml-rs boxes the `SerializationError` into `Error::Error(Box<dyn Error>)`, and the raise path renders it with `format!("{:?}", e)`. So this one surfaces in OCaml as `Failure "InvalidData"` — arkworks' `Debug` output — while the bigint one surfaces as `Failure "deserialization error"`.

This faithfully mirrors the existing `of_bytes` pair, so it isn't a new inconsistency. But these readers are about to sit on Mina's untrusted-gossip decode path (MinaProtocol/mina#19373), where a caller may reasonably want to tell malformed input from a bug, and a message that tracks whatever arkworks' `Debug` happens to print is a thin thing to key on. I'd make all six consistent — `Error::Message` naming the function reads best:

```rust
let x = Fp::deserialize_compressed(src)
    .map_err(|_| ocaml::Error::Message("caml_pasta_fp_of_bytes_from: not a canonical field element"))?;
```

Also avoids the `format!` allocation on a path a peer can trigger.

kimchi-stubs/src/arkworks/bigint_256.rs
```diff
@@ -228,6 +228,44 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<CamlBigInteger256, ocaml::Er
...
+    buf: CamlBigstring,
+    pos: ocaml::Int,
+) -> Result<CamlBigInteger256, ocaml::Error> {
+    let src = buf.slice(
+        pos,
+        core::mem::size_of::<BigInteger256>(),
+        "caml_bigint_256_of_bytes_from",
+    )?;
+    let result = BigInteger256::deserialize_compressed(src)
+        .map_err(|_| ocaml::Error::Message("deserialization error"))?;
```
glyh: Counterpart to `pasta_fp.rs:362`: this one produces `Failure "deserialization error"` while fp/fq produce `Failure "InvalidData"` for the same failure. Whichever shape you pick, worth making all six agree.

glyh:
I'm not very familiar with this codebase. Is there a good reason almost identical funcs getting implemented 3 times here?

kimchi-stubs/src/arkworks/bigint_256.rs
martyall: Done in 8e16305: renamed to *_blit_to_bigstring / *_of_bigstring for bigint_256, pasta_fp and pasta_fq.

kimchi-stubs/src/arkworks/pasta_fp.rs
martyall: Done in 8e16305 for all three: map_err to Error::Message("<fn>: serialization failed"), no unwrap left.

kimchi-stubs/src/arkworks/pasta_fp.rs
martyall: Done in 8e16305. All six readers now raise Error::Message naming the function. fp/fq (of_bytes and of_bigstring): "<fn>: not a canonical field element". bigint_256: "<fn>: deserialization error", since a BigInteger256 is not a field element, and with an exact 32-byte window that arm is unreachable anyway.

martyall:
> I'm not very familiar with this codebase. Is there a good reason almost identical funcs getting implemented 3 times here?

yeah, soemtimes there's a macro to get around this, but in this case I just made a common implementation and wrote out all the bindings to use it

